Issue #8, opened 2019-02-12
high level, low level - confusing naming


DanielMazurkiewicz:
As it is separate topic I've made this issue as a continuation to this post: https://github.com/webmachinelearning/webnn/issues/3#issuecomment-462452394

@anssiko wrote:

> Re low-level and high-level, I observe much of the confusion arises from the inconsistent use of these adjectives in different contexts. Low-level & high-level APIs and low-level & high-level use cases do not map and that causes us talking past each other. We need to add definitions of there terms to the spec, or come up with better names.

Totally agree, easy to mix these terms, this should be more self describing in my opinion, more-less like this:

Predefined ML models:
* predefined models (or as @anssiko mentioned "pre-canned models")
* predefined models API

Generic ML:
* ML API - top level ML api for NN preparing, modifying, running, training, data preparing
* ML operators API - operators specific api
* operators list
* graph format

And each of above will have naturally its own use cases.

anssiko:
I submitted #10 as a partial fix. To me, _application-level_ and _API-level_ are more self-documenting than generic _low-level_ and _high-level_. Let the bikeshedding begin.

anssiko:
It seems #10 addressed the immediate concern with naming, and the spec is now consistent with the charter language.

PR #10, opened 2019-02-13
Rename use case subsections

anssiko:
'Low-level' and 'high-level' are ambiguous in this context.

Partial fix to #8.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/10.html" title="Last updated on Feb 18, 2019, 10:42 AM UTC (e48c3ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/10/94c6171...e48c3ce.html" title="Last updated on Feb 18, 2019, 10:42 AM UTC (e48c3ce)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -18,7 +18,7 @@ Introduction here.
 Use cases {#usecases}
 =====================
 
-## High-Level Use Cases ## {#usecases-highlevel}
+## Application Use Cases ## {#usecases-application}
 
 This section illustrates application-level use cases for neural network
 inference hardware acceleration. All applications in those use cases can be
@@ -116,10 +116,10 @@ it needs to reduce recorded video data to be stored. The application generates
 the short version of the recorded video by using a machine learning model for
 video summarization such as [[Video-Summarization-with-LSTM]].
 
-## Low-Level Use Cases ## {#usecases-lowlevel}
+## Framework Use Cases ## {#usecases-framework}
 
-This section collects API-level use cases for a dedicated low-level API for
-neural network inference hardware acceleration. It is expected that Machine
+This section collects framework-level use cases for a dedicated low-level API
+for neural network inference hardware acceleration. It is expected that Machine
 Learning frameworks will be key consumers of the Web Neural Network API (WebNN
 API) and the low-level details exposed through the WebNN API are abstracted out
 from typical web developers. However, it is also expected that web developers
--- a/index.html
+++ b/index.html
@@ -1214,7 +1214,7 @@ Possible extra rowspan handling
 </style>
   <meta content="Bikeshed version 0da7328bb90ef81993146377e4e0fed236969c4c" name="generator">
   <link href="https://webmachinelearning.github.io/webnn/" rel="canonical">
-  <meta content="1fdaa40ef1f3aa409021ee5b4d83456853561c04" name="document-revision">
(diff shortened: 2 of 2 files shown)


gregwhitworth:
Hey @anssiko - 

Thanks for taking a look at this, Application and API seem just as ambigous to me as you'll need to utilize a Javascript API in order to build any web application that uses this API.

Might I propose the following, as it highlights the need for various capabilities of the API itself:

* Operator
* Load Model

The operator will require operation based level access in order to make it possible, the load model level is merely saying that they'll need access to an inference engine in order to evaluate against a model, the author doesn't necessarily need access to the operations in order to achieve their use case (eg: face or object detection).

tomoyukilabs:
IMO, "application-level" and "API-level" might still look ambiguous as @gregwhitworth has pointed out.

I would like to suggest *application-level* and *execution-level*. In this case, application-level use cases would illustrate typically required operators (linear, convolutions, activations, normalizations, merging, etc.), and execution-level use cases would clarify requirements of capabilities such as operation-level access and loading entire NN models.

huningxin:
Thanks for making this PR @anssiko !

The "application-level" looks good to me. But I agree with @anssiko and @tomoyukilabs that "API-level" might be ambiguous.

By re-reading the text, my suggestion is "framework-level" because the existing low-level use cases section says ["Machine Learning frameworks will be key consumers of the Web Neural Network API (WebNN API)"](https://webmachinelearning.github.io/webnn/#usecases-lowlevel). 

Or just use "application use cases" and "framework use cases".

tomoyukilabs:
@huningxin I can agree with "framework-level" if we don't consider load-model API. I guess that load-model API can be easily called by front-end web apps without NN frameworks.

> the existing low-level use cases section says ["It is expected that Machine Learning frameworks will be key consumers of the Web Neural Network API (WebNN API) and the low-level details exposed through the WebNN API are abstracted out from typical web developers."](https://webmachinelearning.github.io/webnn/#usecases-lowlevel).

This is because load-model API was not considered yet at that moment ;-)

DanielMazurkiewicz:
Guys, isn't it predefined models and use cases for predefined models? Can't we call it like that? Would sound very straight forward and not overcompilcating these names.

And calling low level a "framework" sounds also much closer to what it is (at least to me :-) ). I wouldn't mind calling it also just a "Generic ML API use cases"

anssiko:
Thanks for your comments and suggestions.

Naming is hard, but it seems "Application Use Cases" and "Framework Use Cases" is something that resonates with most people, so I updated the PR accordingly. Further PRs are welcome to improve the prose to explain e.g. what we mean with "application" and "framework" in this context.

I believe many of the current "Application Use Cases" can be implemented with both an "operations API" as well as "load model API" . The latter provides web developers a simpler API at the cost of expressiveness. The former is more expressive, but comes with a steeper learning curve that is smoothened by an ML framework that consumes the said API. Each API design should be optimized for its consumer and key use cases.

I suggest we discuss the "load model API" use cases in a separate issue.

@huningxin as the editor, you should be able to merge this PR :-)

tomoyukilabs:
LGTM. I agree that we can discuss the "load model API" use cases in a separate issue, of course.

huningxin:
LGTM. Thanks for the PR and review! I am going to merge it and feel free to follow up "load model" use case in a separate issue.

PR #15, opened 2019-04-22
Add Intel WebNN POC API as foundation spec

### LINKED ISSUE #14: Add WebNN prototype POC Intel as foundation spec to evolve
Based on today's [call](https://www.w3.org/2019/04/11-webmachinelearning-minutes.html) the following resolution was made:

> RESOLVED: Evolve WebNN API specification using https://github.com/intel/webml-polyfill/blob/master/docs/api.md as foundation specification

@huningxin - please begin the process of moving this proposal over to spec text so that issues can be filed against it.

Thank you.

_[Edited by @anssiko: fix broken link]_

huningxin:
Fix #14

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/15.html" title="Last updated on Apr 24, 2019, 12:26 AM UTC (2f27431)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/15/5766ff9...huningxin:2f27431.html" title="Last updated on Apr 24, 2019, 12:26 AM UTC (2f27431)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -159,7 +159,282 @@ After several weeks, she has developed a tiny DNN model that can even run on
 CPU. In order to accommodate CPU execution, she modifies the application
 so that the application loads the tiny model in the case of CPU-only devices.
 
+API {#api}
+=====================
+
+## Navigator ## {#api-navigator}
+<script type=idl>
+partial interface Navigator {
+  readonly attribute ML ml;
+};
+</script>
+
+## ML ## {#api-ml}
+<script type=idl>
+interface ML {
+  NeuralNetworkContext getNeuralNetworkContext();
+};
+</script>
+
+## NeuralNetworkContext ## {#api-neuralnetworkcontext}
+<script type=idl>
+interface NeuralNetworkContext {
+  // Operand types.
+  const long FLOAT32 = 0;
+  const long INT32 = 1;
+  const long UINT32 = 2;
+  const long TENSOR_FLOAT32 = 3;
+  const long TENSOR_INT32 = 4;
+  const long TENSOR_QUANT8_ASYMM = 5;
+
+  // Operation types.
+  const long ADD = 0;
+  const long AVERAGE_POOL_2D = 1;
+  const long CONCATENATION = 2;
+  const long CONV_2D = 3;
+  const long DEPTHWISE_CONV_2D = 4;
+  const long DEPTH_TO_SPACE = 5;
+  const long DEQUANTIZE = 6;
+  const long EMBEDDING_LOOKUP = 7;
+  const long FLOOR = 8;
+  const long FULLY_CONNECTED = 9;
+  const long HASHTABLE_LOOKUP = 10;
+  const long L2_NORMALIZATION = 11;
+  const long L2_POOL_2D = 12;
+  const long LOCAL_RESPONSE_NORMALIZATION = 13;
+  const long LOGISTIC = 14;
+  const long LSH_PROJECTION = 15;
+  const long LSTM = 16;
+  const long MAX_POOL_2D = 17;
+  const long MUL = 18;
+  const long RELU = 19;
+  const long RELU1 = 20;
+  const long RELU6 = 21;
+  const long RESHAPE = 22;
+  const long RESIZE_BILINEAR = 23;
+  const long RNN = 24;
+  const long SOFTMAX = 25;
+  const long SPACE_TO_DEPTH = 26;
+  const long SVDF = 27;
+  const long TANH = 28;
+  const long BATCH_TO_SPACE_ND = 29;
+  const long TRANSPOSE = 37;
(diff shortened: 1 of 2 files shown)


huningxin:
@gregwhitworth, @nsthorat and @anssiko , please take a look. Thanks!

anssiko:
Thanks @huningxin.

I’ll propose we’ll discuss and review this PR on our next call that’s this Thursday. I’ll send meeting logistics tomorrow.

anssiko:
See also: [HTML preview](https://cdn.staticaly.com/gh/webmachinelearning/webnn/0e5adb1a66d7ea144607ccbdd1806210772cbdc8/index.html) of the spec with this PR integrated.

(I've installed [pr-preview](https://github.com/tobie/pr-preview) webhook to autogenerate spec previews and visual diffs for PRs for convenience of PR review. It should work for branches pushed to this repo.)

huningxin:
Thanks @anssiko . The pr-preview is very helpful. Through the HTML view, I found two typos and fixed them by pushing a new commit.

gramalingam:
I think it is useful to make the (graph-building) syntax simpler for the user. In fact, I think we can make it look almost exactly like an API for executing the ops directly (and the &quot;eager&quot; API), at least for the non-control-flow ops, as explained below.

(1) In the proposal, operands are represented by integers, implicitly associated with the order in which &quot;addOperand&quot; is called. It would be cleaner to have &quot;addOperand&quot; return an object/value that represents the operand. So, this should allow us to replace

    let tensor1 = operandIndex++;
    model.addOperand(float32TensorType);
by

    let tensor1 = model.addOperand(float32TensorType);

(2) Instead of separately creating operands to represent the outputs of an op and separately adding the op (connecting the inputs and outputs up), it is better to have addOperation create the operands representing the outputs and returning them. This would allow us to replace

    // intermediateOutput0 is the output of the first ADD operation.
    let intermediateOutput0 = operandIndex++;
    model.addOperand(float32TensorType);
    // Add the first ADD operation.
    model.addOperation(nn.ADD, [tensor0, tensor1, fusedActivationFuncNone], [intermediateOutput0]);
by

    // Add the first ADD operation.
    // intermediateOutput0 is the output of the first ADD operation.
    let intermediateOutput0 = model.addOperation(nn.ADD, [tensor0, tensor1, fusedActivationFuncNone]);
I omitted the type float32TensorType since it can be inferred from the operands. But where necessary, we can add the type as an extra parameter. This approach can still support operations that return multiple outputs.

(3) Furthermore, instead of representing operations by a constant like nn.ADD, it seems better to encapsulate them as a method ADD, allowing us to simplify

    let intermediateOutput0 = model.addOperation(nn.ADD, [tensor0, tensor1, fusedActivationFuncNone]);
to

    let intermediateOutput0 = model.ADD([tensor0, tensor1, fusedActivationFuncNone]);

(4) Similarly, for constants, instead of separately creating an operand and then setting its value, we can use a single method to create a constant that does both. This would allow us to simplify:

    // Add the operand for the NONE activation function, and set its value to FUSED_NONE.
    let fusedActivationFuncNone = operandIndex++;
    model.addOperand(scalarInt32Type);
    model.setOperandValue(fusedActivationFuncNone, newInt32Array([nn.FUSED_NONE]));
to:

    // Add the operand for the NONE activation function, and set its value to FUSED_NONE.
    let fusedActivationFuncNone = model.Constant(scalarInt32Type, newInt32Array([nn.FUSED_NONE]));
(As discussed earlier, the type can be omitted if it can be inferred.)

gregwhitworth:
Thanks @gramalingam - since that is an issue regarding the API itself, this issue and PR is scoped to moving in the Intel POC API into the spec. They have the note that it is not ready for implementation and we stipulated that we already know of a few issues we'd like to raise regarding the API itself so I think this is sufficient to land the PR based on our resolution from 2 weeks ago. To keep this issue scoped I've forked your issue to a new one for easier telecon communication and tracking comments specifically around that topic here https://github.com/webmachinelearning/webnn/issues/16

@huningxin The spec porting looks good. Expect - as @gramalingam pointed out, a few issues from us on what we feel will improve the API.

anssiko:
The ideal would be to land this PR as-is to form a baseline that is implemented experimentally and then in subsequent PRs improve the API based on discussion in separate topic-specific issues such as #16. 

On tomorrow’s call I’d like us to resolve to merge this PR and continue discussion on API improvements.

gramalingam:
Thanks @gregwhitworth … yes, my comment was not about this PR, but wasn't sure where to post the suggestions. Thanks for creating the separate thread.

huningxin:
Thanks @gregwhitworth @gramalingam @anssiko for review and comments. As the resolution in CG meeting, I merged it. So we can start evolve from the foundation.

PR #22, opened 2019-07-02
Simplify the graph building API

### LINKED ISSUE #16: graph-building syntax simpler for web developers
I think it is useful to make the (graph-building) syntax simpler for the user. In fact, I think we can make it look almost exactly like an API for executing the ops directly (and the &quot;eager&quot; API), at least for the non-control-flow ops, as explained below.

(1) In the proposal, operands are represented by integers, implicitly associated with the order in which &quot;addOperand&quot; is called. It would be cleaner to have &quot;addOperand&quot; return an object/value that represents the operand. So, this should allow us to replace

    let tensor1 = operandIndex++;
    model.addOperand(float32TensorType);
by

    let tensor1 = model.addOperand(float32TensorType);

(2) Instead of separately creating operands to represent the outputs of an op and separately adding the op (connecting the inputs and outputs up), it is better to have addOperation create the operands representing the outputs and returning them. This would allow us to replace

    // intermediateOutput0 is the output of the first ADD operation.
    let intermediateOutput0 = operandIndex++;
    model.addOperand(float32TensorType);
    // Add the first ADD operation.
    model.addOperation(nn.ADD, [tensor0, tensor1, fusedActivationFuncNone], [intermediateOutput0]);
by

    // Add the first ADD operation.
    // intermediateOutput0 is the output of the first ADD operation.
    let intermediateOutput0 = model.addOperation(nn.ADD, [tensor0, tensor1, fusedActivationFuncNone]);
I omitted the type float32TensorType since it can be inferred from the operands. But where necessary, we can add the type as an extra parameter. This approach can still support operations that return multiple outputs.

(3) Furthermore, instead of representing operations by a constant like nn.ADD, it seems better to encapsulate them as a method ADD, allowing us to simplify

    let intermediateOutput0 = model.addOperation(nn.ADD, [tensor0, tensor1, fusedActivationFuncNone]);
to

    let intermediateOutput0 = model.ADD([tensor0, tensor1, fusedActivationFuncNone]);

(4) Similarly, for constants, instead of separately creating an operand and then setting its value, we can use a single method to create a constant that does both. This would allow us to simplify:

    // Add the operand for the NONE activation function, and set its value to FUSED_NONE.
    let fusedActivationFuncNone = operandIndex++;
    model.addOperand(scalarInt32Type);
    model.setOperandValue(fusedActivationFuncNone, newInt32Array([nn.FUSED_NONE]));
to:

    // Add the operand for the NONE activation function, and set its value to FUSED_NONE.
    let fusedActivationFuncNone = model.Constant(scalarInt32Type, newInt32Array([nn.FUSED_NONE]));
(As discussed earlier, the type can be omitted if it can be inferred.)

_Originally posted by @gramalingam in https://github.com/webmachinelearning/webnn/pull/15#issuecomment-486310862_

huningxin:
fix #16

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/22.html" title="Last updated on Jul 2, 2019, 8:11 AM UTC (bdc0e8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/22/635339e...huningxin:bdc0e8f.html" title="Last updated on Jul 2, 2019, 8:11 AM UTC (bdc0e8f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -176,78 +176,24 @@ interface ML {
 };
 </script>
 
-## NeuralNetworkContext ## {#api-neuralnetworkcontext}
+## OperandDescriptor ## {#api-operanddescriptor}
 <script type=idl>
-interface NeuralNetworkContext {
-  // Operand types.
-  const long FLOAT32 = 0;
-  const long INT32 = 1;
-  const long UINT32 = 2;
-  const long TENSOR_FLOAT32 = 3;
-  const long TENSOR_INT32 = 4;
-  const long TENSOR_QUANT8_ASYMM = 5;
-
-  // Operation types.
-  const long ADD = 0;
-  const long AVERAGE_POOL_2D = 1;
-  const long CONCATENATION = 2;
-  const long CONV_2D = 3;
-  const long DEPTHWISE_CONV_2D = 4;
-  const long DEPTH_TO_SPACE = 5;
-  const long DEQUANTIZE = 6;
-  const long EMBEDDING_LOOKUP = 7;
-  const long FLOOR = 8;
-  const long FULLY_CONNECTED = 9;
-  const long HASHTABLE_LOOKUP = 10;
-  const long L2_NORMALIZATION = 11;
-  const long L2_POOL_2D = 12;
-  const long LOCAL_RESPONSE_NORMALIZATION = 13;
-  const long LOGISTIC = 14;
-  const long LSH_PROJECTION = 15;
-  const long LSTM = 16;
-  const long MAX_POOL_2D = 17;
-  const long MUL = 18;
-  const long RELU = 19;
-  const long RELU1 = 20;
-  const long RELU6 = 21;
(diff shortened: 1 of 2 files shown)


huningxin:
This PR is based https://github.com/webmachinelearning/webnn/issues/16#issuecomment-504305802 with considering of the feedback on  [WebML CG Teleconference - 27 June 2019](https://www.w3.org/2019/06/27-webmachinelearning-minutes.html#x01).

@gramalingam @dsmilkov @RafaelCintron and others who have interests, please help review. Thanks.

huningxin:
Based on the resolution of [22 Aug CG meeting](https://www.w3.org/2019/08/22-webmachinelearning-minutes.html#x03)

> RESOLVED: Merge PR #22

I am going to merge it. Thanks.

PR #30, opened 2019-10-15
Add target hardware section to explainer

anssiko:
At [F2F](https://www.w3.org/2019/09/20-webmachinelearning-minutes.html#x02) we agreed it'd be good to start evolve the [explainer](https://github.com/webmachinelearning/webnn/blob/master/explainer.md) document in parallel with the [formal spec](https://webmachinelearning.github.io/webnn/).

Here's my first stab at the hardware support section of that document, PTAL @huningxin @nsthorat and others.

We should consider the explainer document a place where we write down our collective thoughts without needing to worry too much if those will change. By this definition, the explainer is expected evolve and change significantly and everyone is encouraged to contribute. It is a markdown file, so low barrier to entry :-)

To contract with the spec, I'd propose we put things in the formal spec that we think have reached adequate stability, since changing the formal spec is more costly (in terms of editors' time, also in terms of implementations that implement the spec in the future, let alone web developers that write to the API direct).


huningxin:
Looks good. Thanks @anssiko .

Issue #26, opened 2019-09-07
Float16 type support


BenjaminPoulain:
The current spec defines the tensor types "float32", "int32" and "8 bits quantized".

We should consider "float16" for a few reasons:

- Certain hardware are significantly slower handling float32 compared to float16. This includes but is not limited to GPUs.
- Storing the intermediate values in float16 can be beneficial to reduce memory footprint and memory bandwidth. This is true even if all the operations are done on float32 in the FPUs.
- Supporting constant tensors in float16 would allow clients to half the size of the model's parameters. This can reduce models size in memory and reduce network bandwidth when loading the page.

Implementations without native float16 support should be allowed to use float32 internally. This would help maximizing performance on float32 hardware while keeping things simple for the authors.

huningxin:
Thanks for proposing this @BenjaminPoulain . Adding "float16" support makes sense to me.

anssiko:
Resolution from https://www.w3.org/2020/01/09-webmachinelearning-minutes.html#x04
>Add "float16" to OperandType enum in WebNN API and define a way for the API to respond when float16 is not natively supported

@huningxin feel free to craft a PR for the group to review. Thanks!

wchao1115:
Agreed with @BenjaminPoulain on the stated benefits of supporting float16 in the API. Models with float16 weights and features have become increasingly popular in real-world use cases especially on the GPU e.g. Nvidia TesorCore in the Pascal or newer generation is specifically float16-based. Float16 is also a primary native data type on most DSP-based AI hardware chips.

That said, in absence of native support in the underlying hardware, internal simulation such as up-casting the data type on the fly could lead to severe performance penalty and sometimes unexpected computation errors. The API should be designed such that it could fail the operation with unsupported error to allow the client an opportunity to gracefully fallback either by switching to a different target computing device, or to a model variation that simply processes float32 weights and feature sets.

huningxin:
> @huningxin feel free to craft a PR for the group to review. Thanks!

Put together #35. @BenjaminPoulain @wchao1115 @anssiko , please take a look. Thanks.

BenjaminPoulain:
Closing. @huningxin resolved the issue in https://github.com/webmachinelearning/webnn/pull/35.

Issue #19, opened 2019-05-09
Define set of exceptions for WebNN API


jvenkatn:
WebNN API would need exception handlers which report appropriate error status for an API (other than the default JavaScript runtime exceptions).
These exceptions would be used by developers to catch specific states of error in WebNN API.

anssiko:
This issue has been addressed. The specification currently throws one of the following in applicable places:

"NotSupportedError" DOMException
"SecurityError" DOMException
"OperationError" DOMException
"DataError" DOMException

The WG believes this provides an appropriate level of detail without disclosing too much information to become a privacy issue.

Thanks for filing this issue!

PR #35, opened 2020-01-10
Add float16 and tensor-float16 into OperandType

huningxin:
***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/35.html" title="Last updated on Jan 24, 2020, 2:11 PM UTC (6d1deee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/35/63e8ae8...huningxin:6d1deee.html" title="Last updated on Jan 24, 2020, 2:11 PM UTC (6d1deee)">Diff</a>


huningxin:
Address the resolution from https://www.w3.org/2020/01/09-webmachinelearning-minutes.html#x04
>Add "float16" to OperandType enum in WebNN API and define a way for the API to respond when float16 is not natively supported

https://github.com/webmachinelearning/webnn/issues/26#issuecomment-572630792

BenjaminPoulain:
@huningxin Thank you for taking care of the float16 addition.

> Sorry it isn't clear to me at which point the API can fail with appropriate error code if float16 is not natively supported by the executor. I presume the failure would come from the execution.startCompute(...) call but not sure how it'll get back to the client, and how is it supposed to be handled by it.

Given the behavior decided for float16, I agree the error handling should be made explicit in the spec.

Presumably, Compilation.createExecution() sets up device placement and can figure if any operation cannot be executed on hardware. What do you think @wchao1115 ?

anssiko:
For API design considerations, see also the W3C TAG's design guideline for promise-returning API and errors: https://www.w3.org/2001/tag/doc/promises-guide#errors

wchao1115:
> @huningxin Thank you for taking care of the float16 addition.
> 
> > Sorry it isn't clear to me at which point the API can fail with appropriate error code if float16 is not natively supported by the executor. I presume the failure would come from the execution.startCompute(...) call but not sure how it'll get back to the client, and how is it supposed to be handled by it.
> 
> Given the behavior decided for float16, I agree the error handling should be made explicit in the spec.
> 
> Presumably, Compilation.createExecution() sets up device placement and can figure if any operation cannot be executed on hardware. What do you think @wchao1115 ?

I guess createExecution() could work depending on how Compilation.finish() works. Is device selection supposed to happen at finish() or createExecution()? Compilation.finish() as defined now returns no error code. Assuming device selection happens at createExecution() then why do we need finish()? If the API fails to create an Execution instance due to a device error, should the caller attempt to create a new Execution from a new Compilation instance (since the failing compilation is "finished") or reuse the existing Compilation with a difference device preference policy to create a new Execution instance that works? I think the way Compilation.finish() is defined (as an async call no less) creates a set of complex interaction with the Execution instance in practice.

BenjaminPoulain:
Uh, I forgot Compilation.finish() existed :)
I am not sure why there is a need for both Compilation.finish() can Compilation.createExecution(). I agree, it is not clear where it is best to report the error with the APIs.

Then, what about dropping Compilation.finish() (compiling implicitly on createExecution()) and reporting failure in createExecution()'s catch() result?

wchao1115:
> Uh, I forgot Compilation.finish() existed :)
> I am not sure why there is a need for both Compilation.finish() can Compilation.createExecution(). I agree, it is not clear where it is best to report the error with the APIs.
> Then, what about dropping Compilation.finish() (compiling implicitly on createExecution()) and reporting failure in createExecution()'s catch() result?

I think that would work yes. I'm generally averse to stateful API especially one that exposes a terminal state, since it isn't clear what would be considered a legal calling pattern once the terminal state has been reached. Compilation.finish() exposes a terminal state, and if it's truly terminal, any subsequent calls to Compilation.setPreference(...) must also be able to return a failure code (because it's illegal), but again having what appears to be a simple setter return a failure code makes the API that much harder to use.

On a separate topic, I'm also wondering about the CompilationPreference enum on whether it is the preference to the compilation process or the execution process. If the model is said to be compiled at createExecution() where a device is selected for that compilation, then is it the compilation of the model or the subsequent execution of the compiled model that is to be expected "fast answer"?

Further, assuming a PC is equipped with both a dedicated VPU and an Intel integrated GPU, both of which require lower power level than a discrete GPU, which device e.g. the VPU or integrated GPU will be selected for "low power" option? How do we make sure that this enum options will not become increasingly hard to deciphered into the future as the landscape for AI capable hardwares evolves?

huningxin:
Thanks @wchao1115 @BenjaminPoulain @anssiko !

I agree that the float16 type unsupported error should be spec'ed. However, as you know, in current spec, the whole errors/exceptions are to be defined (#19). So my initial thought of this PR is to only fix the missing float16 data type and leave its error handling to a dedicated issue that can be fixed in the following PR. 

WDYT?

(The discussions about Compilation/Execution interfaces are so valuable. I will share my two cents in following comments. And I would also suggest we can create specific issues for each topic.)

huningxin:
@wchao1115 @BenjaminPoulain , my comments are based on [POC implementation](https://github.com/otcshare/chromium-src) of current spec. Feel free to open new issues for discussion.

> Is device selection supposed to happen at finish() or createExecution()?

The device selection happens in Compilation.finish(). Once device selected, the device code compilation/generation is also in Compilation.finish().

> Compilation.finish() as defined now returns no error code.

Compilation.finish returns a promise. So if there are any errors during compilation, e.g. unsupported data type / operation in the model, app is supposed to get the error from promise.catch(onFailure). 

> Compilation.finish() exposes a terminal state, and if it's truly terminal, any subsequent calls to Compilation.setPreference(...) must also be able to return a failure code (because it's illegal)

Yes. This is current behavior of Compilation.

> is it the compilation of the model or the subsequent execution of the compiled model that is to be expected "fast answer"?

The compiled model is expected "fast answer" as the selected device is supposed to have low latency. The Execution represents an inference request that is created from the compiled model (Compilation). Multiple execution requests can be created from one compiled model.

> which device e.g. the VPU or integrated GPU will be selected for "low power" option?

The current compilation preference is just a hint set by app. The browser decides the device(s) that meet "low power" preference. This is not ideal for VPU/iGPU case. I think we need a solution to enumerate devices and allow explicit device selection. Would you like to open a new issue?

wchao1115:
Thanks for the explanation. Can a client calls finish() again with a new preference after the first call succeeds? If yes, will the new compilation state reflect in the existing Execution instance, or are they expected to call createExecution() to get a new execution state every time after a succeeded finish() call? What happens if a createExecution() call is made before a succeeded finish() call is made? I think some samples on what calling patterns are expected from the API could be helpful here.

huningxin:
> Can a client calls finish() again with a new preference after the first call succeeds?

No. That would be illegal and get an error.

> or are they expected to call createExecution() to get a new execution state every time after a succeeded finish() call?

Yes. For each inference request, it is expected to call Compilation.createExecution() to get a new execution instance on a succeeded finished compilation object.

> What happens if a createExecution() call is made before a succeeded finish() call is made?

It would get an error.

> I think some samples on what calling patterns are expected from the API could be helpful here.

There are basic examples in current spec: https://webmachinelearning.github.io/webnn/#examples. Example 3 and 4 are about model compilation and execution. If you think anything is missing, feel free to file an issue.

wchao1115:
Thanks for your answer @huningxin. My 2c is that Compilation.finish() is redundant and provides no obvious value beyond what Compilation.createExecution already provides. I can create a separate issue to track this.

huningxin:
> So my initial thought of this PR is to only fix the missing float16 data type and leave its error handling to a dedicated issue that can be fixed in the following PR.

Open https://github.com/webmachinelearning/webnn/issues/36 

@BenjaminPoulain @wchao1115 , are you fine with that we keep this PR only for adding float16 and follow up the unsupported float16 error (with other unsupported operand types) in #36 ?

huningxin:
> My 2c is that Compilation.finish() is redundant and provides no obvious value beyond what Compilation.createExecution already provides. I can create a separate issue to track this.

Creating a separate issue for Compilation/Execution design sounds great to me. Thanks in advance!

RafaelCintron:
I agree with @wchao1115's feedback about eliminating `finish` as a method and confusion around the compilation preference. 

I presume that calling `Compilation.setPreference` only affects executions created from that point forward?  If so, I think we should eliminate `setPreference` in favor of passing the compilation preference as a parameter to `CreateExecution`.  This way, it is more clear that the compilation option only affects new executions and doesn't invalidate existing ones.  

I am OK with merging this PR as is as dealing with compilation/execution design as a separate issue.

anssiko:
I resolved the merge conflict. It was just due to build tool generated metadata clash.

@wchao1115 or @huningxin feel free to merge.

huningxin:
Thanks @anssiko for resolving conflict and thanks @BenjaminPoulain @wchao1115 @RafaelCintron for review.

BenjaminPoulain:
I am a bit late but: Thanks @huningxin for this patch :)

PR #40, opened 2020-02-03
Fixed a critical typo. The convolution forward pass always uses cross…

wchao1115:
…-correlation in DML.


huningxin:
Looks good. Thanks for the fix.

Issue #28, opened 2019-09-16
[op compatibility] conv2d


nsthorat:
This issue will track op compatibility resolution for conv2d.

**Signature:**
conv2d(x, filter, padding, strides, dilations)

**Arguments:**
**_x:_** 4-dim tensor with logical shape [N, H, W, in_channels] 
_**filter:**_ 4-dim tensor with logical shape [H, W, in_channels, out_channels]
**_padding:_** array of ints. Padding for the beginning and ending along each spatial axis.
_**strides:**_ array of ints that have length 1, 2 or 4. The stride of the sliding window for each dimension of input.
_**dilations:**_ array of ints of length 2: [dil_height, dil_width] The dilation factor for each spatial dimension of the input.

**Docstring:**
Computes a 2-D convolution given 4-D input and filter tensors.

_Notes:_
- Tensor layout should be a single logical type to leave transpose graph optimizations up to the browser (this is much more portable because the model author does not have to think about the hardware of the device).
- Tensor layout should be NHWC because HTMLMediaElements are already of this format and we want to remove the burden on the user to remove transposing.

_To be discussed:_
- Compatibility with underlying APIs / hardware

huningxin:
> conv2d(x, filter, padding, strides)

Is the `dilations` parameter missed in the signature?

nsthorat:
Fixed, thanks :)

huningxin:
> * Compatibility with underlying APIs / hardware

The following table is based on [WebNN POC native mapping](https://docs.google.com/spreadsheets/d/1nthZOwgIKsj34EB-SymEwoNTPsxo4X8Pxavm-JaBwME/edit?usp=sharing). It captures the mapping to NNAPI, MPS and DirectML that can be used as the base for native APIs compatibility study.

|  WebNN | NNAPI | MPS | DML |
|------|---------|------|------------|
|  **conv2d** | [ANEURALNETWORKS_CONV_2D](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0a34a73b5eaf458b67db5eda71557d1d01)  | [MPSCNNConvolution](https://developer.apple.com/documentation/metalperformanceshaders/mpscnnconvolution) | [DML_OPERATOR_CONVOLUTION](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_convolution_operator_desc) |
[...]

PR #42, opened 2020-02-06
Add Paddle-Lite column into conv2d table

huningxin:


huningxin:
Address comment https://github.com/webmachinelearning/webnn/issues/28#issuecomment-582709990. @smilejames @wchao1115 , please take a look.

huningxin:
Thanks for review @smilejames @anssiko 

> We'd need to convert to HTML for that.

@anssiko , do you have any experiences to share?

anssiko:
Let's stick with markdown format for now. HTML table comes with its own issues. If someone has better ideas, let us know.

PR #45, opened 2020-02-14
Add NCHW support of NNAPI into conv2d table

huningxin:
Address comment in https://github.com/webmachinelearning/webnn/pull/43#issuecomment-586073308

> According to NNAPI ANEURALNETWORKS_CONV_2D doc, it supports "NHWC" or "NCHW" data layout, with the default data layout NHWC. I will make another PR to update the table.


huningxin:
@wchao1115 thanks for review!

Issue #27, opened 2019-09-16
[op compatibility] matMul


nsthorat:
This issue will track op compatibility resolution for matMul.

**Signature:**
matmul(a, b)

**Arguments:**
**_a:_** n-dim tensor
_**b:**_ n-dim tensor

**Docstring:**
If both a and b are 2-D they are multiplied like conventional matrices. If one of a or b are 1-D this is treated as a matrix times vector dot product.

If either argument is N dimensional, N>2, it is treated as a stack of matrices (rank-3) with dimensions corresponding to the inner two indices. The matrix multiplication will be broadcasted accordingly.

Example:
If a has shape [2, 3, 4, 5] and b has shape [5, 4], the resulting tensor will have a shape of [2, 3, 4, 4] as a is treated as a size 2 * 3 = 6 stack of [4, 5] matrices. These get broadcast multiplied over the [5, 4] matrix creating 6 [4, 4] matrices. They keep the original shape of a's outer dimensions, resulting in a shape of [2, 3, 4, 4].

_Notes:_
- Does not support fused bias (this will be taken care of by graph optimizer)
- Does not support transpose arguments (this is taken care of by graph optimizer)

_To be discussed:_
- Compatibility with underlying APIs / hardware

BenjaminPoulain:
> If either argument is N dimensional, N>2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.

Can you please refine this into a precise definition?

BenjaminPoulain:
CoreML & MPS expose the concept of accumulator precision. That is useful to strike a balance between precision and performance.

By default, the accumulator precision could be the same as the type (e.g. a float16 accumulator for float16 matmul). A high precision accumulator makes a best effort to increase precision (e.g. float32 accumulator for float16 inputs).

What do you think about having a an optional 3rd argument to provide the precision?

    matmul(a, b, options)

e.g.:

    matmul(a, b, {'precision': 'high'})

nsthorat:
> > If either argument is N dimensional, N>2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.
> 
> Can you please refine this into a precise definition?
Updated the description a bit but this is the definition from numpy -- I added an example that hopefully makes this clearer.

> CoreML & MPS expose the concept of accumulator precision. That is useful to strike a balance between precision and performance.
> 
[...]

PR #48, opened 2020-03-05
Fix up DirectML definition for matmul.

wchao1115:
A question I have is: what is the defined behavior of webnn matmul when N > 2? The description in the NNAPI row suggests that it'll slice off the ND matrices to 2D along axis 0. This is not a numpy matmul behavior, which is to treat N=3 as a stack of 2D matrics. Which definition is honored here? Or do we want to limit it just to 2D for now?


huningxin:
Change looks good.

> what is the defined behavior of webnn matmul when N > 2?

According to the proposal https://github.com/webmachinelearning/webnn/issues/27#issue-494290232, webnn matmul should behave like numpy matmul.

> The description in the NNAPI row suggests that it'll slice off the ND matrices to 2D along axis 0.

The NNAPI [ANEURALNETWORKS_FULLY_CONNECTED](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0aaada7a3dbaf4676aba560c933ff610c5) only supports 2-D weights. According to the investigation of tflite-converter's handling of [tf.matmul](https://www.tensorflow.org/api_docs/python/tf/linalg/matmul), when N > 3, it first reshapes to 3-D tensor, then slices the 3-D tensor to multiple 2-D tensors along the 0 axis (left most).

E.g. when matmul input tensor is [2, 2, 2, 2], it first reshapes to [4, 2, 2], then slices to 4 x [1, 2, 2] tensors as fully connected input.

Does this behavior look right?

Since this is only my investigation, I put an open in the table. Please correct me if I am wrong.

anssiko:
Thanks @wchao1115. I'll merge this PR now so everyone has the latest correct compatibility data ahead the call.

PR #49, opened 2020-03-11
Add matmul

huningxin:
This PR is based on matmul [signature proposal](https://github.com/webmachinelearning/webnn/issues/27#issue-494290232) and [compatibility table](https://github.com/webmachinelearning/webnn/blob/master/op_compatibility/matmul.md).

 @wchao1115 @nsthorat @BenjaminPoulain and @anssiko, please take a look. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/49.html" title="Last updated on Mar 25, 2020, 2:25 AM UTC (2fd2baf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/49/30d4117...huningxin:2fd2baf.html" title="Last updated on Mar 25, 2020, 2:25 AM UTC (2fd2baf)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -309,14 +309,20 @@ partial interface NeuralNetworkContext {
 
     Computes the matrix product of two input tensors. It behaves as following:
         - If both *a* and *b* are 2-D, they are multiplied like conventional
-            matrices.
+            matrices and produce a 2-D tensor as the output.
         - If either *a* or *b* is N-D, N > 2, it is treated as a stack of
-            matrices (rank-3) with dimensions corresponding to the last two
-            indices. The matrix multiplication will be broadcasted accordingly.
+            matrices with dimensions corresponding to the last two indices. The
+            matrix multiplication will be broadcasted accordingly by following
+            [[!numpy-broadcasting-rule]]. The output is a N-D tensor whose rank
+            is the maximum rank of the input tensors. For each dimension, except
+            the last two, of the output tensor, its size is the maximum size
+            along that dimension of the input tensors.
         - If *a* is 1-D, it is converted to a 2-D tensor by prepending a 1 to
             its dimensions.
         - If *b* is 1-D, it is converted to a 2-D tensor by by appending a 1 to
             its dimensions.
+        - If both *a* and *b* are 1-D, the operation is a vector dot-product,
+            which produces a scalar output.
 </div>
 
 ### mul ### {#api-neuralnetworkcontext-mul}
@@ -463,6 +469,12 @@ Thanks to Tomoyuki Shimizu, Ningxin Hu, and Zhiqiang Yu for the use cases.
 
 <pre class="biblio">
 {
+  "numpy-broadcasting-rule": {
+    "href": "https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules",
+    "title": "General Broadcasting Rules of NumPy",
+    "authors": ["The SciPy community"],
+    "date": "July 2019"
+  },
   "SSD": {
     "href": "https://arxiv.org/abs/1512.02325",
     "title": "SSD: Single Shot MultiBox Detector",
--- a/index.bs
+++ b/index.bs
@@ -292,6 +292,33 @@ partial interface NeuralNetworkContext {
     Computes a 2-D convolution given input and filter 4-D tensors.
 </div>
 
+### matmul ### {#api-neuralnetworkcontext-matmul}
+<script type=idl>
+partial interface NeuralNetworkContext {
+  Operand matmul(Operand a, Operand b);
+};
+</script>
+
+<div algorithm=matmul>
+    **Arguments:**
+        - *a*: an {{Operand}}. The first input N-D tensor.
+        - *b*: an {{Operand}}. The second input N-D tensor.
+
+    **Returns:** an {{Operand}}. The output N-D tensor that contains the matrix
+    product of two input tensors.
+
+    Computes the matrix product of two input tensors. It behaves as following:
+        - If both *a* and *b* are 2-D, they are multiplied like conventional
+            matrices.
+        - If either *a* or *b* is N-D, N > 2, it is treated as a stack of
+            matrices (rank-3) with dimensions corresponding to the last two
+            indices. The matrix multiplication will be broadcasted accordingly.
+        - If *a* is 1-D, it is converted to a 2-D tensor by prepending a 1 to
+            its dimensions.
+        - If *b* is 1-D, it is converted to a 2-D tensor by by appending a 1 to
+            its dimensions.
+</div>
+
 ### mul ### {#api-neuralnetworkcontext-mul}
 <script type=idl>
 partial interface NeuralNetworkContext {
--- a/index.html
+++ b/index.html
@@ -1221,7 +1221,7 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version 5200e023b5b76bb01f8f0ff295f799cb9f4a94f0" name="generator">
+  <meta content="Bikeshed version 576ef4f5126e3fb83bf07d9bbe1a64f99aa4944e" name="generator">
   <link href="https://webmachinelearning.github.io/webnn/" rel="canonical">
 <style>/* style-md-lists */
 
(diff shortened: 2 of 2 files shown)


wchao1115:
@huningxin numpy behavior is that if only one of the two operands is 1D, the promotion of the 1D tensor to 2D takes place. But if both are 1D, the operation is a dot product, returning a scalar tensor. 
1. It might be more clear to add the case when both operands are 1D. 
2. As for the scalar output, can Operand be a scalar value? What's the convention?

huningxin:
@wchao1115 , thanks for the review.

> 1. It might be more clear to add the case when both operands are 1D.

Good idea. I'll add this case.



> 2. As for the scalar output, can Operand be a scalar value? What's the convention?

Operand can be scalar value. In that case, its corresponding OperandDescriptor has scalar type, e.g. `float32`.

For example, to create a scalar float32 input:
```
const tensor = nn.input({{type: 'float32'});
```

gramalingam:
Note that numpy does not allow inputs to be scalar. https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.matmul.html

huningxin:
Thanks @gramalingam for comment.

> Note that numpy does not allow inputs to be scalar. https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.matmul.html

Yes. We follow that behavior. Do you think we need to mention it specifically?

gramalingam:
@huningxin , that's good. I was just confused by the previous two comments that "Operand can be a scalar" meant the input could be a scalar. It may be worth being explicit in the documentation, thanks.

wchao1115:
@huningxin How about this wording? What do you think?

    **Returns:** an {{Operand}}. The output N-D tensor that contains the matrix
    product of two input tensors.

    Computes the matrix product of two input tensors. It behaves as following:
        - If both *a* and *b* are 2-D, they are multiplied like conventional
            matrices and produce a 2-D tensor as the output.
        - If either *a* or *b* is N-D, N > 2, it is treated as a stack of matrices
            (rank-3) with dimensions corresponding to the last two indices. The
            other input broadcasts its dimensions accordingly to allow for matrix 
            multiplications of the last two dimensions of both inputs. The output
            is an N-D tensor of the same shape as the N-D input.
        - If only *a* is 1-D, it is converted to a 2-D tensor by prepending a 1 to
            its dimensions.
        - If only *b* is 1-D, it is converted to a 2-D tensor by by appending a 1 to
            its dimensions.
        - If both *a* and *b* are 1-D, the operation is a vector dot-product, which
            produces a scalar output.

index.bs
```diff
@@ -292,6 +292,33 @@ partial interface NeuralNetworkContext {
...
+    Computes the matrix product of two input tensors. It behaves as following:
+        - If both *a* and *b* are 2-D, they are multiplied like conventional
+            matrices.
+        - If either *a* or *b* is N-D, N > 2, it is treated as a stack of
+            matrices (rank-3) with dimensions corresponding to the last two
+            indices. The matrix multiplication will be broadcasted accordingly.
+        - If *a* is 1-D, it is converted to a 2-D tensor by prepending a 1 to
+            its dimensions.
+        - If *b* is 1-D, it is converted to a 2-D tensor by by appending a 1 to
+            its dimensions.
```
wchao1115: Please consider this wording instead. Hopefully its more spelled out how the API would behave.

    **Returns:** an {{Operand}}. The output N-D tensor that contains the matrix
    product of two input tensors.

    Computes the matrix product of two input tensors. It behaves as following:
        - If both *a* and *b* are 2-D, they are multiplied like conventional
            matrices and produce a 2-D tensor as the output.
        - If either *a* or *b* is N-D, N > 2, it is treated as a stack of matrices
            (rank-3) with dimensions corresponding to the last two indices. The
            other input broadcasts its dimensions accordingly to allow for matrix 
            multiplications of the last two dimensions of both inputs. The output
            is an N-D tensor of the same shape as the N-D input.
        - If only *a* is 1-D, it is converted to a 2-D tensor by prepending a 1 to
            its dimensions.
        - If only *b* is 1-D, it is converted to a 2-D tensor by by appending a 1 to
            its dimensions.
        - If both *a* and *b* are 1-D, the operation is a vector dot-product, which
            produces a scalar output.

gramalingam: What does "rank-3" mean above? Shouldn't it be (rank-2)? How about saying
```
... it is treated as a stack of matrices of rank (N-2) with matrix dimensions corresponding to the last two dimensions
```

wchao1115: > What does "rank-3" mean above? Shouldn't it be (rank-2)? How about saying
> 
> ```
> ... it is treated as a stack of matrices of rank (N-2) with matrix dimensions
> corresponding to the last two dimensions
> ```

That was the original wording from @huningxin. I believe it refers to the stacking part of the sentence and not the matrix part of the sentence i.e. the stacking is considered a third rank. But I agree it's confusing. I like your suggestion but also think the sentence is already clear without it. How about we just remove the '(rank-3)' part?

gramalingam: Since this is the first use of broadcasting, we should also include some general description of broadcasting. The description above is a little imprecise if the broadcasting is bidirectional ... I assume it is bidirectional?

gramalingam: How about replacing the line
```
The output is an N-D tensor of the same shape as the N-D input.
```
by
```
If the two tensors have dimensions [a1, ..., ap, M, N] and [b1, ..., bq, N, K], then the output is a tensor of dimensions [c1, ..., cq, M, K] where [c1, ..., cq] is the result of broadcasting [a1, ... ap] and [b1, ..., bq].
```

gramalingam: Removing the (rank-3) part is also fine.

wchao1115: Since we decided to follow the [numpy broadcasting rule](https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules), I think it's a good idea to add a link referencing the numpy rule here in the API spec. ONNX broadcasting rule is compatible with the numpy rule but further classifies the type of broadcasting to two categories -- either uni or multi-directional. In this case since it only concerns 2 operands and that either one of them can be N-D, this distinction, I believe, is not relevant here i.e. the broadcasting behavior of the N-D operand simply follows the numpy rule as previously stated.

huningxin: @wchao1115 @gramalingam thanks for your comments.

> What does "rank-3" mean above?

 '(rank-3)' refers to the stack. I agree it is confusing. I update the PR to remove it.

> Since we decided to follow the [numpy broadcasting rule](https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules), I think it's a good idea to add a link referencing the numpy rule here in the API spec.

It sounds good to me. I will add that link as a normative reference into the PR. How about saying
```
The matrix multiplication will be broadcasted by using [numpy-broadcasting-rule]. 
```

> The output is an N-D tensor of the same shape as the N-D input.

This may not be true. For example, if a has shape [2, 3, 4, 5] and b has shape [5, 4], the resulting tensor will have a shape of [2, 3, 4, 4].

As we refer to numpy broadcasting rule, how about we say

```
The output is a N-D tensor whose rank is same as the maximum rank of input tensors.
```

gramalingam: Regarding the output dimension: yes, we need to change "The output is an N-D tensor of the same shape as the N-D input." @huningxin 's proposal looks good, except that "the maximum size along that dimension of the input tensors" does not apply to the last two dimension, which are subject to the matrix-multiplication rules. (Which is why I suggested a slightly different wording above; but we can adapt Ningxin's words also accordingly.)

huningxin: @gramalingam, your suggestion sounds good. I'll incorporate that. Thanks.

huningxin:
@wchao1115 @gramalingam , I pushed a new commit https://github.com/webmachinelearning/webnn/pull/49/commits/2fd2bafefb41fe0180d3a03216d0e1f1067a7abe to address your comments. Please take another look. Thanks!

huningxin:
@wchao1115 @gramalingam , thanks for the review and approval.

anssiko:
This PR has received adequate review and all the feedback has been addressed.

@huningxin @wchao1115 feel free to merge this PR.

Issue #36, opened 2020-01-15
Handling unsupported OperandType


huningxin:
Current spec defines [`OperandType` enum](https://webmachinelearning.github.io/webnn/#enumdef-operandtype) 
```webidl
enum OperandType {
  "float32",
  "int32",
  "uint32",
  "tensor-float32",
  "tensor-int32",
  "tensor-quant8-asymm"
};
```
The `float16` and `tensor-float16` are being added #35 . 

However, as mentioned by @wchao1115 in https://github.com/webmachinelearning/webnn/issues/26#issuecomment-572863966, there are situations that the selected device doesn't have native support of an `OperandType`. For example, some CPUs may not support `tensor-float16`, some GPUs may not support `tensor-quant8-asymm` and some AI accelerators may not support `tensor-float32`. To allow the app gracefully handle these situations, e.g. select a different device, or use different model with supported operand type, the API should report the unsupported `OperandType` error.

Open this issue to explore the definition of unsupported OperandType error and the API behavior to return that error.

Thoughts?

wchao1115:
@huningxin  Should "tensor-quant8-asymc" simply be "tensor-int8"? Does the current naming imply that the content in the tensor should be coupled with quantization data e.g. zero-point? It might be cleaner to just define the data type here independent of the operation (e.g. quantization) that have originated it.

huningxin:
> Does the current naming imply that the content in the tensor should be coupled with quantization data e.g. zero-point?

Yes. It is supposed to be used together with `scale` and `zeroPoint` of [`OpearndDescriptor`](https://webmachinelearning.github.io/webnn/#dictdef-operanddescriptor).

> It might be cleaner to just define the data type here independent of the operation (e.g. quantization) that have originated it.

I notice there are different quantization support ways in native API, e.g. ONNX defines quantization operations, such as [QLinearConv](https://github.com/onnx/onnx/blob/master/docs/Operators.md#QLinearConv), but NNAPI defines [quantized tensors](https://developer.android.com/ndk/guides/neuralnetworks#quantized_tensors). 

We may need to create a new issue for compatibility investigation among them. WDYT?

wchao1115:
[...]

PR #50, opened 2020-03-19
How to handle a failing compilation

wchao1115:
Update CompilationPreference to follow WebGL convention for context creation. Update the code sample to show how a certain compilation could fail such as when the model uses data type not supported by the hardware, and how to recover from the failure. #36

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/50.html" title="Last updated on Apr 4, 2020, 7:33 AM UTC (af16a69)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/50/b341ec2...af16a69.html" title="Last updated on Apr 4, 2020, 7:33 AM UTC (af16a69)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -337,6 +337,8 @@ partial interface NeuralNetworkContext {
 ## Model ## {#api-model}
 <script type=idl>
 enum PowerPreference {
+  // Let the user agent decide the most suitable behavior. This is the default value.
+  "default",
   // Prioritizes execution speed over other considerations e.g. power consumption
   "high-performance",
   // Prioritizes power consumption over other consideraions e.g. execution speed
@@ -345,7 +347,7 @@ enum PowerPreference {
 
 dictionary CompilationOptions {
   // Compilation preference as related to power consumption level
-  PowerPreference powerPreference;
+  PowerPreference powerPreference = "default";
 };
 
 interface Model {
--- a/index.bs
+++ b/index.bs
@@ -336,20 +336,15 @@ partial interface NeuralNetworkContext {
 
 ## Model ## {#api-model}
 <script type=idl>
-enum CompilationPreference {
-  // No preference. Let the user agent decide the type of compilation most 
-  // suitable and broadly available. This is the default option.
-  "default",
-  // Indicate a preference for a type of compilation that prioritizes power 
-  // consumption over performance.
-  "low-power",
-  // Indicate a preference for a type of compilation that prioritizes high 
-  // performance over power consumption.
-  "high-performance"
+enum PowerPreference {
+  // Prioritizes execution speed over other considerations e.g. power consumption
+  "high-performance",
+  // Prioritizes power consumption over other consideraions e.g. execution speed
+  "low-power"
 };
 
 interface Model {
-  Promise<Compilation> createCompilation(CompilationPreference preference);
+  Promise<Compilation> createCompilation(optional PowerPreference preference);
 };
 </script>
 
@@ -432,17 +427,11 @@ const model = await nn.createModel([output]);
 </div>
 
 <div class="example">
-The following code compiles the graph according to a compilation preference of speed over power consumption.
-The compilation may be unsuccessful when the preference cannot be fulfilled on a specified model due to hardware
-or other platform limitation, in which case a retry may be attempted with the 'default' option. The 'default' 
-option provides the broadest availability of the compilation result and is expected to succeed.
+The following code compiles the model by prioritizing lower level of power consumption
+over time. This option could be particularly useful for long-running models.
 <pre highlight="js">
-// Create a Compilation object for the constructed model with 'high-performance'
-// compilation preference.
-var compilation = await model.createCompilation('high-performance');
-if (!compilation) {
-  compilation = await model.createCompilation('default');
-}
+// Create a Compilation object for the constructed model.
+const compilation = await model.createCompilation('low-power');
 </pre>
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -343,8 +343,13 @@ enum PowerPreference {
   "low-power"
 };
 
+dictionary CompilationOptions {
+  // Compilation preference as related to power consumption level
+  PowerPreference powerPreference;
+};
+
 interface Model {
-  Promise<Compilation> createCompilation(optional PowerPreference preference);
+  Promise<Compilation> createCompilation(optional CompilationOptions options = {});
 };
 </script>
 
@@ -431,7 +436,8 @@ The following code compiles the model by prioritizing lower level of power consu
 over time. This option could be particularly useful for long-running models.
 <pre highlight="js">
 // Create a Compilation object for the constructed model.
-const compilation = await model.createCompilation('low-power');
+const options = { powerPreference: 'low-power' };
+const compilation = await model.createCompilation(options);
 </pre>
 </div>
 
--- a/index.bs
+++ b/index.bs
@@ -337,9 +337,15 @@ partial interface NeuralNetworkContext {
 ## Model ## {#api-model}
 <script type=idl>
 enum CompilationPreference {
+  // No preference. Let the user agent decide the type of compilation most 
+  // suitable and broadly available. This is the default option.
+  "default",
+  // Indicate a preference for a type of compilation that prioritizes power 
+  // consumption over performance.
   "low-power",
-  "fast-answer",
-  "sustained-speed"
+  // Indicate a preference for a type of compilation that prioritizes high 
+  // performance over power consumption.
+  "high-performance"
 };
 
 interface Model {
@@ -426,11 +432,17 @@ const model = await nn.createModel([output]);
 </div>
 
 <div class="example">
-The following code compiles the graph.
+The following code compiles the graph according to a compilation preference of speed over power consumption.
+The compilation may be unsuccessful when the preference cannot be fulfilled on a specified model due to hardware
+or other platform limitation, in which case a retry may be attempted with the 'default' option. The 'default' 
+option provides the broadest availability of the compilation result and is expected to succeed.
 <pre highlight="js">
-// Create a Compilation object for the constructed model with 'fast-answer'
+// Create a Compilation object for the constructed model with 'high-performance'
 // compilation preference.
-const compilation = await model.createCompilation('fast-answer');
+var compilation = await model.createCompilation('high-performance');
+if (!compilation) {
+  compilation = await model.createCompilation('default');
+}
 </pre>
 </div>
 
--- a/index.html
+++ b/index.html
@@ -1222,7 +1222,7 @@ Possible extra rowspan handling
 		}
 	}
 </style>
-  <meta content="Bikeshed version c80ee3ab, updated Thu Mar 26 10:21:28 2020 -0700" name="generator">
+  <meta content="Bikeshed version c1dbacab, updated Mon Mar 30 18:56:12 2020 -0700" name="generator">
   <link href="https://webmachinelearning.github.io/webnn/" rel="canonical">
 <style>/* style-autolinks */
 
@@ -1472,6 +1472,7 @@ c-[il] { color: #000000 } /* Literal.Number.Integer.Long */
    <h1 class="p-name no-ref" id="title">Web Neural Network API</h1>
 <<<<<<< HEAD
 <<<<<<< HEAD
+<<<<<<< HEAD
    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Draft Community Group Report, <time class="dt-updated" datetime="2020-03-24">24 March 2020</time></span></h2>
 =======
    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Draft Community Group Report, <time class="dt-updated" datetime="2020-03-18">18 March 2020</time></span></h2>
@@ -1479,6 +1480,9 @@ c-[il] { color: #000000 } /* Literal.Number.Integer.Long */
 =======
    <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Draft Community Group Report, <time class="dt-updated" datetime="2020-04-01">1 April 2020</time></span></h2>
 >>>>>>> Update according to feedbacks and to align more closely with WebGL.
+=======
+   <h2 class="no-num no-toc no-ref heading settled" id="subtitle"><span class="content">Draft Community Group Report, <time class="dt-updated" datetime="2020-04-03">3 April 2020</time></span></h2>
+>>>>>>> Add the 'default' option for PowerPreference.
    <div data-fill-with="spec-metadata">
     <dl>
      <dt>This version:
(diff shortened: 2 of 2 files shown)


index.bs
```diff
@@ -304,9 +304,15 @@ partial interface NeuralNetworkContext {
 ## Model ## {#api-model}
 <script type=idl>
 enum CompilationPreference {
+  // No preference. Let the user agent decide the type of compilation most 
+  // suitable and broadly available. This is the default option.
+  "default",
```
huningxin: Do you think whether we can make the compilation preference as an optional argument? Like
```webidl
interface Model {
  Promise<Compilation> createCompilation(optional CompilationPreference preference);
};
```

Similar to WebGPU's [`requestAdapter`](https://gpuweb.github.io/gpuweb/#gpu-interface).

The "no preference" can be fulfilled by calling `createCompilation()`. With that, the example code would look like
```
const compilation = await model.createCompilation('fast-answer');
if (!compilation) {
  compilation = await model.createCompilation();
}
```

index.bs
```diff
@@ -304,9 +304,15 @@ partial interface NeuralNetworkContext {
...
+  // Indicate a preference for a type of compilation that prioritizes power 
+  // consumption over performance.
   "low-power",
-  "fast-answer",
-  "sustained-speed"
+  // Indicate a preference for a type of compilation that prioritizes high 
+  // performance over power consumption.
+  "high-performance"
```
huningxin: "fast-answer" prioritizes maximizing the responsiveness of a single answer. "sustained-speed" prioritizes maximizing the throughput of successive frames.

I propose to create a separate issue to discuss about substituting them with "high-performance".

index.bs
wchao1115: I like the optional arg idea. Thanks.

index.bs
wchao1115: Thanks for your clarification. I had some difficulties distinguishing them just by interpreting the words alone. If the intention is to distinguish between the user's preferences between performance and throughput, which dose make sense, my advice is to spell it out directly e.g. 'high-performance' and 'high-throughput' to avoid having interpret the more abstract words. Could that work? 

Could you also give a few sample scenarios where 'high-throughput' is preferred over 'high-performance'? Concrete examples will be very handy for the description text.

huningxin: > I had some difficulties distinguishing them just by interpreting the words alone.

Apologize for the confusing.

> If the intention is to distinguish the user's preferences between performance and throughput, which does make sense

The intention is to distinguish between latency and throughput. Would it be clearer by using "low-latency" and "high-throughput"?

wchao1115: 'low-latency' sounds fine too.

gramalingam: The compiler may need to a number of different tradeoffs. "Latency" vs. "throughput" is one tradeoff. "Power consumption" vs "performance" is a different tradeoff. What are the ones we need to support? If we want more than one, then preference should not be a single enum: it should perhaps be a list of options?

huningxin: > If we want more than one, then preference should not be a single enum: it should perhaps be a list of options?

I agree with you. We can solve it by introducing `CompilationOptions` dictionary that could contain multiple preferences. For example
```webidl
interface Model {
  Promise<Compilation> createCompilation(optional CompilationOptions options = {});
};

dictionary CompilationOptions {
    PowerPreference powerPreference;
    PerformancePreference performancePreference;
};

enum PowerPreference {
  "low-power",
  "high-performance"
};

enum PerformancePreference {
  "low-latency",
  "high-throughput"
};
```

@gramalingam, what do you think?

wchao1115: @huningxin In general, I think we should think through how a browser or the OS may implement these options. It may not be very useful to define options that cannot be achieved in reality. Take for example, I can see how the power preferences can be supported in the OS today as there is already a way to select a low-power adapter if the hardware supports multiple adapters. I can't quite yet imagine how to accomplish the distinction between speed and throughput. This is much harder in reality as it requires special treatments of graph execution process e.g. batching etc, to maximize the graph compute throughput and avoid stutters.

How about we stick with just PowerPreference for now? This will also be consistent with WebGL _WebGLPowerPreference_ options.

anssiko: I tend to agree. Quoting Joshua Bloch: When in doubt, leave it out.

If there is a fundamental theorem of API design, this is it. It applies equally to functionality, classes, methods, and parameters. Every facet of an API should be as small as possible, but no smaller. You can always add things later, but you can't take them away. Minimizing conceptual weight is more important than class- or method-count.

wchao1115: @anssiko I was just discussing the WebGL _failIfMajorPerformanceCaveat_ flag with @rafaelcintron and how it could apply to our ML audience. We think we should leave that flag out for now too for simplicity. 

Unlike in typical graphics scenario with WebGL where a GPU is almost always a clear winner over a CPU when it comes to rendering perf of heavy graphics, ML perceived performance could be model-dependent. GPU is of clear advantage with deep networks that require significant compute power while there are also many simpler regressions or classifiers that using GPU may never be significantly faster than a CPU, or sometimes even slower.

huningxin: > I can't quite yet imagine how to accomplish the distinction between speed and throughput.

It is actually about how fast the app can get the first response back. For example, when I use WinMLRunner to infer a model, when I choose run it on CPU device, the first iteration time is much faster than GPU device. However, when I run the model repeatedly, the GPU device outperforms CPU device in average performance excluding the first one.  So one  implementation may select CPU device for `fast-answer` and select GPU device for `sustained-speed`. Does that make sense?

I am pasting the WinMLRunner logs of SqueezeNet inference on my machine for your reference. According the results, the first iteration time on CPU is about 31ms, on GPU is about 60ms. However, the average performance on CPU is about 6.9ms, on GPU is 4.37ms.

CPU:
```
First Iteration Performance (load, bind, session creation, and evaluate):
  Load: 10.9661 ms
  Bind: 0.2029 ms
  Session Creation: 11.2625 ms
  Evaluate: 8.8001 ms

Average Performance excluding first iteration. Iterations 2 to 20. (Iterations greater than 1 only bind and evaluate)
  Average Bind: 0.126079 ms
  Average Evaluate: 6.83456 ms
```
GPU:
```
First Iteration Performance (load, bind, session creation, and evaluate):
  Load: 10.1944 ms
  Bind: 1.3036 ms
  Session Creation: 42.8985 ms
  Evaluate: 5.6366 ms

Average Performance excluding first iteration. Iterations 2 to 20. (Iterations greater than 1 only bind and evaluate)
  Average Bind: 0.296279 ms
  Average Evaluate: 4.0754 ms
```

wchao1115: @huningxin Thanks for the example. I think I might have misunderstood the intent behind 'fast-answer' and 'sustained-speed' yet again. In our past conversation, I was led to believe that the distinction between the two is simply latency vs. throughput. What you are now describing is latency vs. first-inference-latency, which is a rather different metrics.

GPU is known to have high initial setup cost from necessary resource creation processes e.g. device and resource creations, pipeline state and shader compilation etc. But once the pipeline state is set up, subsequent executions are extremely fast. A customary practice for GPU utilization is to avoid repeating this setup cost by keeping the pipeline state around and not tearing it down after every single inference call. Historically, the browser is known to do a pretty good job at managing this cost while leveraging GPU for both Canvas2d and WebGL context. But the first-iteration latency itself is often mitigated at the app level by simply discarding the first run.

Anyhow, I hope to illustrate a point that performance-based preference is a broad topic and often scenario-based. Within this thread we have now identified at least a handful of them and I'm sure there may be more.

wchao1115: @huningxin @RafaelCintron @gramalingam I've pushed out a new commit based on the feedbacks from this thread so far. Please take a look.

gramalingam:
I am fine with the proposal. My main concern was to avoid mixing potentially different concerns/options into one. It is fine to add more options later on as and when appropriate.

huningxin:
To allow adding more options later, I propose to change the argument into a dictionary, it also aligns to the WebGL [WebGLContextAttributes](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2) that contains `WebGLPowerPreference`.

For example:
```webidl
interface Model {
  Promise<Compilation> createCompilation(optional CompilationOptions options = {});
};

dictionary CompilationOptions {
    PowerPreference powerPreference;
};

enum PowerPreference {
  "low-power",
  "high-performance"
};
```

huningxin:
Another question is in the latest commit, I failed to find the sample code that handles a failing compilation. Are there any reasons to drop it?

index.bs
```diff
@@ -303,14 +303,15 @@ partial interface NeuralNetworkContext {
 
 ## Model ## {#api-model}
 <script type=idl>
-enum CompilationPreference {
-  "low-power",
-  "fast-answer",
```
huningxin, line 308: @wchao1115 , as you mentioned,

> GPU is known to have high initial setup cost from necessary resource creation processes e.g. device and resource creations, pipeline state and shader compilation etc. 

According to our [WebNN POC](https://github.com/otcshare/chromium-src), we map `fast-answer` to CPU that has low initial setup cost. With that, the app can request a fast first inference for one-shot query scenario. If we drop that, how does an app select the device with fast-first-inference capability like CPU?

wchao1115:
> Another question is in the latest commit, I failed to find the sample code that handles a failing compilation. Are there any reasons to drop it?

As stated in the commit's comment, with this change the browser is expected to handle failing case due to unsupported data type, most likely by falling back to the default option.

index.bs
wchao1115: Typical users may not be able to intuitively infer that `fast-answer` means CPU. I think that's too far of a gap. I previously objected to the term `fast-answer` due to its vagueness. I think there's a benefit of adhering to the WebGL wordings in this case.

wchao1115:
> dictionary CompilationOptions {
>      PowerPreference powerPreference;
> };
> 
I think the dictionary has merit if we believe we need something similar to `WebGLContextAttributes` in the future. For instance, a Boolean field similar to WebGL `failIfMajorPerformanceCaveat` flag could be useful in the future if we need to give the caller a control over how to handle the API failure themselves.

index.bs
```diff
@@ -303,14 +303,20 @@ partial interface NeuralNetworkContext {
...
+enum PowerPreference {
+  // Prioritizes execution speed over other considerations e.g. power consumption
+  "high-performance",
+  // Prioritizes power consumption over other consideraions e.g. execution speed
+  "low-power"
+};
+
+dictionary CompilationOptions {
+  // Compilation preference as related to power consumption level
+  PowerPreference powerPreference;
```
huningxin, line 315: It looks good. Thanks for adding it.

index.bs
huningxin: I agree that the `fast-answer` has some vagueness. To solve this issue, how about we introduce the `DevicePreference` enum that contains 'cpu', 'gpu', 'accelerator' explicitly. The proposal would be
```webidl
enum DevicePreference {
  'cpu',
  'gpu',
  'accelerator'
};

dictionary CompilationOptions {
  PowerPreference powerPreference;
  DevicePreference devicePreference;
};
```

What do you think?

wchao1115: @huningxin By adding a device option to the mix, my concern is that we may be mixing explicitness (e.g CPU, GPU, etc.) with constraints (low power etc.) in the API. This approach rarely produces consistent behaviors in all cases. It's just hard for the implementers of the API to know what's the right thing to do in a given circumstance, and for the consumers  to know what to expect.

Take for example, a low-power profile device on a machine could be an ARM-based CPU, a DSP-based accelerator such as Intel Movidius, or an integrated GPU on a hybrid PC with both integrated and dedicated GPU on board.

In general, it is just hard for a constraint-based API to succeed, especially for one that mixes between a requirement and a preference. This is why defining a very restricted set of preferences is more often a good compromise. WebGL does it this way and DirectX does it this way.

huningxin: > In general, it is just hard for a constraint-based API to succeed, especially one that mixes between requirements and preferences.

I agree. Thanks for sharing your insights.

> Typical users may not be able to intuitively infer that `fast-answer` means CPU. I think that's too far of a gap. 

I agree. The `fast-answer` could be supported by devices other than CPU.

> I previously objected to the term `fast-answer` due to its vagueness.

How about we name it as `fast-first-inference`? Is it more intuitive?

> I prefer to keep just the power preference for now and let the browser and/or the OS pick which device it wants to go with.

It sounds good. If we look from power consumption perspective, the `fast-first-inference` could mean the app prefers the UA to do the first inference as fast as possible, even if this causes more power consumption. So does it make sense to add it into `PowerPreference`?

wchao1115: Regardless of what wording we use, if `fast-answer` is to mean "use CPU" then it doesn't belong to a power preference enum. 

Does "use CPU" always imply "fast first inference"? Is it fast relative to the subsequent inference calls or relative to the first inference call of a different device in the system? 

Let's presume for the sake of argument that we need a way for the developer to specify "I want CPU" option. I believe in the current design that would be the same as specifying "I have no preference, use the default behavior" i.e. just call `createCompilation` with no argument.

huningxin: > Does "use CPU" always imply "fast first inference"?

From the platforms I tested so far, it is true.



> Is it fast relative to the subsequent inference calls or relative to the first inference call of a different device in the system?

It is relative to the first inference call of a different device in the system. For example I shared the WinMLRunner examples on CPU and GPU in previous [comment](https://github.com/webmachinelearning/webnn/pull/50#discussion_r401705495).

> I believe in the current design that would be the same as specifying "I have no preference, use the default behavior" i.e. just call `createCompilation` with no argument.

It sounds fine. As we made `CompilationOptions` as a dictionary that might have other members, how about we add "default" into `PowerPreference` enum. That also aligns to [`WebGLPowerPreference`](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.1). i.e.

```webidl
enum PowerPreference {
  "default",
  "high-performance",
  "low-power"
};
```

What do you think?

wchao1115: @huningxin You previously suggested that we removed `default` and instead making the argument optional (see above). Are you reverting that?

huningxin: Apologize for the confusing. I still suggest making the `CompilationOptions` argument as optional. Now as `powerPreference` is a dictionary member, I am thinking it would be more clear by providing its a default value as 'default'. 

The full idea is
```webidl
enum PowerPreference {
  "default",
  "high-performance",
  "low-power"
};

dictionary CompilationOptions {
  PowerPreference powerPreference = "default";
};

interface Model {
  Promise<Compilation> createCompilation(optional CompilationOptions options = {});
};
```

wchao1115: @huningxin I'm fine with your proposal as that's my original thinking too. That would make `PowerPreference` exactly identical to `WebGLPowerPreference` enum. I like that consistency. For reference, WinML also defines a default option, which in practice uses CPU for inference. If we're good with this, I'll send out a commit with this adjustment and hopefully we can complete this thread.

huningxin: It looks great. Thanks for that!

huningxin:
@wchao1115 , please help resolve the conflicts due to merging PR #49 . Thanks.

wchao1115:
The change is reviewed and merged. I believe we can now close this issue. In summary, we decided that it's the UA responsibility to handle compilation failure for a given power preference. If an unrecoverable failure is triggered, fatal exception will then be thrown.

huningxin:
@wchao1115 , thanks for rebasing and merging.

PR #53, opened 2020-05-02
Add element-wise binary op

huningxin:
The first PR is to add element-wise add and multiply according to [r01](https://www.w3.org/2020/04/30-webmachinelearning-minutes.html#r01) of WebML CG Teleconference – 30 April 2020.

> RESOLUTION: Add elementwise add, elementwise multiply, concatenation, and reshape ops to the WebNN API

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/53.html" title="Last updated on May 4, 2020, 3:39 AM UTC (d926781)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/53/3e09f7b...huningxin:d926781.html" title="Last updated on May 4, 2020, 3:39 AM UTC (d926781)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -236,12 +236,19 @@ interface NeuralNetworkContext {
 };
 </script>
 
-### add ### {#api-neuralnetworkcontext-add}
+### binary ### {#api-neuralnetworkcontext-binary}
 <script type=idl>
+enum BinaryOperation {
+  "add",
+  "sub",
+  "mul",
+  "div",
+  "max",
+  "min"
+};
+
 partial interface NeuralNetworkContext {
-  // Create an Operand object that represents the result of an element-wise
-  // binary addition operation with input operands a and b.
-  Operand add(Operand a, Operand b);
+  Operand binary(BinaryOperation op, Operand lhs, Operand rhs);
 };
 </script>
 
@@ -324,15 +331,6 @@ partial interface NeuralNetworkContext {
         - If both *a* and *b* are 1-D, the operation is a vector dot-product,
             which produces a scalar output.
 </div>
-
-### mul ### {#api-neuralnetworkcontext-mul}
-<script type=idl>
-partial interface NeuralNetworkContext {
-  // Create an Operand object that represents the result of an element-wise
-  // binary multiplication operation with input operands a and b.
-  Operand mul(Operand a, Operand b);
-};
-</script>
 
 ## Model ## {#api-model}
 <script type=idl>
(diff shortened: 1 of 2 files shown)


huningxin:
The first commit is introducing a high-order element-wise `binary` method that can support multiple binary arithmetic operations defined by `BinaryOp` enum. This is based on the idea of @gramalingam. Please take a look. If the shape looks good, I'll add more spec text about arguments and result. Thanks!

anssiko:
Thanks @huningxin! Since minimizing conceptual weight is important when designing APIs, I wonder whether hanging these binary arithmetic operations off of the `NeuralNetworkContext` would make for an easier to use API? That is, use and extend the shape we had initially. That'd also allow autocomplete feature to work, and make it easier to feature detect what ops are supported (in partial implementations).

Current:
>const intermediateOutput0 = nn.add(tensor0, tensor1);

This PR:
>const intermediateOutput0 = nn.binary('add', tensor0, tensor1);

Is the latter a convention that is established? I feel in Web APIs the preference is toward smaller parameter sets, which would suggest the current shape.

I'd like to hear others' feedback, and maybe @gramalingam you can elaborate on your proposal from our call since I failed to minute it in enough detail.

When we're settled on this API shape detail, the editors can then complete this PR.

huningxin:
Thanks @anssiko !

> That'd also allow autocomplete feature to work, and make it easier to feature detect what ops are supported (in partial implementations).

This is a good point.

Actually I am creating this PR is to check whether I got Rama's inputs in [30 April call](https://www.w3.org/2020/04/30-webmachinelearning-minutes.html#t02) correctly.

> <Rama> Elementwise ops can also benefit by introducing a single higher-order op "elementwise"
> <Rama> which takes a scalar function f as a parameter

@gramalingam, could you please clarify?

At meanwhile, I create another #54 that is still using current style and will add other element-wise ops under `NeuralNetworkContext`, say `nn.add`, `nn.sub`, `nn.mul` and `nn.div` etc.,. Feedbacks are welcome. Thanks!

wchao1115:
I second @anssiko feedback. Making an explicit method off the context object is also verifiable, particularly important in the format conversion process. If we are concerned with the first-level bloatedness of the API, a namespace could be used for segmentation e.g. `nn.math.add`.

huningxin:
> `nn.math.add`.

That sounds good.

If there is no objection from @gramalingam, I am going to close this PR.

@wchao1115 @anssiko , please take a look at #54 instead. Thanks!

wchao1115:
> > `nn.math.add`.
> 
> That sounds good.
> 
On a second thought, `nn.math.add` would introduce a regular attribute to the nn interface, making the interface harder to implement because of the additional indirection. It might not be worth a little more tidiness. I retract my previous feedback on this.

gramalingam:
Sorry for the belated response. If we are limited to a fixed-enumeration of the binary-operations, then I agree there is not much difference between the two approaches, and it is simpler to use
```
   const z = nn.add(x, y);
```
Having an element-wise operation like "binary" is useful only if the user is able to define new binary-operations. This would require a somewhat richer interface to be able to construct new scalar binary-operations. This is somewhat related to the other discussions about MLIR-like primitives. I think it is okay to postpone this discussion until we get to that topic.

huningxin:
Thanks @gramalingam for the clarification. I agree with you that the `binary` method with fixed-enumeration of this PR doesn't make much difference. 

As the core binary operators, like `add`, `sub`, `mul` and `div` are commonly supported in both ONNX and XLA-HLO. I would propose to add them under `NeuralNetworkContext` interface as the first step. We can define the extensible higher-order binary op as the next step.

With that, I'll close this PR.

PR #55, opened 2020-05-06
Add concat

huningxin:
This PR is to add concat according to [r01](https://www.w3.org/2020/04/30-webmachinelearning-minutes.html#r01) of WebML CG Teleconference – 30 April 2020.

> RESOLUTION: Add elementwise add, elementwise multiply, concatenation, and reshape ops to the WebNN API

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/55.html" title="Last updated on May 14, 2020, 3:18 PM UTC (7192c61)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/55/97be978...huningxin:7192c61.html" title="Last updated on May 14, 2020, 3:18 PM UTC (7192c61)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -263,6 +263,29 @@ partial interface NeuralNetworkContext {
     is the maximum size along that dimension of the input tensors.
 </div>
 
+### concat ### {#api-neuralnetworkcontext-concat}
+<script type=idl>
+partial interface NeuralNetworkContext {
+  Operand concat(sequence<Operand> inputs, long axis);
+};
+</script>
+
+<div algorithm=concat>
+    **Arguments:**
+        - *inputs*: a sequence of {{Operand}}. All input tensors must have the
+            same shape, except for the size of the dimension to concatenate on.
+        - *axis*: a {{long}}. The axis that the inputs concatenate along, with
+            the value in the interval [0, N) where N is the rank of all the
+            inputs.
+        
+    **Returns:** an {{Operand}}. The concatenated tensor of all the inputs along
+    the *axis*. The output tensor has the same shape except on the dimension
+    that all the inputs concatenated along. The size of that dimension is
+    computed as the sum of all the input sizes of the same dimension.
+
+    Concatenates the input tensors along a given axis.
+</div>
+
 ### conv2d ### {#api-neuralnetworkcontext-conv2d}
 <script type=idl>
 partial interface NeuralNetworkContext {
--- a/index.html
+++ b/index.html
@@ -1223,7 +1223,7 @@ Possible extra rowspan handling
 </style>
   <meta content="Bikeshed version 70ffa934, updated Fri May 1 10:12:04 2020 -0700" name="generator">
   <link href="https://webmachinelearning.github.io/webnn/" rel="canonical">
-  <meta content="6c740a807b2d1702c0a97a21bbdc3e5cae863738" name="document-revision">
+  <meta content="27241e9161b8dfb1da14bcf97d496e73d6b7d1de" name="document-revision">
 <style>/* style-autolinks */
 
 .css.css, .property.property, .descriptor.descriptor {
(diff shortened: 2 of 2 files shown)


index.bs
```diff
@@ -245,6 +245,30 @@ partial interface NeuralNetworkContext {
...
+partial interface NeuralNetworkContext {
+  Operand concat(sequence<Operand> inputs, long dimension);
+};
+</script>
+
+<div algorithm=concat>
+    **Arguments:**
+        - *inputs*: a sequence of {{Operand}}. All input tensors must have the
+            same shape, except for the size of the dimension to concatenate on.
+        - *dimension*: a {{long}}. It specifies the dimension to concatenate on.
```
anssiko, line 259: The dimension to concatenate on.

index.bs
```diff
@@ -245,6 +245,30 @@ partial interface NeuralNetworkContext {
...
+            Its value is in the interval [0, N) where N is the rank of input
```
anssiko, line 260: The value is in the internal

index.bs
huningxin: thanks. fixed in new commit.

index.bs
huningxin: thanks. fixed in the new commit.

index.bs
wchao1115: I think the `dimension` argument should instead be named `axis`. It is a known term across many ML API. I think webnn should be consistent in this regard. Also since this won't be the first API that needs `axis`, what name we choose here will act as a precedence for the remaining of the API.

wchao1115: To help with the description of the `axis` argument, here is a suggestion: "The axis that the inputs concatenate along, with the value in the interval [0, N), default to 0, where N is the rank of all the inputs."

index.bs
```diff
@@ -245,6 +245,30 @@ partial interface NeuralNetworkContext {
...
+            tensors.
+        
+    **Returns:** an {{Operand}}. The output tensor is concatenated from all
+    input tensors along the *dimension*. The output tensor has the same shape of
+    the input tensors, except for the concatenation dimension. For the
+    concatenation dimension of the output tensor, its size is the sum of size
+    along that dimension of all input tensors.
+
+    Concatenates the input tensors along the given dimension.
+</div>
```
wchao1115, line 269: Suggestion: "The concatenated tensor of all the inputs along the `axis`. The output tensor has the same shape except on the dimension that all the inputs concatenated along."

index.bs
huningxin: The `axis` name looks good. 

> default to 0

Do you mean it is an optional argument with default value as 0. It would change the signature to `Operand concat(sequence<Operand> inputs, optional long axis = 0);`. I am not aware any native ML APIs, e.g. ONNX or XLA-HLO, follow this. Could you please share some pointers? Thanks.

index.bs
huningxin: Your suggestion sounds good. I'll incorporate it. Thanks.

index.bs
wchao1115: I don't feel strongly about it being an optional argument. It's up to you. Just note that TF.js has axis as optional.

huningxin: Thanks for TF.js reference. As most APIs don't use default value 0, I'll follow that.

huningxin: Fixed in new commit. Please take a look.

index.bs
huningxin: Fixed in the new commit. Please take a look.

huningxin:
@wchao1115 , pushed a new commit that fixes your comments. Please take another look. Thanks!

wchao1115:
> @wchao1115 , pushed a new commit that fixes your comments. Please take another look. Thanks!

Looks good to me. Thanks!

index.bs
gramalingam: Looks good to me. One approach common in many frameworks is the use of negative axis values to count backwards from the last axis. Thus, axis=-1 would mean last axis, and axis=-2 would mean second axis from last. This may be worth adopting, as it will simplify translation from other frameworks to this. What do you think?

huningxin: Thanks @gramalingam for proposing the negative axis values. AFAIK, there are some OS ML APIs, e.g. [join operator of DirectML](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_join_operator_desc) and [concatenation operation of NNAPI](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0a44cbea825c4b224dd3ea757e9b1f65ed), don't support the negative axis values. If we decided to support that, it would increase the complexity of browser implementation for those APIs. Instead I would prefer to leaving it to JavaScript frameworks. Does it make sense?

huningxin: @gramalingam, are you fine with non supporting negative axis values in this PR?

gramalingam: It is fine to merge this PR as is. (Note that one situation where the graph-builder, say javascript frameworks, cannot do this statically is when the rank of the tensors is not known statically.)

huningxin: Thanks @gramalingam for pointing out this situation. As this is not required by [first wave models](https://github.com/webmachinelearning/webnn/blob/master/op_compatibility/first_wave_models.md), let's merge it. If there are new requirement, we can create PR to amend it.

huningxin:
Resolved the conflicts. I am going to merge it. Thanks all for review!

PR #58, opened 2020-05-10
Gemm and transpose specification

wchao1115:
Defining a specification for `gemm` as a static method on the `nn` interface instead of a regular interface method. This is because in its most generic form, the definition of gemm can be entirely implemented in terms of other regular operations as shown in the sample code next to the specification text. By defining it as a static method to the interface, the `gemm` operation acts as a mere helper function for the caller of WebNN. And since WebIDL permits the implementation of static methods, the browser is free to implement `gemm` however they choose to suit their needs, including treating it as a predefined fused operation in the operating system layer for better perf.

This proposal provides a compromised solution for the high-level vs. low-level primitive discussion, as the caller can choose whether to use the static helper functions or to explicitly call into regular interface operations, while the browser can implement the former as fused operators for better perf.

In addition to `gemm`, this change also define a specification for `transpose`. The specification for `transpose` should be non-controversial as it's consistently defined in virtually all formats including numpy, XLA, TF.JS and ONNX. It should be noted that since `transpose` is needed in the generic definition of `gemm`, it must also be defined as a regular operation in the `nn` interface for completeness.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/58.html" title="Last updated on May 14, 2020, 12:10 AM UTC (03a9a99)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/58/6a12df8...03a9a99.html" title="Last updated on May 14, 2020, 12:10 AM UTC (03a9a99)">Diff</a>


index.bs
```diff
@@ -292,6 +292,42 @@ partial interface NeuralNetworkContext {
...
+    This static operation can be generically implemented in term of other interface operations as follow. However, the user agent is free to implement the operation differently to suit their needs.
+
+    <pre highlight="js">
+    if (aTranspose)
+      a = nn.transpose(a);
+
+    if (bTranspose)
+      b = nn.transpose(b);
+
+    return nn.add(nn.matmul(nn.mul(alpha, a), b), nn.mul(beta, c));
```
huningxin, line 327: The current `mul` operator doesn't accept a float argument, e.g `alpha` and `beta` here.

index.bs
```diff
@@ -292,6 +292,42 @@ partial interface NeuralNetworkContext {
...
+                      optional float alpha = 1.0, optional float beta = 1.0, 
+                      optional boolean aTranspose = false, optional boolean bTranspose = false);
+};
+</script>
+<div algorithm=gemm>
+    **Arguments:**
+        - *nn*: The neural network context.
+        - *a*: an {{Operand}}. The first input 2-D tensor.
+        - *b*: an {{Operand}}. The second input 2-D tensor.
+        - *c*: an {{Operand}}. The third input 2-D tensor.
```
huningxin, line 308: May `c` be an optional argument?

index.bs
wchao1115: Thanks for the feedback. You're right, the float value needs to first be converted to a constant tensor. Since we don't support scalar tensors (do we?), I assume we need to turn it into a single-value tensor. I've added a constant method that helps with this. In ONNX, there is the `constant` op that does this, and in TF.JS there is `tf.scalar'. Please take a look at my new commit.

index.bs
wchao1115: Yes. Please see the change in my new commit.

index.bs
huningxin: We do support scalar tensors. In current spec, the scalar tensor with value `alpha` can be created by:
```js
nn.constant({type: ‘float32'}, new Float32Array([alpha]))
```

index.bs
```diff
@@ -231,6 +231,9 @@ interface NeuralNetworkContext {
   // Create an Operand object that represents a model constant.
   Operand constant(OperandDescriptor desc, ArrayBufferView value);
 
+  // Create a single-value tensor of the specified type from the specified value.
+  Operand constant(float value, optional OperandType type = "tensor-float32");
```
huningxin, line 235: The default value of `type` could be `"float32"`. 

If the `type` is specified as `"int32"`, the `float` type of `value` would not match the `type` parameter.

To create a float32 or int32 scalar tensor, it can be implemented in JavaScript as a helper:
```js
function createFloat32Scalar(value) {
  return nn.constant({type: "float32"}, new Float32Array([value]));
}

function createInt32Scalar(value) {
  return nn.constant({type: "int32"}, new Int32Array([value]));
}
```

index.bs
huningxin: It looks good. Thanks!

index.bs
wchao1115: I just think that webidl does allows overloads, and that having to allocate an array of 1 element is unnecessary overhead. Existing frameworks allow creating either a scalar tensor or a constant tensor from a single value just fine. And I would like to avoid requiring the dev to have to provide their own helper function on top of the API if we could make it easier for them. 

For my choice, I went with the most obvious way that works with the current way the `OperandDescriptor` is defined, which I think can still be improved e.g. I'm not entirely clear on why there is a `"float32"` and a `"tensor-float32"` at the same time. A cleaner way could be the type-bound overloads e.g.  `nn.constant(float value)', 'nn.constant(int value)', etc. But that would require that we remove one of the two data types and make data type "float32" to just mean either float value or float tensor. I'm still open to suggestion on this.

index.bs
wchao1115: I just think `nn.constant(float value)` is a lot more succinct. Also having to create an array of one element seems like an unnecessary overhead.

index.bs
gramalingam: I prefer that the returned Operand's type be determined by the type of the input: e.g., `nn.constant(float value)', 'nn.constant(int value)', as you suggest. I don't follow why this requires changing "OperandType". Whether we should have both "float32" and "tensor-float32" seems a separate discussion to me (with pros/cons both sides; I prefer having the two types).

index.bs
```diff
@@ -292,6 +295,43 @@ partial interface NeuralNetworkContext {
     Computes a 2-D convolution given input and filter 4-D tensors.
 </div>
 
+### gemm ### {#api-neuralnetworkcontext-gemm}
+<script type=idl>
+partial interface NeuralNetworkContext {
+  static Operand gemm(NeuralNetworkContext nn, Operand a, Operand b, optional Operand c = null,
```
gramalingam, line 301: I don't know IDL conventions enough, but is it possible to make this a non-static method? It seems to me that has the advantage of preserving the style in which operations are specified (that is, no difference between how you invoke "gemm" and "matmul" from user's perspective.)

gramalingam: Note that we can always have a separate static-method, if needed, to define the expansion.

index.bs
```diff
@@ -334,6 +374,22 @@ partial interface NeuralNetworkContext {
...
+### transpose ### {#api-neuralnetworkcontext-transpose}
+<script type=idl>
+partial interface NeuralNetworkContext {
+  Operand transpose(Operand input, optional sequence<long> permutation);
+};
+</script>
+<div algorithm=transpose>
+    **Arguments:**
+        - *input*: an {{Operand}}. The input N-D tensor.
+        - *permutation*: an optional sequence of {{long}} values. The values used to permute the output shape. When it's not specified, it's set to `[N-1...0]`, where `N` is the rank of the input tensor. This causes the output to become a transposed tensor of the input. When specified, the number of values in the sequence must be the same as the rank of the input tensor.
```
gramalingam, line 386: How about adding ", which must be a permutation of the sequence [1,2,...,N]." after "an optional sequence of long values"?

anssiko:
As a general comment, it often makes sense to design an API in such a way it makes the common case(s) easy and uncommon case(s) possible if not as straightforward. Said differently, the API should be easy to use and hard to misuse. Folks who need to do uncommon or more complex things refer to the API docs, while the simple common things should be easy.

index.bs
huningxin: According to [the restriction of WebIDL overloading](https://heycam.github.io/webidl/#idl-overloading), the numeric types, including `float` and `long`, are not distinguishable. So I am afraid the idea that defining `nn.constant(float value)` and `nn.constant(long value)` would not work.

For scalar constant, based on @wchao1115 's proposal, we may change the `float` to `double` and set default type as `"float32"`. 
```webidl
Operand constant(double value, optional OperandType type = "float32");
```

With that, the code to create a float32 scalar would be:
```js
nn.constant(alpha);
```

To create an int32 scalar:
```js
nn.constant(beta, "int32");
```

It can overload the tensor constant method, as  `double` and `dictionary` are distinguishable.
```
Operand constant(OperandDescriptor desc, ArrayBufferView value);
```

index.bs
wchao1115: It's always possible to make this a regular method instead of a static one. The WebIDL convention allows both. At the conceptual level though, a static method is considered part of the interface while a regular one is part of an instance of the interface (the object). It implies that a static method is in fact a helper function that requires no state within the interface instance to function properly.

From the caller's point of view, calling into a static method is like calling into a function, and not to an interface object's method. So the semantic is a bit different. 

I propose this distinction in the API to highlight the fact that there are regular operations of the interface (small ops) vs. helper functions to construct subgraphs made of regular operations (big ops).

A nice twist here though is that WebIDL allows static methods to be implementable by the interface implementer. This effectively means the browsers can now override the implementations of these helper functions to their own choosing. For instance, they can implement gemm as a fused operation at the OS layer, etc.

index.bs
wchao1115: Thanks @huningxin. I've updated it per your suggestion. Please take a look at my new commit.

index.bs
wchao1115: Thanks @gramalingam. I understand your feedback and have updated it differently in my new commit. Please have a look if that's ok.

index.bs
wchao1115: I dropped the static method for the gemm operation and instead creating a non-normative note section in the operation's description. Thanks @rafaelcintron for the WebGL sample on non-normative section of a standard spec. @gramalingam please would you have a look at my new commit and let me know if it addresses your concern.

index.bs
huningxin: It makes sense.

index.bs
huningxin: It looks good. Thanks!

index.bs
```diff
@@ -292,6 +297,43 @@ partial interface NeuralNetworkContext {
...
+    <div class="note">
+    The behavior of this operation can be generically emulated from the usage of other operations as follow. However, user agents typically have a more efficient implementation for it, therefore its usage is encouraged from the performance standpoint.
+    <pre highlight="js">
+    if (aTranspose)
+      a = nn.transpose(a);
+
+    if (bTranspose)
+      b = nn.transpose(b);
+
+    var ab = nn.matmul(nn.mul(nn.constant(alpha), a), b);
```
huningxin, line 331: It might be better to use `const` or `let` instead of `var` to declare `ab`.

PR #65, opened 2020-06-03
Support grouped conv2d

huningxin:
Fix [resolution 01](https://www.w3.org/2020/05/28-webmachinelearning-minutes.html#r01) of [WebML CG Teleconference – 28 May 2020](https://www.w3.org/2020/05/28-webmachinelearning-minutes.html) that

> RESOLUTION: Add grouped convolution support to existing conv2d definition in WebNN API

This PR incorporates the enhancement of conv2d in #64. 

@wchao1115 @gramalingam, please take a look.

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/65.html" title="Last updated on Jun 9, 2020, 4:43 AM UTC (d38e104)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/65/a68905d...huningxin:d38e104.html" title="Last updated on Jun 9, 2020, 4:43 AM UTC (d38e104)">Diff</a>


index.bs
```diff
@@ -295,50 +295,61 @@ partial interface NeuralNetworkContext {
...
             layout format of the input, output and filter tensor.
             Specifically,
 
             "nchw":
                 - input tensor: [batches, input_channels, height, width]
-                - filter tensor: [output_channels, input_channels, height,
-                    width]
+                - filter tensor: [output_channels, input_channels,
+                    height, width] when *groups* is equal to 1;
+                    [groups, output_channels/groups, input_channels/groups,
```
wchao1115, line 338: For grouped convolution, the filter shape should still be 4D of `[groups, input_channels/groups, H, W]` because the number of output channels will be derived by the process to be the number of groups, so the output tensor shape is `[batches, groups, H, W]`.

For `nhwc` (*channel last* format), the filter shape is `[H, W, groups, input_channels/groups]`.

Note that to implement depthwise in MobileNet, the group count is set to the number of input channels, so if the input channels is, say, 3, the filter shape will be `[3, 1, H, W]` and the output `[batches, 3, H, W]` accordingly (for 'nchw' layout).

wchao1115: @huningxin for `nchw` I believe we are saying the same thing. When group count is > 1, we are basically breaking up the filter that used to be of `input_channels` depth into groups. But because the total number of channels across all the groups must still be equal to `input_channels` , the filter shape therefore is `[groups, input_channels/groups, H, W]` as the total number of elements in the tensor will still be `groups * input_channels/groups * H * W` = `input_channels * H * W`. The `output_channels` is irrelevant in this case because it's *depthwise*, the number of output channels is only relevant when there is just 1 group of filter channels in use. Hope it makes sense. 

As for `nhwc`, it is `[H, W, groups, input_channels/groups]` because again the total number of tensor elements of the filter across all groups is still `H * W * input_channels'.

huningxin: > The `output_channels` is irrelevant in this case because it's _depthwise_, the number of output channels is only relevant when there is just 1 group of filter channels in use. Hope it makes sense.

Thanks for the explanation @wchao1115 . I have no questions about _depthwise_ case. But for general cases, the filter shape as `[groups, input_channels/groups, H, W]` seems not work. For instance, in [AlexNet ONNX model](https://github.com/onnx/models/tree/master/vision/classification/alexnet), the conv with output "conv4_1" has `input_channels == 384` and `groups == 2`. According to `[groups, input_channels/groups, H, W]` definition, the filter shape would be `[2, 192, 3, 3]`. However,  the actual filter shape is `[384, 192, 3, 3]` where the `output channels == 384`. So I propose to define filter shape as `[output_channels, input_channels/groups, H, W]`.

I post a netron screenshot for your reference. Hope it helps.
image

wchao1115: I checked with our implementation in DirectML and you're right @huningxin. Depthwise is just a special case when `groups` = `input_channels` = `output_channels`. The more generalized use case of grouped convolution has the filter shape for `nchw`  and `nhwc` as `[output_channels, input_channels/groups, H, W]` and `[H, W, input_channels/groups, output_channels]` respectively. Does that sound right? Thanks for the sample on AlexNet.

Maybe we should also call out the depthwise special case in a note section to be more specific about how it can be achieved using grouped convolution.

huningxin: @wchao1115 , thanks for the checking. I pushed a new commit that changes the filter tensor to 4D and adds depthwise case in a note. Please take another look.

huningxin:
Rebased to master after #64 merged. @wchao1115 @gramalingam, please take another look. Thanks!

index.bs
```diff
@@ -313,25 +314,34 @@ partial interface NeuralNetworkContext {
...
     **Returns:** an {{Operand}}. The output 4-D tensor that contains the
     result of the convolution. The logical shape is interpreted according to the
     value of *layout*.
+
+    <div class="note">
+    A *depthwise* conv2d operation is a variant of grouped conv2d where the
+    *groups* = input_channels = output_channels and the shape of filter tensor
+    is [groups, 1, height, width] for "nchw" layout or [height, width, 1,
+    groups] for "nhwc" layout.
+    </div>
```
wchao1115, line 344: Nit. "A *depthwise* conv2d operation is a variant of grouped conv2d, **used in models like the MobileNet**, where the..."

index.bs
wchao1115: Looks good @huningxin. I just have a nit suggestion in the commit to make the note section a bit more clear.

index.bs
huningxin: Good idea. I added that into a new commit. Please take another look.

huningxin:
With two approvals, I am going to merge it. Thanks for the review!

Issue #66, opened 2020-06-04
Evaluate noise suppression models for required ops


anssiko:
Noise suppression use case was adopted by WebNN API in #61 (thanks @ibelem!):

https://webmachinelearning.github.io/webnn/#usecase-noise-suppression

This issue is to discuss the next steps:
* Evaluate RNNoise model for required ops, to be added to the [first-wave models](https://github.com/webmachinelearning/webnn/blob/master/op_compatibility/first_wave_models.md)
* Gauge PoC interest, possible collaboration with related groups e.g. WebRTC
* Sharing resources, e.g. [RNNoise paper](https://arxiv.org/pdf/1709.08243.pdf) and [demo](https://jmvalin.ca/demo/rnnoise/#music_player)

kdavis-mozilla:
Hey @jmvalin your work is coming up yet again!

@wchao1115, the WebNN API spec co-editor, was interested in exploring RNNoise a bit further in this context. So, to connect the dots, I thought it apropos to ping you here if any RNNoise questions arise, as, after all, you know more about RNNoise than anyone.

jmvalin:
@wchao1115  @kdavis-mozilla @anssiko Sure, let me know how I can help.

wchao1115:
Thanks @anssiko, @kdavis-mozilla, and @jmvalin. In addition to RNNoise I'd also like to propose [NSNet](https://arxiv.org/pdf/2001.10601.pdf), the baseline model for the [Interspeech DNS Challenge](https://dns-challenge.azurewebsites.net/).

wchao1115:
@jmvalin. From the implementation standpoint, both RNNoise and NSNet are hybrid DSP/ML based solution. Given that webnn is focusing on the deep learning part, how do you see the solution integrated into the web platform, especially on the signal processing part? 

I'm not up to date on the state of Web Audio implementation in the web browsers; do you think the level of support for FFT windowing in the browser is sufficient for RNNoise?

jmvalin:
I don't actually know the state of DSP in the browser. Being able to compute an FFT would certainly be nice, but it's not a showstopper. Similarly, having fast kernels for convolutions/correlations would be equally useful since RNNoise performs pitch analysis. Honestly what I've like to see at some point is a WebBLAS (plus FFT and convolution/correlation). That would probably cover most use cases -- including a big chunk of WebML.

wchao1115:
To date, we still need to define `split` and `clamp` to satisfy the model's requirements.

[...]

PR #83, opened 2020-08-16
Add GRU and corresponding operators

wchao1115:
Fill the operator gaps to support noise suppression first-wave models as outlined [here](https://github.com/webmachinelearning/webnn/blob/master/op_compatibility/first_wave_models.md). 

Adding GRU and GRUcell operators to support GRU recurrent network. Defining a cell operator in addition to the network operator for added customization flexibility e.g. to support stacked cell recurrent network, etc.

#66

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/83.html" title="Last updated on Sep 1, 2020, 8:14 PM UTC (aa02309)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/83/876b67a...aa02309.html" title="Last updated on Sep 1, 2020, 8:14 PM UTC (aa02309)">Diff</a>


index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
     </div>
 </div>
 
+### gru ### {#api-neuralnetworkcontext-gru}
+Gated Recurrent Unit (GRU) recurrent network using an update gate and a reset gate to compute the hidden state that rolls into the output across the temporal sequence of the network, as outlined in this [paper](https://arxiv.org/pdf/1406.1078.pdf).
```
anssiko, line 494: I suggest you make this link a biblio reference (add into `<pre class="biblio">`) and point to it instead. This allows the reader to see all the references at https://webmachinelearning.github.io/webnn/#references and makes a clear distinction between normative (`[[!foo]]`) and informative references (`[[foo]]`). 

(I notice this "in this paper" convention is used in batchNormalization prose as well, so that could be aligned similarly as well.)

anssiko:
Other than the referencing style comment, LGTM.

I'll defer to @huningxin for the API arguments and returns definitions.

wchao1115:
@pyu10055 please have a look also.

anssiko:
@pyu10055 now has the superpower to assign himself as a reviewer for PRs. Also the editors and the chair can assign him. Thank you @pyu10055 for your help in reviewing changes to the API spec.

Note to my future self: in order to be able to assign folks as reviewers with the built-in feature, an individual must be invited as a member of the GH org (here `webmachinelearning`) and granted at minimum read permissions to the repository in question. See https://docs.github.com/en/github/managing-your-work-on-github/assigning-issues-and-pull-requests-to-other-github-users

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+                  Operand recurrentWeight, Operand hidden,
+                  optional Operand bias, optional Operand recurrentBias,
+                  optional boolean resetAfter = true,
+                  optional RecurrentNetworkWeightLayout layout = "zrn",
+                  optional sequence<RecurrentNetworkActivation> activations);
+};
+</script>
+<div algorithm=grucell>
+    **Arguments:**
+        - *hiddenSize*: a {{long}} scalar. The value of the second dimension of the output tensor shape. It indicates the number of features in the hidden state.
```
huningxin, line 608: As other methods, it might be better to move this (attribute) argument after the last `Operand` arguments (`recurrentBias`).

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *input*: an {{Operand}}. The input 2-D tensor of shape [batch_size, input_size]. 
+        - *weight*: an {{Operand}}. The 2-D input weight tensor of shape [3 * hidden_size, input_size]. The ordering of the weight vectors in the first dimension of the tensor shape is specified according to the *layout* argument.
```
huningxin, line 610: Do we need to mention the `hidden_size` is specified by the *hiddenSize* argument?

huningxin: name it `weights`?

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *recurrentWeight*: an {{Operand}}. The 2-D recurrent weight tensor of shape [3 * hidden_size, hidden_size]. The ordering of the weight vectors in the first dimension of the tensor shape is specified according to the *layout* argument.
```
huningxin, line 611: ditto

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *hidden*: an {{Operand}}. The 2-D input hidden state tensor of shape [batch_size, hidden_size].
```
huningxin, line 612: According to the paper, would `initialHiddenState` be a better name?

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *bias*: an optional {{Operand}}. The 1-D input bias tensor of shape [3 * hidden_size]. The ordering of the bias vectors in the first dimension of the tensor shape is specified according to the *layout* argument.
+        - *recurrentBias*: an optional {{Operand}}. The 1-D recurrent bias tensor of shape [3 * hidden_size]. The ordering of the bias vectors in the first dimension of the tensor shape is specified according to the *layout* argument.
+        - *resetAfter*: an optional {{boolean}} indicating whether to apply the reset gate after or before matrix multilication. Default to true.
```
huningxin, line 615: `resetAfterLinear` that aligns to [ONNX GRU attribute - `linear_before_reset`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#GRU)?

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *layout*: an optional {{RecurrentNetworkWeightLayout}}. The ordering of the weight and bias vectors for the internal gates of GRU, specifically the *update (z)*, *reset (r)*, and *new (n)* gate, as indicated in the first dimension of the weight and bias tensor shapes. When not specified, the default layout is *"zrn"*.
+        - *activations*: an optional sequence of {{RecurrentNetworkActivation}}. A pair of activation functions with the first function used for the update and reset gate, and the second used for the new gate. When not specified, it's default to the sigmoid ("sigmoid") and the hyperbolic tangent ("tanh") function respectively.
+
+    **Returns:** an {{Operand}}. The 2-D tensor of shape [batch_size, hidden_size], the cell output of a single time step of the recurrent network.
```
huningxin, line 619: "the cell output *hidden state*"?

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+
+    <div class="note">
+    The behavior of this operation with default argument values can be generically emulated from the usage of other operations as follow. However, user agents typically have a more efficient implementation for it, therefore its usage is encouraged from the performance standpoint.
+    <pre highlight="js">
+    let one = nn.constant(1);
+    let zero = nn.constant(0);
+
```
huningxin, line 626: Would you like to illustrate the handling of `layout` and `activations`?

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *activations*: an optional sequence of {{RecurrentNetworkActivation}}. A pair of activation functions with the first function used for the update and reset gate, and the second used for the new gate. When not specified, it's default to the sigmoid ("sigmoid") and the hyperbolic tangent ("tanh") function respectively.
+
+    **Returns:** an {{Operand}}. The 2-D tensor of shape [batch_size, hidden_size], the cell output of a single time step of the recurrent network.
+
+    <div class="note">
+    The behavior of this operation with default argument values can be generically emulated from the usage of other operations as follow. However, user agents typically have a more efficient implementation for it, therefore its usage is encouraged from the performance standpoint.
+    <pre highlight="js">
+    let one = nn.constant(1);
+    let zero = nn.constant(0);
```
huningxin, line 625: `const one = nn.constant(1)`?

index.bs
huningxin: It makes sense. Thanks for the change.

index.bs
huningxin: It makes sense. Then current naming looks good.

index.bs
huningxin: Thanks for the explanation and reference. It sounds good.

index.bs
huningxin: > You mean in the informative sample code section?

Yes. 

Or at least we add some comments in the sample code that mentions this implementation is just for default `layout` and `activations` case. What do you think?

index.bs
```diff
@@ -571,32 +572,31 @@ partial interface NeuralNetworkContext {
...
 
 ### gruCell ### {#api-neuralnetworkcontext-grucell}
-A single time step of the Gated Recurrent Unit (GRU) recurrent network using an update gate and a reset gate to compute the hidden state that rolls into the output across the temporal sequence of a recurrent network.
+A single time step of the Gated Recurrent Unit [[GRU]] recurrent network using an update gate and a reset gate to compute the hidden state that rolls into the output across the temporal sequence of a recurrent network.
 <script type=idl>
 partial interface NeuralNetworkContext {
-  Operand gruCell(long hiddenSize, Operand input, Operand weight, 
-                  Operand recurrentWeight, Operand hidden,
+  Operand gruCell(Operand input, Operand weight,
+                  Operand recurrentWeight, Operand hiddenState, long hiddenSize,
```
huningxin, line 599: As you use `initialHiddenState` in `gru`, would you like to use `initialHiddenState` for `gruCell`?

index.bs
huningxin: it makes sense.

index.bs
huningxin: Looks good. Thanks!

anssiko (approved):

anssiko:
@pyu10055 I took the liberty to assign you as a reviewer to this PR without your explicit consent. Do you prefer to be added as a reviewer for the WebNN API changes going forward? I'd prefer not to burden you too much with review requests if you have limited bandwidth for reviews currently.

One option is to keep you in the review loop in a non-blocking manner i.e. if we don't hear any concerns from you we consider you're OK with the proposed changes. Would that work for you? Thanks for your help!

index.bs
huningxin: Another reason is I am thinking about referring to the sample code for webnn-polyfill implementation. I guess it would also help other implementers.

For example the handling of `activations`,
```js
let activation1 = nn.sigmoid;
let activation2 = nn.tanh;
if (activations) {
  activation1 = nn[activations[0]];
  activation2 = nn[activations[1]];
}

let z = activation1(....);

let r = activation1(....);

let n = activation2(....);
```

index.bs
wchao1115: `gruCell` only handles a single time step that accepts just a single input hidden state and produces just a single output hidden state while `gru` runs through the cell for the entire temporal sequence with multiple hidden states produced for and accepted by the cell in each time step. That's why the prefix *initial* is only applicable to `gru` and not `gruCell`. Let me know if that makes sense.

wchao1115:
@gramalingam

index.bs
huningxin: > Or at least we add some comments in the sample code that mentions this implementation is just for default `layout` and `activations` case.

I just realized that I missed the text "The behavior of this operation with default argument values". Apologize for that. It looks good to me.

index.bs
huningxin: It makes sense. Thanks for the explanation.

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+      cellBias.push(bias ? (nn.squeeze(nn.slice(bias, [slot, 0], [slot + 1, -1]), [0])) : null);
+      cellRecurrentBias.push(recurrentBias ? (nn.squeeze(nn.slice(recurrentBias, [slot, 0], [slot + 1, -1]), [0])) : null);
+    }
+
+    for (let step = 0; step < steps; ++step) {
+      let cellHidden = [];
+      let cellOutput = null;
+
+      for (let slot = 0; slot < slots; ++slot) {
+        cellHidden.push(nn.squeeze(nn.slice(hidden, [slot, 0, 0], [slot + 1, -1, -1]), [0]));
```
huningxin, line 567: `hidden` is not defined. Is it `hiddenState`?

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+        - *activations*: an optional sequence of {{RecurrentNetworkActivation}}. A pair of activation functions with the first function used for the update and reset gate, and the second used for the new gate. When not specified, it's default to the sigmoid ("sigmoid") and the hyperbolic tangent ("tanh") function respectively.
+
+    **Returns:** a sequence of {{Operand}}. The first element of the sequence is a 3-D tensor of shape [num_directions, batch_size, hidden_size], the cell output from the last time step of the network. Additionally, if *returnSequence* is set to true, the second element is the 4-D output tensor of shape [steps, num_directions, batch_size, hidden_size] containing every cell outputs from each time step in the temporal sequence.
+
+    <div class="note">
+    The behavior of this operation with default argument values can be generically emulated from the usage of other operations as follow. However, user agents typically have a more efficient implementation for it, therefore its usage is encouraged from the performance standpoint.
+    <pre highlight="js">
+    let hiddenState = initialHiddenState;
+    let slots = (direction == "both" ? 2 : 1);
```
huningxin, line 547: Would it be good to be `const slots`?

index.bs
wchao1115: Super. I did miss renaming that one.

index.bs
wchao1115: Yes. Fixing.

index.bs
```diff
@@ -452,11 +386,79 @@ partial interface NeuralNetworkContext {
...
+        - *sigmoid*: Compute the sigmoid function of the input tensor, element-wise.
+        - *sin*: Compute the sine of the input tensor, element-wise.
+        - *sqrt*: Compute the square root of the input tensor, element-wise.
+        - *tan*: Compute the tangent of the input tensor, element-wise.
+        - *tanh*: Compute the hyperbolic tangent of the input tensor, element-wise.
+</div>
+
 ### gemm ### {#api-neuralnetworkcontext-gemm}
-Calculate the <a href="https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms#Level_3">general matrixmultiplication of the Basic Liner Algebra Subprograms</a>. The calculation follows the expression `alpha * A * B + beta *C`, where `A`, `B`, and `C` are matrices, and `A` and `B` may optionally be transposed prior to the calculation.
+Calculate the [general matrixmultiplication of the Basic Liner Algebra Subprograms](https://en.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms#Level_3). The calculation follows the expression `alpha * A * B + beta *C`, where `A`, `B`, and `C` are matrices, and `A` and `B` may optionally be transposed prior to the calculation.
```
gramalingam, line 458: Is having matrixmultiplication as one word intentional? Two words seems better to me.

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+
+enum RecurrentNetworkActivation {
+  "relu",
+  "sigmoid",
+  "tanh"
+};
+
+partial interface NeuralNetworkContext {
+  sequence<Operand> gru(Operand input, Operand weight, 
+                        Operand recurrentWeight, Operand initialHiddenState,
```
gramalingam, line 515: Should we make the initialHiddenState optional (with a value 0, if not specified)?

index.bs
wchao1115: its a typo from the past. i'll fix it.

index.bs
wchao1115: good point. i'll look into it.

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+};
+</script>
+<div algorithm=gru>
+    **Arguments:**
+        - *input*: an {{Operand}}. The input 3-D tensor of shape [steps, batch_size, input_size]. 
+        - *weight*: an {{Operand}}. The 3-D input weight tensor of shape [num_directions, 3 * hidden_size, input_size]. The ordering of the weight vectors in the second dimension of the tensor shape is specified according to the *layout* argument.
+        - *recurrentWeight*: an {{Operand}}. The 3-D recurrent weight tensor of shape [num_directions, 3 * hidden_size, hidden_size]. The ordering of the weight vectors in the second dimension of the tensor shape is specified according to the *layout* argument.
+        - *initialHiddenState*: an {{Operand}}. The 3-D initial hidden state tensor of shape [num_directions, batch_size, hidden_size].
+        - *hiddenSize*: a {{long}} scalar. The value of the third dimension of the cell output tensor shape. It indicates the number of features in the hidden state.
+        - *steps*: a {{long}} scalar. The number of time steps in the recurrent network.
```
gramalingam, line 532: What about varying length sequences? Do we want to support them?

index.bs
```diff
@@ -696,11 +922,28 @@ partial interface NeuralNetworkContext {
...
+partial interface NeuralNetworkContext {
+  Operand slice(Operand input, sequence<long> starts, sequence<long> ends);
+};
+</script>
+<div algorithm=slice>
+    **Arguments:**
+        - *input*: an {{Operand}}. The input tensor.
+        - *starts*: a sequence of {{long}}. The starting indices of the corresponding axes of the input shape.
+        - *ends*: a sequence of {{long}}. The ending indices of the corresponding axes of the input shape.
+            The ending index value of -1 selects all the remaining tensor values from the starting index of the given axis.
```
gramalingam, line 942: If we want to allow -1, it may be convenient to allow any negative value, interpreted as counting back from the end, and to allow it for starts as well.

index.bs
```diff
@@ -696,11 +922,28 @@ partial interface NeuralNetworkContext {
...
+    **Returns:** an {{Operand}}. The output tensor. The values of the output
     tensor are the same as values of the input tensor. The shape of the output
     tensor is specified by the *newShape* argument.
 </div>
 
+### slice ### {#api-neuralnetworkcontext-slice}
+Produce a slice of the input tensor.
+<script type=idl>
+partial interface NeuralNetworkContext {
+  Operand slice(Operand input, sequence<long> starts, sequence<long> ends);
```
gramalingam, line 934: Adding a sequence<long> axes may be helpful.

index.bs
```diff
@@ -488,13 +490,238 @@ partial interface NeuralNetworkContext {
...
+
+      hiddenState = cellOutput;
+
+      if (returnSequence) {
+        cellOutput = nn.reshape(cellOutput, [1, slots, -1, hiddenSize]);
+        sequence = (sequence ? nn.concat([sequence, cellOutput], 0) : cellOutput);
+      }
+    }
+
+    return (sequence ? [hiddenState, sequence] : [hiddenState]);
```
gramalingam, line 589: nit: may be use "returnSequence ? [hiddenState, sequence] : [hiddenState]"? There will be a difference only in the case where steps=0.

gramalingam: I guess that does raise another question: the returned value should probably be an empty tensor (of size 0) rather than null.

index.bs
wchao1115: I can't find other use of that in PyTorch or TensorFlow. It seems something specific to ONNX. Do you have more context of why it is needed besides it is defined in ONNX?

index.bs
wchao1115: steps == 0 should not be allowed. I'll add the clarification text on the argument.

index.bs
wchao1115: @gramalingam I did some research and I couldn't find variable length sequence supported in any other popular frameworks but ONNX. I looked at TensorFlow (and its JS sibling), PyTorch, CoreML and MXNet. No other framework supports this but ONNX. Do you remember why ONNX added this? Can we leave it for now for compactness?

index.bs
wchao1115: @gramalingam I think we're going to have to choose which semantic we want here, the [numpy](https://numpy.org/doc/stable/reference/arrays.indexing.html#basic-slicing-and-indexing)'s one or the [TensorFlow](https://js.tensorflow.org/api/latest/#slice)'s one. ONNX follows numpy's convention, which is what you suggested.

The TensorFlow's convention is `(start, length)` where `start` is 0-based and `length` is 1-based. Here `length = -1` means the remaining number of element to the end of the axis dimension.

The numpy's convention is `(start, end)` where both are 0-based indices, and both are allowed negative values. Here `end = -1` means `end = length - 1`, so a slice of `(0, -1)` selects all the elements in an axis except the last one. The drawback of this convention is that to select to the end of the axis dimension without knowing the dimension size (length *n*), one would need to pass an artificially big number like INT_MAX i.e. `(0, INT_MAX)`. That's fine but we now need to agree on what that big number maybe e.g. 32 or 64 bit, signed or unsigned, etc. Numpy has a benefit of a Python's missing argument idiom, so one could simply do `(0,)` as absence implies length *n*.

The `(0,-1)` use case of the TensorFlow's convention is quite handy and is used a lot in practice. I'm leaning more toward that convention. Note that PyTorch `narrow` operator follows the `(start, length)` convention as well but the spec does not explicitly state that it supports passing -1 as a size.

index.bs

index.bs
wchao1115: Please take a look. `initialHiddenState` is now optional.

index.bs
wchao1115: Added

index.bs
gramalingam: I agree using -1 to go until the end is convenient. If we treat `(start,end)` as being inclusive of `end`, we could support both features. But it could have the disadvantage of causing some confusion if people are used to particular interpretations.

index.bs
gramalingam: It possibly comes from CNTK.

index.bs
gramalingam: Either way: if the convention is different from ONNX, it will complicate translating between the two. Sigh! I guess we can't make it work for both Tensorflow and ONNX. A compromise is unavoidable. So, I am okay with the proposal

huningxin (approved):
Thanks for the update. Looks good to me.

PR #92, opened 2020-09-23
Replaced interface with partial interface in WebIDL

wonsuk73:
- For all of WebIDL definition in this spec, partial interface probably need to be used rather then interface according to WebIDL spec [1]
- Because the definition of an interface is separated over more than one section of the document
[1] https://heycam.github.io/webidl/

### DIFF
--- a/index.html
+++ b/index.html
@@ -1709,7 +1709,7 @@ so that the application loads the tiny model in the case of CPU-only devices.</p
 };
 </pre>
    <h3 class="heading settled" data-level="3.2" id="api-ml"><span class="secno">3.2. </span><span class="content">ML</span><a class="self-link" href="#api-ml"></a></h3>
-<pre class="idl highlight def"><c- b>interface</c-> <dfn class="dfn-paneled idl-code" data-dfn-type="interface" data-export id="ml"><code><c- g>ML</c-></code></dfn> {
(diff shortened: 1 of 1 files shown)


anssiko:
Thanks for reviewing this spec @wonsuk73! Please note the duplicate occurrences of the following interfaces you flagged are found in the [IDL Index section](https://webmachinelearning.github.io/webnn/#idl-index):

interface ML
interface Operand
interface Model
interface Compilation
interface Execution

The IDL Index section is Bikeshed-generated boilerplate that concatenates all IDL blocks from the spec together. That is a Bikeshed feature and not a bug :) I'll let spec editors @huningxin @wchao1115 confirm and check our IDL usage is correct in this regard.

We use partial interface extensively for NeuralNetworkContext that hosts the definitions of a number of operations, and that usage seems valid to me.

wonsuk73:
Thanks for quick review @anssiko. I think most of your comments are right. and I misunderstood some part of WebDIL spec.
But I think [interface NeuralNetworkContext](https://webmachinelearning.github.io/webnn/#neuralnetworkcontext) need to be  replaced with partial interface because that define part of NeuralNetworkContext interface. Am I right?

anssiko:
I think that is the first occurrence of the NeuralNetworkContext interface, so only subsequent occurrences of the same interface need partial. We should have those covered, but better doublecheck.

wonsuk73:
@anssiko Thanks! You are completely right! I think I need to withdraw this PR.

PR #96, opened 2020-09-30
void -> undefined for methods of 3.8. Execution

wonsuk73:
- Regarding to [Web IDL](https://heycam.github.io/webidl/), void should be changed with undefined.
- Additionally you can refer this, [void -> undefined, and allow it anywhere] (https://github.com/heycam/webidl/pull/906/)


wonsuk73:
@huningxin @wchao1115 @anssiko could you review this?

anssiko:
@wonsuk73 could you update index.bs and generate index.html per https://github.com/webmachinelearning/webnn#generating-the-specification and update this PR accordingly including both the index.bs and index.html changes in the commit? Thanks!

wonsuk73:
@anssiko Thanks for your comment! I generated an index.html file using bikeshed as you said. But I think it updated the part of css of index.html file. So when I render index.html in chrome, spec's background color is black. It's strange. It's okay? or I need to do something else?

wonsuk73:
I would like to close this commit and re-commit index.bs and index.html through bikeshed with new PR.

PR #97, opened 2020-10-01
Adding extended attributes to Web IDL definition

wonsuk73:
- Web NN API is to access the hardware acceleration for machine learning. So it is only need to be exposed in secure contexts. It means that this API is only available over HTTPS.


wonsuk73:
@huningxin @wchao1115 @anssiko could you review and share your view?

anssiko:
@wonsuk73 similarly, could you including both the index.bs and index.html changes in the commit? Thanks!

Instructions: https://github.com/webmachinelearning/webnn#generating-the-specification

wonsuk73:
I would like to close this commit and re-commit index.bs and index.html through bikeshed with new PR.

PR #95, opened 2020-09-28
Add clamp and split

huningxin:
fix https://github.com/webmachinelearning/webnn/issues/66#issuecomment-699670404

@wchao1115 @pyu10055 @gramalingam @anssiko , PTAL. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/95.html" title="Last updated on Oct 9, 2020, 3:00 AM UTC (8c18b5c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/95/65772f1...huningxin:8c18b5c.html" title="Last updated on Oct 9, 2020, 3:00 AM UTC (8c18b5c)">Diff</a>

PR #101, opened 2020-10-11
Added extended attributes to Web IDL definition

wonsuk73:
- Web NN API is to access the hardware acceleration for machine learning. So it is only need to be exposed in secure contexts. It means that this API is only available over HTTPS.
- removed "Models" from biblio because it isn't unused any more

***
<a href="https://pr-preview.s3.amazonaws.com/wonsuk73/webnn/pull/101.html" title="Last updated on Oct 13, 2020, 1:21 PM UTC (20dc95c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/101/d070638...wonsuk73:20dc95c.html" title="Last updated on Oct 13, 2020, 1:21 PM UTC (20dc95c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -181,12 +181,13 @@ API {#api}
 ## Navigator ## {#api-navigator}
 <script type=idl>
 partial interface Navigator {
-  readonly attribute ML ml;
+  [SecureContext] readonly attribute ML ml;
 };
 </script>
 
 ## ML ## {#api-ml}
 <script type=idl>
+[SecureContext, Exposed=Window]
 interface ML {
   NeuralNetworkContext getNeuralNetworkContext();
 };
@@ -220,6 +221,7 @@ dictionary OperandDescriptor {
 
 ## Operand ## {#api-operand}
 <script type=idl>
+[SecureContext, Exposed=Window]
 interface Operand {};
 </script>
 
@@ -228,6 +230,7 @@ interface Operand {};
 The {{NeuralNetworkContext}} interface represents a global state of neural network compute workload and execution processes.
 
 <script type=idl>
+[SecureContext, Exposed=Window]
 interface NeuralNetworkContext {
   ModelBuilder createModelBuilder();
 };
@@ -240,6 +243,7 @@ The {{ModelBuilder}} interface defines a set of operations as identified by the 
 <script type=idl>
 typedef record<DOMString, Operand> NamedOperands;
 
+[SecureContext, Exposed=Window]
 interface ModelBuilder {
   // Create an operand that represents a model input.
   Operand input(DOMString name, OperandDescriptor desc);
@@ -1093,6 +1097,7 @@ dictionary CompilationOptions {
   PowerPreference powerPreference = "default";
 };
 
+[SecureContext, Exposed=Window]
 interface Model {
   Promise<Compilation> compile(optional CompilationOptions options = {});
 };
@@ -1115,6 +1120,7 @@ dictionary Output {
 typedef record<DOMString, Input> NamedInputs;
 typedef record<DOMString, Output> NamedOutputs;
 
+[SecureContext, Exposed=Window]
 interface Compilation {
   Promise<NamedOutputs> compute(NamedInputs inputs, optional NamedOutputs outputs);
 };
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1153,12 +1159,6 @@ Benjamin Poulain for their contributions to the API specification.
 
 <pre class="biblio">
 {
-  "Models": {
-      "href": "https://github.com/webmachinelearning/webnn/blob/master/op_compatibility/first_wave_models.md",
-      "title": "The first-wave models",
-      "authors": ["Machine Learning for the Web Community Group"],
-      "date": "2020"
-  },
```
huningxin, line 1161: Thanks @wonsuk73 for raising this issue. I also found it when rebasing #95 . My fix is to refer the models instead of deleting it, please take a look at https://github.com/webmachinelearning/webnn/pull/95/commits/ee9f4405e58479ee2c9ab2571da45d2a6ba7b8d0.

huningxin (commented):
The extended attributes look good to me. Thanks @wonsuk73 !

index.bs
wonsuk73: @huningxin Thanks for your comment. Sorry that I didn't check your rebasing work^^; There is the reference for Models then we have to keep this.
But I don't know how can I bring this modification out from my PR because we use Bikteshed. Is there any solution to handle this?

huningxin: If you are fine the fix within #95, we can merge that PR first. Then you can rebase this PR and eliminate this change. Does it sound good to you?

wonsuk73: @huningxin I completely agree with your suggestion. But I don't know how can I rebase my PR for both files. Could you guide this? Thanks in advance!

huningxin: Sure. As #95 merged, you can rebase on top of the master branch. The index.bs should be fine. The index.html would have conflict, you can ignore that and finish the rebase. After that, you can regenerate index.html by running bikeshield. Then you may push the new index.bs and index.html into your branch. You may need `-f` to force update your branch as its base changes.

huningxin: My steps for your reference:
1. `git fetch origin`: fetch the latest origin/master
2. `git checkout <your_branch>`
3. `git rebase origin/master`: the index.html would have conflicts
4. `git restore --staged index.html`
5. `git restore index.html`: discard the index.html
6. `git restore --staged index.bs`
7. edit index.bs and revert the removing the models from biblio
8. `git add index.bs`
9. `git rebase --continue`
10. `make`: regenerate the index.html
11. `git add index.html`
12. `git commit`
13. `git push <your_remote> <your_branch>`

wonsuk73: @huningxin Thanks a lot for your very kind guide. I will try to do that with your guide! :)

wonsuk73: @huningxin When I push after rebase, I got error below. Is it correct?
remote: Permission to webmachinelearning/webnn.git denied to wonsuk73. fatal: unable to access 'https://github.com/webmachinelearning/webnn.git/': The requested URL ret
urned error: 403"

If this process is complicated, then I might withdraw this PR and re-do with new PR. I think this is simple way to do that :)

huningxin: > When I push after rebase, I got error below. Is it correct?

You need to push to your own forked repo (https://github.com/wonsuk73/webnn.git) and the branch (master) for this PR.

> If this process is complicated, then I might withdraw this PR and re-do with new PR. I think this is simple way to do that :)

This also works. Thanks!

wonsuk73: @huningxin Thanks for comment! I did push to my forked repo. It seems it works! Could you review this? :)

huningxin: Great. It looks good. Thanks.

huningxin (approved):

wchao1115 (approved):

wonsuk73:
@huningxin This PR is conflict again. So could you give a guide for this? I might need to do the same work I did before. right?

huningxin:
> @huningxin This PR is conflict again. So could you give a guide for this? I might need to do the same work I did before. right?

Please resolve the conflicts as you did before. It would be easier. Since #113 implements a deploy script that will generate the index.html, you can simply drop your changes of index.html.

wonsuk73:
@huningxin, there is errors as below. It might be occurred because of the change of branches. So I would like to withdraw this PR and make a new PR.
error: src refspec — does not match any
error: src refspec force does not match any

Issue #90, opened 2020-09-17
Suggestion: use an options dictionary for functions with lots of args


zolkis:
A syntax-only question/suggestion. There may be reasons for the current design (disclaimer: I am long C) but wouldn't the JavaScript code look better with shorter signature containing the operands and the rest would be named in a dictionary?

For instance, now there is 

```javascript 
partial interface NeuralNetworkContext {
  Operand gemm(Operand a, Operand b, optional Operand c, 
               optional float alpha = 1.0, optional float beta = 1.0, 
               optional boolean aTranspose = false, optional boolean bTranspose = false);
};
```
and it could be also written (following the pattern used in many other web APIs) as

```javascript
partial interface NeuralNetworkContext {
  Operand gemm(Operand a, Operand b, optional Operand c, optional GemmOptions options);
};

dictionary GemmOptions {
   float alpha = 1.0;
   float beta = 1.0; 
   boolean aTranspose = false; 
   boolean bTranspose = false;
};
```
while the code using defaults would stay the same, and code using non-defaults would look more clear:

```javascript
let r = nn.gemm(a, b, c);

let q = nn.gemm(a, b, c, { alpha: 1.1, beta: 1.1, aTranspose:  true });
```
as opposed to 
```javascript
let q = nn.gemm(a, b, c, 1.1, 1.1, true);
```
Maybe not so big improvement for `gemm`, but bigger opportunities for the likes of `gru` ops, improving clarity of the function semantics and code readability as well.

I am aware this doesn't add much value to counter the arguments for the original design, if there is more to that than to make it look like a C API :).

wchao1115:
The suggestion looks reasonable. The only concern I have would be that if we apply this style of API to `gemm` we would probably need to deliver the same treatment across all other API for consistency, which would likely introduce many more *options* types. Secondly, it might be a bit ambiguous to differentiate what property should be in the options and what not.

anssiko:
I discussed this proposal with the editors and there is agreement this proposal has merit.

In order to make timely progress, and given no call next week, I mandate the editors to work with @zolkis to craft a PR for the group to review.

@zolkis introduced this proposal on 1 Oct call and we recorded no concerns at that time.

wchao1115:
Fixed with PR #112.

PR #112, opened 2020-10-26
Use optional dictionaries for operator's optional parameters.

wchao1115:
Addressing issue #90. Using optional dictionaries simplifies the calling code and allows the API to evolve more easily over time since adding a new optional parameter would not require adding a new method to the model builder interface.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/112.html" title="Last updated on Oct 27, 2020, 6:06 AM UTC (9e71e31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/112/77ff4ee...9e71e31.html" title="Last updated on Oct 27, 2020, 6:06 AM UTC (9e71e31)">Diff</a>


index.bs
```diff
@@ -258,36 +258,38 @@ interface ModelBuilder {
...
+  long axis = 1;
+  float epsilon = 1e-5;
+};
+
 partial interface ModelBuilder {
-  Operand batchNormalization(Operand input, Operand mean, Operand variance, 
-                             optional Operand scale, optional Operand bias, 
-                             optional long axis = 1, optional float epsilon = 1e-5);
+  Operand batchNormalization(Operand input, Operand mean, Operand variance,
+                             optional BatchNormalizationOptions options = {});
```
huningxin, line 270: The usage of `scale`, `bias`, `axis` and `epsilon` in the sample code of `batchNormalization` would be changed to `options.scale`, `options.bias`, `options.axis` and `options.epsilon`.

index.bs
```diff
@@ -309,15 +311,21 @@ partial interface ModelBuilder {
...
 Clamp the input tensor element-wise within a range specified by the minimum and maximum values.
 <script type=idl>
+dictionary ClampOptions {
+  Operand minValue;
+  Operand maxValue;
+};
+
 partial interface ModelBuilder {
-  Operand clamp(Operand x, optional Operand minValue, optional Operand maxValue);
+  Operand clamp(Operand x, optional ClampOptions options = {});
```
huningxin, line 320: ditto, the sample code would be changed.

index.bs
```diff
@@ -494,31 +497,38 @@ partial interface ModelBuilder {
...
+  float beta = 1.0;
+  boolean aTranspose = false;
+  boolean bTranspose = false;
+};
+
 partial interface ModelBuilder {
-  Operand gemm(Operand a, Operand b, optional Operand c, 
-               optional float alpha = 1.0, optional float beta = 1.0, 
-               optional boolean aTranspose = false, optional boolean bTranspose = false);
+  Operand gemm(Operand a, Operand b, optional GemmOptions options = {});
```
huningxin, line 509: ditto for sample code.

anssiko (approved):
Thank you @wchao1115!

@zolkis PTAL, this fixes #90.

zolkis:
Nice, thanks! That was a lot to figure out on where to draw the line for options, for each algorithm in particular. Looks much better.

index.bs
wchao1115: Do you mean samples in webnn_samples repo? I believe I've updated all the samples in this spec and in the explainer to reflect this change. The webnn_samples change would be on a separate PR to that repo.

wchao1115:
> Nice, thanks! That was a lot to figure out on where to draw the line for options, for each algorithm in particular. Looks much better.

Thanks to your suggestion! This change simply moves every optional params to optional fields of separate dictionaries. It makes the API contract more resistant to changes, and also simplifies how the browsers would implement it in C++ classes. In practice it doesn't simplify the user's code as much though, but is still worth it from the other reasons mentioned.

huningxin:
> Thanks to your suggestion! This change simply moves every optional params to optional fields of separate dictionaries.

+1, This change allows developers not to specify all preceding optional parameters if they only want to specify some later ones. e.g. previously if developers just want to set layout to 'nhwc' of conv2d, they have to fill all preceding optional parameters with default values:
```js
builder.conv2d(input, filter, [0,0,0,0], [1,1], [1,1], 1, 'nhwc');
```
Now they can only set layout:
```js
builder.conv2d(input, filter, {layout: 'nhwc'});
```
And specifying the optional parameter name also improves the code readability.

index.bs
huningxin: I mean the sample/pseudo code of batchNormalization in the spec
```js
let shape = [1,-1,1,1];
return builder.add(
    builder.mul(
      builder.reshape(scale, shape),
      builder.div(
        builder.sub(input, builder.reshape(mean, shape)),
        builder.sqrt(builder.add(builder.reshape(variance, shape), builder.constant(epsilon)))
        )
      ),
    builder.reshape(bias, shape)
  );
```
Once the spec change landed, I can take care of the update of webnn-polyfill and webnn-samples.

index.bs
```diff
@@ -721,7 +744,7 @@ partial interface ModelBuilder {
             builder.mul(
               r,
               builder.add(
-                (recurrentBias ? builder.slice(recurrentBias, [2 * hiddenSize], [hiddenSize]) : zero),
+                (options.recurrentBias ? builder.slice(options.recurrentBias, [2 * hiddenSize], [hiddenSize]) : zero),
                 builder.matmul(
                   hiddenState, 
                   builder.transpose(builder.slice(recurrentweight, [2 * hiddenSize, 0], [hiddenSize, -1]))
```
huningxin, line 750: There is a old typo, the `recurrentweight` should be `recurrentWeight` in this pseudo code. I found that in recent implementation of gru in webnn-polyfill.

index.bs
wchao1115: Fixed.

index.bs
wchao1115: Fixed

index.bs
wchao1115: Fixed.

index.bs
wchao1115: BTW, can we just delete index.html from the repo since we already build it off index.bs? That way we don't need to keep updating the file manually for every change we make.

index.bs

index.bs
huningxin: +1. We can use github actions to build the index.html and deploy it to gh-pages branch.

huningxin (approved):

index.bs
huningxin: Created #113 for this. Please take a look.

wchao1115: Thanks @huningxin

PR #114, opened 2020-10-27
Add instructions for PRs

anssiko:


huningxin (approved):
Thanks @anssiko !

wchao1115 (approved):

anssiko:
Thanks!

We as a group want to encourage direct contributions to the spec from all participants. I hope these concrete instructions on how to do that lowers the threshold for doing so. If there are any questions, folks can always reach out to me or the spec editors for advise.

PR #117, opened 2020-10-31
Fix the gru pseudo sample code

huningxin:
It should reshape the 2-D output of gruCell to 3-D tensor.

The fix is verified by the webnn-polyfill gru implementation: https://github.com/huningxin/webnn-polyfill/blob/gru/src/nn/ops/gru.ts#L113

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/117.html" title="Last updated on Oct 31, 2020, 5:11 AM UTC (bcdb16b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/117/3dd5bbf...huningxin:bcdb16b.html" title="Last updated on Oct 31, 2020, 5:11 AM UTC (bcdb16b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -634,11 +634,13 @@ partial interface ModelBuilder {
         let slice = (slot == 1 || options.direction == "backward" ? steps - step - 1 : step);
         let cellInput = builder.squeeze(builder.slice(input, [slice, 0, 0], [1, -1, -1]), { axes: [0] });
 
-        let result = builder.gruCell(
-          cellInput, cellWeight[slot], cellRecurrentWeight[slot], 
-          cellHidden[slot], hiddenSize, { bias: cellBias[slot], 
-          recurrentBias: cellRecurrentBias[slot], resetAfter: options.resetAfter, 
-          layout: options.layout, activations: options.activations });
+        let result = builder.reshape(
+          builder.gruCell(
+            cellInput, cellWeight[slot], cellRecurrentWeight[slot],
+            cellHidden[slot], hiddenSize, { bias: cellBias[slot],
+            recurrentBias: cellRecurrentBias[slot], resetAfter: options.resetAfter,
+            layout: options.layout, activations: options.activations }),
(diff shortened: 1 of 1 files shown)


wchao1115 (commented):
`gruCell` returns 2D tensor of shape `[batch_size, hidden_size]`. In this line `concat` is there to merge up to 2 results depending on the `numDirections` value. In the subsequent lines inside the `if (options.returnSequence)` block, the concatenated cell result will be reshaped to 4D and merged with each step result up to `steps` value. I don't see why the existing pseudo sample code would be wrong. 

The proposed change would instead incorrectly flatten the shape of the cell result to a batch size of 1.

huningxin:
> I don't see why the existing pseudo sample code would be wrong.

The problem is when `steps` > 1, the concatenated 2D output `cellOutput` will be assigned to `hiddenState`. Then the for-loop will feed this 2D `hiddenState` to `builder.slice(hiddenState, [slot, 0, 0], [1, -1, -1])` which expects a 3D tensor and throws an error.


> The proposed change would instead incorrectly flatten the shape of the cell result to a batch size of 1.

The change would reshape the `result` in shape [1, batch, hidden_size], after `concat`, the `cellOutput` will be in shape [numDirections, batch, hidden_size]`.

The issue and the fix were verified by [the test case of gru with 2 steps](https://github.com/huningxin/webnn-polyfill/blob/gru/test/ops/gru.js#L96) in my local webnn-polyfill (I'll send the PR soon).

wchao1115:
Thanks @huningxin I got it now. You're absolutely right. The sample misses adjusting the output of `gruCell` to incorporate the batch size in a 3D tensor, which later incorrectly updates the `hiddenState` for the next step in the sequence. Your fix looks good.

wchao1115 (approved):

huningxin:
Thanks for review and approval!

PR #121, opened 2020-11-09
Correct mismatched references in NSNet baseline sample

wchao1115:
Correct mismatched references between the original version of NSNet and the link to the revisioned NSNet2 in the scenario section. Adjust some wordings in the alternatives section.


anssiko (approved):

huningxin (commented):
Looks good, thanks!

Issue #108, opened 2020-10-21
WebNN API is not support Mirrorpad, SquaredDifference, Pow, TransposeConv


calvinlcchen:
As a use case about style transfer, source is [https://intel.github.io/webml-polyfill/examples/style_transfer/?prefer=none&b=WebGL&m=fast_style_transfer_onnx&s=image&d=0&f=WebNN,](https://intel.github.io/webml-polyfill/examples/style_transfer/?prefer=none&b=WebGL&m=fast_style_transfer_onnx&s=image&d=0&f=WebNN,)
Some ops style transfer used is not be supported in current WebML API spec.
Please add following ops to support list. thanks.
mirrorpad, SquaredDifference, Pow, TransposeConv  for tflite format models.
mirrorpad, Pow, ConvTranspose for onnx format models.

huningxin:
According to [29 Oct CG call](https://www.w3.org/2020/10/29-webmachinelearning-minutes.html#t03),  this is related to [the style transfer use case](https://webmachinelearning.github.io/webnn/#usecase-style-transfer) and it would be good to support. Thanks for the proposal @calvinlcchen !  

> As a use case about style transfer, source is https://intel.github.io/webml-polyfill/examples/style_transfer/?prefer=none&b=WebGL&m=fast_style_transfer_onnx&s=image&d=0&f=WebNN,

The models are based on [Fast Style Transfer in TensorFlow](https://github.com/lengstrom/fast-style-transfer) that is based on "a combination of Gatys' [A Neural Algorithm of Artistic Style](https://arxiv.org/abs/1508.06576), Johnson's [Perceptual Losses for Real-Time Style Transfer and Super-Resolution](http://cs.stanford.edu/people/jcjohns/eccv16/), and Ulyanov's [Instance Normalization](https://arxiv.org/abs/1607.08022)." 

The `InstanceNormalization` could be defined as a high-level function that can be decomposed to `add`, `sub`, `mul`, `div`, `reduceMean` and `sqrt`. (The `SquaredDifference` and `Pow` are used to convert `InstanceNormalization`, we can skip them).

The `mirroredpad` can be supported by defining a general `pad` operation with `symmetric` padding mode.

The `TransposeConv` of tflite model and `ConvTranspose` of onnx model can be supported by defining a `convTranspose2d`

[...]

PR #123, opened 2020-11-28
Support style-transfer models with changes/additions to the following operations

wchao1115:
#108 
- Extend the `conv2d` operation to support transposed convolution, an essential upsample tool for encoder-decoder models.
- Add `instanceNormalization` in addition to batch-normalization. Instance-normalization is a fused operation for a normalization subgraph that computes the mean and variance values per-feature instance on the fly.
- Replace the `sqrt` unary operation with a more generic `pow` binary operation used by the normalization process.
- Add `pad` operation that supports all 4 padding modes found in various frameworks.
- Add `resample` operation to support both upsampling and downsampling of feature instances. This operation is used in the ONNX version of the style-transfer models.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/123.html" title="Last updated on Dec 16, 2020, 5:09 AM UTC (5601203)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/123/c028ff1...5601203.html" title="Last updated on Dec 16, 2020, 5:09 AM UTC (5601203)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -260,7 +260,7 @@ interface ModelBuilder {
 </script>
 
 ### batchNormalization ### {#api-modelbuilder-batchnorm}
-Normalize the tensor values across dimensions in a batch through a specialized [[BatchNorm]] [transform](https://en.wikipedia.org/wiki/Batch_normalization#Batch_Normalizing_Transform). The *mean* and *variance* tensors are previously calculated during model training pass.
+Normalize the tensor values of input features across the batch dimension using [[Batch-Normalization]]. For each input feature, the mean and variance values of that feature are computed across the batch dimension during training.
 <script type=idl>
 dictionary BatchNormalizationOptions {
   Operand scale;
@@ -277,33 +277,32 @@ partial interface ModelBuilder {
 <div algorithm=batchnorm>
     **Arguments:**
         - *input*: an {{Operand}}. The input N-D tensor.
-        - *mean*: an {{Operand}}. The 1-D tensor of the batch mean values 
-            whose length is equal to the size of the input dimension denoted by *options.axis*.
-        - *variance*: an {{Operand}}. The 1-D tensor of the batch variance values
-            whose length is equal to the size of the input dimension denoted by *options.axis*.
+        - *mean*: an {{Operand}}. The 1-D tensor of the mean values of the input features across the batch whose length is equal to the size of the input dimension denoted by *options.axis*.
+        - *variance*: an {{Operand}}. The 1-D tensor of the variance values of the input features across the batch whose length is equal to the size of the input dimension denoted by *options.axis*.
         - *options*: an optional {{BatchNormalizationOptions}}. The optional parameters of the operation.
               - *scale*: an {{Operand}}. The 1-D tensor of the scaling values whose length is equal to the size of the input dimension denoted by *options.axis*.
               - *bias*: an {{Operand}}. The 1-D tensor of the bias values whose length is equal to the size of the input dimension denoted by *options.axis*.
-              - *axis*: a {{long}} scalar. The index for the feature dimension of the input shape for which the normalizing batch is defined and that the mean and variance values are computed. When it's not specified, the default value is 1.
-              - *epsilon*: a {{float}} scalar. The Epsilon value to prevent computational error due to divide-by-zero. The default value is 0.00001 when not specified.
+              - *axis*: a {{long}} scalar. The index to the feature count dimension of the input shape for which the mean and variance values are. When it's not specified, the default value is 1.
+              - *epsilon*: a {{float}} scalar. A small value to prevent computational error due to divide-by-zero. The default value is 0.00001 when not specified.
         
-    **Returns:** an {{Operand}}. The batch normalized N-D tensor of the same shape as the input tensor.
+    **Returns:** an {{Operand}}. The batch-normalized N-D tensor of the same shape as the input tensor.
 
-    When *input* is a 4-D tensor of the *"nchw"* or *"nhwc"* layout, *axis* should be set to 1 or 3 respectively, 
-    to designate the feature dimension of the input tensor.
+    When *input* is a 4-D tensor of the *"nchw"* or *"nhwc"* layout, *options.axis* should be set to 1 or 3 respectively. The axis value designates the feature or channel count dimension of the input tensor.
 
     <div class="note">
     The behavior of this operation when the input tensor is 4-D of the *"nchw"* layout can be generically emulated from 
     the usage of other operations as follow. However, user agents typically have a more efficient implementation for it, 
     therefore its usage is encouraged from the performance standpoint.
     <pre highlight="js">
-    let shape = [1,-1,1,1];
+    const shape = [1,-1,1,1];
     return builder.add(
         builder.mul(
           builder.reshape(options.scale, shape),
           builder.div(
             builder.sub(input, builder.reshape(mean, shape)),
-            builder.sqrt(builder.add(builder.reshape(variance, shape), builder.constant(options.epsilon)))
+            builder.pow(
+              builder.add(builder.reshape(variance, shape), builder.constant(options.epsilon)),
+              builder.constant(0.5))
             )
           ),
         builder.reshape(options.bias, shape)
@@ -384,8 +383,10 @@ Compute a 2-D convolution given 4-D input and filter tensors
 <script type=idl>
 dictionary Conv2dOptions {
   sequence<long> padding;
+  sequence<long> outputPadding;
   sequence<long> strides;
   sequence<long> dilations;
+  boolean transpose = false;
   long groups = 1;
   OperandLayout layout = "nchw";
 };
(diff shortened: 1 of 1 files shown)


anssiko (approved):
Thanks! Added to the [2020-12-10 agenda](https://github.com/webmachinelearning/meetings/blob/master/telcons/2020-12-10-agenda.md) for discussion.

As usual, if we get all the reviews completed ahead the call we can merge this earlier than that and recap the changes on the call.

wchao1115:
@gramalingam

index.bs
```diff
@@ -788,6 +801,63 @@ partial interface ModelBuilder {
     </div>
 </div>
 
+### instanceNormalization ### {#api-modelbuilder-instancenorm}
+Normalize the input features per feature instance using [[Instance-Normalization]]. Unlike [[#api-modelbuilder-batchnorm]] where the mean and variance values are computed across the batch dimension during training, the mean and variance values of instance normalization is computed from each individual feature instance on the fly.
+<script type=idl>
+dictionary InstanceNormalizationOptions {
+  Operand scale;
```
gramalingam, line 808: I wonder if it would be better to not allow "Operands" in "Options", and make "Operands" explicit (even if optional). I also think it is useful to restrict "Options" to values that are known at the graph-building time (while "Operands" represent values known at run-time).

wchao1115: @gramalingam That's a good observation and I have thought about this exact point but ultimately went for the idea that any optional params, whether it's an operand or not, ought to be placed in the options struct because the original motivation for this is to simplify future changes to the API so that no overload will ever be needed in the future. And, that a future change may end up being an additional operand. 

So even if we are willing to draw an arbitrary line right now barring a non-compile time param from entering the options struct, we will be forced to break that rule anyway in the future at the very first instance of the need to extend the API to support an additional optional operand.

index.bs
```diff
@@ -402,8 +403,10 @@ partial interface ModelBuilder {
...
             - *strides*: a sequence of {{long}} of length 2. The stride of the sliding window for each spatial dimension of *input*, [stride_height, stride_width]. If not present, the values are assumed to be [1,1].
             - *dilations*: a sequence of {{long}} of length 2. The dilation factor for each spatial dimension of *input*, [dilation_height, dilation_width]. If not present, the values are assumed to be [1,1].
+            - *transpose*: a {{boolean}} indicating that a transposed convolution operation is performed. Transposed convolution is used in upsampling networks to increase the resolution of a feature as opposed to the typical convolution process that reduces the feature's resolution. When transposed convolution is performed, *options.outputPadding* may be needed to disambiguate the output tensor shape. If not present, this option is assumed to be false.
```
huningxin, line 409: I am not sure whether the transposed convolution needs to allow specifying the output shape explicitly. For example, [`conv2d_transpose`](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_transpose) of TensorFlow uses `output_shape` and [`ConvTranspose`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvTranspose) supports both `output_padding` and `output_shape`. It would require the input size when calculating the output padding from output shape. However, the input size might not be known at the graph build time.

wchao1115: We only need either the output paddings or the output shape and never both. This is true for TensorFlow, PyTorch, and ONNX. The output paddings are values specified by the model's author and not inferred from the input shape. And since the output shape can be fully inferred at runtime for the majority case of non-transposed convolutions, I'm more inclined to define only the output padding values as an optional param needed only for transposed convolutions with stride values greater than 1.

huningxin: > We only need either the output paddings or the output shape and never both.

That's correct.

> And since the output shape can be fully inferred at runtime for the majority case of non-transposed convolutions,

I mean the output shape of transposed convolution.

For example, in the ONNX port of fast-transfer [models](https://github.com/microsoft/Windows-Machine-Learning/tree/master/Samples/FNSCandyStyleTransfer/UWP/cs/Assets), the ConvTranspose op sets `output_shape` attribute, e.g. [722, 722]. If we only support output paddings, developers might have to calculate by themselves when building the graph. And according the formula `output padding = output size - ((input size - 1) * stride + filter size - beginning padding - ending padding)`, the input size might need to be inferred from previous layers.

index.bs
```diff
@@ -991,6 +1129,34 @@ partial interface ModelBuilder {
...
+Resample the tensor values from the source to the destination dimensions according to the scaling factors.
+<script type=idl>
+enum InterpolationMode {
+  "nearest-neighbor",
+  "linear"
+};
+
+dictionary ResampleOptions {
+  InterpolationMode mode = "nearest-neighbor";
+  sequence<float> scales;
```
huningxin, line 1142: I am not sure whether it needs to allow specifying the new size. With that developers would be able to set either scales or size for when working with different frameworks.

index.bs
wchao1115: I added `outputSizes` as an alternative to `outputPadding` for transposed convolution as suggested here. Please take a look.

index.bs
```diff
@@ -788,6 +801,63 @@ partial interface ModelBuilder {
...
+    **Arguments:**
+        - *input*: an {{Operand}}. The input N-D tensor.
+        - *options*: an optional {{InstanceNormalizationOptions}}. The optional parameters of the operation.
+              - *scale*: an {{Operand}}. The 1-D tensor of the scaling values whose length is equal to the size of the input dimension denoted by *options.axis*.
+              - *bias*: an {{Operand}}. The 1-D tensor of the bias values whose length is equal to the size of the input dimension denoted by *options.axis*.
+              - *epsilon*: a {{float}} scalar. A small value to prevent computational error due to divide-by-zero. The default value is 0.00001 when not specified.
+        
+    **Returns:** an {{Operand}}. The instance-normalized N-D tensor of the same shape as the input tensor.
+
+    When *input* is a 4-D tensor of the *"nchw"* or *"nhwc"* layout, *options.axis* should be set to 1 or 3 respectively. The axis value designates the feature or channel count dimension of the input tensor.
```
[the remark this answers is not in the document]
wchao1115: I added the `layout` param instead, since both the feature and spatial dimensions are needed to be identified for the operation. Also limit the input to only 4-D as opposed to N-D; both CoreML and ONNX only support 4-D.

index.bs
```diff
@@ -788,6 +801,63 @@ partial interface ModelBuilder {
     </div>
 </div>
 
+### instanceNormalization ### {#api-modelbuilder-instancenorm}
+Normalize the input features per feature instance using [[Instance-Normalization]]. Unlike [[#api-modelbuilder-batchnorm]] where the mean and variance values are computed across the batch dimension during training, the mean and variance values of instance normalization is computed from each individual feature instance on the fly.
```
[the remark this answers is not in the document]
wchao1115: Fixed

index.bs
wchao1115: I think the target sizes can be easily converted to `scales`. It could be redundant to do both.

index.bs
huningxin: looks good now, thanks!

index.bs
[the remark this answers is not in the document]
huningxin: looks good, thanks!

index.bs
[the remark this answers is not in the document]
huningxin: thanks

index.bs
huningxin: I suppose the usage of target sizes is similar to the `outputSizes` of transposed conv2d. It is used when the input size is unknown when graph building time.

wchao1115: Good point. The input to `resample` can indeed be a free dimension input to the model. I added the `sizes` option as an alternative to `scales` as suggested.

huningxin: thanks, looks good now.

huningxin (approved):
Looks good, thanks for your great efforts @wchao1115 !

index.bs
gramalingam: "Any optional param ought to be placed in the options struct" seems a bit extreme. In the end, either is functionally equivalent, and it is mostly a matter of style as to which is more readable. I personally prefer making the operand explicit, but am okay with this

gramalingam:
One general question relating to padding in convolution etc.: ONNX and TF have options like SAME for padding. I don't see something similar here, is there a way to achieve that?

wchao1115:
> One general question relating to padding in convolution etc.: ONNX and TF have options like SAME for padding. I don't see something similar here, is there a way to achieve that?

Algorithmic padding can be handled by the padding array. Here, the SAME padding simply means whether the padding array will pad to the left (lower end) or to the right (upper end) of the input spatial dimensions. ONNX has more flexibility in specifying which side of the input dimensions is prioritized to receive the padding while TensorFlow *SAME* padding has a fixed *pad upper* logic.

That said if the input dimensions aren't known at graph building time (in models with free dimensions), then obviously the padding array can't be calculated ahead of time either, similar to the argument that @huningxin made earlier in the thread about the need for `outputSizes` in the transposed case. We can add an `autoPad` option to mitigate that issue at runtime.

wchao1115:
@gramalingam Can you please take a look at the latest commit? Based on your feedback, an `autoPad` option is added to handle the case where the input shape isn't known at graph construction time.

PR #126, opened 2020-12-10
Use ML compute API of macOS/iOS in the explainer

huningxin:
[ML Compute](https://developer.apple.com/documentation/mlcompute) API is a computation graph based API that can accelerate ML frameworks on macOS and iOS, such as [Mac-optimized TensorFlow](https://blog.tensorflow.org/2020/11/accelerating-tensorflow-performance-on-mac.html).

Underlying, it supports both CPU and GPU, as the doc mentions

> ML Compute uses high performance BNNS primitives from the Accelerate framework for the CPU, and Metal Performance Shaders for the GPU. 

From the architecture perspective, it looks like closer to WebNN mapping, so I propose to use ML compute in the explainer.

@anssiko @wchao1115 , please take a look. CC @BenjaminPoulain


anssiko (approved):

anssiko:
(To clarify, this is just informative text and as such does not bind implementers to use any particular platform API for their implementation of this Web API. As usual with web specs, implementation details such as this are out of scope.)

wchao1115 (approved):

huningxin:
Thanks for review and comments. Let' me merge this PR.

Issue #141, opened 2021-02-04
Prefix generic interface names?


anssiko:
> 11. A lot of the names are very generic (Operand, Compilation) - this feels like something we might want to prefix with something or synchronize with TC39 about.

via https://github.com/w3ctag/design-reviews/issues/570#issuecomment-768875996

(WebIDL namespaces https://heycam.github.io/webidl/#idl-namespaces for  regular operations and read only regular attributes only.)

cynthia:
I thought a bit about this, and figured it would be worth to be on record stating that prefixing the names with anything that indicates this is only for neural networks is probably not a good idea. (It's a compute graph that can be used for other things too)

wchao1115:
I made a proposal in PR #149 to prefix all WebNN types with "ML", taking a cue from the `ML` attribute off the `Navigator` interface.

anssiko:
For spec archeology purposes, linking https://github.com/heycam/webidl/pull/425 here that added [`[LegacyNamespace]` extended attribute](https://heycam.github.io/webidl/#LegacyNamespace) to WebIDL to allow defining an interface in a namespace. As the "Legacy" in the name suggests, the use of this extended attribute is not encouraged at this time.

Issue #136, opened 2021-02-04
Constructor or builder pattern for model building?


anssiko:
> 5. Wouldn't it make sense to have a constructor rather than a builder pattern for createModelBuilder()? (e.g. new ModelBuilder(navigator.ml.getNNContext());

via https://github.com/w3ctag/design-reviews/issues/570#issuecomment-768875996

Early discussion suggests preference for the constructor. Some points brought up on the call today suggests we'll evaluate pros and cons here before committing.

wchao1115:
@cynthia @anssiko A key benefit of a factory method over concrete construction e.g. `X.createY()` as opposed to `new Y(X)` is that the instantiation of the concrete type of `Y` can be internally parameterized and hidden away from the caller. This generally makes the API more robust over time since the implementor of `X` is free to instantiate `Y` as it sees fit. More specifically if we need `Y` to have multiple specializations, exposing `Y` as a constructible concrete type would then require the caller to know the types of every specializations of `Y` as `Y` evolves. In the context of the WebNN API, we believe that the notion of the *model builder* could have multiple concrete specializations depending on its execution mode (e.g. eager vs. deferred), and so the factory method makes it more flexible to evolve it over time.

dontcallmedom:
for background, [design patterns for constructors in Web APIs](https://w3ctag.github.io/design-principles/#constructors)

wchao1115:
Thanks @dontcallmedom for the background on the Web Platform design guidelines.

According to the section concerning the [usage of constructor](https://w3ctag.github.io/design-principles/#constructors). It claims:

> This won’t be appropriate in all cases. For example:
> - Some objects represent access to privileged resources, so they need to be constructed by factory methods which can access those resources.

In our case, I could imagine an implementation of the `ModelBuilder` construction needing access to the underlying device encapsulated by but not publicly exposed through the `NeuralNetworkContext` interface, and so, `context.CreateModelBuilder()` would be a more flexible contract for the implementer. Hope I clarify enough of my concern on this topic.

@cynthia

wchao1115:
It appears that it should be safe for the implementer of webnn to internally cast the context type down to the underlying device. This addresses my concern of using factory method. We can use the webidl interface ctor to achieve the same goal. The PR #149 addresses this feedback.

Issue #135, opened 2021-02-04
getNeuralNetworkContext() and createModelBuilder() params


anssiko:
> 4. getNeuralNetworkContext() and createModelBuilder() seem strange (no parameters, for one thing) - is this expected to accept parameters/configs at some point? If so, we'd like to see what is intended here.

via https://github.com/w3ctag/design-reviews/issues/570#issuecomment-768875996

Per discussion the future eager mode would make use of the parameters.

wchao1115:
A more substantial design point about the context vs. the model builder is that the context holds the states that apply globally across topologies that would be subsequently created while the model builder only holds the states that apply specifically to the current topology being constructed. Examples of the context states are device contexts, resource allocators, graph compilers, managed resource caches, etc. Examples of builder states are current topologies, graph designated inputs, transient or cached compilation states, cached compiled states, etc. Regardless of whether the methods are parameterless, this distinction persists.

Potential parameterization of the context includes device types (e.g. CPU, GPU, NPU, adaptive, etc.) For the model builder includes builder's state management (e.g. auto-resetting topology once construction finished), deferred or immediate execution (eager mode), hints for free-dimension inputs, etc.

anssiko:
Nudge @cynthia :-)

cynthia:
While I understand the intent, the pattern itself seems somewhat different from patterns in the wild. Tensorflow is the only framework that has this kind of design, where the closure in which the compute graph is constructed uses a context manager  (e,g. with tf.device) - other implementations construct and "transfer". (Pytorch, MXNet, and PaddlePaddle all seem to do this.)

I'm imagining something like this:

    const m = new ModelBuilder();
    
    // Do things with m... all of this probably happens in CPU-land.
    
    let context = getContext({device: "cuda0"'}) // If this fails, fallback to a different context?
    m.setContext({context: context}); // Setup transfer
    m.compile(); // If this fails (e.g. not enough device memory for co-processor), then probably need to fallback

wchao1115:
Thanks @cynthia. We do not expose the notion of device explicitly in the API because in the context of the browser it could be challenging for the developers to know what type of device to target and which hardware platform the users will be having underneath.

[...]

PR #149, opened 2021-03-01
WebGL and WebGPU interops

wchao1115:
- Create context from external sources e.g. WebGLRenderingContext and WebGPU device (#135)
- Make power preference part of context creation options.
- Constant operands can be created from either WebGL or WebGPU buffers
- Model inputs and outputs can be bound with WebGL or WebGPU textures
- Prefix all types with "ML". Simplify "NeuralNetworkContext" to just "MLContext" (#141)
- Switch to use constructor for MLModelBuilder instead of factory method (#136)

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/149.html" title="Last updated on Mar 15, 2021, 6:06 AM UTC (c43ed52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/149/4db4c69...c43ed52.html" title="Last updated on Mar 15, 2021, 6:06 AM UTC (c43ed52)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -273,7 +273,7 @@ The {{MLModelBuilder}} interface defines a set of operations as identified by th
 typedef record<DOMString, MLOperand> MLNamedOperands;
 
 dictionary MLBuffer {
-  required (ArrayBufferView or WebGLBuffer or GPUBuffer) resource;
+  required (ArrayBuffer or WebGLBuffer or GPUBuffer) resource;
   unsigned long long offset = 0;
   unsigned long long size;
 };
@@ -692,7 +692,7 @@ partial interface MLModelBuilder {
     if (!hiddenState) {
       const desc = { type: 'float32', dimensions: [numDirections, 1, hiddenSize] };
       const totalSize = numDirections * hiddenSize;
-      hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0));
+      hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0).buffer);
     }
 
     let sequence = null;
@@ -1037,11 +1037,11 @@ partial interface MLModelBuilder {
     <pre highlight="js">
     // input: [[1,2,3], [4,5,6]]
     const input = builder.constant(
-      { type: 'float32', dimensions: [2,3] }, new Float32Array([1,2,3,4,5,6]));
+      { type: 'float32', dimensions: [2,3] }, new Float32Array([1,2,3,4,5,6]).buffer);
 
     // padding: [[1,1], [2,2]]
     const padding = builder.constant(
-      { type: 'float32', dimensions: [2,2] }, new Float32Array([1,1,2,2]));
+      { type: 'float32', dimensions: [2,2] }, new Float32Array([1,1,2,2]).buffer);
 
     // "constant" padded:
     //    [[0,0,0,0,0,0,0],
@@ -1367,7 +1367,7 @@ partial interface MLModelBuilder {
 </div>
 
 ## Compilation ## {#api-mlcompilation}
-The {{MLCompilation}} interface represents a compiled version of the computational graph of a {{MLModel}} according to the compilation options. A compilation of a graph once constructed is immutable and cannot be subsequently changed.
+The {{MLCompilation}} interface represents a compiled computational graph. A compilation of a graph once constructed is immutable and cannot be subsequently changed.
 
 <script type=idl>
 dictionary MLInput {
@@ -1422,14 +1422,14 @@ const builder = new MLModelBuilder(context);
 const desc = {type: 'float32', dimensions: TENSOR_DIMS};
 
 // constant1 is a constant MLOperand with the value 0.5.
-const constantBuffer1 = new Float32Array(TENSOR_SIZE).fill(0.5);
+const constantBuffer1 = new Float32Array(TENSOR_SIZE).fill(0.5).buffer;
 const constant1 = builder.constant(desc, constantBuffer1);
 
 // input1 is one of the input MLOperands. Its value will be set before execution.
 const input1 = builder.input('input1', desc);
 
 // constant2 is another constant MLOperand with the value 0.5.
-const constantBuffer2 = new Float32Array(TENSOR_SIZE).fill(0.5);
+const constantBuffer2 = new Float32Array(TENSOR_SIZE).fill(0.5).buffer;
 const constant2 = builder.constant(desc, constantBuffer2);
 
 // input2 is another input MLOperand. Its value will be set before execution.
@@ -1458,8 +1458,8 @@ const compilation = await builder.compile({'output': output});
 The following code executes the compiled model.
 <pre highlight="js">
 // Setup the input buffers with value 1.
-const inputBuffer1 = new Float32Array(TENSOR_SIZE).fill(1);
-const inputBuffer2 = new Float32Array(TENSOR_SIZE).fill(1);
+const inputBuffer1 = new Float32Array(TENSOR_SIZE).fill(1).buffer;
+const inputBuffer2 = new Float32Array(TENSOR_SIZE).fill(1).buffer;
 
 // Asynchronously execute the compiled model with the specified inputs.
 const inputs = {
@@ -1471,7 +1471,7 @@ const outputs = await compilation.compute(inputs);
 // Log the shape and computed result of the output operand.
 console.log('Output shape: ' + outputs.output.dimensions);
 // Output shape: 1,2,2,2
-console.log('Output value: ' + outputs.output.buffer);
+console.log('Output value: ' + new Float32Array(outputs.output.resource));
 // Output value: 2.25,2.25,2.25,2.25,2.25,2.25,2.25,2.25
 </pre>
 </div>
--- a/index.bs
+++ b/index.bs
@@ -187,20 +187,62 @@ partial interface Navigator {
 
 ## ML ## {#api-ml}
 <script type=idl>
+// WebGL specification:
+// (https://www.khronos.org/registry/webgl/specs/latest/2.0/#2.1)
+[SecureContext, Exposed=Window]
+interface WebGLRenderingContext {
+  // ...
+};
+
+// WebGPU specification:
+// (https://gpuweb.github.io/gpuweb/#gpu-device)
+[SecureContext, Exposed=Window]
+interface GPUDevice {
+  // ...
+};
+
+enum MLPowerPreference {
+  // Let the user agent decide the most suitable behavior
+  "default",
+  // Prioritizes execution speed over power consumption
+  "high-performance",
+  // Prioritizes power consumption over other considerations such as execution speed
+  "low-power"
+};
+
+dictionary MLContextOptions {
+  // Preference as related to power consumption
+  MLPowerPreference powerPreference = "default";
+};
+
 [SecureContext, Exposed=Window]
 interface ML {
-  NeuralNetworkContext getNeuralNetworkContext();
+  // Create a context with options
+  MLContext createContext(optional MLContextOptions options = {});
+
+  // Create a context from WebGL rendering context
+  MLContext createContext(WebGLRenderingContext glContext);
+
+  // Create a context from WebGPU device
+  MLContext createContext(GPUDevice gpuDevice);
 };
 </script>
 
-## OperandDescriptor ## {#api-operanddescriptor}
+## MLContext ## {#api-mlcontext}
+The {{MLContext}} interface represents a global state of neural network compute workload and execution processes.
 <script type=idl>
-enum InputOperandLayout {
+[SecureContext, Exposed=Window]
+interface MLContext {};
+</script>
+
+## MLOperandDescriptor ## {#api-mloperanddescriptor}
+<script type=idl>
+enum MLInputOperandLayout {
   "nchw",
   "nhwc"
 };
 
-enum OperandType {
+enum MLOperandType {
   "float32",
   "float16",
   "int32",
@@ -209,9 +251,9 @@ enum OperandType {
   "uint8"
 };
 
-dictionary OperandDescriptor {
+dictionary MLOperandDescriptor {
   // The operand type.
-  required OperandType type;
+  required MLOperandType type;
 
   // The dimensions field is only required for tensor operands.
   // The negative value means an unknown dimension.
@@ -219,73 +261,87 @@ dictionary OperandDescriptor {
 };
 </script>
 
-## Operand ## {#api-operand}
+## MLOperand ## {#api-mloperand}
 <script type=idl>
 [SecureContext, Exposed=Window]
-interface Operand {};
+interface MLOperand {};
 </script>
 
-## NeuralNetworkContext ## {#api-neuralnetworkcontext}
+## MLModelBuilder ## {#api-mlmodelbuilder}
 
-The {{NeuralNetworkContext}} interface represents a global state of neural network compute workload and execution processes.
+The {{MLModelBuilder}} interface defines a set of operations as identified by the [[#usecases]] that can be composed into a computational graph. It also represents the intermediate state of a graph building session.
 
 <script type=idl>
+// WebGL specification:
+// (https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.4)
 [SecureContext, Exposed=Window]
-interface NeuralNetworkContext {
-  ModelBuilder createModelBuilder();
+interface WebGLBuffer {
+  // ...
 };
-</script>
 
-## ModelBuilder ## {#api-modelbuilder}
+// WebGPU specification:
+// (https://gpuweb.github.io/gpuweb/#buffer-interface)
+[SecureContext, Exposed=Window]
+interface GPUBuffer {
+  // ...
+};
 
-The {{ModelBuilder}} interface defines a set of operations as identified by the [[#usecases]] that can be composed into a computational graph. It also represents the intermediate state of a graph building session.
-
-<script type=idl>
-typedef record<DOMString, Operand> NamedOperands;
+typedef record<DOMString, MLOperand> MLNamedOperands;
 
 [SecureContext, Exposed=Window]
-interface ModelBuilder {
-  // Create an operand that represents a model input.
-  Operand input(DOMString name, OperandDescriptor desc);
+interface MLModelBuilder {
+  // Construct the model builder from the context.
+  constructor(MLContext context);
 
-  // Create an operand that represents a model constant.
-  Operand constant(OperandDescriptor desc, ArrayBufferView value);
+  // Create an operand for a model input.
+  MLOperand input(DOMString name, MLOperandDescriptor desc);
+
+  // Create an operand for a model constant.
+  MLOperand constant(MLOperandDescriptor desc, ArrayBufferView bufferView);
+
+  // Create an operand for a model constant from a WebGL buffer.
+  MLOperand constant(MLOperandDescriptor desc, WebGLBuffer glBuffer, 
+                     optional long offset = 0, optional long size);
+
+  // Create an operand for a model constant from a WebGPU buffer.
+  MLOperand constant(MLOperandDescriptor desc, GPUBuffer gpuBuffer, 
+                     optional long offset = 0, optional long size);
 
   // Create a single-value operand from the specified number of the specified type.
-  Operand constant(double value, optional OperandType type = "float32");
+  MLOperand constant(double value, optional MLOperandType type = "float32");
 
   // Create a model that encapsulates the composition of operands by identifying the output operands.
-  Model createModel(NamedOperands outputs);
+  MLModel createModel(MLNamedOperands outputs);
 };
 </script>
 
-### batchNormalization ### {#api-modelbuilder-batchnorm}
+### batchNormalization ### {#api-mlmodelbuilder-batchnorm}
 Normalize the tensor values of input features across the batch dimension using [[Batch-Normalization]]. For each input feature, the mean and variance values of that feature supplied in this calculation as parameters are previously computed across the batch dimension of the input during the model training phrase of this operation.
 <script type=idl>
-dictionary BatchNormalizationOptions {
-  Operand scale;
-  Operand bias;
+dictionary MLBatchNormalizationOptions {
+  MLOperand scale;
+  MLOperand bias;
   long axis = 1;
   float epsilon = 1e-5;
 };
 
-partial interface ModelBuilder {
-  Operand batchNormalization(Operand input, Operand mean, Operand variance,
-                             optional BatchNormalizationOptions options = {});
+partial interface MLModelBuilder {
+  MLOperand batchNormalization(MLOperand input, MLOperand mean, MLOperand variance,
+                             optional MLBatchNormalizationOptions options = {});
 };
 </script>
 <div algorithm=batchnorm>
     **Arguments:**
-        - *input*: an {{Operand}}. The input N-D tensor.
-        - *mean*: an {{Operand}}. The 1-D tensor of the mean values of the input features across the batch whose length is equal to the size of the input dimension denoted by *options.axis*.
-        - *variance*: an {{Operand}}. The 1-D tensor of the variance values of the input features across the batch whose length is equal to the size of the input dimension denoted by *options.axis*.
-        - *options*: an optional {{BatchNormalizationOptions}}. The optional parameters of the operation.
-              - *scale*: an {{Operand}}. The 1-D tensor of the scaling values whose length is equal to the size of the input dimension denoted by *options.axis*.
-              - *bias*: an {{Operand}}. The 1-D tensor of the bias values whose length is equal to the size of the input dimension denoted by *options.axis*.
+        - *input*: an {{MLOperand}}. The input N-D tensor.
+        - *mean*: an {{MLOperand}}. The 1-D tensor of the mean values of the input features across the batch whose length is equal to the size of the input dimension denoted by *options.axis*.
+        - *variance*: an {{MLOperand}}. The 1-D tensor of the variance values of the input features across the batch whose length is equal to the size of the input dimension denoted by *options.axis*.
+        - *options*: an optional {{MLBatchNormalizationOptions}}. The optional parameters of the operation.
+              - *scale*: an {{MLOperand}}. The 1-D tensor of the scaling values whose length is equal to the size of the input dimension denoted by *options.axis*.
+              - *bias*: an {{MLOperand}}. The 1-D tensor of the bias values whose length is equal to the size of the input dimension denoted by *options.axis*.
               - *axis*: a {{long}} scalar. The index to the feature count dimension of the input shape for which the mean and variance values are. When it's not specified, the default value is 1.
               - *epsilon*: a {{float}} scalar. A small value to prevent computational error due to divide-by-zero. The default value is 0.00001 when not specified.
         
-    **Returns:** an {{Operand}}. The batch-normalized N-D tensor of the same shape as the input tensor.
+    **Returns:** an {{MLOperand}}. The batch-normalized N-D tensor of the same shape as the input tensor.
 
     When *input* is a 4-D tensor of the *"nchw"* or *"nhwc"* layout, *options.axis* should be set to 1 or 3 respectively. The axis value designates the feature or channel count dimension of the input tensor.
 
@@ -311,26 +367,26 @@ partial interface ModelBuilder {
     </div>
 </div>
 
-### clamp ### {#api-modelbuilder-clamp}
+### clamp ### {#api-mlmodelbuilder-clamp}
 Clamp the input tensor element-wise within a range specified by the minimum and maximum values.
 <script type=idl>
-dictionary ClampOptions {
-  Operand minValue;
-  Operand maxValue;
+dictionary MLClampOptions {
+  MLOperand minValue;
+  MLOperand maxValue;
 };
 
-partial interface ModelBuilder {
-  Operand clamp(Operand x, optional ClampOptions options = {});
+partial interface MLModelBuilder {
+  MLOperand clamp(MLOperand x, optional MLClampOptions options = {});
 };
 </script>
 <div algorithm=clamp>
     **Arguments:**
-        - *x*: an {{Operand}}. The input tensor.
-        - *options*: an optional {{ClampOptions}}. The optional parameters of the operation.
-            - *minValue*: an {{Operand}}. Specifies the minimum values of the range. It is either a scalar, or of the shape that is unidirectionally broadcastable to the shape of *x* according to [[!numpy-broadcasting-rule]]. When it is not specified, the clamping is not performed on the lower limit of the range.
-            - *maxValue*: an {{Operand}}. Specifies the maximum values of the range. It is either a scalar, or of the shape that is unidirectionally broadcastable to the shape of *x* according to [[!numpy-broadcasting-rule]]. When it is not specified, the clamping is not performed on the upper limit of the range.
+        - *x*: an {{MLOperand}}. The input tensor.
+        - *options*: an optional {{MLClampOptions}}. The optional parameters of the operation.
+            - *minValue*: an {{MLOperand}}. Specifies the minimum values of the range. It is either a scalar, or of the shape that is unidirectionally broadcastable to the shape of *x* according to [[!numpy-broadcasting-rule]]. When it is not specified, the clamping is not performed on the lower limit of the range.
+            - *maxValue*: an {{MLOperand}}. Specifies the maximum values of the range. It is either a scalar, or of the shape that is unidirectionally broadcastable to the shape of *x* according to [[!numpy-broadcasting-rule]]. When it is not specified, the clamping is not performed on the upper limit of the range.
 
-    **Returns:** an {{Operand}}. The output tensor of the same shape as *x*.
+    **Returns:** an {{MLOperand}}. The output tensor of the same shape as *x*.
 
     Clamp the input tensor element-wise within a range specified by *minValue* and *maxValue*. The calculation follows the expression min(max(x, minValue), maxValue). When *minValue* is not specified, the clamping is not performed on the lower limit. When *maxValue* is not specified, the clamping is not performed on the upper limit.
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -187,20 +187,62 @@ partial interface Navigator {
...
+interface WebGLRenderingContext {
+  // ...
+};
+
+// WebGPU specification:
+// (https://gpuweb.github.io/gpuweb/#gpu-device)
+[SecureContext, Exposed=Window]
+interface GPUDevice {
+  // ...
+};
```
anssiko, line 202: We should use Bikeshed's [custom definitions](https://tabatkins.github.io/bikeshed/#custom-dfns) for adding anchors to other IDL that is not yet part of the autolinking database.

To do that, you can add the  following additional `<pre class="anchors">` block e.g. below the `<pre class='metadata'>` block and remove the `interface WebGLRenderingContext` and  `interface GPUDevice` declarations. Bikeshed is picky about the whitespace here, so indentation must be exactly a multiple of 4 spaces (or tabs):
```
</pre>
<pre class="anchors">
urlPrefix: https://www.khronos.org/registry/webgl/specs/latest/2.0/; spec: WEBGL-2
    type: interface
        text: WebGLRenderingContext; url: 2.1
        text: WebGLBuffer; url: 5.4
        text: WebGLTexture; url: 5.9
urlPrefix: https://gpuweb.github.io/gpuweb/; spec: WEBGPU
    type: interface
        text: GPUDevice; url: gpu-device
        text: GPUBuffer; url: buffer-interface
        text: GPUTexture; url: texture-interface
</pre>
```

index.bs
```diff
@@ -209,83 +251,97 @@ enum OperandType {
...
 
-## ModelBuilder ## {#api-modelbuilder}
-
-The {{ModelBuilder}} interface defines a set of operations as identified by the [[#usecases]] that can be composed into a computational graph. It also represents the intermediate state of a graph building session.
+// WebGPU specification:
+// (https://gpuweb.github.io/gpuweb/#buffer-interface)
+[SecureContext, Exposed=Window]
+interface GPUBuffer {
+  // ...
+};
```
anssiko, line 287: You can remove these lines after having added the custom definitions.

index.bs
```diff
@@ -1288,100 +1344,133 @@ partial interface ModelBuilder {
...
 };
 
+// WebGPU specification:
+// (https://gpuweb.github.io/gpuweb/#texture-interface)
 [SecureContext, Exposed=Window]
-interface Model {
-  Promise<Compilation> compile(optional CompilationOptions options = {});
+interface GPUTexture {
+  // ...
 };
```
anssiko, line 1413: Similarly these can be removed.

anssiko (changes requested):
Thanks @wchao1115!

Added some comments on how to make Bikeshed link to externally defined IDL that is not yet in the autolinking database.

index.bs
wchao1115: Thanks a lot @anssiko. I'm expecting exactly this type of feedback from others who are more familiar with Bikeshed than I am. Will look into this right away.

wchao1115:
@rafaelcintron

index.bs
anssiko: Happy to help. For future reference, [definitions data model](https://tabatkins.github.io/bikeshed/#dfn-contract) explains what metadata is needed for cross-spec autolinking to work. If we wish to export definitions to be used by other specs, we should comply with this data model. I believe IDL definitions are automatically exported.

index.bs
```diff
@@ -1288,100 +1344,133 @@ partial interface ModelBuilder {
...
 
-typedef record<DOMString, Input> NamedInputs;
-typedef record<DOMString, Output> NamedOutputs;
+dictionary MLWebGLOutput {
+  WebGLTexture glTexture;
+  sequence<long> dimensions;
+};
+
+dictionary MLGPUInput {
+  required GPUTexture gpuTexture;
```
huningxin, line 1436: Should we consider to support `GPUBuffer` as input? Given TensorFlow.js WebGPU backend uses GPUBuffer as input and output: https://github.com/tensorflow/tfjs/blob/master/tfjs-backend-webgpu/src/backend_webgpu.ts#L42. @pyu10055 for comments.

index.bs
```diff
@@ -1288,100 +1344,133 @@ partial interface ModelBuilder {
...
+  sequence<long> dimensions;
+};
+
+dictionary MLGPUOutput {
+  GPUTexture gpuTexture;
```
huningxin, line 1441: ditto.

index.bs
```diff
@@ -209,83 +251,97 @@ enum OperandType {
...
+  MLOperand constant(MLOperandDescriptor desc, GPUBuffer gpuBuffer, 
+                     optional long offset = 0, optional long size);
 
   // Create a single-value operand from the specified number of the specified type.
-  Operand constant(double value, optional OperandType type = "float32");
+  MLOperand constant(double value, optional MLOperandType type = "float32");
 
   // Create a model that encapsulates the composition of operands by identifying the output operands.
-  Model createModel(NamedOperands outputs);
+  MLModel createModel(MLNamedOperands outputs);
```
huningxin, line 314: Should we still keep `MLModel` building device agnostic and leave the device association to compile time? Say to support `model.compile(glContext)` and `model.compile(gpuDevice)`. With that, developer can build the same model and compile for different devices.

index.bs
```diff
@@ -1288,100 +1344,133 @@ partial interface ModelBuilder {
...
+typedef record<DOMString, MLGPUOutput> MLNamedGPUOutputs;
 
 [SecureContext, Exposed=Window]
-interface Compilation {
-  Promise<NamedOutputs> compute(NamedInputs inputs, optional NamedOutputs outputs = {});
+interface MLCompilation {
+  Promise<MLNamedOutputs> compute(MLNamedInputs inputs, 
+                                  optional MLNamedOutputs outputs = {});
+
+  Promise<MLNamedWebGLOutputs> compute(MLNamedWebGLInputs glInputs,
```
huningxin, line 1459: Should the `MLCompilation` compiled for `WebGLRenderingContext` only accept `WebGLTexture` as input and output? Otherwise, there might be expensive data transferring between devices?

index.bs
```diff
@@ -1288,100 +1344,133 @@ partial interface ModelBuilder {
...
-interface Compilation {
-  Promise<NamedOutputs> compute(NamedInputs inputs, optional NamedOutputs outputs = {});
+interface MLCompilation {
+  Promise<MLNamedOutputs> compute(MLNamedInputs inputs, 
+                                  optional MLNamedOutputs outputs = {});
+
+  Promise<MLNamedWebGLOutputs> compute(MLNamedWebGLInputs glInputs, 
+                                       optional MLNamedWebGLOutputs glOutputs = {});
+
+  Promise<MLNamedGPUOutputs> compute(MLNamedGPUInputs gpuInputs,
```
huningxin, line 1462: ditto for `MLCompilation` compiled for `GPUDevice`

index.bs
wchao1115: I think it's a good idea. Thanks @huningxin

index.bs
wchao1115: There are interop use cases where the model constants (e.g. weights, etc.) are already uploaded to the GPU buffers before the model is created. In that case it will be inefficient to require the caller to make a roundtrip of that data back to the CPU memory just to construct a device-agnostic model.

As we position WebNN also as a backend API for frameworks, the notion of device-agnostic graph could be in a way of the interop efficiency since by the time the backend graph is initiated (by the frameworks), the model constants may already be mapped to the memory hosted by the frameworks, and for the frameworks already dealing with device resources, those constants could already be uploaded to the device.

index.bs
pyu10055: +1 WebGPU will be replacing WebGL, providing support for WebGPU from the start will future-proof the API.

index.bs
huningxin: It makes sense, thanks for the explanation. Given the `MLModel` is now device specific, would we consider merging it into the `MLCompilation`? Let's say replace the `MLModel createModel(MLNamedOperands outputs)` with `Promise<MLCompilation> compile(MLNamedOperands outputs)`.

index.bs
wchao1115: Fixed. Please have a look.

index.bs
wchao1115: Fixed.

index.bs
wchao1115: Yes, it will fail. Cross-device, cross-adapter support should not occur at this level. If the context is created with one device/adapter and the input is a resource from another, the API should fail. The callers of WebNN needs to ensure that the all the device-dependent resources are from the device that is bound to the context.

index.bs
```diff
@@ -209,83 +249,83 @@ enum OperandType {
...
-  Operand input(DOMString name, OperandDescriptor desc);
+  // Create an operand for a model constant from a WebGL buffer.
+  MLOperand constant(MLOperandDescriptor desc, WebGLBuffer glBuffer, 
+                     optional long offset = 0, optional long size);
 
-  // Create an operand that represents a model constant.
-  Operand constant(OperandDescriptor desc, ArrayBufferView value);
+  // Create an operand for a model constant from a WebGPU buffer.
+  MLOperand constant(MLOperandDescriptor desc, GPUBuffer gpuBuffer, 
+                     optional long offset = 0, optional long size);
```
huningxin, line 292: Would it make sense to introduce the `MLBuffer` that supports `ArrayBuffer`, `WebGLBuffer` and `GPUBuffer` by WebIDL [union type](https://heycam.github.io/webidl/#idl-union)? The `offset` and `size` probably need to be `unsigned long long` as `GPUBuffer` uses it.

e.g.
```js
dictionary MLBuffer {
  required (ArrayBuffer or WebGLBuffer or GPUBuffer) buffer;
  unsigned long long offset = 0;
  unsigned long long size;
};

MLOperand constant(MLOperandDescriptor desc, MLBuffer buffer);
```

The `compute` method can also use `MLBuffer`.

index.bs
huningxin: thanks!

index.bs
```diff
@@ -1288,100 +1328,159 @@ partial interface ModelBuilder {
...
+                                       optional MLNamedWebGLTextureOutputs glTextureOutputs = {});
+
+  Promise<MLNamedWebGLOutputs> compute(MLNamedWebGLBufferInputs glBufferInputs, 
+                                       optional MLNamedWebGLBufferOutputs glBufferOutputs = {});
+
+  Promise<MLNamedGPUOutputs> compute(MLNamedGPUTextureInputs gpuTextureInputs,
+                                     optional MLNamedGPUTextureOutputs gpuTextureOutputs = {});
+
+  Promise<MLNamedGPUOutputs> compute(MLNamedGPUBufferInputs gpuBufferInputs,
+                                     optional MLNamedGPUBufferOutputs gpuBufferOutputs = {});
```
huningxin, line 1473: With `MLBuffer`, we may also simplify the `compute` method by union types.

e.g.
```js
dictionary MLInput {
  required (MLBuffer or WebGLTexture or GPUTexture) data;
  sequence<long> dimensions;
};

dictionary MLOutput {
  (MLBuffer or WebGLTexture or GPUTexture) data;
  sequence<long> dimensions;
};

typedef record<DOMString, MLInput> MLNamedInputs;
typedef record<DOMString, MLOutput> MLNamedOutputs;

Promise<MLNamedOutputs> compute(MLNamedInputs inputs, 
                                optional MLNamedOutputs outputs = {});
```

index.bs
huningxin: It makes sense. We may need to specify this behavior in the following PR.

index.bs
anssiko: LGTM.

anssiko (approved):

index.bs
wchao1115: Good idea. I'll make that change.

index.bs
wchao1115: Likewise. I'll make that change.

index.bs
huningxin: @wchao1115 , any thoughts about this idea?

wchao1115: I can see the logic to it. I think it makes sense.

huningxin: I think it would also help the use case of [operation-specific API](https://github.com/webmachinelearning/proposals/issues/2#issuecomment-773441501). For example, the `webml_create_conv2d` could be just compiling a `conv2d`. e.g.
```js
const conv2d_op = await builder.compile({output: builder.conv2d(input, filter)});
```
 For fused conv2d, it could be compiling `relu`, `add` and `conv2d`. e.g.
```js
const fused_conv2d_op = await builder.compile({output: builder.relu(builder.add(builder.conv2d(input, filter), bias))});
```

@pyu10055

wchao1115: Yes, that looks much more compact. I've made the suggested changes, please take a look. 

However, since `MLModel` is removed, calling the class `MLModelBuilder` may no longer be appropriate. We could rename `MLModelBuilder` to something more relevant, or rename `MLCompilation` to `MLModel` (as in *compiled model*)

pyu10055: I see your point that device specific model can improve overall performance, there are cases the data is on CPU but the user want to execute it on GPU, does this API requires framework to upload the data to GPU as Texture and call WebNN API?
Will it be more efficient for WebNN to copy CPU data directly to its backend data structure without going through the Texture intermediate layer?
If it is really tied to the Web APIs (WebGL, WebGPU, WASM), an op level API would make much more sense, since most ML frameworks already has a graph building APIs. And the WebNN API can be stateless and functional API.

wchao1115: No, it doesn't require the framework to upload. The constants, inputs, and outputs can still be in `ArrayBufferView` in which case the implementation of webnn should still upload the resources themselves if it is running on a hardware device. This change just adds an additional capacity for the API to deal with uploaded resources.

index.bs
```diff
@@ -209,83 +249,81 @@ enum OperandType {
...
-
-## ModelBuilder ## {#api-modelbuilder}
+typedef record<DOMString, MLOperand> MLNamedOperands;
 
-The {{ModelBuilder}} interface defines a set of operations as identified by the [[#usecases]] that can be composed into a computational graph. It also represents the intermediate state of a graph building session.
-
-<script type=idl>
-typedef record<DOMString, Operand> NamedOperands;
+dictionary MLBuffer {
+  required (ArrayBufferView or WebGLBuffer or GPUBuffer) resource;
```
huningxin, line 276: Should we use `ArrayBuffer` instead of `ArrayBufferView`? The `offset` and `size` would means byte offset and size that align to WebGPUBuffer.

huningxin:
And there is a bikeshield [build error](https://github.com/webmachinelearning/webnn/pull/149/checks?check_run_id=2063026140)
```
LINK ERROR: No 'idl' refs found for 'MLModel'.
{{MLModel}}
```
The spec still uses {{MLModel}} in Compilation section.

index.bs
huningxin: Looks good, thanks!

index.bs
huningxin: Looks good with an open, commented inline. Thanks!

index.bs
huningxin: > However, since `MLModel` is removed, calling the class `MLModelBuilder` may no longer be appropriate. We could rename `MLModelBuilder` to something more relevant, or rename `MLCompilation` to `MLModel` (as in _compiled model_)

Agree. I am fine with renaming `MLCompilation` to `MLModel`. 

Actually, in this spec, a _model_ means a _graph_ and we use _graph_ everywhere. To avoid this translation, probably we could rename 
 * `MLModelBuilder` -> `MLGraphBuilder`
 * `MLModelBuilder.compile` -> `MLGraphBuilder.build`
 * `MLCompilation` -> `MLGraph`

e.g.
```js
const graph = await builder.build({c: builder.matmul(a, b)});
const outputs = await graph.compute(inputs);
```

huningxin: > If it is really tied to the Web APIs (WebGL, WebGPU, WASM), an op level API would make much more sense, since most ML frameworks already has a graph building APIs. And the WebNN API can be stateless and functional API.

@pyu10055 , the `compute` API of WebNN is stateless. The graph can contain just a single op (e.g. `conv2d`)  or a few decomposed ops that can be fused by native ML APIs (e.g. `conv2d` + `add` + `relu`). With this PR, now the `compute` can accept input and output resources on GPU device without copying between GPU and CPU. 

Does this meet the op level execution requirement of the ML frameworks mentioned in https://github.com/webmachinelearning/proposals/issues/2#issuecomment-773441501?

wchao1115:
> MLModelBuilder -> MLGraphBuilder
MLModelBuilder.compile -> MLGraphBuilder.build
MLCompilation -> MLGraph

I like that!

index.bs
wchao1115: If I'm not mistaken, the offset and length of the `ArrayBuffer` is defined as 32-bit integer while those of the GPU types are 64 bits.

huningxin: I am not sure whether the 32-bit integer length is an implementation behavior. But regarding to [`AllocateArrayBuffer`](https://tc39.es/ecma262/#sec-allocatearraybuffer) of ecma262 spec, the `byteLength` is not restricted to 32-bit integer. And the V8 seems to define `maxByteLength` of `ArrayBuffer` to [`Number.MAX_SAFE_INTEGER`](https://tc39.es/ecma262/#sec-number.max_safe_integer) (2^53 - 1) on 64-bit systems.

huningxin:
> The constants, inputs, and outputs can still be in `ArrayBufferView` in which case the implementation of webnn should still upload the resources themselves if it is running on a hardware device.

We may need to define which resource type is acceptable for a specific type of device. For example, if the device is CPU,  the `WebGLBuffer` or `WebGPUBuffer` resources may not be acceptable?

```c++
typedef (WebGLBuffer or GPUBuffer or WebGLTexture or GPUTexture) GPUResource;
```

| device type | resource type | acceptable |
|---|---|---|
| cpu | ArrayBuffer | ✔️ |
| cpu | GPUResource | ❌ (?) |
| gpu | ArrayBuffer | ✔️ |
| gpu | GPUResource (on the same adapter) | ✔️ |
| gpu | GPUResource (on the different adapter) | ❌ |
| accelerator | ArrayBuffer | ✔️ |
| accelerator | GPUResource | ❌ (?) |

@wchao1115 , I put together above table with two opens. Please take a look and feel free to add anything I missed.

index.bs
huningxin: Additionally, to support the eager execution of frameworks (e.g. TF.js) efficiently, we may consider the point made by @wchao1115 in the comment https://github.com/webmachinelearning/proposals/issues/2#issuecomment-781057233 that

> 2\. When an operation is invoked from a framework to the underlying platform, the data type and layout format must be in a form that can be readily usable by the operation without incurring additional copies or conversions. 

The framework may compute multiple ops without accessing the intermediate results. So if we allow passing the output of a compute to the next compute without making the data available, it would help the underlying implementation hide the additional copies and conversions. The idea is illustrated by following pseudo framework integration code:

```js
// assume conv2d_op, add_op, relu_op are compiled single-op graphs
// input = tf.tensor(values);
// conv2d = tf.conv2d(input, filter);
conv2d.mlOutput = conv2d_op.compute({'input': {resource: input.arrayBuffer}).output; // webnn backend
// add = tf.add(conv2d, bias);
add.mlOutput = add_op.compute({'a': conv2d.mlOutput}).output; // webnn backend
// relu = tf.relu(add);
relu.mlOutput = relu_op.compute({'input': add.mlOutput}).output;  // webnn backend
// await relu.data();
await relu.mlOutput.readDataAsync({resource: relu.arrayBuffer});  // webnn backend
```

It may require the following changes:
 * change the `MLOutput` to an interface that can read the output data (`MLOutputData`) back asynchronously
 * `compute` can accept `MLInput` or `MLOutput` in `MLNamedInputs`
 * Because `MLOutput` can read output data back asynchronously, the `compute` method can be synchronous. 

```js
dictionary MLInput {
  required (MLBuffer or WebGLTexture or GPUTexture) resource;
  sequence<long> dimensions;
};
dictionary MLOutputData {
  (MLBuffer or WebGLTexture or GPUTexture) resource;
  sequence<long> dimensions;
};
interface MLOutput {
  Promise<MLOutputData> readDataAsync(optional MLOutputData data); 
};
typedef record<DOMString, (MLInput or MLOutput)> MLNamedInputs;
typedef record<DOMString, MLOutput> MLNamedOutputs;
[SecureContext, Exposed=Window]
interface MLCompilation {
  MLNamedOutputs compute(MLNamedInputs inputs, optional sequence<DOMString> requiredOutputs);
};
```

Thoughts?

index.bs
```diff
@@ -692,7 +692,7 @@ partial interface MLModelBuilder {
     if (!hiddenState) {
       const desc = { type: 'float32', dimensions: [numDirections, 1, hiddenSize] };
       const totalSize = numDirections * hiddenSize;
-      hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0));
+      hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0).buffer);
```
huningxin, line 695: According to `dictionary MLBuffer`, it would require the `resource` field, e.g.
```js
hiddenState = builder.constant(desc, {resource: new Float32Array(totalSize).fill(0).buffer});
```

Or should we define the `MLBuffer` as a union of buffer type and buffer range?, e.g.
```js
typedef (ArrayBuffer or WebGLBuffer or GPUBuffer) MLBufferType;

dictionary MLBufferRange {
  required MLBufferType buffer;
  unsigned long long offset = 0;
  unsigned long long size;
};

typedef (MLBufferRange or MLBufferType) MLBuffer;
```

With that, the original code works.
```js
hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0).buffer);
```

WDYT?

index.bs
```diff
@@ -1037,11 +1037,11 @@ partial interface MLModelBuilder {
     <pre highlight="js">
     // input: [[1,2,3], [4,5,6]]
     const input = builder.constant(
-      { type: 'float32', dimensions: [2,3] }, new Float32Array([1,2,3,4,5,6]));
+      { type: 'float32', dimensions: [2,3] }, new Float32Array([1,2,3,4,5,6]).buffer);
```
huningxin, line 1040: ditto.

index.bs
```diff
@@ -1037,11 +1037,11 @@ partial interface MLModelBuilder {
...
 
     // padding: [[1,1], [2,2]]
     const padding = builder.constant(
-      { type: 'float32', dimensions: [2,2] }, new Float32Array([1,1,2,2]));
+      { type: 'float32', dimensions: [2,2] }, new Float32Array([1,1,2,2]).buffer);
```
huningxin, line 1044: ditto.

index.bs
```diff
@@ -273,7 +273,7 @@ The {{MLModelBuilder}} interface defines a set of operations as identified by th
 typedef record<DOMString, MLOperand> MLNamedOperands;
 
 dictionary MLBuffer {
-  required (ArrayBufferView or WebGLBuffer or GPUBuffer) resource;
+  required (ArrayBuffer or WebGLBuffer or GPUBuffer) resource;
```
huningxin, line 276: BTW, i think the field name `buffer` seems better. Because it is used in the `MLInput` as the field name is also `resource`. There will be double 'resource' to create a `MLInput`, e.g. `{'input': {resource: {resource: arrayBuffer}}`.

index.bs
wchao1115: We should discuss eager execution in a separate issue. Please feel free to post an issue for that. 

I think in general eager execution is there to aid model development by making the computed result immediately available. I feel it's an issue separated from how to make the operation running as efficiently as possible because the former is for debugging while the latter is for perf. 

The general steps of operation execution is build > compile > dispatch > execute. The compile step is where activities like constant folding and fusions occur. Fusion in particular is the result of fusing multiple semantic operations that were previously built as a subgraph into an atomic executable unit. Additionally for compute platforms like the GPU, the compile step is the time that JITing of shader fragments and preprocessing of constant weight data (aka. weight-packing) can occur.

Once the subgraph is compiled down, the compiled operations can be dispatched or placed into the command buffers. The placement of dispatches can be done by multiple CPU threads simultaneously while the GPU is executing the command buffers that were previously dispatched, preventing stalls and bubbles while appropriately saturating the GPU. 

Under the hood, eager execution is all these 4 steps sequentially executed in one go.

huningxin: Thanks for the explanation @wchao1115 !

What I propose here is to support the op-level execution of a framework, like TensorFlow.js, based on current graph API. The framework may rely on the op-level execution capability to enable the eager mode. I am not proposing to add the eager execution capability into WebNN API itself. I agree the later one should be discussed in a separate issue.

As @pyu10055 mentioned in above https://github.com/webmachinelearning/webnn/pull/149#discussion_r591039364, the framework can execute ops by computing single-op graph or fused-ops graph of WebNN. However, current `compute` method couples the graph execution and output data reading-back/conversion together. This design would not be optimal when frameworks execute multiple ops (multiple WebNN graphs) without accessing the intermediate results. Some frameworks, like TensorFlow.js, infer a model by executing all its ops but only access the output data of the final ops. If using today's `compute` API, it would convert or read back output data after each op's compute that is not necessary. So I propose to decouple the graph execution and output data reading-back into two steps. `compute` just returns an opaque `MLOutput` object. The `MLOutput` can be passed to the next `compute` as input that doesn't cause data reading-back or conversion. Framework only invokes `readDataAsync` when need accessing the output data.

Does it make sense?

index.bs
wchao1115: @huningxin The `MLBufferRange` is really a buffer view while `MLBufferType` is a buffer resource type. Calling either a buffer resource or a buffer view a `MLBuffer` could be confusing.

I propose in my latest commit a notion of `MLBufferView` which could either be a standard array buffer view or a buffer resource view.

``` javascript
dictionary MLBufferResourceView {
  required (WebGLBuffer or GPUBuffer) resource;
  unsigned long long offset = 0;
  unsigned long long size;
};
 
typedef (ArrayBufferView or MLBufferResourceView) MLBufferView;

interface MLModelBuilder {
  ...
  MLOperand constant(MLOperandDescriptor desc, MLBufferView bufferView);
};
```

This way this code can just look like this:

``` javascript
      hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0));
```
It then follows that for the input/output, the types are the following: Note that the field's name is now simply `data` as opposed to `resource` to avoid the confusion with the GPU resource types.

``` javascript
dictionary MLInput {
  required (MLBufferView or WebGLTexture or GPUTexture) data;
  sequence<long> dimensions;
};

dictionary MLOutput {
  (MLBufferView or WebGLTexture or GPUTexture) data;
  sequence<long> dimensions;
};

...
// 3. Bind inputs to the model and execute for the result.
const bufferA = new Float32Array(4).fill(1.0);
const bufferB = new Float32Array(4).fill(0.8);
const inputs = {'A': {data: bufferA}, 'B': {data: bufferB}};
const outputs = await compilation.compute(inputs);
...
```
If this all looks good, I'll prepare the last commit that renames everything to:

> MLModelBuilder -> MLGraphBuilder
MLModelBuilder.compile -> MLGraphBuilder.build
MLCompilation -> MLGraph

huningxin: Thanks @wchao1115 . Your proposal in the latest commit https://github.com/webmachinelearning/webnn/pull/149/commits/be764ba4c32c40f77ec7c45cb70a12137e95c5c2 looks good to me.

huningxin (approved):
LGTM. Thanks much for the great work.

index.bs
wchao1115: This is a good issue. I'll work on this after completing this PR. 

I understand the scenario and I certainly agree that `compute` should become a sync all. But I think the fact that `MLOutput` now specifies the union of different output resource types including the hardware resources, and that the validity of the use case of those resources are already dependent upon the same input resource type i.e. a `WebGLBuffer` output resource is only available when the input is also of `WebGLBuffer` type, should already mean that `compute` will produce its result on a corresponding resource type. 

Basically if the concern is over the repeated downloads cost of intermediate results across multiple disconnected single operations, the answer is if your input resource is a GPU buffer, the output is going to have to be in a GPU buffer, and not something that is automatically downloaded to an array buffer. This is true for whether the execution is over just 1 operator or an entire graph i.e. if you want to download the result, you will need to download from the output resource type the execution has placed its result into. If all the executions is happening on an array buffer, then obviously no download would be needed, and all that lies outside of the WebNN API.

huningxin: My concern is about if the input and output are array buffers but the compute device is a gpu or an accelerator, the implementation needs to upload and download the data repeatedly. 

Another aspect is about memory relayout, even for cpu, some native APIs, like oneDNN, use [optimized memory layout](https://oneapi-src.github.io/oneDNN/dev_guide_understanding_memory_formats.html) for hardware acceleration. Since the memory layout/relayout brings significant overhead, so it would be good to avoid that if the intermediate results are not accessed. I am not sure whether gpu device has similar memory relayout overhead. Any insights @wchao1115 ?

wchao1115:
@anssiko @huningxin It looks like `deploy.yaml` doesn't run on the merged change. So the generated html file is now stale. Is there a way to manually trigger it to run?

huningxin:
> @anssiko @huningxin It looks like `deploy.yaml` doesn't run on the merged change. So the generated html file is now stale. Is there a way to manually trigger it to run?

I am not aware the manual way to trigger a github action. Hopefully PR #151 would fix it. Please take a look.

index.bs
wchao1115: > My concern is about if the input and output are array buffers but the compute device is a gpu or an accelerator, the implementation needs to upload and download the data repeatedly.

When the call creates an `MLContext` they have an option to either create a default context with preference options e.g. power preference, etc. -or- create a context from an existing device. This distinction is important because if a default context is created, it is implied that the API implementation chooses and completely owns the device it uses for subsequent execution. In this use case, the caller has no way to know if the underlying device is a HW resource or a CPU memory because -- it is up to the API to decide. So, the output must be in a form that can readily be used, which is an array buffer view. If the underlying device happens to be a GPU, the output is automatically downloaded. This contract is important because it is how the API is used in a device-agnostic way -- from the caller's point of view. 

But if the context is explicitly created by an existing device, it means the device is under the control of the caller, which implies that the caller is the one who would manage the lifetime of the resources created by that device. In this use case, the output will be on the device's resource, and that the caller would own download the data from the resource at the right time in their scenario. This use case is what being described as the one that needs the atomic operations (aka the op-level execution) because its when the caller would have the control on the device resource including data uploads/downloads, layout assignment, and block layout processing etc.

So, in short, yes, for the default context, the data is automatically uploaded/downloaded if the underlying device needs them, but that is not the scenario that would call for the efficient op-level execution where the explicit resource management is a big implicit part of the said "efficiency" to be had.

huningxin: My concern is about the efficient op-level execution on CPU. This scenario was mentioned by @pyu10055 in https://github.com/webmachinelearning/proposals/issues/2#issuecomment-773441501 for framework's CPU - WebAssembly backend. 

The CPU can only be selected for default context. However as the default context would automatically upload/download data, it may bring unnecessary memory re-layout overhead as I mentioned. So my proposal is to allow the user code to control when to relayout/download data to array buffer for default context. I suppose this won't change the device-agnostic view of the default context. And it would also enable the efficient op-level execution for framework's CPU - WebAssembly scenario.

anssiko:
The merge of #151 triggered a rebuild that deployed properly.

@wchao1115 @huningxin great work with this PR. This is on our agenda for [this week's call](https://github.com/webmachinelearning/meetings/blob/master/telcons/2021-03-18-agenda.md) (note: it is 1 hour later than usual in the US/Canada!), and that's a good opportunity to summarize the key changes and design considerations discussed in this PR.

index.bs
pyu10055: I agree with @huningxin that, the returned tensors should be tensor handles that can be downloaded by a separate API.

wchao1115: @huningxin and @pyu10055, on the topic of wasm interop, I had a chance to chat with people who are more familiar with wasm, and they pointed me to this [`WebAssembly.Memory`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory) object, which allows for an efficient interop with JS `ArrayBuffer`. In short, the memory passed between wasm and JS is always part of a wasm `Memory` object that can be the backing store for a JS ArrayBuffer. Also, take a look at a more elaborate sample on JS/WASM interop [here](https://rob-blackbourn.github.io/blog/webassembly/wasm/array/arrays/javascript/c/2020/06/07/wasm-arrays.html)

Thru the wasm memory object, a wasm library can always interop with a JS ArrayBuffer with no copy required, and likewise in the opposite direction -- an ArrayBuffer can also be created from a wasm memory and used as an output buffer of a JS API call. Hope this helps clarify this topic.

huningxin: Thanks for the pointers @wchao1115 ! It really helps.

With the `ArrayBufferView` support by `MLGraphBuilder.constant` and `MLGraph.compute`, a wasm library can interop with WebNN API with `ArrayBufferView` backed by the `WebAssembly.Memory.buffer`. There are no issues.

However, as I mentioned in https://github.com/webmachinelearning/webnn/issues/156, there would be other two requirements for an efficient wasm interops:
1. a wasm library could create a `MLContext` from the CPU device, otherwise there might be across device data copies that a wasm library doesn't want. 
2. for op-level execution (by single-op `MLGraph`), a wasm library could control when the output data is available, otherwise there might be tensor memory re-layout overhead between native ML API and WebNN for each op's execution that are not always needed.

huningxin: > adding a device preference option in the `MLContextOptions`. This option will be complementary to the existing power preference.

I like it!

> what prevents the caller from compiling all the 3 `conv2d` ops into a single compiled graph before executing it in one go?

That's because the framework provides the op API. When executing a graph, the user code just calls the op API to execute the operations of the graph one-by-one. The framework doesn't have the knowledge of the graph, it just executes ops.

For this scenario, the framework doesn't know the user code will execute 3 `conv2d`. When user code calls a `conv2d`, the framework just computes that `conv2d`. The framework may use a platform-specific tensor layout in the `conv2d` implementation. Until the user code calls `h.data()`, the framework starts to convert and copy the tensor data into a `ArrayBufferView` in plain layout. 

So if the framework uses WebNN to compute `conv2d`, it would hurt the performance if WebNN always converts and copies the output tensor data to `ArrayBufferView` for each compute.

> The point is that the end result of executing a publicly defined API must not be platform-specific

I agree with you, at the interops point, it should use the plain/standard tensor layout.

wchao1115: > That's because the framework provides the op API. When executing a graph, the user code just calls the op API to execute the operations of the graph one-by-one. 

That doesn't sound like what a framework does. Are you saying that there are frameworks that never compiles a graph internally and just relying on individual op execution?

huningxin: Framework may provide both op API and model API, for example TensorFlow.js provides [Operations API](https://js.tensorflow.org/api/latest/#Operations) and [Models/Loading API](https://js.tensorflow.org/api/latest/#Models-Loading). It would depend on what API the user code uses. If the user code uses the op API to execute a graph, the framework won't get the graph but individual ops.

Actually, webnn-polyfill uses TensorFlow.js op API. I suppose it is a common scenario. cc @pyu10055 .

pyu10055: @wchao1115 @huningxin Thank you Ningxin for brought up this point, op level execution is common for eager execution, especially in JS environment, users could creating graph on the fly. There will be no specific graph boundary, framework would not be able to know which are the intermediate tensors that users would not need to access.

PR #154, opened 2021-03-20
Add the operation level execution use case

huningxin:
***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/154.html" title="Last updated on Mar 23, 2021, 12:42 AM UTC (ac3f8a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/154/8a564a1...huningxin:ac3f8a2.html" title="Last updated on Mar 23, 2021, 12:42 AM UTC (ac3f8a2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -189,7 +189,7 @@ so that the application loads the tiny model in the case of CPU-only devices.
 
 ### Operation Level Execution ### {#usecase-op-level-exec}
 
-A JavaScript ML framework is responsible for loading, interpreting and executing a ML model. During the model execution phase, the framework iterates through the operations of the model and executes each operation on the hardware device, like CPU or GPU. To avoid the unnecessary data copying across devices, the framework selects the same device to execute the operations. For a compute intensive operation, such as convolution 2D or matrix multiplication, the framework uses WebNN API to execute it with the ML-specific acceleration available on that selected device.
+A JavaScript ML framework is responsible for loading, interpreting and executing a ML model. During the model execution phase, the framework iterates through the operations of the model and executes each operation on the hardware device, like CPU, GPU or ML accelerator. To avoid the unnecessary data copying across devices, the framework selects the same device to execute the operations. For a compute intensive operation, such as convolution 2D or matrix multiplication, the framework uses WebNN API to execute it with the ML-specific acceleration available on that selected device.
 
 API {#api}
 =====================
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -187,6 +187,10 @@ After several weeks, she has developed a tiny DNN model that can even run on
 CPU. In order to accommodate CPU execution, she modifies the application
 so that the application loads the tiny model in the case of CPU-only devices.
 
+### Operation Level Execution ### {#usecase-op-level-exec}
+
+A JavaScript ML framework is responsible for loading, interpreting and executing a ML model. During the model execution phase, the framework iterates through the operations of the model and executes each operation on the hardware device, like CPU or GPU. To avoid the unnecessary data copying cross devices, the framework selects the same device to execute the operations. For a compute intensive operation, such as convolution 2D or matrix multiplication, the framework uses WebNN API to execute it with the ML-specific acceleration available on that selected device.
```
wchao1115, line 192: "an ML graph."

wchao1115 (approved):

index.bs
huningxin: Thanks. May I know which word you suggest to replace with "an ML graph"?

anssiko, line 192: s/cross devices/across devices/

anssiko (approved):
With a typo fix.

index.bs
huningxin: Thanks! Fixed it in the new commit https://github.com/webmachinelearning/webnn/pull/154/commits/2af623828791bf6ea310fd3b5467e10e687e84b7

wchao1115: just a wording suggestion: "a ML model" -> "an ML graph"

pyu10055 (commented):
Do we want to limited the op execution to just CPU and GPU devices?
When ML accelerators are used in WebNN, are we recommending not to use op level execution?

huningxin:
@pyu10055 , thanks for your comments. Although some existing ML accelerators are designed to execute graph, I agree we should not limit the use case for that.

index.bs
huningxin: Thanks for the suggestion. I think a ML model here refers to what the JS framework handles, e.g. TensorFlow models of TensorFlow.js, ONNX models of ONNX.js. I don't want to blur it with ML graph of WebNN that is a framework agnostic in-memory representation. Does it make sense?

pyu10055 (approved):

huningxin:
Thanks for the review and approval!

Issue #156, opened 2021-03-22
Support CPU - WebAssembly scenario of the op level execution use case


huningxin:
Open this issue to follow up the operation-specific APIs discussion of [3/18 WebML CG call](https://www.w3.org/2021/03/18-webmachinelearning-minutes.html#t02). @pyu10055 @wchao1115 @anssiko @jbingham, please take a look.

### Use case

This is one scenario of the framework's [op level execution use case](https://github.com/webmachinelearning/webnn/pull/154) (more details can be found in [operation-specific API proposal](https://github.com/webmachinelearning/proposals/issues/2)). A JavaScript ML framework executes ops on the CPU device with WebAssembly. For the compute intensive ops, such as conv2d or matmul, the framework also wants to use WebNN API to execute the op (by a single-op MLGraph) with the ML-specific instructions, such as [Vector Neural Network Instructions (VNNI)](https://en.wikichip.org/wiki/x86/avx512_vnni#:~:text=AVX%2D512%20Vector%20Neural%20Network%20Instructions%20(AVX512%20VNNI)%20is,convolutional%20neural%20network%2Dbased%20algorithms), on the same CPU device.

### Requirements

WebNN should allow frameworks create a `MLContex` for CPU device. This would avoid the unnecessary data copying cross devices when frameworks use WebAssembly - CPU to execute other ops.

WebNN should allow frameworks control when the output data is available for access. This would avoid the unnecessary tensor layout conversions between native ML API and the WebNN. Some background:
 - Some native ML APIs use hardware dependent memory layout for acceleration, for example oneDNN uses different [blocked memory layouts](https://oneapi-src.github.io/oneDNN/dev_guide_understanding_memory_formats.html) for better vectorization and cache reuse on different platforms.
 - The memory layout conversions are expensive.
 - Frameworks may use WebNN API to execute multiple ops (via multiple single-op MLGraphs) without access the intermediate results between them. 

For example, a user of TensorFlow.js may execute 3 `conv2d` but only access the output of the last one:
```js
c = tf.conv2d(a, b);
e = tf.conv2d(c, d);
h = tf.conv2d(f, g);
output = await h.data();
```
A potential WebNN implementation would only need to do the memory layout conversion and put the data into `ArrayBufferView` when `h.data()` is invoked.

anssiko:
(Cross-linking @wchao1115’s comment on WebAssembly.Memory object https://github.com/webmachinelearning/webnn/pull/149#discussion_r603700378.)

[...]

Issue #105, opened 2020-10-20
Support the execution of the sub-graph scenario


huningxin:
Opening this issue to follow up the discussion of https://github.com/webmachinelearning/webnn/pull/94#discussion_r498289215.

@pyu10055 mentioned:

> With this API, the computation is tied with the compilation. should the compile method have the inputs/outputs pair and execution can only execution the graph compiled with the input/output pair? Otherwise, the compilation might not support the execution of the sub-graph scenario?

Actually, the existing `builder.createModel` allows to specify the outputs. With that, developers could create different sub-graphs (models) from a topology within a builder and compile/compute them individually. For example:

```js
const builder = nn.createModelBuilder();
const a = builder.input('a', descA);
const b = builder.constant(descB, bufferB);
const c = builder.constant(descC, bufferC);
const d = builder.mul(a, b);
const e = builder.add(d, c);
const model1 = builder.createModel({d}); // d = a * b
const model2 = builder.createModel({e}); // e = a* b + c
```

@pyu10055 , please let us know any gaps of the existing API and share more details of your proposal. Thanks!

huningxin:
Hi @pyu10055 , this issue was discussed in [10/29 WebML CG call](https://www.w3.org/2020/10/29-webmachinelearning-minutes.html#t02). The participants would like to know more about your use case and understand the gaps. Could you please elaborate it a bit more? Thanks!

pyu10055:
This API allows users to specify two models with different output nodes. I think it should be ok for the purpose of transfer learning. But this API does not seem to follow the conventional builder pattern, where the final export API would not change the structure of the built object. It feels more natural to build a single model, but compile into two execution plans using the compile method.

anssiko:
Per [earlier discussion](https://www.w3.org/2020/11/12-webmachinelearning-minutes.html#t03) it seems the major use cases are addressed with the existing API. I'll close this issue but please feel free to re-open if new information is made available that suggests otherwise.

I have [updated the 2021-01-07 agenda](https://github.com/webmachinelearning/meetings/commit/e69f99e2dd398b349022acb9433e85ec0573c772#diff-a017f0094a1702062feec0648735a3905047c7e309bf075e5549eb30d807d014) accordingly.

PR #162, opened 2021-04-13
Add support for device selection

wchao1115:
- Add a device preference in the context option in addition to the power preference. This change allows the app to have more control over what type of execution device should be used. It enables an effective interop with WASM (#156).
- Follow up on the issue discussed in [PR#149](https://github.com/webmachinelearning/webnn/pull/149#discussion_r604542024) around supporting operation-level scenario in frameworks that support asynchronous data downloads. 
- Further clarify level of support for GPU resources from WebGPU and WebGL in relation to the underlying execution device the MLContext is based on.

@RafaelCintron

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/162.html" title="Last updated on May 6, 2021, 12:14 AM UTC (806cd20)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/162/47a8e0c...806cd20.html" title="Last updated on May 6, 2021, 12:14 AM UTC (806cd20)">Diff</a>


index.bs
```diff
@@ -447,16 +490,30 @@ partial interface Navigator {
...
 <script type=idl>
+enum MLDevicePreference {
+  // Let the user agent decide the most suitable device to use.
+  "default",
+  // Prefer a graphics processor or GPU.
+  "graphics",
+  // Prefer a compute-specific processor such as a neural engine or NPU.
+  "compute",
+  // Prefer a traditional software-based device such as a CPU.
+  "software"
```
huningxin, line 501: I prefer to naming it as `cpu`. It would be a clear device name that the Wasm libs could select to avoid the unnecessary data copying cross devices, like the [Support CPU - WebAssembly scenario of the op level execution use case](https://github.com/webmachinelearning/webnn/issues/156) describes.

wchao1115: Other than the API name prefix, most API naming conventions generally discourage the use of abbreviations as identifier names. If we were to call it "cpu" we would also need a "gpu" value to be consistent, and then what's for AI accelerators? "npu", "vpu", "tpu"? I think that would start a naming discussion to no end.

huningxin: I suppose the "cpu" and "gpu" abbreviations would be well known by developers because they are already widely used by native ML APIs, e.g. NNAPI [`DeviceTypeCode`](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ga61f16b495d2e04593c819c12e42ac9f2) defines `ANEURALNETWORKS_DEVICE_CPU` and `ANEURALNETWORKS_DEVICE_GPU`, ML Compute [`MLCDeviceType`](https://developer.apple.com/documentation/mlcompute/mlcdevicetype) defines `gpu` and `cpu` and oneDNN [`dnnl_engine_kind_t`](https://oneapi-src.github.io/oneDNN/group__dnnl__api__engine.html#ga04b3dd9eba628ea02218a52c4c4363a2) also defines `dnnl_cpu` and `dnnl_gpu`.

For AI accelerators, probably we can just call them as "accelerator" (in the context of WebNN). This is also used by native APIs, such as NNAPI [`DeviceTypeCode`](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ga61f16b495d2e04593c819c12e42ac9f2) defines `ANEURALNETWORKS_DEVICE_ACCELERATOR` for "Dedicated accelerator for Machine Learning workloads.". Another example is OpenCL [`cl_device_type`](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#CL_DEVICE_TYPE) defines `CL_​DEVICE_​TYPE_​CPU`, `CL_DEVICE_TYPE_GPU` and `CL_​DEVICE_​TYPE_​ACCELERATOR` for heterogeneous hardware.

Actually, today's modern CPUs already have ML specific instructions, such as Vector Neural Network Instruction (VNNI) of X86 and Scalable Vector Extension 2 (SVE2) of Arm, and are adding even more advanced instructions, e.g., for matrix multiplication. My major concern is "software" seems not to reflect these new capabilities that would accelerate the ML workloads for web apps.

index.bs
```diff
@@ -1616,27 +1673,37 @@ partial interface MLGraphBuilder {
...
+  constructor(MLResource resource, optional sequence<long> dimensions);
+
+  // The tensor dimensions
+  sequence<long> dimensions();
+
+  // The underlying resource of the tensor
+  MLResource resource();
+
+  // Asynchronously download the result of a computation onto a typed array.
+  Promise<ArrayBufferView> data();
```
huningxin, line 1694: I would propose to support downloading the result into a pre-allocated `ArrayBufferView`. e.g.,
```js
Promise<(ArrayBufferView or undefined)> data(optional ArrayBufferView buffer);
```

If `buffer` is not provided, the returned promise resolves with a newly created `ArrayBufferView` contains the downloaded result.
If `bufer` is provided, the returned promise resolves with `undefined` after downloading the result into user provided `buffer`.

This would allow Wasm libs download the data to the existing Wasm memory instead of making another copy, as @pyu10055 mentioned in the [CPU - WebAssembly example](https://github.com/webmachinelearning/proposals/issues/2#issuecomment-773441501) of operation-level execution use case.

index.bs
```diff
@@ -1616,27 +1673,37 @@ partial interface MLGraphBuilder {
...
+The {{MLGraph}} interface represents a compiled computational graph. A compiled graph once constructed is immutable and cannot be subsequently changed.
+
+<script type=idl>
+typedef record<DOMString, MLTensor> MLNamedTensors;
 
 [SecureContext, Exposed=Window]
 interface MLGraph {
-  Promise<MLNamedOutputs> compute(MLNamedInputs inputs, 
-                                  optional MLNamedOutputs outputs = {});
+  MLNamedTensors compute(MLNamedTensors inputs, optional MLNamedTensors outputs = {});
```
huningxin, line 1706: If the returned `MLTensor` supports downloading the result into the pre-allocated `ArrayBufferView` as proposed in https://github.com/webmachinelearning/webnn/pull/162/files#r613402691, then the `outputs` argument could be just the names of the required outputs, e.g.,
```js
MLNamedTensors compute(MLNamedTensors inputs, optional sequence<DOMString> outputNames = []);
```

index.bs
```diff
@@ -1817,11 +1884,12 @@ const c = builder.matmul(a, b);
...
 const bufferA = new Float32Array(sizeOfShape(descA.dimensions)).fill(0.5);
-const inputs = {'a': {data: bufferA}};
+const inputs = {'a': new MLTensor(bufferA)};
 // Pre-allocate output buffer for c.
-const outputs = {'c': {data: new Float32Array(sizeOfShape([3, 3]))}};
-await graph.compute(inputs, outputs);
-console.log(&#96;values: ${outputs.c.data}&#96;);
+const outputs = {'c': new MLTensor(new Float32Array(sizeOfShape([3, 3])))};
+graph.compute(inputs, outputs);
+const data = await outputs.c.data();
```
huningxin, line 1891: I am not sure whether the `MLTensor` is a right solution for this pre-allocated outputs scenario. Will `outputs.c.data()` turn a newly created `ArrayBufferView` instead of downloading the result into the pre-allocated `ArrayBufferView`?

index.bs
```diff
@@ -1845,21 +1913,23 @@ const e = builder.add(d, c);
...
+const inputs = {'a': new MLTensor(bufferA)};
 
 // Compute both d and e.
-let outputs = await graph.compute(inputs);
+let outputs = graph.compute(inputs);
 console.log(&#96;outputs include ${Object.keys(outputs)}&#96;);
 
 // Compute d.
-outputs = await graph.compute(inputs, {d});
+outputs = graph.compute(inputs, {d});
```
huningxin, line 1923: For the optional output scenario, user code only needs to specify the required output names. Previous `MLOutput` has all optional fields, so it can be an empty dictionary and not bound to any resources. However with the new `MLTensor`, it looks like user code has to create from a resource which is not necessary for this usage.

index.bs
wchao1115: I'm open for naming suggestion if others have input on this. In my line of work designing Windows API at Microsoft, we generally stay away from using abbreviations because its meaning and perceived freshness could change over time. (Windows API is well known for longevity).

index.bs
wchao1115: Yes, in this case `{d}` is a dictionary with a single key associated with a null (`MLTensor`) value. Am I missing something here?

huningxin: I may make mistake for `record<DOMString, MLOutput>` case, it should be `{'d': {}}`. However, for `record<DOMString, MLTensor>`, I am afraid the `{'d': null}` is invalid.

index.bs
gramalingam: My 2 cents worth: "compute" and "software" sound too generic, and I wouldn't be able to guess what they meant.  On the other hand, I do think most people can guess what "cpu" or "gpu" mean. I agree custom neural accelerators don't have as obvious a choice. "npu" or "accelerator" or "custom" sound fine to me.

index.bs
huningxin, line 1694: According to the table of device types and the resource types, the `MLTensor` that downloads the result to `ArrayBufferView` probably needs to be separated from `MLTensor` for GPU resources (say `MLGPUTensor`). It is because that the GPU resources have their own data download methods. e.g.,

```js
interface MLTensor {
  constructor(ArrayBufferView buffer, optional sequence<long> dimensions);
  sequence<long> dimensions();
  Promise<ArrayBufferView> data();
};

typedef (MLResourceBufferView or WebGLTexture or GPUTexture) MLGPUResource;

interface MLGPUTensor {
  constructor(MLGPUResource resource, optional sequence<long> dimensions);
  sequence<long> dimensions();
  MLGPUResource resource();
};
```

index.bs
huningxin, line 1706: The `compute` method probably could be overrode for different device types and different output allocations (newly allocated or pre-allocated), e.g.,
```js
typedef record<DOMString, MLTensor> MLNamedTensors;
typedef record<DOMString, MLGPUTensor> MLNamedGPUTensors;

interface MLGraph {
  // MLContext with any device types, allocate output `MLTensor`s for listed output names.
  MLNamedTensors compute(MLNamedTensors inputs, optional sequence<DOMString> outputNames);

  // MLContext with any device types, use pre-allocated output 'MLTensor`s.
  void compute(MLNamedTensors inputs, MLNamedTensors outputs);

  // MLContext with GPUDevice or WebGLRenderingContext device type, allocate output `MLGPUTensor`s for listed output names.
  MLNamedGPUTensors compute(MLNamedGPUTensors inputs, optional sequence<DOMString> outputNames);

  // MLContext with GPUDevice or WebGLRenderingContext device type, use pre-allocated output 'MLTensor`s.
  void compute(MLNamedGPUTensors inputs, MLNamedGPUTensors outputs);
};
```

index.bs
huningxin: As the outputs will be returned by compute, probably we can optimize this usage by allowing user to only specify the output names in an array. e.g.,
```js
outputs = graph.compute(inputs, ['d']);
```

index.bs
huningxin: As the outputs are given, I would suggest `compute` not to return the results. The calling `data()` of the given output `MLTensor` should download the results to the pre-allocated `ArrayBufferView` and return it (the same object). e.g.,
```js
const preallocatedBuffer = new Float32Array(sizeOfShape([3, 3]);
const outputs = {'c': new MLTensor(preallocatedBuffer))};
graph.compute(inputs, outputs); // compute does not return
const returnedBuffer = await outputs.c.data();
// returnedBuffer === preallocatedBuffer 
```

index.bs
huningxin: For the CPU - WebAssembly scenario of [Operation-specific API](https://github.com/webmachinelearning/proposals/issues/2#issuecomment-773441501), the framework could create `MLTensor`s with pre-allocated Wasm memory, e.g.,

```js
const inputs = {'x': new MLTensor(input_tensor_info.f32())};
const outputs = {'y': new MLTensor(output_tensor_info.f32_write())};
conv2d.compute(inputs, outputs);
await outputs.y.data()
// The results have been downloaded into the memory pointed by output_tensor_info.f32_write().
```

For GPU scenario, the framework could create `MLGPUTensor`s with WebGLTextures, e.g.,
```js
const inputs = {'x': new MLGPUTensor(inputTextureData.texture)};
const outputs = {'y': new MLGPUTensor(outputTextureData.texture)};
conv2d.compute(inputs, outputs);
// Post-processing with outputTextureData.texture
```

@wchao1115 @pyu10055 , does it look good?

pyu10055: @huningxin In WebGL, there is no way to share textures across contexts, is the input textures limited to a GPU context created during compilation or users can specify a context during graph creation?

index.bs
pyu10055: +1 on cpu and gpu

index.bs
```diff
@@ -1616,27 +1673,37 @@ partial interface MLGraphBuilder {
...
-dictionary MLOutput {
-  (MLBufferView or WebGLTexture or GPUTexture) data;
-  sequence<long> dimensions;
+[SecureContext, Exposed=Window]
+interface MLTensor {
+  // Construct a tensor from a resource
+  constructor(MLResource resource, optional sequence<long> dimensions);
+
+  // The tensor dimensions
+  sequence<long> dimensions();
```
pyu10055, line 1688: should tensor also have dtype method

index.bs
wchao1115: @pyu10055 the assumption is that the app must create an MLContext with the WebGL context that creates the texture, which later passed back as an input to the graph created by the MLContext. It is generally known that a resource should not be shared across different GPU contexts. In DirectX, you can achieve cross-adapter sharing of resources, but you need to explicitly create a shared resource handle and manage the lifetime of the shared resource explicitly. But at the Web API level, I agree that it is best to make this simpler and simply not allow cross-adapter sharing.

huningxin: @pyu10055 

> users can specify a context during graph creation?

yes, user can create a `MLContext` by specifying a `WebGLRenderingContext`, as [ML.createContext](https://webmachinelearning.github.io/webnn/#api-ml)
```js
MLContext createContext(WebGLRenderingContext glContext);
```

index.bs
huningxin: +1 to add data type, probably reuse the [`MLOperandType`](https://webmachinelearning.github.io/webnn/#enumdef-mloperandtype)?

explainer.md
```diff
@@ -44,12 +44,13 @@ const graph = await builder.build({'C': C});
 // 3. Bind inputs to the graph and execute for the result.
 const bufferA = new Float32Array(4).fill(1.0);
 const bufferB = new Float32Array(4).fill(0.8);
-const inputs = {'A': {data: bufferA}, 'B': {data: bufferB}};
-const outputs = await graph.compute(inputs);
+const inputs = {'A': new MLTensor(bufferA), 'B': new MLTensor(bufferB)};
+const outputs = graph.compute(inputs);
+const data = await outputs.C.data();
```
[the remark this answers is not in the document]
huningxin: I understand it is, the returned promise resolves with an `ArrayBufferView`.

index.bs
```diff
@@ -1657,52 +1724,51 @@ interface MLGraph {
...
+                |outputs|: an optional {{MLNamedTensors}}. The names and pre-allocated resources of required outputs for the compute request. Default to be an empty [=record=] which means that the compute request is for all outputs.
             </pre>
 
-            **Returns:** {{Promise}}&lt;{{MLNamedOutputs}}&gt;. The dimensions and data of outputs returned by the compute request.
+            **Returns:** {{MLNamedTensors}}. The dimensions and data of outputs returned by the compute request.
 
-            1. Let |promise| be [=a new promise=].
             <!-- Validate inputs and outputs -->
-            1. If any of the following requirements are unmet, then [=reject=] |promise| with a {{TypeError}} and stop.
+            1. If any of the following requirements are unmet, then stop.
```
anssiko, line 1747: Do we want to signal to the web developer the executing has stopped?

There's a leftover promise reference on L1801 that should be either removed or repurposed for signaling this, this one:

reject promise with an OperationError and stop.

index.bs
```diff
@@ -1744,18 +1810,18 @@ interface MLGraph {
...
-                                            1. Set the content of |results|[|outputName|].{{MLOutput/data}} to the content of output tensor returned by |this|.{{MLGraph/[[implementation]]}}.
+                                            1. Let |results|[|outputName|].{{MLTensor/resource}} be a new {{ArrayBufferView}} of size |outputSize| and kind that is compatible to output tensor according to [this table](#appendices-mloperandtype-arraybufferview-compatibility).
+                                            1. Set the content of |results|[|outputName|].{{MLTensor/resource}} to the content of output tensor returned by |this|.{{MLGraph/[[implementation]]}}.
                                     1. Remove |outputName| from |remainingOutputNames|.
-                                    1. If |remainingOutputNames| is empty, then resolve |promise| with |results| and stop.
+                                    1. If |remainingOutputNames| is empty, then stop.
```
anssiko, line 1824: I think L1828 should be updated to either:

1. Return |results|.

Or remove this step if the preferred design is to use `data()` for data download per https://github.com/webmachinelearning/webnn/pull/162#discussion_r613776751

anssiko (changes requested):

index.bs
```diff
@@ -1657,52 +1724,51 @@ interface MLGraph {
...
+                            1. The kind of |value|.{{MLTensor/resource}} must be compatible to |inputOperand|.{{MLOperandDescriptor/type}} according to [this table](#appendices-mloperandtype-arraybufferview-compatibility).
+                        1. If |value|.{{MLTensor/dimensions}} was given, then:
+                            1. The length of |value|.{{MLTensor/dimensions}} must be the same as the length of |inputOperand|.{{MLOperandDescriptor/dimensions}}.
                             1. Let |i| be 0.
                             1. While true:
-                                1. Let |dimension| be |value|.{{MLInput/dimensions}}[|i|].
+                                1. Let |dimension| be |value|.{{MLTensor/dimensions}}[|i|].
                                 1. |dimension| must be greater than 0.
                                 1. If |inputOperand|.{{MLOperandDescriptor/dimensions}}[|i|] is greater than 0, then |dimension| must be equal to |inputOperand|.{{MLOperandDescriptor/dimensions}}[|i|].
                                 1. Set |i| to |i| + 1.
```
anssiko, line 1762: As a drive-by comment, many web specs tend to use a more prose-like form:

Increment |i| by 1.

And in many cases if it makes the algorithm more readable, instead of "i" they may say "foobar counter" instead. Not sure if that applies here, just noting in general in web spec algorithms it is recommended to be self-documenting rather than concise. That said, "i" is pretty widely understood :-)

index.bs
wchao1115: I couldn't find any reference or recommendations regarding usage of abbreviations in the Web API definition. Anyone knows where to look? @anssiko. If it is allowed, should it be used in lowercase or uppercase? etc.

DirectX API in general avoids abbreviations, and terms like "graphics" or even "software" have been used for quite some time. e.g. 

https://docs.microsoft.com/en-us/windows/win32/direct3d11/atoc-dx-graphics-direct3d-11
https://docs.microsoft.com/en-us/windows/win32/api/d3dcommon/ne-d3dcommon-d3d_driver_type

index.bs
wchao1115: > probably needs to be separated from ones for GPU resources (say MLGPUTensor). It is because that the GPU resources have their own data download methods. 

If the tensor has a backing GPU resource, the execution output will be on that backing resource. The `await data()` call is only when the user wants to download the data from the GPU resource to a CPU buffer. 

If the user wants to make a copy from the tensor resource, they should do it themselves outside of the WebNN call.

index.bs
wchao1115: @huningxin I suggest only to have 1 method that requires the caller to pre-allocate the output buffer. It's much simpler that way i.e.

```javascript
typedef record<DOMString, MLTensor> MLNamedTensors;

interface MLGraph {
  // MLContext with any device types, use pre-allocated output 'MLTensor`s.
  void compute(MLNamedTensors inputs, MLNamedTensors outputs);
};
```

I wonder how practical the optional output case (Example 5) really is. For a multi-output topology, not requiring all of the outputs to be produced at execution is uncommon. The optimization gain from not having to specify an optional output is going to be minimal relative to the cost of executing the entire topology anyway. We should probably remove this use case and simplify the API.

Also, there is no need to special-case the GPU resources tensor as the `MLTensor' already wraps those types nicely. See the 'Device Selection` section for more detail as to when different type of resource is allowed.

index.bs
wchao1115: Actually that wouldn't be needed. Remember that `MLTensor` is just a resource holder type. The data type of the resource is already specified when an operand is created either explicitly through the `MLOperandDescriptor` or implicitly through the execution of the relevant operation in the graph. 

The fact that there is an optional `dimensions` param in its ctor may have confused the matter, but it's only there to support the graph with free dimensions input where the input shape isn't known until the graph execution time. In a normal use case, it won't be needed, hence optional.

This conversation makes me realize that calling the type `MLTensor` is itself misleading because when you call something a tensor, you would naturally think that it has both a shape and a data type association. A more appropriate name is probably `MLResource` or `MLGraphResource`.

index.bs
wchao1115: We should remove this use case and simplify the API. See more details in my other reply below.

index.bs
huningxin: > I wonder how practical the optional output case (Example 5) really is.

The use case came from https://github.com/webmachinelearning/webnn/issues/105 @pyu10055 . With the new `MLTensor` interface, user code can only download the data of a specific output. This would help avoid the overhead of downloading other outputs that user code doesn't need.

index.bs
huningxin: > If the tensor has a backing GPU resource, the execution output will be on that backing resource. The `await data()` call is only when the user wants to download the data from the GPU resource to a CPU buffer.

We probably need to coordinate with other Web API, e.g. if the resource is a `GPUBuffer`, WebGPU defines [`mayAsync`](https://gpuweb.github.io/gpuweb/#dom-gpubuffer-mapasync) and [`getMappedRange`](https://gpuweb.github.io/gpuweb/#dom-gpubuffer-getmappedrange) for the data downloading.

My another question is if the tensor is created with an `ArrayBufferView`, what the `MLResource resource()` would return? If it returns the `ArrayBufferView`, what's the semantic difference from `Promise<ArrayBufferView> data()`?

index.bs
wchao1115: Sure. And this would still work, right? 

```Javascript
const d = builder.matmul(a, b);
const e = builder.add(d, c);
const graph = await builder.build({d, e});
const output = {'d': new MLTensor(new Float32Array(sizeOfShape([3, 3])))};
graph.compute(inputs, output);
let data = await output.d.data();
```

huningxin: right.

index.bs
wchao1115: > My another question is if the tensor is created with an ArrayBufferView, what the MLResource resource() would return? If it returns the ArrayBufferView, what's the semantic difference from Promise<ArrayBufferView> data()?

`resource()` returns the backing resource of the `MLTensor` while `await data()` asynchronously download and layout-convert the content to a standard layout to a separate buffer. I believe you proposed this semantic for the WASM backend when an operation is executed stand alone.

wchao1115: > We probably need to coordinate with other Web API, e.g. if the resource is a GPUBuffer, WebGPU defines mayAsync and getMappedRange for the data downloading.

We define `await data()` as a contract to asynchronously download the content to a CPU buffer. How exactly that download is implemented from a GPU resource to a CPU buffer is up to the implementor of this contract. 

If the caller of WebNN wants to make a copy from one GPU resource to the other, then that is entirely outside the scope of WebNN.

index.bs
wchao1115: Syntax-wise, are you saying that the following is also invalid? 

```javascript
const d = builder.matmul(a, b);
const e = builder.add(d, c);
--> (?) const graph = await builder.build({d, e});
```

Can you tell me exactly how `record<K,V>` should be used and how should we write the above line in question? I've never found a usage sample for it in the WebIDL spec.

index.bs
anssiko: @wchao1115 Web Platform Design Principles > Naming principles https://w3ctag.github.io/design-principles/#naming-is-hard is probably the best resource. Naming is always a topic of active discussion in API design, and there's often no absolute rights or wrongs.

Given there's a Web API called WebGPU, I suspect "gpu" would pass if you prefer it that way. Similarly, we can probably agree that "cpu" is familar to web developers of today.

That said, I'd challenge the group to think this from the perspective of future-proofing. How do the possible future architectures fit into the device buckets we choose? We could make this extensible with new buckets in the future, or we could make the initial ones abstract enough to accommodate future devices. I think there's always a trade-off whatever we choose. What I think we want to achieve is to be able to provide web developers with knobs they can use to give hints to the implementation so their workloads can be executed in a manner that yields the best user experience.

index.bs
huningxin: > ```js
> --> (?) const graph = await builder.build({d, e});
> ```

This code is valid. It is equivalent to `await builder.build({'d': d, 'e': e});`

They are two styles of [object properties definition](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#property_definitions). The former one is so called "shorthand property names" of  ES2015.

The `outputs = graph.compute(inputs, {d});` is invalid is because the object `d` is a type of `Operand` in that example. If `d` is a `MLOutput`, it will be valid.

index.bs
huningxin: > `resource()` returns the backing resource of the `MLTensor` while `await data()` asynchronously downloads and layout-converts the content to a standard layout onto a separate buffer. I believe you proposed this semantic for the WASM backend when an operation is executed stand alone.

If the `MLTensor` is created with an `ArrayBufferView`, e.g. a view of the existing Wasm memory, I suppose `await data()` should download and relayout the results to the given buffer instead of creating a separate one.

For example, in the following code snippet, the `preallocatedBuffer`, `backingBuffer` and `returnedBuffer` should be the same buffer, and `await outputTensor.data()` downloads and relayouts the results to that buffer. Is my understanding correct?
```js
const preallocatedBuffer = new Float32Array(sizeOfShape([3, 3]);
const outputTensor = new MLTensor(preallocatedBuffer);
graph.compute(inputs, {'output': outputTensor});
const backingBuffer = outputTensor.resource();
// backingBuffer === preallocatedBuffer and the results are not available in preallocatedBuffer.
const returnedBuffer = await outputTensor.data();
// returnedBuffer === preallocatedBuffer and the results are available in preallocatedBuffer.
```

index.bs
wchao1115: Thanks @anssiko. I'll take this advice in the design principles: *Please consult widely on names in your APIs.* 

The hardest is probably the accelerator type. Let's discuss this naming in our next CG call.

index.bs
huningxin: So according to this semantic, the assumptions would be:
1. `backingBuffer` and `preallocatedBuffer` are same object
2. `returnedBuffer` is different from `preallocatedBuffer`
3. before `await data()`, the results are not available in `preallocatedBuffer`
4. after `await data()`, the results are available in `returnedBuffer` and `preallocatedBuffer`

Is my understanding correct? 

Then, the `returnedBuffer` would not be used by the Wasm backend, because it needs another copy to move the results of `returnedBuffer` into its existing memory. The Wasm backend provides the `preallocatedBuffer` of its existing memory and would expect the results are downloaded and layout converted into that buffer.

Actually, to meet the requirement that downloads or layout converts into a pre-allocated buffer, how about overriding `MLTensor.data()` to accept an `ArrayBufferView` as a target, say

```js
interface MLTensor {
  constructor(ArrayBufferView buffer, optional sequence<long> dimensions);
  sequence<long> dimensions();
  Promise<ArrayBufferView> data();
  Promise<void> data(ArrayBufferView buffer)
};
```

Then,
```js
const preallocatedBuffer = new Float32Array(sizeOfShape([3, 3]);
const outputs = graph.compute(inputs);
await outputs.c.data(preallocatedBuffer);
// the results are downloaded and layout converted into preallocatedBuffer.
```

huningxin: My another question is about `MLTesnor.resource()`, what's the usage for the context created with `MLContextOptions`?

According to the "device types and the resource types" table, for the device type "default", "graphics", "compute" or "software", the only supported resource type is `ArrayBufferView`. It means the `MLTensor.resource()` would return an `ArrayBufferView`. However it seems not useful because the user code would eventually call `await data()`.

Did I miss anything?

index.bs
huningxin: Good suggestion, thanks! Acknowledged, will update the compute algorithm once this PR merged.

index.bs
wchao1115: @huningxin You're giving 2 requirements that are a bit contradicting. They need to be reconciled before this design can go on. 

On one hand, you said that the operation-specific calls from TF.js would need to stage its intermediate result in the output tensor after the compute call, and that the subsequent download and layout conversion should happen as an explicit step afterward.

On another hand, you also said here that the WASM backend wants to avoid additional copy, would try to avoid a separate layout conversion step, and would want layout conversion to happy in-place. 

I personally do not prefer this 2-stage design as it complicates the API semantic for the compute step quite a bit. It's confusing to have many overload methods that are actually carrying out different API semantics and not simply extending the existing semantic. 

I need to think about this some more as it's really messy now. I would like the compute semantic that can be defined as clearly as possible so that the caller won't end up depending on the implementation-specific behaviors full of caveats and assumptions.

gramalingam: This does look like a complex design issue. On one hand, we have data located in different devices, and the need to copy them. On the other hand, we also have asynchrony, which requires some mechanism to indicate when data is available (handled by "Promise" here). Should we separate these two aspects (instead of bundling them together for a specific usage)? Here are a few related points: (a) How do users reuse the same underlying memory (in a GPU, for example) over multiple calls? Do they create distinct MLTensors for each use? (b) How would users know if the previous use of some memory/buffer is now complete and that the memory is free to be reused? (c) Do we want the ability to pass the asynchronous data ("Promise") produced by one execution to be passed in as input to another execution, without creating extra copy/wait?

anssiko: This design discussion merits its own issue.

@wchao1115 mind opening one?

wchao1115: Since this PR hasn't been merged, I don't think a new issue is appropriate. I think we should resolve it either in this PR (if it pulls through) or as a new PR (if this is abandoned).

anssiko: The convention is to document self-contained design discussions in their own issues. It helps discoverability and future generations who look for information on a certain design choice.

If you feel it is better to have these discussions (soon 50 comments) here in this PR e.g. due to complex interactions and to retain context that’s fine.

This PR can fix multiple issues at the same time, if you opt in to the idea.

huningxin: @wchao1115 , let me try to clarify

> On one hand, you said that the operation-specific calls from TF.js would need to stage its intermediate result in the output tensor after the compute call, and that the subsequent download and layout conversion should happen as an explicit step afterward.

Yes. This is good for performance. The implementation can keep the data on device and optimal internal layout as long as possible. `MLTensor` could be passed between graph computes. The implementation only needs to do data download and layout convert upon `await data()`

> On another hand, you also said here that the WASM backend wants to avoid additional copy, would try to avoid a separate layout conversion step, and would want layout conversion to happen in-place in the output buffer.

It's only about the target buffer of the results. The Wasm backend wants to download the results in standard layout to a pre-allocated buffer (usually the Wasm memory) instead of getting a newly returned buffer. Otherwise, Wasm backend needs to copy it back to its memory. 

For example, with current API, Wasm backend needs to copy the data of `returnedBuffer` to `wasmMemoryView` by `wasmMemoryView.set`.
```js
const wasmMemoryView = new Float32Array(wasmMemory, offset, length);
const outputs = graph.compute(inputs);
const returnedBuffer = await outputs.c.data();
wasmMemoryView.set(returnedBuffer );  // memory copy
```

If `data()` accepts the pre-allocated buffer, as I proposed, it would avoid that copy, for example
```js
const wasmMemoryView = new Float32Array(wasmMemory, offset, length);
const outputs = graph.compute(inputs);
await outputs.c.data(wasmMemoryView );
```

I don't think the requirement that downloads the results to a pre-allocated buffer conflicts with the separate data download and layout convert step. They are both useful for effective Wasm backend integration.

huningxin: @wchao1115 

> I need to think about this some more as it's really messy now. I would like the compute semantic that can be defined as clearly as possible so that the caller won't end up depending on the implementation-specific behaviors full of caveats and assumptions.

Agree. 

If you look at the device types and resource types table in this PR (I am copying it here for reference):

<table>
  <tr><th>Device Type<th>ArrayBufferView<th>GPUBuffer<th>GPUTexture<th>WebGLBuffer<th>WebGLTexture
  <tr><td>GPUDevice<td>Yes<td>Yes<td>Yes<td>No<td>No
  <tr><td>WebGLRenderingContext<td>Yes<td>No<td>No<td>Yes<td>Yes
  <tr><td>default<td>Yes<td>No<td>No<td>No<td>No
  <tr><td>graphics<td>Yes<td>No<td>No<td>No<td>No
  <tr><td>compute<td>Yes<td>No<td>No<td>No<td>No
  <tr><td>software<td>Yes<td>No<td>No<td>No<td>No
</table>

the `ArrayBufferView` is supported by all device types. On the other hand, the GPU resource types are context specific and may need coordination with other corresponding GPU APIs. That's the reason I propose to separate `MLTensor` for `ArrayBufferView` from GPU resource types. The `MLTensor` for `ArrayBufferView` would make this spec self-contained.

My refined proposal of `MLTensor` and `compute` API is:
```js
interface MLTensor {
  constructor(ArrayBufferView buffer, optional sequence<long> dimensions);
  sequence<long> dimensions();
  Promise<ArrayBufferView> data(); // return a newly allocated buffer containing the results.
  Promise<void> data(ArrayBufferView buffer); // download the results to pre-allocated buffer.
};

typedef record<DOMString, MLTensor> MLNamedTensors;

interface MLGraph {
  MLNamedTensors compute(MLNamedTensors inputs);
};
```

For the GPU resources, I think the requirement is a binding mechanism for compute inputs/outputs. So I suppose the [`MLInput`](https://webmachinelearning.github.io/webnn/#dictdef-mlinput) and [`MLOutput`](https://webmachinelearning.github.io/webnn/#dictdef-mloutput) of the current spec are still good for this purpose. Probably we can make it more GPU resources specific, e.g.,

```js
typedef  (WebGLBuffer or GPUBuffer or WebGLTexture or GPUTexture) MLGPUResource;

dictionary MLGPUInput {
  required MLGPUResource resource;
  sequence<long> dimensions;
};

dictionary MLGPUOutput {
  MLGPUResource resource;
  sequence<long> dimensions;
};

typedef record<DOMString, MLGPUInput> MLNamedGPUInputs;
typedef record<DOMString, MLGPUOutput> MLNamedGPUOutputs;

partial interface MLGraph {
  (MLNamedGPUOutputs or undefined) compute(MLNamedGPUInputs inputs, optional MLNamedGPUOutputs outputs = {});
};
```
The `compute` can be sync as well, because now the data downloading to `ArrayBufferView` are handled by `MLTensor`. For these GPU resources, there are corresponding GPU APIs to download the data asynchronously.

Any thoughts?

wchao1115: Let's step back a bit and really think about this requirement for a sec:

> The framework may compute multiple ops without accessing the intermediate results. So if we allow passing the output of a compute to the next compute without making the data available, it would help the underlying implementation hide the additional copies and conversions. The idea is illustrated by following pseudo framework integration code:

```javascript
// assume conv2d_op, add_op, relu_op are compiled single-op graphs
// input = tf.tensor(values);
// conv2d = tf.conv2d(input, filter);
conv2d.mlOutput = conv2d_op.compute({'input': {resource: input.arrayBuffer}).output; // webnn backend
// add = tf.add(conv2d, bias);
add.mlOutput = add_op.compute({'a': conv2d.mlOutput}).output; // webnn backend
// relu = tf.relu(add);
relu.mlOutput = relu_op.compute({'input': add.mlOutput}).output;  // webnn backend
// await relu.data();
await relu.mlOutput.readDataAsync({resource: relu.arrayBuffer});  // webnn backend
```

i.e. if a framework graph `[input]->conv1->conv2->[output]` has each of the two `conv` operations compiled separately on the backend, we want the result out of `conv1` to be an opaque handle with implementation-specific layout format that can later be "fixed" in a final async call (e.g. `await relu.data()` in the pseudo code above).

On a second thought, this requirement can't be allowed. No matter where a convolution operation is defined, whether at the framework level or at webnn backend, an output of a public operation must be in a known layout format. 

While I understand the motive of potentially reducing copies at the backend, from the API standpoint, it simply cannot be allowed or we'll end up with a spec that says that the layout of an output of an operation is in an unknown hardware-specific format. 

This means that for a framework graph `[input]->conv1->conv2->[output]`, if a user (of the framework) executes the entire graph in one atomic action (load-and-run), the framework backend must compile the two convolution ops together in order to achieve the optimization that would allow the intermediate result between the two ops to stay in a native format. And that only the final output after `conv2` is converted back to a known layout format. This is precisely the value of graph compilation -- to optimize by hiding the intermediate results within the graph that aren't exposed to the user.

For the case where the framework graph is actually exposed to the user and that the user chooses to execute each op individually (eager execution), the execution result from each op must be produced and converted into a known layout format every step of the way. Yes, it won't be as efficient but that's exactly the trade-off the user is making when using eager execution -- to favor inspection of  the output of each operator in the graph over end-to-end graph execution performance.

huningxin: @wchao1115 

> whether at the framework level or at webnn backend, an output of a public operation must be in a known layout format.

Agree. So the result of `MLTensor.data()` is always the standard layout format.

> we'll end up with a spec that says that the layout of an output of an operation is in an unknown hardware-specific format.

The internal format is up to the implementation, the spec won't define that.

> For the case where the framework graph is actually exposed to the user and that the user chooses to execute each op in the graph individually (eager execution), the execution result from each op must be produced and converted into a known layout format every step of the way.

The conversion would only be needed when user code calls `MLTensor.data()`. Otherwise, it can be kept in the internal format and be consumed by the next op compute as an input. 

As a summary, if user codes wants to access the output data, call `MLTensor.data()`. The result is always in the standard layout format.

wchao1115: > Agree. So the result of `MLTensor.data()` is always the standard layout format.

Not only that. The output of `MLGraphBuilder.conv2d` must also be in standard format. See detail in the [conv2d API spec](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-conv2d), both the size and the shape of the output operand are clearly defined, because this operation might be the last in the graph, and so its output in that case is the graph's output, which must be defined.

huningxin: > The output of `MLGraphBuilder.conv2d` must also be in standard format. See detail in the [conv2d API spec](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-conv2d), both the size and the shape of the output operand are clearly defined, because this operation might be the last in the graph, and so its output in that case is the graph's output, which must be defined.

Now with `MLTensor` as the graph output, the user code can only access the results by calling `MLTensor.data()`. So if `MLTensor.data()` always returns the standard layout format, it would not break the spec definition.

Did I miss anything?

wchao1115: The output of the `conv2d` will be in the output buffer that is bound to the graph output in the `compute` call. Before this PR, there was the `MLOutput` dictionary where the `data` field holds the output buffer. In this PR, that would be the `resource()` call. This is true whether the output buffer is a GPU resource or an ArrayBufferView. I now think it's probably best to leave the async download step up to the caller, and not tying it to the WebNN API since the caller would know how to deal with data downloads from a resource type they specify. As for the operation itself, the output buffer needs to always be in standard formats to keep the spec well defined.

anssiko:
From [today's call](https://www.w3.org/2021/04/29-webmachinelearning-minutes.html#r01):

>RESOLUTION: Open new issues for 1) data UL/DL 2) layout format conversion discussions, and finish PR #162 addressing device selection

wchao1115:
Based on the resolution in the CG call, I've pulled back on the asynchronous data download call and only left the change with the addition of the device selection note. 

Also, based on our discussion, we will leave the accelerator device preference option out of the spec for now since the resource models around different kinds of accelerators is still not well defined. We'll deal with these issues separately from this PR.

Please have a look.

index.bs
```diff
@@ -435,6 +449,30 @@ They are represented by callbacks and promises in JavaScript.
...
+
+An {{MLContext}} interface represents a global state of neural network execution. One of the important context states is the underlying execution device that manages the resources and facilitates the compilation and the eventual execution of the neural network graph. An {{MLContext}} could be created from a specific GPU device such as {{GPUDevice}} or {{WebGLRenderingContext}} that is already in use by the application, in which case the corresponding {{GPUBuffer}} or {{WebGLBuffer}} resources used as graph constants, as well as the {{GPUTexture}} and {{WebGLTexture}} as graph inputs must also be created from the same device. In a multi-adapter configuration, the device used for {{MLContext}} must be created from the same adapter as the device used to allocate the resources referenced in the graph.
+
+In a situation when a GPU context executes a graph with constants or inputs given as {{ArrayBufferView}}, the content of such constant or input is automatically uploaded from the system memory to the GPU memory. Likewise, the content of the execution result is downloaded to the system memory as {{ArrayBufferView}} upon requested.
```
huningxin, line 456: Should we info the two aspects of a CPU context when executing a graph with `ArrayBufferView`:
1. no data moving across device
2. may still have layout format conversions

huningxin (commented):
Thanks @wchao1115 !

anssiko (approved):

index.bs
wchao1115: Updated. Please have a look.

huningxin (approved):
Looks good. Thanks much @wchao1115 !

PR #166, opened 2021-05-07
Support download data asynchronously

### LINKED ISSUE #156: Support CPU - WebAssembly scenario of the op level execution use case

huningxin:
Follow up on the issue discussed in #162 for operation-level scenario in frameworks that support asynchronous data downloads. The changes include:

- Introduce `MLTensor` interface that supports download data asynchronously.
- Add an overloaded version of `MLGraph.compute` that returns `MLTensor` synchronously.
- Add a new example that computes multiple graphs without downloading the intermediate results.

Fix #156.

@RafaelCintron

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/166.html" title="Last updated on May 11, 2021, 8:03 AM UTC (a0f8b31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/166/0c847fb...huningxin:a0f8b31.html" title="Last updated on May 11, 2021, 8:03 AM UTC (a0f8b31)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -441,9 +441,9 @@ depends on work that happens on any timeline other than the [=Content timeline=]
 They are represented by callbacks and promises in JavaScript.
 
 <div class="example">
-{{MLGraph/compute()|MLGraph.compute()}}:
+{{MLGraph/compute(inputs, outputs)|MLGraph.compute(inputs, outputs)}}:
 
-  1. User issues a compute request by calling {{MLGraph/compute()|MLGraph.compute()}} on the [=Content timeline=] and gets a promise in return.
+  1. User issues a compute request by calling {{MLGraph/compute(inputs, outputs)|MLGraph.compute(inputs, outputs)}} on the [=Content timeline=] and gets a promise in return.
   2. User agent processes the compute request on the [=Device timeline=] by calling the OS ML API.
   3. After the ML device operating on [=Queue timeline=] is done, the user agent makes the results ready to be consumed by user and [=resolves=] the promise.
 
@@ -1701,23 +1701,30 @@ partial interface MLGraphBuilder {
 The {{MLGraph}} interface represents a compiled computational graph. A compiled graph once constructed is immutable and cannot be subsequently changed.
 
 <script type=idl>
+typedef (MLBufferView or WebGLTexture or GPUTexture) MLData;
+
 dictionary MLInput {
-  required (MLBufferView or WebGLTexture or GPUTexture) data;
+  required MLData data;
   sequence<long> dimensions;
 };
 
 dictionary MLOutput {
-  (MLBufferView or WebGLTexture or GPUTexture) data;
-  sequence<long> dimensions;
+  required MLData data;
 };
 
-typedef record<DOMString, MLInput> MLNamedInputs;
+interface MLTensor {
+  Promise<MLData> data();
+  sequence<long> dimensions();
+};
+
+typedef record<DOMString, (MLInput or MLTensor)> MLNamedInputs;
 typedef record<DOMString, MLOutput> MLNamedOutputs;
+typedef record<DOMString, MLTensor> MLNamedTensors;
 
 [SecureContext, Exposed=Window]
 interface MLGraph {
-  Promise<MLNamedOutputs> compute(MLNamedInputs inputs, 
-                                  optional MLNamedOutputs outputs = {});
+  MLNamedTensors compute(MLNamedInputs inputs);
+  Promise<undefined> compute(MLNamedInputs inputs, MLNamedOutputs outputs);
 };
 </script>
 
@@ -1742,22 +1749,85 @@ interface MLGraph {
 </dl>
 
 <dl dfn-type=method dfn-for=MLGraph>
+    : <dfn>compute(inputs)</dfn>
+    ::
+        Issue a compute request of the {{MLGraph}} given {{MLNamedInputs}}. Return {{MLNamedTensors}}.
+
+        <div algorithm="MLGraph.compute(inputs)">
+            **Called on:** {{MLGraph}} |this|.
+
+            **Arguments:**
+            <pre class=argumentdef for="MLGraph/compute(inputs)">
+                |inputs|: an {{MLNamedInputs}}. The inputs for the compute request.
+            </pre>
+
+            **Returns:** {{MLNamedTensors}}.
+
+            1. Let |outputs| be a new {{MLNamedTensors}}.
+
+            <!-- Validate inputs -->
+            1. If any of the following requirements are unmet, then throw a {{TypeError}} and stop.
+
+                <div class=validusage>
+                    1. For each |key| -> |value| of |inputs|:
+                        1. |this|.{{MLGraph/[[inputOperands]]}}[|key|] must exist.
+                        1. Let |inputOperand| be |this|.{{MLGraph/[[inputOperands]]}}[|key|].
+                        1. If |value| is an {{MLInput}}, then:
+                            1. If |value|.{{MLInput/data}} is an {{ArrayBufferView}}, then:
+                                1. The kind of |value|.{{MLInput/data}} must be compatible to |inputOperand|.{{MLOperandDescriptor/type}} according to [this table](#appendices-mloperandtype-arraybufferview-compatibility).
+                            1. If |value|.{{MLInput/dimensions}} was given, then:
+                                1. The length of |value|.{{MLInput/dimensions}} must be the same as the length of |inputOperand|.{{MLOperandDescriptor/dimensions}}.
+                                1. Let |i| be 0.
+                                1. While true:
+                                    1. Let |dimension| be |value|.{{MLInput/dimensions}}[|i|].
+                                    1. |dimension| must be greater than 0.
+                                    1. If |inputOperand|.{{MLOperandDescriptor/dimensions}}[|i|] is greater than 0, then |dimension| must be equal to |inputOperand|.{{MLOperandDescriptor/dimensions}}[|i|].
+                                    1. Increment |i| by 1.
+                                    1. If |i| if equal to the length of |value|.{{MLInput/dimensions}}, then break.
+                            1. Else:
+                                1. For each |dimension| of |inputOperand|.{{MLOperandDescriptor/dimensions}}:
+                                    1. The value of |dimension| must be greater than 0.
+                </div>
+
+            <!-- Compute -->
+            1. Issue the following steps on the [=Device timeline=] of |this|.{{MLGraph/[[implementation]]}}:
+                <div class=device-timeline>
+                    1. For each |key| -> |value| of |inputs|:
+                        1. Let |inputTensor| be the input tensor of |this|.{{MLGraph/[[implementation]]}} for |key|.
+                        1. Set the dimensions of |inputTensor| to |value|.{{MLInput/dimensions}}.
+                        1. Set the content of |inputTensor| to the content of |value|.{{MLInput/data}}.
+                    1. For each |key| -> |value| of |this|.{{MLGraph/[[outputOperands]]}}:
+                        1. Issue a compute request of |this|.{{MLGraph/[[implementation]]}} for |key|.
+                        1. If there is an error returned by |this|.{{MLGraph/[[implementation]]}}, then:
+                            1. Throw an {{OperationError}} and stop.
+                        1. Else:
+                            1. Let |outputTensor| be the output tensor returned by |this|.{{MLGraph/[[implementation]]}}.
+                            1. Let |output| be a new {{MLTensor}}.
+                            1. Associate |output| with |outputTensor|.
+                            1. Set |outputs|[|key|] be |output|.
+                </div>
+
+            1. Return |outputs|.
+
+        </div>
+
     : <dfn>compute(inputs, outputs)</dfn>
     ::
-        Issue a compute request of the {{MLGraph}} given {{MLNamedInputs}} and optional {{MLNamedOutputs}}. The returned {{Promise}} resolves when the results in {{MLNamedOutputs}} are ready to be consumed.
+        Issue a compute request of the {{MLGraph}} given {{MLNamedInputs}} and {{MLNamedOutputs}}. The returned {{Promise}} resolves when the results in {{MLNamedOutputs}} are ready to be consumed.
 
-        <div algorithm=MLGraph.compute>
+        <div algorithm="MLGraph/compute(inputs, outputs)">
             **Called on:** {{MLGraph}} |this|.
 
             **Arguments:**
             <pre class=argumentdef for="MLGraph/compute(inputs, outputs)">
-                |inputs|: a {{MLNamedInputs}}. The data and optional dimensions of inputs for the compute request.
-                |outputs|: an optional {{MLNamedOutputs}}. The names and pre-allocated resources of required outputs for the compute request. Default to be an empty [=record=] which means that the compute request is for all outputs.
+                |inputs|: an {{MLNamedInputs}}. The inputs for the compute request.
+                |outputs|: an {{MLNamedOutputs}}. The pre-allocated outputs for the compute request.
             </pre>
 
-            **Returns:** {{Promise}}&lt;{{MLNamedOutputs}}&gt;. The dimensions and data of outputs returned by the compute request.
+            **Returns:** {{Promise}}&lt;{{undefined}}&gt;.
 
             1. Let |promise| be [=a new promise=].
+
             <!-- Validate inputs and outputs -->
             1. If any of the following requirements are unmet, then [=reject=] |promise| with a {{TypeError}} and stop.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1742,24 +1745,23 @@ interface MLGraph {
...
+                |outputNames|: an optional [=sequence=]&lt;{{DOMString}}&gt. The names of required outputs for the compute request. Default to be an empty [=sequence=] which means that the compute request is for all outputs.
             </pre>
 
-            **Returns:** {{Promise}}&lt;{{MLNamedOutputs}}&gt;. The dimensions and data of outputs returned by the compute request.
+            **Returns:** {{MLNamedOutputs}}.
 
-            1. Let |promise| be [=a new promise=].
             <!-- Validate inputs and outputs -->
-            1. If any of the following requirements are unmet, then [=reject=] |promise| with a {{TypeError}} and stop.
+            1. If any of the following requirements are unmet, then throw a {{TypeError}} and stop.
```
anssiko, line 1764: Would it be helpful to use different exception types to signal different ways `compute()` can fail? Do we expect a web developer to write control flow for error handling that could recover if it'd know more about the reason of failure? If that'd be a valid use case, then:

We could either use multiple simple exceptions (these match ECMAScript error objects):
https://heycam.github.io/webidl/#dfn-simple-exception

Or DOMExceptions with error names:
https://heycam.github.io/webidl/#idl-DOMException-error-names

For example, we could throw a new "DataError" DOMException to signal value.data errors.

index.bs
```diff
@@ -1804,45 +1806,22 @@ interface MLGraph {
...
-                                            1. Set the content of |outputs|[|outputName|].{{MLOutput/data}} to the content of output tensor returned by |this|.{{MLGraph/[[implementation]]}}.
-                                        1. Else:
-                                            1. Let |results|[|outputName|].{{MLOutput/data}} be a new {{ArrayBufferView}} of size |outputSize| and kind that is compatible to output tensor according to [this table](#appendices-mloperandtype-arraybufferview-compatibility).
-                                            1. Set the content of |results|[|outputName|].{{MLOutput/data}} to the content of output tensor returned by |this|.{{MLGraph/[[implementation]]}}.
-                                    1. Remove |outputName| from |remainingOutputNames|.
-                                    1. If |remainingOutputNames| is empty, then resolve |promise| with |results| and stop.
-                            </div>
+                        1. If there is an error returned by |this|.{{MLGraph/[[implementation]]}}, then:
+                            1. Throw an {{OperationError}} and stop.
```
anssiko, line 1814: Given "OperationError" is a [DOMException](https://heycam.github.io/webidl/#idl-DOMException) error name, we should say:

Throw an "OperationError" DOMException and stop.

(For [simple exceptions](https://heycam.github.io/webidl/#dfn-simple-exception), we can simply say e.g. "Throw a TypeError".)

anssiko (changes requested):
Thanks! I made a few comments about error handling.

index.bs
```diff
@@ -1895,14 +1874,15 @@ const descB = {type: 'float32', dimensions: [4, 3]};
...
 
 const bufferA = new Float32Array(sizeOfShape(descA.dimensions)).fill(0.5);
 const inputs = {'a': {data: bufferA}};
 // Pre-allocate output buffer for c.
-const outputs = {'c': {data: new Float32Array(sizeOfShape([3, 3]))}};
-await graph.compute(inputs, outputs);
-console.log(&#96;values: ${outputs.c.data}&#96;);
+const bufferC = new Float32Array(sizeOfShape([3, 3]));
+const outputs = graph.compute(inputs);
+await outputs.c.data(bufferC);
```
wchao1115, line 1884: This pattern implies that for every `compute` call, the browser must allocate an intermediate resource for the execution result, since the only pre-allocated output buffer allowed is the buffer used as the download destination. I'm thinking in the case of an implementation on the GPU where the execution result will need to be staged in a GPU buffer that can't be pre-allocated according to this change.

index.bs
```diff
@@ -1701,23 +1701,26 @@ partial interface MLGraphBuilder {
...
+  required MLData data;
   sequence<long> dimensions;
 };
 
-dictionary MLOutput {
-  (MLBufferView or WebGLTexture or GPUTexture) data;
-  sequence<long> dimensions;
+interface MLOutput {
+  Promise<MLData> data();
+  Promise<undefined> data(MLData data);
```
wchao1115, line 1713: I think this complicates the GPU case and makes it less efficient because it essentially requires 2 buffers to get a GPU execution result, one hidden from the caller and another as a destination of an async download call. This may not be unusual if the scenario calls for an execution result be in a CPU memory like object detection, etc., but there are also many other scenarios where the output should be in a GPU buffer like the style transfer or GAN, which may eventually require not 2 but 3 buffers to accomplish it; first is the internal one, then a copy caused by a download, and the last one after the conversion from a tensor layout to a video frame texture.

wchao1115: On the semantic front, if the `data` call is going to mean more than just a simple data download operation, say, including layout conversion in some implementation, then I will have a hard time understanding how the browser would be able to implement this call behind the scene i.e. how could an implementation of a mere `MLOutput` struct know how to layout-convert every possible output of every operation defined which has this `MLOutput` struct bound to at execution time? It would also mean that any operation can produce an output operand of any layout format since the real output data can only be obtained from the eventual `data` call after the graph is executed.

index.bs
huningxin: Thanks @wchao1115 . I think you are right, the pre-allocated output buffers should be provided at compute call. I revert the `MLOutput` for the pre-allocated output buffers and only for that usage. By following the discussion in #162, I add the `MLTensor` interface for the newly allocated output buffers that are returned by compute call. Please take another look.

index.bs
huningxin: Thanks @anssiko . I'll fix that once the API shape discussion is settled.

index.bs
wchao1115: The new change seems to introduce an additional `MLTensor` into the mix with another `compute` overload that has a different semantic. The problem with this change is that the caller will now need to know how webnn is implemented in order to choose the right calling pattern for the `compute` method. Generally speaking, the hard part of designing an abstraction API, like webnn, is precisely to make it so that the caller doesn't need to have an intimate knowledge of how the API is implemented in order to call it correctly.

The point I was making earlier about how an implementation of `MLOutput` could possibly know how to implement it properly when the layout conversion semantic is baked into its definition is about all of the possible outcomes it must produce, not about what platform API, the browser must call. 

For example, an `MLOutput` could be bound to a graph that has one of the various activations, a convolution, a pooling, a batchnorm, or an RNN, etc as the graph's output layer.  So if we define `MLOutput's` behavior so that it knows how to perform layout conversion *after* any graph is executed, then it effectively must know how *any* graph produces its output through *any* operator at the end of the graph. It must also know how to post-process any graph output of any such operator. I think this will be very hard for the implementer to get right on any platform. As a data point, DirectML, for instance, has no concept of layout post-processing post-graph execution.

huningxin: > The problem with this change is that the caller will now need to know how webnn is implemented in order to choose the right calling pattern for the `compute` method.

It's based on pre-allocating output buffers or not. If the caller pre-allocates the output buffers, call `compute(inputs, outputs)`. If the caller doesn't, call `outputs = compute(inputs)`. In current spec, the `compute` also has the optional outputs argument that is for the two usages. The difference is only for the newly allocated outputs returned by `compute`.

> The point I was making earlier about how an implementation of `MLOutput` could possibly know how to implement it properly when the layout conversion semantic is baked into its definition is about all of the possible outcomes it must produce, not about what platform API, the browser must call.

As I mentioned, the layout conversion should be an implementation choice. From the API surface, the results returned by `MLTensor.data()` are always in standard layout defined in the spec. An implementation would not care about the layout conversion, if the native API it uses doesn't have such concept, such as DirectML as you mentioned, actually the [webnn-native DirectML backend](https://github.com/webmachinelearning/webnn-native/tree/main/src/webnn_native/dml) implements in this way. However, this change would allow another implementation to optimize if the native API exposes the platform optimized layout, such as oneDNN that I mentioned.

wchao1115: There is nothing wrong with having multiple use cases defined for the API. There is also nothing wrong with having multiple implementations. But when the caller must know how to call the API correctly for a specific implementation of the API, that's when the abstraction becomes leaky. For instance, in your example the caller must always allocate an output buffer if the implementation is based on DirectML, but must not if the implementation is WASM. Or that it must use `MLTensor` and not `MLOutput` an a way to get the data out. 

In practice, the caller likely won't be able to know that since that is a browser's choice and one that also depends on which platform the browser is running on. Even if there is a way for the caller to know that, it would have been very hard to get it right in every single situation.

huningxin: > For instance, in your example the caller must always allocate an output buffer if the implementation is based on DirectML, but must not if the implementation is WASM. Or that it must use `MLTensor` and not `MLOutput` as a way to get the data out.

I suppose not. The callers would choose `MLTensor` or `MLOutput` based on the usage other than depending on a specific implementation. If the callers have the pre-allocated output buffers, they would bind the buffers via `MLOutput` and call `compute(inputs, outputs)`. If the callers don't allocate output buffers, they would call `compute(inputs)` that returns `MLTensor`. Then the callers will get the output buffers by `MLTensor.data()` asynchronously.

On the other hand, an implementation should support both `MLTensor` and `MLOutput` for the two usages. According to a DirectML based implementation, if a `GPUBuffer` is pre-allocated and bound to the `MLOutput`, the implementation would be able to get the `ID3D12Resource` of the `GPUBuffer` and bind it as the op dispatching output. For `MLTensor` case, the implementation would allocate a new `ID3D12Resource` as the op dispatching output. The implementation would also create a new `MLTensor` for the `ID3D12Resource` and return it. When the caller calls `MLTensor.data()`, the implementation would download the data from the underlying `ID3D12Resource`, copy its content to a new `ArrayBufferView` and return it. This is based on my knowledge of DirectML and experience of webnn-native DirectML backend, please correct me if I am wrong.

huningxin: This would be also useful for GPU device if the caller would like to execute multiple graphs in a stream. As the example 5 in this PR illustrates, with `MLTensor`, the data of `output1` and `output2` could be kept in GPU without downloading to CPU, only the data of `output3` will be downloaded to CPU until calling `output3.data()`.
```js
const context = navigator.ml.createContext({devicePreference: 'gpu'});
const builder = new MLGraphBuilder(context);

async function buildConv2d(inputShape, filterShape) {
  const input = builder.input('input', {type: 'float32', dimensions: inputShape});
  const filter = builder.constant({type: 'float32', dimensions: filterShape},
                                  new Float32Array(sizeOfShape(filterShape)).fill(0.5));
  const output = builder.conv2d(input, filter);
  return await builder.build({'output': output});
}

// Build three graphs that each one contains a conv2d op.
const conv2dOp1 = await buildConv2d([1, 1, 9, 9], [1, 1, 3, 3]);
const conv2dOp2 = await buildConv2d([1, 1, 7, 7], [1, 1, 3, 3]);
const conv2dOp3 = await buildConv2d([1, 1, 5, 5], [1, 1, 3, 3]);

// Compute the graphs and access the final result.
const inputBuffer = new Float32Array(9*9).fill(0.5);
const output1 = conv2dOp1.compute({'input': {data: inputBuffer}).output;
const output2 = conv2dOp2.compute({'input': output1}).output;
const output3 = conv2dOp3.compute({'input': output2}).output;
console.log(`shape: [${output3.dimensions()}], values: ${await output3.data()}`);
```

anssiko:
(We discussed this PR on today's call https://www.w3.org/2021/05/13-webmachinelearning-minutes.html#t11 and agreed to continue and consolidate related discussion in here.)

index.bs
wchao1115: Thanks @huningxin for explaining the use case in more detail. The way I would summarize this requirement is that we want to allow the compiled graph to be able to produce the result in native (platform-specific) memory format to accommodate frameworks that uses webnn in an op-by-op calling pattern. Although I would claim that frameworks that are serious about execution performance should use graph API, as all the platform-specific tasks such as memory format and others can be freely implemented within the graph execution session as an implementation detail, the way an existing code base like TF.js already works could be hard to argue with.

I think the key to support discrete use cases such as this is to treat it as a special case so that it doesn't pollute the entire API and ambiguates the more mainstream and recommended use case. This way not only that the special case can be treated more attentively in the API, it also makes it much easier to document the use case more clearly.

In this case, I would propose that we allow native memory format graph output as a special case through an optional flag in the graph builder and coupled it with an explicit conversion API. Please take a look below and see if this makes sense.

```javascript
const builder = new MLGraphBuilder(context, {allowNativeFormatOutput: true});
...
const op1 = await builder.build(...);
const op2 = await builder.build(...);
...
const output1 = op1.compute({'input': {data: inputBuffer}).output;
const output2 = op2.compute({'input': output1}).output;
// If output is in native format, convert it to standard format
const outputData = (output2.isNativeFormat ? builder.convertToStandardFormat(output2.data) : output2.data);
```

Note that I also propose making both `compute` and `convertTo...` sync calls to simplify threading and resource management, to avoid non-trivial resource barrier logic in the Javascript thread. 

I also prefer that we eliminate `compute`'s return value altogether and require that it accepts pre-allocated output resources. As stated in my previous reply, it is prohibitively expensive to create GPU resources on the fly as a function call's return value. It's also inefficient for the browser to internally manage output buffer reuse for webnn. The best policy is often to leave it to the caller.

huningxin: Thanks @wchao1115 for the summary and the native format proposal.

In general, I would agree with you that we should treat this use case as a special case without making the mainstream use case harder to implement. My concern about the the proposal is that the data in native format would be eventually exposed in the `output.data` even JavaScript code is not expected to use it before `convertToStandardFormat`. In this way, it would expose the implementation details and might cause fingerprinting issue.

On the other hand, as I updated in https://github.com/webmachinelearning/webnn/issues/156#issuecomment-846828170, according to the WebNN conv2d prototype within TF.js Wasm backend, there are some findings:

> 1. TF.js Wasm backend expects the input and output data of an op execution to be in standard layout.
> 2. TF.js Wasm backend pre-allocates input and output buffers for an op execution.
> 3. TF.js Wasm backend executes an op synchronously.

Regarding to #1, I think probably we can leave the native format support out of the scope for the current spec. For #2 and #3, actually they align to @wchao1115 's proposal that makes the `MLGraph.compute` to be sync call and accept pre-allocated output resources. As the `MLGraphBuilder.build` is invoked for op execution when the [`operator_cache`](https://github.com/huningxin/tfjs/blob/webnn_op/tfjs-backend-wasm/src/cc/conv2d_impl.cc#L223) misses, it would also need to be sync call.

So I propose:
1. leave the native format support in a separate issue.
2. change `MLGraph.compute` to `undefined compute(MLNamedInputs inputs, MLNamedOutputs outputs);`
3. change `MLGraphBuilder.build` to `MLGraph build(MLNamedOperands outputs);`

With these changes, I suppose #156 can be satisfied.

wchao1115: Fingerprinting is a good callout. To mitigate that, you could null out `output2.data` with only a flag `isNativeFormat` set on it, and have `convertTo..` takes the graph output name instead of the actual data content in native format. The reason I put `convertTo..` on the graph builder is precisely because the function call would need access to the operator graph to work.

So, there could be some tweaking to hide the native content if we want to. The main part of the proposal is to make the caller do something extra and a bit unusual to signify the discrete use case. 

I'm also fine if we want to skip that for now and only focused on standard format use cases in V1. I will look at your other changes later in the day.

wchao1115: @huningxin Why do you still want to keep the async version of both `build` and `compute` call? Which is the recommended one for the caller?

huningxin: It is because the `build` and `compute` methods may take time to complete. If the callers are on the browser main thread, they can use the async version to avoid freezing the UI. 

There is W3C TAG recommendations for [sync](https://w3ctag.github.io/design-principles/#synchronous) and [async](https://w3ctag.github.io/design-principles/#promises) API design where it mentions:
> An API might need to be asynchronous if:
> * the user agent may need to do a significant amount of work on another thread, or in another process, before returning the result.

wchao1115: I think sync and async is a good discussion point. Supporting both may complicate the implementation in the browser as they carry different guarantees and require different setup with different implications to the caller. I personally still prefer that we stick with just one model for consistency.

wchao1115: @RafaelCintron has a great suggestion w.r.t. sync/async API in the CG call this morning. 

> RafaelCintron: if we must have async version, perhaps we can have that be strongly type to only take ArrayBuffer and return them

I think it's a good idea. Not sure how to translate that into an API proposal. @huningxin Do you want to try that?

huningxin: Agree, it is a good idea. Thanks @RafaelCintron . 

However, as the origin intent of this PR is for layout conversion and it has touched so many other important topics, I would suggest we handle them in separate issues. What I would like to do is:
1. open an issue about native format support for future spec and close this PR.
2. open an issue for compute API that only takes pre-allocated output buffers.
3. open an issue for sync API for wasm lib usage https://github.com/webmachinelearning/webnn/issues/156#issuecomment-846828170

WDYT? @wchao1115

wchao1115: Are you suggesting that we close this PR with the 3 issues tracked separately? I'm fine with that too. I can probably tackle the second issue in my next PR.

huningxin: > Are you suggesting that we close this PR with the 3 issues tracked separately? I'm fine with that too. 

Yes. Opened #173

> I can probably tackle the second issue in my next PR.

Thanks. I happened to open #174 for second and third issue. Please take a look.

huningxin:
Follow up with #173 and #174. Close this PR. Thanks again for the valuable inputs.

PR #174, opened 2021-05-31
Support sync API and require pre-allocated output buffers

huningxin:
This PR follows the discussion in PR #166 and fixes use case #156. The changes include
1. Change `MLGraphBuilder.build` and `MLGraph.compute` to sync API. 
  a. Sync API is easy to use and required by sync wasm lib implementation. 
  b. The caller can call the sync API within a webworker to achieve asynchronous https://github.com/webmachinelearning/webnn/issues/156#issuecomment-849665061 @pyu10055 
2. Change `MLGraph.compute` to require pre-allocated output buffers. 
  a. For wasm lib, the output buffers are pre-allcoated in wasm memory. 
  b. For GPU, it is prohibitively expensive to create GPU resources on the fly as a function call's return value https://github.com/webmachinelearning/webnn/pull/166#discussion_r637575735 @wchao1115 
3. Simplify the the input / output resources binding by `MLResource`. Leave `MLInput` only for dynamic input shape case.

Please take a look. Thanks.

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/174.html" title="Last updated on Jun 7, 2021, 1:51 AM UTC (1622f2b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/174/93f8753...huningxin:1622f2b.html" title="Last updated on Jun 7, 2021, 1:51 AM UTC (1622f2b)">Diff</a>

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -40,16 +40,17 @@ const A = builder.input('A', operandType);
 const B = builder.input('B', operandType);
 const C = builder.add(builder.mul(A, constant), B);
 // 2. Compile it into an executable.
-const graph = await builder.build({'C': C});
+const graph = builder.build({'C': C});
 // 3. Bind inputs to the graph and execute for the result.
 const bufferA = new Float32Array(4).fill(1.0);
 const bufferB = new Float32Array(4).fill(0.8);
+const bufferC = new Float32Array(4);
 const inputs = {'A': {data: bufferA}, 'B': {data: bufferB}};
-const outputs = await graph.compute(inputs);
+const outputs = {'C': {data: BufferC}};
+graph.compute(inputs, outputs);
 // The computed result of [[1, 1], [1, 1]] is in the buffer associated with
 // the output operand.
-console.log('Output shape: ' + outputs.C.dimensions);
-console.log('Output value: ' + outputs.C.data);
+console.log('Output value: ' + bufferC);
 ```
 
 Check it out in [WebNN Code Editor](https://webmachinelearning.github.io/webnn-samples/code/?example=mul_add.js).
@@ -102,7 +103,7 @@ export class NSNet2 {
     this.hiddenSize = 400;
   }
 
-  async load(baseUrl, batchSize, frames) {
+  async build(baseUrl, batchSize, frames) {
     const context = navigator.ml.createContext();
     const builder = new MLGraphBuilder(context);
     // Create constants by loading pre-trained data from .npy files.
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1887,49 +1768,26 @@ const a = builder.input('a', descA);
 const descB = {type: 'float32', dimensions: [4, -1]};
 const b = builder.input('b', descB);
 const c = builder.matmul(a, b);
-const graph = await builder.build({c});
+const graph = builder.build({'c': c});
 
-async function compute(shapeA, shapeB) {
+function compute(shapeA, shapeB, shapeC) {
```
anssiko, line 1773: For completeness, I propose we add the `sizeOfShape` util function here:
```
function sizeOfShape(array) {
  return array.reduce(
      (accumulator, currentValue) => accumulator * currentValue);
}
```

anssiko, line 1773: Should we also use a different name for this custom `compute` function that wraps the `MLGraph.compute`? It is doing both buffer allocation and compute.

index.bs
```diff
@@ -1771,103 +1678,77 @@ interface MLGraph {
...
+                    1. Let |outputTensor| be the output tensor returned by |this|.{{MLGraph/[[implementation]]}}.
+                    1. If the kind of |value| is not compatible with the value type of |outputTensor|, then throw a {{DataError}} {{DOMException}} and stop.
+                    1. Let |outputSize| be 1.
+                    1. For each |dimension| of dimensions of |outputTensor|:
+                        1. Set |outputSize| to the product of |outputSize| and |dimension|.
+                    1. If |outputSize| is greater than the length of |value|, then:
+                        1. Throw a {{DataError}} {{DOMException}} and stop.
+                    1. Else:
+                        1. Set the values of |value| to the values of |outputTensor|.
+            1. Return {{undefined}}.
```
anssiko, line 1751: It is not common for Web APIs to return undefined. It is more common to return null. I propose we align with that convention.

anssiko (changes requested):
Thanks @huningxin! I submitted some review comments.

I'd encourage @wchao1115 @pyu10055 @RafaelCintron to take an in-depth look at this PR latest before our next call.

index.bs
```diff
@@ -1728,23 +1639,19 @@ partial interface MLGraphBuilder {
...
-dictionary MLOutput {
-  (MLBufferView or WebGLTexture or GPUTexture) data;
+dictionary MLInput {
+  required MLResource resource;
   sequence<long> dimensions;
 };
 
-typedef record<DOMString, MLInput> MLNamedInputs;
-typedef record<DOMString, MLOutput> MLNamedOutputs;
+typedef record<DOMString, (MLResource or MLInput)> MLNamedInputs;
```
wchao1115, line 1649: I think this is a good idea as it removes a level of indirection in the case when just a resource is passed in, but should the `dimensions` field in `MLInput` now be required?

wchao1115 (changes requested):

index.bs
huningxin: Good point. I fixed it in the new commit.

huningxin: Good idea. I renamed it to `allocateAndCompute` in the new commit. Please take a look.

index.bs
huningxin: I follow [WebIDL](https://heycam.github.io/webidl/#idl-undefined) that uses `undefined` as `void` to define a function returning nothing. There are two spec examples of this usage: [forEach](https://heycam.github.io/webidl/#es-forEach) and [ApplyConstraints](https://www.w3.org/TR/mediacapture-streams/#dfn-applyconstraints-algorithm)

index.bs
huningxin: Good catch, it should be required. I fixed it in the new commit. Please take a look.

wchao1115 (approved):

index.bs
anssiko: Thanks! That is correct usage. I misunderstood the semantics.

For reference:
https://github.com/heycam/webidl/pull/906
https://github.com/heycam/webidl/issues/60

anssiko (approved):

anssiko:
@pyu10055 @RafaelCintron we appreciate your quick review of this PR before this Thursday's call.

explainer.md
```diff
@@ -102,7 +103,7 @@ export class NSNet2 {
     this.hiddenSize = 400;
   }
 
-  async load(baseUrl, batchSize, frames) {
+  async build(baseUrl, batchSize, frames) {
```
pyu10055, line 106: you need to remove async keyword if build method is sync.

pyu10055 (approved):

explainer.md
huningxin: This is needed, because `buildConstantByNpy` in this function uses `fetch` to download `.npy` files from network which is async.

huningxin:
Thanks for the review and approval. I am going to merge this PR.

Issue #185, opened 2021-06-24
Conv2d and BatchNorm should support optional fused activation function


wchao1115:
When webnn is used as a target backend of frameworks, it is likely given a convolution or batchnorm operation together with a fused `relu` or other activation function that are widely known fusions supported in most frameworks today including ONNX Runtime and TensorFlow.js. Here is the compiled list of known activation functions that can be specified as a fused activation to these two operations.

``` javascript
enum MLFusedActivation {
  "elu",
  "hard-sigmoid",
  "leaky-relu"
  "linear",
  "relu",
  "selu",
  "sigmoid",
  "softplus",
  "softsign",
  "tanh"
};
```

huningxin:
Thanks @wchao1115 . According to my investigations, the frameworks also usually fuse the bias into conv2d. It would be good if webnn supports the fused bias in conv2d op.

WDYT?

huningxin:
Regarding to the `MLFusedActivation`, should it include the `clamp`? It is useful to support `relu6` of MobileNet.

wchao1115:
Sounds good. I'll add `clamp` and `relu6` in the list as well. Regarding fused bias, can you give an example of the backend API that supports that? Note that frameworks can fuse anything at high level but without a broad backend support, the implementation will be forced to decouple the fused operation anyway. So I think we should consider fused operation first in the context of platform support.

wchao1115:
BTW I'm going to avoid the never-ending debate by not asking "why 6?" ;-)

huningxin:
> Regarding fused bias, can you give an example of the backend API that supports that?

A couple of examples:
 * NNAPI's [`ANEURALNETWORKS_CONV_2D`](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0a34a73b5eaf458b67db5eda71557d1d01) supports fused bias via the third operand, "2: A 1-D tensor, of shape [depth_out], specifying the bias".
 * DirectML's [`DML_CONVOLUTION_OPERATOR_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_convolution_operator_desc) supports fused bias via the member `const DML_TENSOR_DESC     *BiasTensor;`.
 * MLCompute's [`MLCConvolutionLayer`](https://developer.apple.com/documentation/mlcompute/mlcconvolutionlayer) supports fused bias via `biases: MLCTensor?` parameter.
[...]

Issue #138, opened 2021-02-04
String enum for activations


anssiko:
>7. If there are layers that will be taking activations as string enums, there should simply be a string enum for activations rather than have it just in RecurrentNetworkActivation. (One may argue that hyperbolic tangent is RNN specific, but...

via https://github.com/w3ctag/design-reviews/issues/570#issuecomment-768875996

Some activations can not be supported in GRU, will clarify in the spec.

wchao1115:
@cynthia As @anssiko pointed out, there are a lot of activation functions and new ones are discovered from time to time, however not all activations are suitable for recurrent network and other categories of networks, so having a single enum that defines every possible activation functions would create a maintenance burden and that it could be misused i.e. an unsupported activation in a network would then need to be handled as a runtime error, which complicates the caller's code somewhat.

wchao1115:
@anssiko Can this issue be closed? Is there any other aspect of the original feedback?

cynthia:
> so having a single enum that defines every possible activation functions would create a maintenance burden and that it could be misused

Is there any prior art (as in, a neural network API) that restricts this though? I checked and activating a GRU with relu seemed to work fine in all of the frameworks I tried - even if it's nonsensical.

wchao1115:
A problem with defining all the known activations as a single enum that is used everywhere is that every operation that uses it will be forced to either support all of it, or having an ever-changing logic over time that accepts some and rejects some. This is very unpredictable and hard to maintain the right set of expectation over time, from the API longevity standpoint. 

In the use case of `gruCell` there is in fact a known use of `relu` activation in place of `tanh` for the recurrent gate (also known as the 'new' gate), but stretching that to cover many more may be a bridge too far. That is why `relu` is currently part of the [`MLRecurrentNetworkActivation`](https://webmachinelearning.github.io/webnn/#enumdef-mlrecurrentnetworkactivation) enum, but not others.

Remember that the main purpose of defining these networks such as the recurrent ones as standalone operations is to provide a pathway for the implementer to hook them up directly with the optimizations already built into many hardware platforms today. For the apps, these "big" operations are seen as just short hands for the common known cases. But from the implementer's point of view, it is a huge performance boost to be able to tap into the existing silicon blocks or optimized instructions specially built for it.  

[...]

PR #188, opened 2021-06-28
Add activation functions as fused operators to BatchNorm and Conv2d

wchao1115:
As part of this change, I generalize a way to allow an operator to be created before it is connected to the graph. This makes it possible for fused operators, particularly the various activation functions to be expressed by a generic operator type instead of through an ever-expanding list of enum values, which addresses both #185 and #138. 

Mechanically, a notion of `MLOperator` is defined. This is a definition of a type of operation and not the operation itself, as no input operands are needed in the creation of an operator and no operand is expected from it. The operator will then connect with the rest of the graph via the graph builder once the input operands are ready.  The best example of this mechanic is demonstrated in the note section of the `batchNormalization` opeation where an activation function is passed into as an `MLOperator` and not as `MLOperand`. The builder's method `setOperand` connects the operator with the input operands and get an output operand out of it. 

As for #138, much of the discussion in the issue is centered around the need to not prematurely leave out other possible activation functions that could be supported by the future recurrent network. But maintaining an ever-expanding list of enum values for all the possible (present and future) activation functions is error-prone and in a way giving out a wrong and confusing signal to the API implementer who certainly can only base their implementation on the current underlying platform support.

This change addresses the need to keep the list open while avoiding conveying an unrealistic expectation to the implementer by not defining the list of supported activation functions in the description of the operation itself but instead leaving it out to the caller. Note that the graph building operation would not fail even when an unsupported activation function is hooked up, but the subsequent graph compilation step `builder.build` would.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/188.html" title="Last updated on Jul 11, 2021, 12:52 AM UTC (5bdc94b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/188/a913875...5bdc94b.html" title="Last updated on Jul 11, 2021, 12:52 AM UTC (5bdc94b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -529,6 +529,12 @@ dictionary MLOperandDescriptor {
 interface MLOperand {};
 </script>
 
+## MLOperator ## {#api-mloperator}
+<script type=idl>
+[SecureContext, Exposed=(Window, DedicatedWorker)]
+interface MLOperator {};
+</script>
+
 ## MLGraphBuilder ## {#api-mlgraphbuilder}
 
 The {{MLGraphBuilder}} interface defines a set of operations as identified by the [[#usecases]] that can be composed into a computational graph. It also represents the intermediate state of a graph building session.
@@ -571,6 +577,7 @@ dictionary MLBatchNormalizationOptions {
   MLOperand bias;
   long axis = 1;
   float epsilon = 1e-5;
+  MLOperator activation;
 };
 
 partial interface MLGraphBuilder {
@@ -588,6 +595,7 @@ partial interface MLGraphBuilder {
               - *bias*: an {{MLOperand}}. The 1-D tensor of the bias values whose length is equal to the size of the input dimension denoted by *options.axis*.
               - *axis*: a {{long}} scalar. The index to the feature count dimension of the input shape for which the mean and variance values are. When it's not specified, the default value is 1.
               - *epsilon*: a {{float}} scalar. A small value to prevent computational error due to divide-by-zero. The default value is 0.00001 when not specified.
+              - *activation*: an {{MLOperator}}. The optional activation function that immediately follows the normalization operation.
         
     **Returns:** an {{MLOperand}}. The batch-normalized N-D tensor of the same shape as the input tensor.
 
@@ -599,7 +607,9 @@ partial interface MLGraphBuilder {
     therefore its usage is encouraged from the performance standpoint.
     <pre highlight="js">
     const shape = [1,-1,1,1];
-    return builder.add(
+    return builder.setOperand(
+      activation,
+      [builder.add(
         builder.mul(
           builder.reshape(options.scale, shape),
           builder.div(
@@ -607,10 +617,8 @@ partial interface MLGraphBuilder {
             builder.pow(
               builder.add(builder.reshape(variance, shape), builder.constant(options.epsilon)),
               builder.constant(0.5))
-            )
-          ),
-        builder.reshape(options.bias, shape)
-      );
+            )),
+        builder.reshape(options.bias, shape))]);
     </pre>
     </div>
 </div>
@@ -625,6 +633,7 @@ dictionary MLClampOptions {
 
 partial interface MLGraphBuilder {
   MLOperand clamp(MLOperand x, optional MLClampOptions options = {});
+  MLOperator clamp(optional MLClampOptions options = {});
 };
 </script>
 <div algorithm=clamp>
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1499,6 +1640,54 @@ partial interface MLGraphBuilder {
     tensor is specified by the *newShape* argument.
 </div>
 
+### setOperand ### {#api-mlgraphbuilder-set-operand}
```
huningxin, line 1643: I suppose `setOperand` should be the implementation detail. 

It is because when building a graph, if the user code wants to fuse the activation, it would call the `MLOperator` version, e.g. `output = builder.conv2d(input, filter, {activation: builder.relu()})`. Otherwise, the caller would call the `MLOperand` version, e.g. `output = builder.relu(builder.conv2d(input, filter))`. The `setOperand` seems not useful for user code.

Any other usages I missed?

index.bs
```diff
@@ -599,18 +607,18 @@ partial interface MLGraphBuilder {
     therefore its usage is encouraged from the performance standpoint.
     <pre highlight="js">
     const shape = [1,-1,1,1];
-    return builder.add(
+    return builder.setOperand(
+      activation,
```
huningxin, line 611: As the sample code, I suppose it could just use one type of activation here, such as `builder.relu`.

index.bs
```diff
@@ -529,6 +529,12 @@ dictionary MLOperandDescriptor {
 interface MLOperand {};
 </script>
 
+## MLOperator ## {#api-mloperator}
+<script type=idl>
+[SecureContext, Exposed=(Window, DedicatedWorker)]
+interface MLOperator {};
```
huningxin, line 535: For consistency, should we name it as `MLOperation`? This spec uses "operation" term in other places.

index.bs
wchao1115: Yes, in operations like `batchNorm` or `gruCell`. You're right it could be treated as a hidden concept for atomic operations like `conv2d`. For composite operations that happen to accept fused operators, it should be possible to implement them purely as composites of smaller ops, as the pseudo-code in their note sections suggests.

index.bs
wchao1115: Maybe it's just a wording thing but I tend to think of an *operation* as an *operator* with *operands* e.g. A + B is an operation, while the *operator* is just the act imposed on the operands e.g. + is the operator. 

I think of `conv2d` as an operation since it accepts operands as required params, while `relu` could either be an operation or an operator, depending on whether it is created with or without an operand. Does that make sense?

index.bs
wchao1115: The implementation of this operation will not know ahead of time what activation is parameterized here. This is why I think we also need to define `setOperand`.

index.bs
huningxin: Thanks for your explanation. It makes sense to me.

index.bs
huningxin: I suppose an implementation could know the activation type internally. For example, an implementation may add a internal slot into `MLOperator` that keeps the activation type, e.g. the `MLOperator` instance created by `builder.relu()` has the type `relu`. Later, when the implementation composites the `batchNormalization`, it can check the type of the `MLOperator`.

anssiko (approved):
Thank you for this substantial update!

I’ll unblock my review so you can go ahead after addressing all review feedback. Should this PR require more extensive discussion, we can bring this to our post-break call in mid-Aug.

index.bs
gramalingam: How about a name like ```apply``` instead of ```setOperand```? I think it is better if we use a name that indicates the side-effect-free functional behavior rather than one that suggests an imperative behavior with side-effects. I think this is essentially the operation of applying a function (represented by operator) to its arguments.

wchao1115: Thanks for the feedback @gramalingam. I like that.

index.bs
wchao1115: What do you think about moving `setOperand` to the `MLOperator` itself instead of on the builder interface? That way the builder's implementation doesn't need to track the operator type of each operator internally, which could be tedious and ineffective. Instead just have the implementation of the supported operators (e.g. activation functions) support this additional capability themselves without effecting how other operations are implemented. This should address your concern, if I sense it correctly.

huningxin: Moving `setOperand` to `MLOperator` looks better. However my point is either `MLOperator.setOperand/apply` or `MLOperator.type` should be internal to implementation. If there are no use cases for user code, I would suggest exclude it in this PR.

wchao1115: An implementation that doesn't want to handle composite operations like `batchNorm` natively should be able to implement it as if it's a graph of smaller operations, like what being described in the Note section. Remember that that is one of our design principles -- that every big op defined must be publicly implementable by compositing smaller public constructs? 

Although we do expect operations like `batchNorm` to be implemented natively, some implementation on some platform may not have a mean to do that, and would only settle on a default implementation as a graph of public operations.

huningxin: > An implementation that doesn't want to handle composite operations like `batchNorm` natively should be able to implement it as if it's a graph of smaller operations,  like what being described in the Note section.

How about the Note section says "When the input tensor is 4-D of the "nchw" layout and ___the activation is a relu operator___, the behavior of this operation can be generically emulated from the usage of other operations as follow."

```js
const shape = [1,-1,1,1];
return builder.relu(
  builder.add(
    builder.mul(
      builder.reshape(options.scale, shape),
      builder.div(
        builder.sub(input, builder.reshape(mean, shape)),
        builder.pow(
          builder.add(builder.reshape(variance, shape), builder.constant(options.epsilon)),
          builder.constant(0.5))
        )),
    builder.reshape(options.bias, shape)));
```

> some implementation on some platform may not have a mean to do that,

That's true. For example, the webnn-polyfill implementation uses TF.js's [`tf.batchNorm`](https://js.tensorflow.org/api/latest/#batchNorm) that doesn't support fused activation. 

> and would only settle on a default implementation as a graph of public operations

This might not be true. For example, webnn-polyfill implements WebNN spec defined ops as one or multiple TF.js kernels. For fused `batchNormalziation`, it could call the corresponding Tf.js kernel of an `MLOperator`, e.g. `tf.relu`, right after `tf.batchNorm`. That doesn't require the `setOperand` as a public API.

huningxin: @wchao1115 , I put my idea into https://github.com/webmachinelearning/webnn/pull/193, please take a look. Thanks.

wchao1115: Thanks for the feedback @huningxin . Can you please take a look at the new commit? I adapted your suggestion in this update.

huningxin (approved):
LGTM, thanks for the update!

Issue #194, opened 2021-07-16
There is a spelling error in "during the model training phrase"


fujunwei:
Does it means "during the model training `phase`" describing in [instanceNormalization](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-instancenorm) and [batchNormalization](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-instancenorm)?

And is the "[=======](https://github.com/webmachinelearning/webnn/blob/main/index.bs#L474)" delimiter not required?
![Capture2](https://user-images.githubusercontent.com/7856909/125877914-3bbbd76b-07a4-4a98-95d3-74845ba48ff8.PNG)

anssiko:
I believe this is fixed by #195 (thanks @huningxin!).

PR #195, opened 2021-07-19
Support hard-swish operation

### LINKED ISSUE #181: Request for HardSwish op
`HardSwish` is an important activation function in MobileNetV3. Both TensorFlow and ONNX have supported it.

huningxin:
fix #181 , 

This PR also fixes
- some typo raised by #194
- change the local Makefile to die on error that aligns to the [auto-publish.yml](https://github.com/webmachinelearning/webnn/blob/main/.github/workflows/auto-publish.yml).

@wchao1115  @pyu10055 PTAL. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/195.html" title="Last updated on Jul 21, 2021, 2:25 AM UTC (721ed65)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/195/31a71f6...huningxin:721ed65.html" title="Last updated on Jul 21, 2021, 2:25 AM UTC (721ed65)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1233,6 +1233,41 @@ partial interface MLGraphBuilder {
     </div>
 </div>
 
+### hardSwish ### {#api-mlgraphbuilder-hard-swish}
+Computes the hard-swish activation that is introduced by [[MobileNetV3]] on the input tensor element-wise.
+<script type=idl>
+partial interface MLGraphBuilder {
+  MLOperand hardSwish(MLOperand x);
+  MLOperator hardSwish();
+};
+</script>
+<div algorithm=hard-swish>
+    **Arguments:**
+        - *x*: an {{MLOperand}}. The input tensor.
+
+    **Returns:**
+        - an {{MLOperand}}. The output tensor of the same shape as *x*.
+        - an {{MLOperator}}. The operator representing the hard-swish operation.
+
+    <div class="note">
+    The behavior of this operation can be generically emulated from the usage of
+    other operations as follow. However, user agents typically have a more
+    efficient implementation for it, therefore its usage is encouraged from the
+    performance standpoint.
+    <pre highlight="js">
+    return builder.div(
+               builder.mul(
+                   x,
+                   builder.max(
+                       builder.constant(0),
+                       builder.min(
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1233,8 +1232,43 @@ partial interface MLGraphBuilder {
     </div>
 </div>
 
+### hardSwish ### {#api-mlgraphbuilder-hard-swish}
+Computes the hard-swish activation that is introduced by [[MobileNetV3]] on the input tensor element-wise.
```
wchao1115, line 1236: Probably remove the word "hard swish" from the description.

huningxin: How about we say "Computes the h-swish nonlinearity that is introduced by [[MobileNetV3]] ... "? "h-swish nonlinearity" is the term used in the reference paper: https://arxiv.org/pdf/1905.02244.pdf

wchao1115: I don't know enough that the term h-swish is a generic mathematical term of just a random name chosen randomly. But my point is that if there's a credible literature that backs the term and that it is widely known e.g. relu, then it's fine, but if not, it's probably better off just leaving the description as generic. @anssiko may have some opinion. It's just a nit though, you shouldn't be blocked by this.

huningxin: > it's probably better off just leaving the description as generic

Agree. I addressed it in the new commit. Please take another look. Thanks!

wchao1115 (approved):

index.bs
anssiko: The change that landed LGTM.

The best practice is to use generic language where possible and if using specialized language back it up with good reference(s) that define the new terms or concepts. The spec may also define new initially only locally used terms and concepts, and if there's a possibility of reuse these are to be annotated using the [definitions syntax](https://tabatkins.github.io/bikeshed/#definitions). [Autolinking](https://tabatkins.github.io/bikeshed/#autolinking) then creates cross-references to any existing definitions for reuse and checks for name clashes.

This is a good test: a reader who is familiar with the domain is able to understand the text without expanding references too much.

PR #204, opened 2021-08-29
Add default values of scales into resample op

### LINKED ISSUE #192: The scales and sizes of MLResampleOptions can't be both empty
As the spec defined, `scales` and `sizes` are hosted in the `MLResampleOptions`, both of them could be empty. I think when neither `scales` nor `sizes` is defined, the `resample` operation shouldn't be created.

huningxin:
fix #192 

@wchao1115 @pyu10055 @anssiko , PTAL. Thanks.

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/204.html" title="Last updated on Aug 29, 2021, 12:54 AM UTC (006f05f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/204/dc85541...huningxin:006f05f.html" title="Last updated on Aug 29, 2021, 12:54 AM UTC (006f05f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1702,7 +1702,7 @@ partial interface MLGraphBuilder {
         - *options*: an optional {{MLResampleOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLInterpolationMode}}. The interpolation algorithm used to fill the output tensor values.
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
-            - *scales*: a sequence of {{float}} of length 4. Each value represents the scaling factor used to scale in each input dimensions.
+            - *scales*: a sequence of {{float}} of length 4. Each value represents the scaling factor used to scale in each input dimensions. If not set, the values are assumed to be [1.0, 1.0, 1.0, 1.0].
             - *sizes*: a sequence of {{long}} of length 4. The target sizes for each input dimensions. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each input dimension.
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor.


index.bs
```diff
@@ -1702,7 +1702,7 @@ partial interface MLGraphBuilder {
         - *options*: an optional {{MLResampleOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLInterpolationMode}}. The interpolation algorithm used to fill the output tensor values.
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
-            - *scales*: a sequence of {{float}} of length 4. Each value represents the scaling factor used to scale in each input dimensions.
+            - *scales*: a sequence of {{float}} of length 4. Each value represents the scaling factor used to scale in each input dimensions. If not set, the values are assumed to be [1.0, 1.0, 1.0, 1.0].
```
anssiko, line 1705: I'll loop in @dontcallmedom for his advice, given this is an instantiation of https://github.com/webmachinelearning/webnn/pull/205#discussion_r701070910

We have a sequence in an IDL fragment, so we learned we cannot use the WebIDL default value shorthand ("=").

The custom prose the spec has been using is:

>If X is not set, the values are assumed to be Y.

Should we use this or maybe something like the following:

>The [default value](https://heycam.github.io/webidl/#dfn-dictionary-member-default-value) is Y.

Or perhaps a longer version that explicitly specifies the default value is used if the value is "not present" or "is invalid"? (I think we may want to branch into two in some cases.)

>If X is not present or its value invalid, set X to [1.0, 1.0, 1.0, 1.0].

We expect to use the same pattern in many places so would like to identify the most concise way.

Some notes:

- WebIDL spec had a [present](https://www.w3.org/TR/WebIDL-1/#dfn-present) dfn for testing dictionary members, but it was removed at some point?

- For attributes there [invalid value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#invalid-value-default) and [missing value default](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#missing-value-default) microsyntaxes, but nothing for dictionaries?

anssiko (approved):

index.bs
dontcallmedom: > The custom prose the spec has been using is:
> 
> > If X is not set, the values are assumed to be Y.
> 
> Should we use this or maybe something like the following:
> 
> > The [default value](https://heycam.github.io/webidl/#dfn-dictionary-member-default-value) is Y.

This LGTM - paging @domenic and @timothyxu to confirm the wording indeed works, and whether a non-empty array as a default value doesn't create other issues from an IDL perspective.

> Or perhaps a longer version that explicitly specifies the default value is used if the value is "not present" or "is invalid"? (I think we may want to branch into two in some cases.)

The term "present" has been replaced by Infra's [exists](https://infra.spec.whatwg.org/#map-exists) (in WebIDL and in specs that rely on this concept).

An "invalid" value would cause a `TypeError` so there is no need to describe that situation.

anssiko: Thanks @dontcallmedom!

domenic: This doesn't really work, no. You need to have a normative algorithm which checks if input["scales"] exists, and if not, use the defaults. I.e. something like

>1. Let _scales_ be the [list](https://infra.spec.whatwg.org/#list) « 1.0, 1.0, 1.0, 1.0 ».
>2. If _input_["scales"] exists, then set _scales_ to _input_["scales"].

(and maybe also validate that input["scales"] is of length 4, or whatever you need?)

This seems difficult since right now your spec has no algorithms (i.e. no [method steps](https://heycam.github.io/webidl/#method-steps)) for its operations, just a non-normative description of the inputs and return value.

anssiko: Thanks @domenic! Your feedback is now captured in https://github.com/webmachinelearning/webnn/issues/211 so we track and address it properly.

wchao1115 (approved):

index.bs
huningxin: Thanks for your feedback @domenic . We'll fix #211 properly. To unblock the development of #208 , I would suggest to merge this one. @anssiko , what do you think?

anssiko: @huningxin, I was just thinking of suggesting the same, given we track #211 separately.

Please go ahead and merge.

PR #205, opened 2021-08-30
Change resample to resample2d

### LINKED ISSUE #200: Request input layout and resize only on height and width for Resample
OpenVINO only [support NHWC layout and resize on height and width for rank 4](https://github.com/openvinotoolkit/openvino/blob/master/inference-engine/src/mkldnn_plugin/nodes/mkldnn_interpolate_node.cpp#L2047).

And NNAPI support only height and width of the output tensor and has  [the optional of "NHWC" or "NCHW" data layout](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0a42bd92518e273b6716ecd56b571fcd3e) .

huningxin:
fix #200 

This change supports resample2d:
* resize the spatial dimensions
* specify the input layout

@wchao1115 @pyu10055 @anssiko , please take a look.

cc @fujunwei.

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/205.html" title="Last updated on Sep 26, 2021, 2:56 AM UTC (cf99237)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/205/4fc42a4...huningxin:cf99237.html" title="Last updated on Sep 26, 2021, 2:56 AM UTC (cf99237)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1690,7 +1690,7 @@ enum MLInterpolationMode {
 
 dictionary MLResample2dOptions {
   MLInterpolationMode mode = "nearest-neighbor";
-  sequence<float> scales;
+  sequence<float> scales = [1.0, 1.0];
   sequence<long> sizes;
   MLInputOperandLayout layout = "nchw";
 };
@@ -1705,9 +1705,9 @@ partial interface MLGraphBuilder {
         - *options*: an optional {{MLResample2dOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLInterpolationMode}}. The interpolation algorithm used to fill the output tensor values.
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
-            - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width]. If not set, the values are assumed to be [1.0, 1.0].
+            - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width].
             - *sizes*: a sequence of {{long}} of length 2. The target sizes for each spatial dimensions of input, [size_height, size_width]. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each spatial dimension of input.
-            - *layout*: an {{MLInputOperandLayout}}. This option specifies the layout format of the input. The default value is *"nchw"*.
+            - *layout*: an {{MLInputOperandLayout}}. This option specifies the layout format of the input.
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor.
 </div>
--- a/index.bs
+++ b/index.bs
@@ -1692,7 +1692,7 @@ dictionary MLResample2dOptions {
   MLInterpolationMode mode = "nearest-neighbor";
   sequence<float> scales;
   sequence<long> sizes;
-  MLInputOperandLayout layout = "nchw";
+  sequence<long> axes;
 };
 
 partial interface MLGraphBuilder {
@@ -1707,7 +1707,7 @@ partial interface MLGraphBuilder {
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
             - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width]. If not set, the values are assumed to be [1.0, 1.0].
             - *sizes*: a sequence of {{long}} of length 2. The target sizes for each spatial dimensions of input, [size_height, size_width]. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each spatial dimension of input.
-            - *layout*: an {{MLInputOperandLayout}}. This option specifies the layout format of the input.
+            - *axes*: a sequence of {{long}} of length 2. The two consecutive dimensions of the input tensor to which the interpolation algorithm applies. The valid values in the sequence are [0, 1], [1, 2] or [2, 3]. When not specified, the sequence is assumed to be [2, 3].
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor.
 </div>
--- a/index.bs
+++ b/index.bs
@@ -1680,32 +1680,34 @@ partial interface MLGraphBuilder {
     </div>
 </div>
 
-### resample ### {#api-mlgraphbuilder-resample}
-Resample the tensor values from the source to the destination dimensions according to the scaling factors.
+### resample2d ### {#api-mlgraphbuilder-resample2d}
+Resample the tensor values from the source to the destination spatial dimensions according to the scaling factors.
 <script type=idl>
 enum MLInterpolationMode {
   "nearest-neighbor",
   "linear"
 };
 
-dictionary MLResampleOptions {
+dictionary MLResample2dOptions {
   MLInterpolationMode mode = "nearest-neighbor";
   sequence<float> scales;
   sequence<long> sizes;
+  MLInputOperandLayout layout = "nchw";
 };
 
 partial interface MLGraphBuilder {
-  MLOperand resample(MLOperand input, optional MLResampleOptions options = {});
+  MLOperand resample2d(MLOperand input, optional MLResample2dOptions options = {});
 };
 </script>
-<div algorithm=resample>
+<div algorithm=resample2d>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input 4-D tensor.
-        - *options*: an optional {{MLResampleOptions}}. The optional parameters of the operation.
+        - *options*: an optional {{MLResample2dOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLInterpolationMode}}. The interpolation algorithm used to fill the output tensor values.
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
-            - *scales*: a sequence of {{float}} of length 4. Each value represents the scaling factor used to scale in each input dimensions. If not set, the values are assumed to be [1.0, 1.0, 1.0, 1.0].
-            - *sizes*: a sequence of {{long}} of length 4. The target sizes for each input dimensions. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each input dimension.
+            - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width]. If not set, the values are assumed to be [1.0, 1.0].
+            - *sizes*: a sequence of {{long}} of length 2. The target sizes for each spatial dimensions of input, [size_height, size_width]. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each spatial dimension of input.
+            - *layout*: an {{MLInputOperandLayout}}. This option specifies the layout format of the input. The default value is *"nchw"*.
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor.
 </div>
--- a/index.bs
+++ b/index.bs
@@ -1690,7 +1690,7 @@ enum MLInterpolationMode {
 
 dictionary MLResample2dOptions {
   MLInterpolationMode mode = "nearest-neighbor";
-  sequence<float> scales = [1.0, 1.0];
+  sequence<float> scales;
   sequence<long> sizes;
   MLInputOperandLayout layout = "nchw";
 };
@@ -1705,7 +1705,7 @@ partial interface MLGraphBuilder {
         - *options*: an optional {{MLResample2dOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLInterpolationMode}}. The interpolation algorithm used to fill the output tensor values.
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
-            - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width].
+            - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width]. If not set, the values are assumed to be [1.0, 1.0].
             - *sizes*: a sequence of {{long}} of length 2. The target sizes for each spatial dimensions of input, [size_height, size_width]. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each spatial dimension of input.
             - *layout*: an {{MLInputOperandLayout}}. This option specifies the layout format of the input.
 


index.bs
```diff
@@ -1678,32 +1678,34 @@ partial interface MLGraphBuilder {
...
+<div algorithm=resample2d>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input 4-D tensor.
-        - *options*: an optional {{MLResampleOptions}}. The optional parameters of the operation.
+        - *options*: an optional {{MLResample2dOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLInterpolationMode}}. The interpolation algorithm used to fill the output tensor values.
                 If not set, it is assumed to be the *Nearest Neighbor* interpolation.
-            - *scales*: a sequence of {{float}} of length 4. Each value represents the scaling factor used to scale in each input dimensions.
-            - *sizes*: a sequence of {{long}} of length 4. The target sizes for each input dimensions. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each input dimension.
+            - *scales*: a sequence of {{float}} of length 2. Each value represents the scaling factor used to scale in each spatial dimensions of input, [scale_height, scale_width]. If not set, the values are assumed to be [1.0, 1.0].
```
anssiko, line 1706: Alternative to "If not set, the values are assumed to be [1.0, 1.0]." is to use a [dictionary member default value](https://heycam.github.io/webidl/#dfn-dictionary-member-default-value) for scales in IDL:

`sequence<float> scales = [1.0, 1.0];`

What would be the expected behavior when only either scale_height or scale_width is set?

index.bs
```diff
@@ -1678,32 +1678,34 @@ partial interface MLGraphBuilder {
...
+            - *sizes*: a sequence of {{long}} of length 2. The target sizes for each spatial dimensions of input, [size_height, size_width]. When the target sizes are specified, the *options.scales* argument is ignored as the scaling factor values are derived from the target sizes of each spatial dimension of input.
+            - *layout*: an {{MLInputOperandLayout}}. This option specifies the layout format of the input. The default value is *"nchw"*.
```
anssiko, line 1708: Given the default value is defined in IDL, "The default value is *"nchw"*." is redundant.

anssiko (changes requested):
Thanks @huningxin! I had a few minor comments.

index.bs
huningxin: > `sequence<float> scales = [1.0, 1.0];`

Looks good, I'll address it in the new commit.

> What would be the expected behavior when only either scale_height or scale_width is set?

Since the spec says, it should be a sequence of {{float}} of length 2, it would throw error when only either scale_height or scale_width is set.

index.bs
huningxin: addressed in the new commit

index.bs
huningxin: > sequence<float> scales = [1.0, 1.0];

Unfortunately, It triggers a bikeshed error of the bot: https://github.com/webmachinelearning/webnn/pull/205/checks?check_run_id=3495820911

```
 [INFO] Build bikeshed document "index.bs" (gh)…
    $ bikeshed --die-on=link-error spec "index.bs" "index.bs.built.html" 
      FATAL ERROR: IDL SYNTAX ERROR LINE: 9 - expected ";" skipped: " = [1.0, 1.0]"
       ✘  Did not generate, due to fatal errors
    Failed
  Error: Process completed with exit code 2.
```

dontcallmedom: [only an empty array `[]`](https://heycam.github.io/webidl/#index-prod-DefaultValue) is a valid default for a sequence in an IDL fragment; I think the prose version is the best we can do

anssiko: Thanks @dontcallmedom, @huningxin let's revert to the prose definition then.

anssiko (approved):

index.bs
huningxin: Thanks @dontcallmedom , fixed in the new commit.

jbingham:
@pyu10055 can you review? Chai is away. Also, we can check with the Android NN API team.

index.bs
```diff
@@ -1678,32 +1678,34 @@ partial interface MLGraphBuilder {
     </div>
 </div>
 
-### resample ### {#api-mlgraphbuilder-resample}
-Resample the tensor values from the source to the destination dimensions according to the scaling factors.
+### resample2d ### {#api-mlgraphbuilder-resample2d}
+Resample the tensor values from the source to the destination spatial dimensions according to the scaling factors.
```
wchao1115, line 1682: `resample` is defined such that the interpolation can work across any dimensions and not necessarily restricted to just the spatial dimensions of known layouts e.g. NCHW. This is also the intent of the ONNX `resize` operator. 

The DirectML [`DML_RESAMPLE_OPERATOR_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_resample_operator_desc) works the same way i.e. if you intend to interpolate the spatial dimensions, say for NCHW layout, while keeping the batch and channel dimension fixed, then you can either keep the first and second value of the `scales` array at 1 or keep the first and the second value of the `sizes` array as the same values as the input tensor's first and second dimension values.  Hope it helps.

wchao1115 (changes requested):

index.bs
wchao1115: Sorry, I hit the wrong review button. My feedback is non-blocking. Just thought that the current definition of `resample` is already flexible enough to address the issue #200.

huningxin: Thanks @wchao1115 , it is very helpful. 

> Just thought that the current definition of `resample` is already flexible enough to address the issue #200.

I agree the current `resample` is flexible. However, I understand the issue #200 is about other native ML APIs, such as NNAPI and OpenVINO, only support interpolating the spatial dimensions. I suppose it is not easy to be emulated by other small ops. Also, from the use case perspective, the first-wave models only use the spatial dimensions interpolation. So I propose we support resample2d in the v1 spec.

wchao1115: I think that's a fair point, but restricting `resample` to just spatial dimensions will limit its ability to support ONNX. The `resize` operator in ONNX has been defined this way since its inception in opset-10. That said, the majority of cases will be for spatial interpolations; the non-spatial ones should not be very common.

As for whether interpolation across non-spatial dimensions can be emulated in NNAPI, it looks like that is possible through ANEURALNETWORKS_STRIDED_SLICE. It will be less efficient but it should work, just strided slice the input tensor first, then proceed with spatial interpolation. An even better option would be to extend the [`slice`](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-slice) operation to support a `strides` param in addition to `start` and `sizes`. That way the emulation can be done thru webnn operations itself.

huningxin: Thanks @wchao1115 .

> That said, the majority of cases will be for spatial interpolations; the non-spatial ones should not be very common.

That's true. It looks like both [tf.image.resize](https://www.tensorflow.org/api_docs/python/tf/image/resize) and [torch.nn.functional.interpolate](https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html) (when input is 4-D) support spatial resampling. BTW, could you please share the use cases that use non-spatial ones? We probably can check with the group whether there are interests to support them.

> As for whether interpolation across non-spatial dimensions can be emulated in NNAPI, it looks like that is possible through `ANEURALNETWORKS_STRIDED_SLICE`.

According to the NNAPI execution provider of ONNXRuntime, its [`ResizeOpBuilder`](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/nnapi/nnapi_builtin/builders/op_builder.cc#L2411-L2412) seems to just ignore the scales of batch and channel dimension. Could you please detail how to emulate interpolation across non-spatial dimensions with `ANEURALNETWORKS_STRIDED_SLICE`?

index.bs
anssiko: I looped in @dontcallmedom in https://github.com/webmachinelearning/webnn/pull/204#discussion_r702866958 that I believe is superseded by this PR, but the same questions still apply.

I'll unblock my review, since we can make the prose updates separately.

anssiko (approved):

index.bs
wchao1115: > Could you please detail how to emulate interpolation across non-spatial dimensions with ANEURALNETWORKS_STRIDED_SLICE?

Actually it might even be easier to emulate using `split` and `concat` e.g. given the input tensor of shape `[2, 3, 4]` being  resampled across dimension 1 and 2 (instead of 2 and 3), one could split the input along dimension 2, producing a sequence of four `[1, 1, 2, 3]` sub-tensors. Each sub-tensor will then be resampled individually before being concatenated back together, also along dimension 2. Could that work?

huningxin: `split` and `concat` could work for this case. However, if the interpolation would be across three dimensions, for example dimension 1, 2 and 3, e.g. for 3D volumes. How would this work?

wchao1115: No, the `split` and `concat` trick is only to support 2D interpolation across any two dimensions. I don't think we can easily emulate any N-D interpolation using this trick. 

The wording of the ONNX [`resize` operator](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Resize) makes it sound like it can support N-D interpolation across any multiple dimensions, but in reality, ONNX Runtime validates that only 2 scale dimensions with values other than 1 is allowed. See [this thread](https://github.com/microsoft/onnxruntime/issues/2463).

DirectML [`DML_RESAMPLE_OPERATOR_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_resample1_operator_desc) supports up to 4-D interpolation, that is it can handle 2, 3, or 4 scale dimensions with values other than 1. This is necessary in order to support frameworks other than ONNX e.g. PyTorch [`interpolate`](https://pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html), etc.

I think at a minimum WebNN needs to have the same support as ONNX. I think limiting the spec to only support spatial 2D interpolation (of the last two dimensions) would be too limited. And because by far the majority cases are going to be for spatial 2D interpolations, the emulation via `split/concat` shouldn't be too bad.

huningxin: Thanks @wchao1115 for sharing the details. They are very informative.

It looks like ONNX and PyTorch support two use cases:
 1. spatial 2D interpolation for 4-D tensor; 
 2. volumetric 3D interpolation for 5-D tensor.

Given WebNN [`resample`](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-resample-input-options-input) takes 4-D tensor as input, I suppose it would make sense to support use case 1 by this `resample2d` proposal. With this proposal, caller could specify the input layout "nchw" and "nhwc" that align with other ops like `conv2d` and `pool2d`. 

Later we can extend the spec with `resample3d` for 5-D tensor, probably, together with `conv3d` and `pool3d` for volume image. That would support use case 2.

What do you think?

wchao1115: #1 for now sounds reasonable, however I prefer that we don't limit the API with fixed layouts, but rather have the interpolation axes be controlled by just `scales` and `sizes`. If the NNAPI doesn't support interpolating across any two dimensions but only the known two dimensions of known layouts, then the remaining cases should be emulated with split/concat. I just think that adding the `layout` option is probably too heavy-handed.

gramalingam: @wchao1115 : just so I understand, are you suggesting that scales would be an array of same size as the tensor rank (e.g., 4 or 5), but at most two of these values can be different from 1?

wchao1115: Correct. `scales` would be a sequence of floats with the length equal to the input tensor's rank, which if the operation is going to be limited to 4 and 5-D tensors, would allow the scale values of any two dimensions be something other than 1.

gramalingam: Are the two dimensions required to be consecutive? If that is the case, another alternative would be to use an ```axis``` attribute, along with ```scales``` as a sequence of size 2. If the dimensions are not consecutive, then I am not sure if ```split/concat``` would suffice, we might need a ```transpose```, right?

wchao1115: Yes, the two dimensions are consecutive. I tend to think of WebNN as a lower level interface since it is aimed to be a backend interface to other frameworks. In that line of thinking I tend to prefer a more generic and flexible contract e.g. scale array over axis index, etc.

huningxin: > Yes, the two dimensions are consecutive.

For a 4-D tensor, it seems there are three cases of `scales` if it defined as a sequence of floats with the length 4:
1. [x, x, 1, 1]
2. [1, x, x, 1]
3. [1, 1, x, x]

Actually, the more common case (2) and (3) are supported in this `resample2d` proposal for "nhwc" and "nchw" input layout. For the case (1), although it is not common, it could be emulated by `split` and `concat` with the `resample2d`.

I agree the proposal of `scales` with length equal to the input tensor's rank and proposal of the `axis` attribute bring more flexibility. However, regarding to the use cases to be supported, this level of flexibility might be required. And I am afraid the limited usage of such flexible interface, e.g., caller can only set two consecutive dimensions be scales other than 1, might bring more confusion.

wchao1115: @huningxin I think of WebNN as a backend interface to ML frameworks. I think that is how we position it. And because it's a lower level interface than framework formats like ONNX, I prefer to err on the side of having a more flexible contract even if it means it may be a bit harder to use in some cases. 

While the existing `scales` array is flexible and can be used to implement any known layout supported in the framework side, the opposite isn't true. Even for the cases that can be supported, the conversion from frameworks with a more flexible interface e.g. from ONNX to WebNN is not going to be straightforward with the fixed layout approach.

huningxin: @wchao1115 
> I think of WebNN as a backend interface to ML frameworks.

+1

> Even for the cases that can be supported, the conversion from frameworks with a more flexible interface

Agree. The interface should make the conversion from frameworks easy to implement. 

My point is the interface should also help the caller use it right, e.g., the scales can only apply on two consecutive dimensions. That's the reason I propose the scales/sizes to be a sequence of length 2. To support of case (1) [x, x, 1, 1], I suppose the `axis` proposal by @gramalingam would work instead of the fixed layout proposal.

wchao1115: > To support of case (1) [x, x, 1, 1], I suppose the axis proposal by @gramalingam would work instead of the fixed layout proposal.

@gramalingam What will the `axis` value be in this case? Just want to make sure I understand your proposal.

gramalingam: ```axis=0``` in this case.

huningxin: As mentioned in [16 Sep WebML WG call](https://www.w3.org/2021/09/16-webmachinelearning-minutes.html#t02), the `axis` proposal might be extended to `axes` that is a sequence of integer with length 2. It would contain two consecutive numbers that indicate the two spatial axes to be interpolated. In particular, the possible values are [0, 1], [1, 2] or [2, 3] regarding to above three cases. What do you think?

wchao1115: That sounds reasonable. Thanks @huningxin

huningxin: Thanks @wchao1115 . The `axes` option is implemented in the new commit. Please take a look.

wchao1115 (approved):

index.bs
wchao1115: Looks good. Thanks for looking into it.

huningxin:
Per the [resolution](https://www.w3.org/2021/09/30-webmachinelearning-minutes.html#r01) of WebML WG Teleconference – 30 Sep 2021

> RESOLUTION: Merge Change resample to resample2d PR #205

I am going to merge this PR. Thanks for the review and feedback.

Issue #169, opened 2021-05-13
AI accelerator device selection


anssiko:
For discussion, branching from https://github.com/webmachinelearning/webnn/pull/162

anssiko:
Related discussion from our TPAC meeting regarding fingerprinting surface impact of exposing adapter name in WebGPU/GL: https://www.w3.org/2021/10/26-webmachinelearning-minutes.html#t07

The key use case for graphics APIs seems to be to work around GPU driver bugs. That is different from WebNN use case of optimizing the user experience, power, and performance.

If we consider expanding the current WebNN [device selection](https://www.w3.org/TR/webnn/#programming-model-device-selection) with more devices in the future, we should revisit the latest WebGPU/GL discussion to adapt any best practices from there.

anssiko:
This issue was revisited in https://www.w3.org/2021/11/18-webmachinelearning-minutes.html#t04.

Summary: A smart automatic device selection would require workload analysis that is costly and even native platform APIs don’t support that currently. The MLPowerPreference hint allows implementations to make more informed decisions on whether to use purpose-built AI accelerators.

I think investigation into power-performance tradeoffs and requirements of key use cases could help figure out what changes if any are needed to the current hints:

```
enum MLDevicePreference {
  "default",
  "gpu",
  "cpu"
};

enum MLPowerPreference {
  // Let the user agent select the most suitable behavior.
  "default",

  // Prioritizes execution speed over power consumption.
  "high-performance",
  
  // Prioritizes power consumption over other considerations such as execution speed.
  "low-power"
};
```

One rough mapping of hints to use cases might be:

- ”prefer power efficiency / low power”: long-running tasks where energy comsumption is a concern e.g. noise suppression and background segmentation while on a battery-powered device,
- ”prefer high(est) performance”: heavy tasks that should complete as soon as possible e.g. content creation such as video post-processing with super resolution while plugged in (or not worried of battery running out),
- ”prefer fast(est) response”: e.g. high-quality real-time communication, VR/AR see motion-to-photon latency,
- ”prefer to excute when there’s time and energy budget available”: duties that can wait until resources are available to not interfere with competing tasks, e.g. a background task of grouping photos in a library by faces.

Currently we do not have explicit hints for the latter two. Fastest response could be implied by the ”cpu” selection.

To avoid premature optimization, we’d benefit from testing/prototyping with AI accelerators so let’s keep this issue open and collect feedback from implementations.

[...]

PR #207, opened 2021-08-30
Update "Performance Adaptation" use case

spshin3:
- Adding a new use case of neural network deployment for 'Framework  Use Cases'

***
<a href="https://pr-preview.s3.amazonaws.com/spshin3/webnn/pull/207.html" title="Last updated on Dec 15, 2021, 5:54 AM UTC (437de95)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/207/dc85541...spshin3:437de95.html" title="Last updated on Dec 15, 2021, 5:54 AM UTC (437de95)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/spshin3/webnn/pull/207.html" title="Last updated on Feb 18, 2022, 10:51 AM UTC (7b49df4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/207/dc85541...spshin3:7b49df4.html" title="Last updated on Feb 18, 2022, 10:51 AM UTC (7b49df4)">Diff</a>


anssiko:
@spshin3 thanks for the proposal. Is the key requirement needed to enable this proposed use case to allow "checking automatically the hardware capabilities of devices"?

If so, this use case seems similar to https://webmachinelearning.github.io/webnn/#usecase-perf-adapt and we could consider expanding it or merging the two where applicable.

index.bs
```diff
@@ -369,6 +369,19 @@ so that the application loads the tiny model in the case of CPU-only devices.
...
 A JavaScript ML framework is responsible for loading, interpreting and executing a ML model. During the model execution phase, the framework iterates through the operations of the model and executes each operation on the hardware device, like CPU, GPU or ML accelerator. To avoid the unnecessary data copying across devices, the framework selects the same device to execute the operations. For a compute intensive operation, such as convolution 2D or matrix multiplication, the framework uses WebNN API to execute it with the ML-specific acceleration available on that selected device.
 
+### Neural Network Deployment ### {#usecase-neural-net-deploy}
+An application developer wants to develop a DNN model which outperforms in 
+Natural Language Processing such as [[GPT-3]] on the web. The model has pre-trained
+with an enormous and diverse data, and she wants to deploy this DNN model in 
+various devices environments with efficiency. To address this issue, she develops
+a framework using WebNN API for interoperability, so that the DNN model can
+select the best options among CPU, GPU, or ML accelerators by checking
+automatically the hardware capabilities of devices
```
wchao1115, line 379: This wording seems to imply that WebNN is responsible for automatically pick "the best" option to accelerate the given model based on what is available on the platform without necessarily relying on the input from either the web application or the framework that uses it. This is a big requirement that goes beyond simply:

> checking automatically the hardware capabilities of devices

as it also involves analyzing model's complexity and memory bandwidth requirements, etc. Also note that "best" is subjective e.g. is it best for inference latency, data throughput, or battery life? 

I understand that this is a new use case. I just want to flag that this may be a design requirement that we have yet to agree as one of the design goals.

wchao1115: Thanks for addressing my question. But I'm still not sure I fully understand this use case. Are we saying that the use case is for a framework to use WebNN for transfer learning on the user-machine via the web browser?

spshin3: Unfortunately, we missed to consider the web browser scenario when we discussed this issue. We just focused the scenario for implementing DNN model in the mobile devices. Actually, the automatic selection of hardware in mobile devices are being researched in point of view for designing the mobile chip product such as Qualcom's snapdragon.

During several months, we investigated automatic hardware selection in web browser, however, this is very premature issue to implement with the above approach. Just checking the hardware capability is not enough to implement automatic hardware selection function and it seems that this function may be related with ML model-specific matter rather than hardware-specific matter.

To deal with automatic hardware selection, web application developer may need to measure their DNN model performance  by using the benchmark framework such as tensorflow benchmark library. See this reference [Tensorflow's Performance measurement](https://www.tensorflow.org/lite/performance/measurement), and issue #169.

anssiko: Thank you @spshin3 for the revised use case.

Do you think the proposed "Performance Benchmark" use case (as in this PR) would benefit from additional API surface or features in the WebNN API? There's a separate [Battery Status API](url) that may address your requirements related to identifying when the battery of the user's device is charging or the device is plugged in. I think these two APIs used together might help address your use case.

Feel free to follow up with your suggestions in https://github.com/webmachinelearning/webnn/issues/169 for any device preference hints or similar features that'd help this use case. Or if changes would be suggested to the Battery Status API, feel free to file issues for that spec.

Thank you for your continued contributions!

spshin3: Thank you @anssiko for reviewing.

Actually, there are several related works about power efficiency in deep learning, and the simulation result shows that the GPU, and TPU show better power efficiency than CPU.  I think that the trends of the results might be similar in low power situation. Please see this [Benchmarking the Performance and Energy Efficiency of AI Accelerators for AI Training](https://arxiv.org/abs/1909.06842) and [Deep Learning on a Low Power GPU](https://project-archive.inf.ed.ac.uk/ug4/20181261/ug4_proj.pdf)
With this results, in any case, choosing the most sophisticated hardware for DNN model may be the way to achieve best performance and power efficiency by just checking the device spec with WebNN API. And this scenario is related with [Performance Adaptation](https://www.w3.org/TR/webnn/#usecase-perf-adapt) use case. 

However, GPU or TPU and NPU(ML-hardware accelerator) are not always working better than CPU and GPU. As we can see also in @huningxin 's simulation [WebNN ML JS Framework Performance presentation slides](https://lists.w3.org/Archives/Public/www-archive/2021Oct/att-0014/WebNN_ML_JS_Framework_Performance.pdf), some DNN models support the better CPU utilization to achieve good performance in resource-limited situation such as non-GPU device or non-TPU device. The MobileNet model is one of the example to resolve this kind of problem. In this case, I think that designing DNN model with low complexity, optimizing the number of parameters in DNN model, or its input/output data format, such kind of things are the matter for affecting the performance differences among CPU, GPU, and TPU.

I think that it is very premature to conclude that we can choose the best hardware with the power preference. Indeed, the realistic way for choosing the best hardware is profiling the model performance in different devices and using its benchmark results when they implement the DNN model as we described in "Performance Benchmark".

spshin3: Additionally, it might be somewhat unrelated to performance issues......
In the cost efficiency aspect, the benchmark framework can be also useful for exploiting the reasonable hardware for their ML system. [MArk](https://www.usenix.org/conference/atc19/presentation/zhang-chengliang)
Like this case, we may need to consider the hyperparemeter such as batch size when we discuss the best hardware or preference of hardware.

Thanks.

anssiko: Thanks for your insights, @spshin3.

PTAL @huningxin @wchao1115.

wchao1115: Coming back to the updated proposed text here, I'm still a bit confused about the WebNN use case here. It reads like you're talking about either one of the two things here:
1. Using WebNN to benchmark performance of a model (presumably authored as a WebNN graph) across different hardware as an offline exercise to understand the performance characteristic of this model on each of the target hardware. -or-
2. At model inference time, use WebNN as a way to "test" the performance across the CPU, GPU, and other accelerator in an unknown system configuration to understand what kind of compute device the model should target at that time.

Or maybe I'm missing the whole thing?

spshin3: Thanks for carefull reviewing again!
I think that the 'option 2' is similar with our intention for the functionality of 'benchmark'.
If some developers want to test across different hardware like option1, they needs to set each target configuration of ardware offline.

anssiko: Given 'option 2', it looks like we could update the existing use case https://www.w3.org/TR/webnn/#usecase-perf-adapt

My assumptions are that the web developer may want to:
- Change the model to better fit the hardware (current use case)
- Choose the execution device to better fit the model (proposed use case)

Or we can do both.

I tried to incorporate these approaches into the existing use case merging with @spshin3's use case, proposal:

>### Performance Adaptation
>A web application developer has a concern about performance of her DNN model. The model needs to run on both a mobile device with a low power CPU as well as on a laptop with a powerful CPU, GPU and a dedicated AI accelerator.
>
>She has confirmed that the model may run too slow on the mobile device which does not have GPU acceleration. To address this issue, her web application refers to the WebNN API to confirm whether acceleration is available or not, so that the application can display a warning for devices without acceleration. After several weeks, she has developed a tiny DNN model that can even run on a CPU. In order to accommodate CPU execution, she modifies the application so that the application loads the tiny model in the case of CPU-only devices.
>
>When executing the DNN model on a laptop with a more powerful CPU, GPU and a dedicated AI accelerator, she wants to use the execution device that minimizes the inference time. To address this issue, she runs the model on each execution device and measures the inference time for each test run. This information helps her release a web application that provides the best possible user experience given available hardware.

WDYT @spshin3? Suggestions welcome!

spshin3: I fully agreed @anssiko 's comment because we changed our approach now (automatic device selection -> test device performance). 
I reflected the comment by revising our proposal. Please check the new version!

Thanks @anssiko @wchao1115 !

wchao1115: @anssiko The text that you propose here is still unclear to me. The wording makes it hard to understand what it tries to say. I don't know how to help because I may misrepresent it. 

If we go with my wild guess of #2 above, then the issue we should consider is whether WebNN is the right API to "test" that whether a specific type of accelerator is available in the system it currently runs on. And even if it could, the presence of certain accelerator in the system doesn't necessarily imply the performance of the model being run on it. I just can't see the WebNN API as the right API for this purpose.

anssiko: @wchao1115 I think I understand your concerns. My proposal was too idealistic.

A concern that I stated earlier but failed to factor into this latest proposal was that the device selection is a hint in part due to privacy reasons. Now, I think that alone might make this use case not possible to be implemented reliably.

Then, if we in addition consider "the presence of certain accelerator in the system doesn't necessarily imply the performance of the model being run on it" I'm afraid the use case is not practical.

I don't immediately see how we could change the API to better cater for this use case while retaining its privacy-preserving qualities. @spshin3 thoughts?

Should we drop the existing [performance adaption use case](https://www.w3.org/TR/webnn/#usecase-perf-adapt) entirely?

anssiko:
@spshin3 recently we landed a change that made device type normative while defined power preference as a hint: https://github.com/webmachinelearning/webnn/issues/169#issuecomment-1144996210 -- see also other related discussion in that issue as well as related discussion at https://www.w3.org/2022/06/02-webmachinelearning-minutes.html#t07

Please review the latest spec and let us know if this spec change impacts your expectations with regard to this use case. We are trying to freeze the CR scope for the spec and would like to make a decision on this PR soon. Hearing no concerns, we'd be inclined to label this use cases as post-CR. Thank you for your contributions.

spshin3:
Dear Anssi,

I'm sorry for being late for the discussion of my PR.

Actually, I had been away from office for a long time due to the private schedule.....

I will check the latest work of WebML and contributes for the issue.!

Thanks and Regards.

- Sungpil

=======================================================
Sungpil Shin&nbsp;/ Senior&nbsp;ResearcherStandards &amp; Open Source Research Division&nbsp;218 Gajeong-ro, Yuseong-gu, Daejeon, 34129, Republic of KoreaTel 042-860-1379&nbsp;Mobile 010-5051-7536&nbsp;Email&nbsp; ***@***.*** ***@***.***



-----Original Message-----

anssiko:
@spshin3 we have created a dedicated repo for future work: https://github.com/webmachinelearning/proposals

Please see the [readme](https://github.com/webmachinelearning/proposals/blob/master/README.md) for submission instructions.

Based on our discussions in the WG, we'd welcome you to submit this use case to that repo so it can be appropriately considered. Thanks!

anssiko:
@spshin3, thank you for this proposal. After careful consideration, the WG suggested this use case should be submitted as a future work proposal to https://github.com/webmachinelearning/proposals 

Per https://www.w3.org/2022/09/08-webmachinelearning-minutes.html#t08 I'm closing this PR.

PR #208, opened 2021-09-07
Support rounding type of pool2d operations

### LINKED ISSUE #198: Support for configuring rounding type in pooling operations
Hi all, 

I'm a student intern working on GSoC 2021 project "OpenCV.js: Accelerate OpenCV.js DNN via WebNN" (opencv/opencv#20406). [Here](https://docs.google.com/document/d/18PcYuRC2UCKUpustZn3t8beDF1UNfP29ikwPLPKXFuY/edit) is the proposal. In this project, I will improve the performance of OpenCV.js DNN Module using WebNN.

However, when I am implementing OpenCV pooling layer with WebNN backend, I found that OpenCV's pooling layer supports both ceil (with ```ceilMode``` flag) and floor rounding type in https://github.com/opencv/opencv/blob/e9db7fea5b793bbe30ebf730130db24cc29267ea/modules/dnn/include/opencv2/dnn/all_layers.hpp#L255 
and https://github.com/opencv/opencv/blob/e9db7fea5b793bbe30ebf730130db24cc29267ea/modules/dnn/src/layers/pooling_layer.cpp#L145. 

And through the preliminary investigation, ONNX pooling also supports both [ceilMode and floorMode](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-2). 

However, in WebNN pooling spec, we do not support configuring the rounding type in pooling operations. We defaultly use floor rounding type. According to the investigation, both Alexnet and GoogleNet-v1 use ceil rounding type while in the ONNX layer tests, floor rounding type is used. In this case, it is useful to let WebNN pooling spec support configuring rounding type.

huningxin:
fix #198 

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/208.html" title="Last updated on Sep 22, 2021, 6:29 AM UTC (17d1fc5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/208/35b4120...huningxin:17d1fc5.html" title="Last updated on Sep 22, 2021, 6:29 AM UTC (17d1fc5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1566,6 +1566,7 @@ dictionary MLPool2dOptions {
   MLAutoPad autoPad = "explicit";
   MLInputOperandLayout layout = "nchw";
   MLRoundingType roundingType = "floor";
+  sequence<long> outputSizes;
 };
 
 partial interface MLGraphBuilder {
@@ -1601,6 +1602,7 @@ partial interface MLGraphBuilder {
                     - input tensor: [batches, height, width, channels]
                     - output tensor: [batches, height, width, channels]
             - *roundingType*: an {{MLRoundingType}}. The option specifies the rounding function used to compute the output shape.
+            - *outputSizes*: a sequence of long of length 2. The sizes of the two spacial dimensions of the output tensor. When the output sizes are explicitly specified, the options.roundingType is ignored. If not specified, the output sizes are automatically computed.
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor that contains the
     result of the reduction. The logical shape is interpreted according to the
--- a/index.bs
+++ b/index.bs
@@ -1553,6 +1553,11 @@ partial interface MLGraphBuilder {
 ### pooling operations ### {#api-mlgraphbuilder-pool2d}
 Compute a *mean*, *L2 norm*, or *max* reduction operation across all the elements within the moving window over the input tensor. See the description of each type of reduction in [[#api-mlgraphbuilder-reduce]].
 <script type=idl>
+enum MLRoundingType {
+  "floor",
+  "ceil"
+};
+
 dictionary MLPool2dOptions {
   sequence<long> windowDimensions;
   sequence<long> padding;
@@ -1560,6 +1565,7 @@ dictionary MLPool2dOptions {
   sequence<long> dilations;
   MLAutoPad autoPad = "explicit";
   MLInputOperandLayout layout = "nchw";
+  MLRoundingType roundingType = "floor";
 };
 
 partial interface MLGraphBuilder {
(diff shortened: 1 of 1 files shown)


anssiko (approved):

index.bs
```diff
@@ -1553,13 +1553,19 @@ partial interface MLGraphBuilder {
 ### pooling operations ### {#api-mlgraphbuilder-pool2d}
 Compute a *mean*, *L2 norm*, or *max* reduction operation across all the elements within the moving window over the input tensor. See the description of each type of reduction in [[#api-mlgraphbuilder-reduce]].
 <script type=idl>
+enum MLRoundingType {
+  "floor",
+  "ceil"
+};
```
wchao1115, line 1559: See my comment on #198. Rounding mode for tensor size calculation should be done above WebNN, similar to how it is done in ONNX Runtime above DirectML.

huningxin: @wchao1115 , thanks for your comments on #198. 

However a caller may not calculate the tensor shape and rely on WebNN to do that, e.g. the OpenCV WebNN backend mentioned in #198. The [MLOperand](https://webmachinelearning.github.io/webnn/#api-mloperand) interface of WebNN doesn't allow query the shape. So if the caller doesn't infer the shape by itself, it would be hard to calculate the output shape with a desired rounding type from the input shape, especially when the input is an intermediate operand. That's why I propose to let the callers to just configure the rounding type instead of calculating the output shape at all by themselves. 

I suppose this would map ONNX `ceil_mode` well. What do you think?

wchao1115: That's a fair point. The current WebNN design does allow dynamic shape inference, which would make it somewhat easier to call the API directly (since the caller needs not worry about correctly implementing shape inference for all cases themselves), but it does come at a cost of additional work on the implementer's side, and in theory some long-term maintenance cost of the API due to the additional policy that must be implemented for such a caller.

If we want to continue to allow dynamic shape inference in the WebNN API, then adding an optional output size's rounding mode would not be out of line. In that case, I would ask that we also add `outputSizes` so if the framework callers already calculate the output size themselves (most frameworks do), then they can just ignore the rounding mode.

huningxin: > In that case, I would suggest that we also add `outputSizes` so if the framework callers already calculate the output size themselves (most frameworks do), then they can just ignore the rounding mode.

It sounds good to me. 

@wchao1115 , according to DirectML backend implementation, how does pooling ops of DirectMLX, such as [`AveragePooling`](https://github.com/microsoft/DirectML/blob/master/Libraries/DirectMLX.h#L2061), support rounding mode? It calculates the output sizes inside and doesn't allow to configure neither rounding type nor output sizes.

wchao1115: [DirectMLX](https://github.com/microsoft/DirectML/blob/master/Libraries/DirectMLX.h) is just a helper library to DirectML. It doesn't add any meaningful feature to it, just literally reduces typing and makes DirectML easier to access. The library also doesn't do shape inference at runtime, only at construction time, so if the framework that uses it support runtime shape inference, it needs to handle that before passing it down to DirectML or to its helper DirectMLX.

huningxin: Thanks @wchao1115 .

`outputSizes` is added into the latest commit. Please take another look.

anssiko (approved):

wchao1115 (approved):

huningxin:
Thanks for the reviews and approvals.

PR #214, opened 2021-09-29
Split transposed convolution to convTranspose2d

### LINKED ISSUE #212: Support `iohw` filterLayout for `conv2d` op
The fast style transfer ONNX model converted from https://github.com/acerwebai/VangoghCrazyWorld includes `ConvTranspose` op, its filterLayout should be `iohw` while which is not supported in current WebNN spec.

Screenshot (54)

huningxin:
fix #212

cc @Honry

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/214.html" title="Last updated on Oct 25, 2021, 9:14 AM UTC (21a82d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/214/35b4120...huningxin:21a82d8.html" title="Last updated on Oct 25, 2021, 9:14 AM UTC (21a82d8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -761,7 +761,8 @@ enum MLFilterOperandLayout {
   "oihw",
   "hwio",
   "ohwi",
-  "ihwo"
+  "ihwo",
+  "iohw"
 };
 
 enum MLAutoPad {
@@ -827,6 +828,9 @@ partial interface MLGraphBuilder {
 
                 "ihwo":
                     - [input_channels/groups, height, width, output_channels]
+
+                "iohw":
+                    - [input_channels/groups, output_channels, height, width]
 
             - *bias*: an {{MLOperand}}. The additional 1-D tensor with the shape of [output_channels] whose values are to be added to the convolution result.
             - *activation*: an {{MLOperator}}. The optional activation function that immediately follows the convolution operation. 
--- a/index.bs
+++ b/index.bs
@@ -757,12 +757,11 @@ partial interface MLGraphBuilder {
 ### conv2d ### {#api-mlgraphbuilder-conv2d}
 Compute a 2-D convolution given 4-D input and filter tensors
 <script type=idl>
-enum MLFilterOperandLayout {
+enum MLConv2dFilterOperandLayout {
   "oihw",
   "hwio",
   "ohwi",
-  "ihwo",
-  "iohw"
+  "ihwo"
 };
 
 enum MLAutoPad {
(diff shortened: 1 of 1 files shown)


anssiko (approved):
(PR Preview build fail is not related to this change.)

gramalingam:
Hi, conv-transpose is a different operator from conv, right? It is unclear to me how extending the filter-layout alone will suffice, when the webnn spec does not have a conv-transpose op. Conv-transpose is defined to be a separate op from conv in onnx, pytorch, and tensorflow. So, it seems unlikely it can be simulated directly using a conv.

huningxin:
@gramalingam , conv-transpose is supported by webnn [conv2d](https://www.w3.org/TR/webnn/#dom-mlconv2doptions-transpose) op via its `transpose` option

> transpose: a boolean indicating that a transposed convolution operation is performed

index.bs
```diff
@@ -828,6 +829,9 @@ partial interface MLGraphBuilder {
                 "ihwo":
                     - [input_channels/groups, height, width, output_channels]
 
+                "iohw":
+                    - [input_channels/groups, output_channels, height, width]
+
```
wchao1115, line 834: When the `transpose` flag is set to true, "iohw" filter layout should mean `[input_channels, output_channels/groups, height, width]` right? Because transposed convolution up-samples the output as opposed to convolution, which down-samples it. How do we distinguish in the API that the interpretation of "iohw" depends on whether it is a transposed operation? Is this actually the reason to split transposed convolution to its own operations i.e. `convTranspose2d`?

wchao1115: Yeah, guys the more I think about this, the more I feel we need a separate `convTranspose2d` to simplify the two different use cases between convolution (downsampling) and transposed convolution (upsampling). It has real implication how we support the known filter layouts. @huningxin What do you think? If you agree, would you like to take this PR a step further and split the operation?

gramalingam: Make sense to me too to split it.

huningxin: > @huningxin What do you think? If you agree, would you like to take this PR a step further and split the operation?

SGTM. I'll take that. Thanks @wchao1115 @gramalingam for your feedbacks.

wchao1115 (approved):

huningxin:
Thanks! With two approvals, I am going to merge this one to unblock the polyfill and samples update.

Issue #210, opened 2021-09-09
Use modern WebIDL and Infra standard conventions


anssiko:
| ℹ️ **This is a meta issue. See [the (living) task list](https://github.com/webmachinelearning/webnn/issues/210#issuecomment-1326361748) for pointers to specific issues and PRs.**
| --- |

The WebIDL spec recommends algorithmic method steps for defining operations: https://heycam.github.io/webidl/#method-steps

This spec currently uses a WebGPU spec inspired convention that differs from this best practice. We should consider adopting the "method steps" convention.

(Here's one example of method steps in context of the WebNN API spec:
https://webmachinelearning.github.io/webnn/#dom-ml-createcontext)

anssiko:
I labeled this issue with "cr" implying we should address this issue before our Candidate Recommendation publication to satisfy "The document is considered complete and fit for purpose" requirement.

anssiko:
Updated patterns for defining getters, setters, methods and constructors:

https://webidl.spec.whatwg.org/#getter-steps
https://webidl.spec.whatwg.org/#setter-steps
https://webidl.spec.whatwg.org/#method-steps
https://webidl.spec.whatwg.org/#constructor-steps

anssiko:
(API review questions that'll also inform this work on method steps discussed in https://github.com/webmachinelearning/webnn/issues/298)

anssiko:
(Renamed the issue to better reflect the scope of this work.)

@zolkis can you document here the spec sections where we need to do editorial work such as add algorithmic method steps, add internal slots, use getters, setters, define methods and constructors using the modern WebIDL conventions?

This would help the WG understand the scope of this effort and areas where contributions are welcome. I would focus this task on editorial changes and only propose normative changes when they are needed to fix a spec bug, or to clarify an aspect where the spec and implementation disagree.

The list can take a simple form of:

- [ ] Add methods step for [`createContextSync`](https://www.w3.org/TR/webnn/#dom-ml-createcontextsync)
- [ ] ...

If you feel it is easier to create a new issue for each task instead, feel free to do so.

zolkis:
This issue is good aggregate tracker. Issues might be spanned on need, if the discussion is expected to be contained, or too big.

zolkis:
[...]

PR #233, opened 2021-11-23
Change the type of padding input to array

### LINKED ISSUE #224: Use tensor type for the padding rather than MLOperand
The pad in the [spec](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-pad) uses the `MLOperand padding` as its sencond input. But I find that the pad in [DML](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_padding_operator_desc) and in [IE](https://docs.openvino.ai./latest/openvino_docs_ops_movement_Pad_1.html), padding was defined with a tensor type. I think it's better to use tensor type to avoid the gap between spec and implementation.

huningxin:
fix #224

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/233.html" title="Last updated on Nov 24, 2021, 12:27 AM UTC (379149e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/233/aaddda1...huningxin:379149e.html" title="Last updated on Nov 24, 2021, 12:27 AM UTC (379149e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1563,13 +1563,14 @@ dictionary MLPadOptions {
 };
 
 partial interface MLGraphBuilder {
-  MLOperand pad(MLOperand input, MLOperand padding, optional MLPadOptions options = {});
+  MLOperand pad(MLOperand input, sequence<sequence<long>> padding,
+                optional MLPadOptions options = {});
 };
 </script>
 <div algorithm=pad>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input tensor.
-        - *padding*: an {{MLOperand}}. The 2-D Tensor of integer values indicating the number of padding values to add at the beginning and end of each input dimensions. The tensor has shape [*n*, 2] where *n* is the rank of the input tensor. For each dimension *D* of *input*, *padding[D, 0]* indicates how many values to add before the content in that dimension, and *padding[D, 1]* indicates how many values to add after the content in that dimension.
+        - *padding*: a sequence of length 2, where each element is a sequence of {{long}} whose length is equal to the rank of the input tensor. The integer values indicating the number of padding values to add at the beginning and end of each input dimensions. For each dimension *D* of *input*, *padding[D, 0]* indicates how many values to add before the content in that dimension, and *padding[D, 1]* indicates how many values to add after the content in that dimension.
         - *options*: an optional {{MLPadOptions}}. The optional parameters of the operation.
             - *mode*: a {{MLPaddingMode}}. The different ways to pad the tensor. When not set, it's assumed to be "constant".
             - *value*: a {{float}}. The pad value when the *options.mode* is set to *"constant"*. When not set, it's assumed to be 0.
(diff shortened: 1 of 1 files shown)


dontcallmedom:
I don't know why it didn't run nor how to find out why; I guess further commits might trigger it (but I'm not sure)

anssiko:
The padding argument has semantics that cannot be fully expressed in terms of Web IDL (`sequence<sequence<long>>`) so we need the prose as in this PR to define what is a valid argument and what's not. This is one example where using algorithmic prose might improve readability (tracked in https://github.com/webmachinelearning/webnn/issues/210).

To take a concrete example, if a malformed padding argument is provided, do we expect `compute()` to throw a DataError DOMException? Do we want to fail sooner than that? Failing immediately upon invocation of `pad()` would be helpful for eager execution I believe.

Not a blocker for this PR, but a comment for future enhancement.

anssiko (approved):
Unblocking review with a comment on future enhancement.

dontcallmedom:
> The padding argument has semantics that cannot be fully expressed in terms of Web IDL (`sequence<sequence<long>>`) 

As far as I can tell, this is a valid WebIDL construct.

huningxin:
> Can we trigger it?

Retriggered. Now the preview is available.

> To take a concrete example, if a malformed padding argument is provided, do we expect `compute()` to throw a DataError DOMException? Do we want to fail sooner than that? Failing immediately upon invocation of `pad()` would be helpful for eager execution I believe.

+1 to fail upon invocation of `pad()`.

> This is one example where using algorithmic prose might improve readability (tracked in #210).

+1. I'll look into it and probably address it in a separate PR. Thanks @anssiko .

anssiko:
@huningxin I opened https://github.com/webmachinelearning/webnn/issues/234 so that "fail upon invocation" can be addressed separately. I think also some other ops may be affected.

anssiko (approved):

huningxin:
a soft ping @wchao1115 @pyu10055 , thanks!

index.bs
```diff
@@ -1563,13 +1563,14 @@ dictionary MLPadOptions {
 };
 
 partial interface MLGraphBuilder {
-  MLOperand pad(MLOperand input, MLOperand padding, optional MLPadOptions options = {});
+  MLOperand pad(MLOperand input, sequence<sequence<long>> padding,
+                optional MLPadOptions options = {});
```
wchao1115, line 1567: My only holdback on this is that the `padding` data itself may be unknown during graph construction time, and that having a placeholder operand will allow for the dynamic padding at an execution time. ONNX for example has this provision for the pad op.

huningxin: According to the implementation feedback (#224), the major issue is that native ML APIs, like DirectML ([`DML_PADDING_OPERATOR_DESC`](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_padding_operator_desc)), only support static padding data.

wchao1115: This is actually a shortcoming in DirectML, which is being fixed.

huningxin: Thanks @wchao1115 for the inputs. I have passed your feedback to the original issue and proposed to keep the current version. Once we make consensus there, I'll close this PR.

huningxin:
Per feedback of https://github.com/webmachinelearning/webnn/issues/224#issuecomment-1033397241, I'll close this PR and keep the current definition.

PR #237, opened 2021-12-01
Add MLOperand and update MLOperator description

anssiko:
I started filling in a high-level description for the MLOperand that was missing referring to the programming model section for a more elaborate description. As a drive-by, I reworked the MLOperator description a bit to refer to the activation functions explicitly. Perhaps having this list here helps the readers? I hope it is not too hard to maintain if we add more activation functions.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/237.html" title="Last updated on Jan 17, 2022, 3:51 PM UTC (b08444a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/237/fd171b5...b08444a.html" title="Last updated on Jan 17, 2022, 3:51 PM UTC (b08444a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -604,7 +604,7 @@ interface MLOperator {};
 </script>
 
 <div class="note">
-{{MLGraphBuilder}} operations whose `options` dictionary contains an `activation` or `activations` dictionary member accepts these activation function types: {{clamp()}}, {{elu()}}, {{hardSigmoid()}}, {{hardSwish()}}, {{leakyRelu()}}, {{linear()}}, {{relu()}}, {{sigmoid()}}, {{softplus()}}, {{softsign()}}, {{tanh()}}.
+These activations function types are used to create other operations such as [[#api-mlgraphbuilder-conv2d]] or [[#api-mlgraphbuilder-batchnorm]].
 </div>
 
 <div class="note">
--- a/index.bs
+++ b/index.bs
@@ -586,6 +586,9 @@ dictionary MLOperandDescriptor {
 </script>
 
 ## MLOperand ## {#api-mloperand}
+
+The {{MLOperand}} interface represents data that flows within the computational graph. See [[#programming-model]].
+
 <script type=idl>
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface MLOperand {};
@@ -593,7 +596,7 @@ interface MLOperand {};
 
 ## MLOperator ## {#api-mloperator}
 
-The {{MLOperator}} interface defines a type of operation such as the various types of activation function used to create other operations such as [[#api-mlgraphbuilder-conv2d]] or [[#api-mlgraphbuilder-batchnorm]].
+Objects implementing the {{MLOperator}} interface represent activation function types: {{clamp()}}, {{elu()}}, {{hardSigmoid()}}, {{hardSwish()}}, {{leakyRelu()}}, {{linear()}}, {{relu()}}, {{sigmoid()}}, {{softplus()}}, {{softsign()}}, {{tanh()}}.
 
 <script type=idl>
 [SecureContext, Exposed=(Window, DedicatedWorker)]
@@ -601,7 +604,7 @@ interface MLOperator {};
 </script>
 
 <div class="note">
-The implementation of this interface can simply be a struct that holds a string type of the operation along with other properties needed. The actual creation of the operation e.g. a [[#api-mlgraphbuilder-sigmoid]] or [[#api-mlgraphbuilder-relu]] can then be deferred until when the rest of the graph is ready to connect with it such as during the construction of [[#api-mlgraphbuilder-conv2d]] for example.
+The implementation of the {{MLOperator}} interface can simply be a struct that holds a string type of the activation function along with other properties needed. The actual creation of the activation function e.g. a [[#api-mlgraphbuilder-sigmoid]] or [[#api-mlgraphbuilder-relu]] can then be deferred until when the rest of the graph is ready to connect with it such as during the construction of [[#api-mlgraphbuilder-conv2d]] for example.
 </div>
 
 ## MLGraphBuilder ## {#api-mlgraphbuilder}
--- a/index.bs
+++ b/index.bs
@@ -596,12 +596,16 @@ interface MLOperand {};
 
 ## MLOperator ## {#api-mloperator}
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -585,22 +585,25 @@ dictionary MLOperandDescriptor {
...
+
 <script type=idl>
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface MLOperand {};
 </script>
 
 ## MLOperator ## {#api-mloperator}
 
-The {{MLOperator}} interface defines a type of operation such as the various types of activation function used to create other operations such as [[#api-mlgraphbuilder-conv2d]] or [[#api-mlgraphbuilder-batchnorm]].
+Objects implementing the {{MLOperator}} interface represent activation function types: {{clamp()}}, {{elu()}}, {{hardSigmoid()}}, {{hardSwish()}}, {{leakyRelu()}}, {{linear()}}, {{relu()}}, {{sigmoid()}}, {{softplus()}}, {{softsign()}}, {{tanh()}}.
```
huningxin, line 598: Listing the available activation functions would help readers who jump from the operations that support the fused activations, such as `conv2d` or `batchNormalization`. However, I am not sure whether we should keep mentioning these operations. It might be useful for readers who first read the `MLOperator`, so they can jump to `conv2d` or `batchNormalization` to get an idea of how these activation functions are used.

huningxin (commented):
Thanks @anssiko . The `MLOperand` update looks good to me. Left a minor comment for `MLOperator` update, please take a look.

index.bs
anssiko: Thanks for the comment. My idea was to enable navigation from ops supporting fused activations to activation functions. I thought we'd use MLOperator as an intermediary. I'm hearing navigating from MLOperator to supporting ops would also be helpful, and it indeed makes sense to enable cross-linking both ways.

Should we add a complete list of such ops here or is the "such as § 7.7.4 conv2d or § 7.7.1 batchNormalization" good enough?

I pushed an update where the list of activation function becomes an informative note, also noted what type of operations support activations. PTAL.

We could expand that note with a few examples or a complete list of ops that support fused activations.

WDYT?

anssiko:
@huningxin, I pushed an update PTAL.

Here's a preview since PR Preview failed initially and did not rerun on new pushes:

[Preview](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fwebmachinelearning%2Fwebnn%2F4b15fd5559bf36db823a17993a457c5e9fa00b83%2Findex.bs&force=1&md-warning=not%20ready#api-mloperand)

Bikeshed actually has a nice feature that in this case lists activation functions automatically so we don't need to maintain the list manually:
Screen Shot 2021-12-08 at 15 35 37

huningxin:
Thanks @anssiko 

> Bikeshed actually has a nice feature that in this case lists activation functions automatically so we don't need to maintain the list manually:

It seems to list all references, including all activation functions (e.g., relu) and the operations that use them (e.g., conv2d). Is it possible to distinguish the two categories by bikeshed?

anssiko:
Not that I know of. Bikeshed parses the IDL for all references to generate this list.

huningxin (approved):

index.bs
```diff
@@ -585,22 +585,29 @@ dictionary MLOperandDescriptor {
...
+
 <script type=idl>
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface MLOperand {};
 </script>
 
 ## MLOperator ## {#api-mloperator}
 
-The {{MLOperator}} interface defines a type of operation such as the various types of activation function used to create other operations such as [[#api-mlgraphbuilder-conv2d]] or [[#api-mlgraphbuilder-batchnorm]].
+Objects implementing the {{MLOperator}} interface represent activation function types.
```
wchao1115, line 598: I think we might want to leave some room for the future here and not prematurely limiting the notion of an `MLOperator`, which is a generic notion, to just the activation functions. It could also be confusing when the notion of the activation function appears twice in the API in different context, one concretely represented by `MLOperator` and another as standalone operations off the `MLGraphBuilder`. 

The idea behind `MLOperator` is to allow deferred definition of activation functions used during the construction of a fused operation e.g. GRU, etc. while allowing for future extension without changing the fused operation's API signature.

index.bs
```diff
@@ -585,22 +585,29 @@ dictionary MLOperandDescriptor {
...
 
 <script type=idl>
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface MLOperator {};
 </script>
 
 <div class="note">
-The implementation of this interface can simply be a struct that holds a string type of the operation along with other properties needed. The actual creation of the operation e.g. a [[#api-mlgraphbuilder-sigmoid]] or [[#api-mlgraphbuilder-relu]] can then be deferred until when the rest of the graph is ready to connect with it such as during the construction of [[#api-mlgraphbuilder-conv2d]] for example.
+These activations function types are used to create other operations such as [[#api-mlgraphbuilder-conv2d]] or [[#api-mlgraphbuilder-batchnorm]].
```
wchao1115, line 606: We can say something like "one such use of this interface is for when an activation function is fused into another operation such as convolution or batchnorm during a graph construction session.."

index.bs
```diff
@@ -585,22 +585,29 @@ dictionary MLOperandDescriptor {
 </script>
 
 ## MLOperand ## {#api-mloperand}
+
+The {{MLOperand}} interface represents data that flows within the computational graph. See [[#programming-model]].
+
```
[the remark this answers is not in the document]
anssiko: Thanks for this detailed description. That high-level description was borrowed from the programming model section. I updated this PR per your suggestion. Given the programming model links to this description, I think it is fine to have the proper description here with the interface definition.

index.bs
anssiko: Mentioning the extensibility for future-proofing is a good idea, PR updated.

Would it make any sense to specialize `MLActivation : MLOperator`?

index.bs
anssiko: Thanks, updated per your suggestion.

anssiko:
PR updated, PTAL @wchao1115

wchao1115 (approved):

anssiko:
@wchao1115 thanks! Please merge at will coordinating with your and @huningxin's other work to avoid merge conflicts.

huningxin:
With approvals, let's merge this PR and rebase others on top of this.

PR #250, opened 2022-01-24
Update createContext() and MLContext interface prose and algorithms

anssiko:
- Define MLContext internal slots
- Migrate context type, device preference, power preference into MLContext
  (prose sourced from WebIDL comments and device selection section)
- Define createContext() method steps in terms of MLContext internal slots

---
I took a stab at moving some WebIDL comments into prose, and on the same pass defined MLContext internal slots and tied those to the existing getContext() algorithm. This internal state may be useful also in other algorithms. There's probably room for improvement, so looking for your comments.

Please check in particular that the createContext() behaviour is as we'd like it to be (you can set either context, or device and power preference, not all at the same time).

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/250.html" title="Last updated on Jan 24, 2022, 4:07 PM UTC (a1a9c31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/250/fd171b5...a1a9c31.html" title="Last updated on Jan 24, 2022, 4:07 PM UTC (a1a9c31)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -451,10 +451,7 @@ An {{MLContext}} interface represents a global state of neural network execution
 
 In a situation when a GPU context executes a graph with a constant or an input in the system memory as an {{ArrayBufferView}}, the input content is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of the graph execution. This data upload and download cycles will only occur whenever the execution device requires the data to be copied out of and back into the system memory, such as in the case of the GPU. It doesn't occur when the device is a CPU device. Additionally, the result of the graph execution is in a known layout format. While the execution may be optimized for a native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert the content back to a known layout format at the end of the graph in order to maintain the expected behavior from the caller's perspective.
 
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM, thanks @anssiko !

index.bs
```diff
@@ -531,33 +515,86 @@ The {{ML/createContext()}} method steps are:
 1. Switch on the method's first argument:
     <dl class=switch>
     <dt>{{MLContextOptions}}
-    <dd>Set |context|'s [=context type=] to [=default-context|default=].
+    <dd>Set |context|.{{[[contextType]]}} to [=default-context|default=].
+    <dd>Set |context|.{{[[devicePreference]]}} to the value of {{MLContextOptions}}'s {{devicePreference}} member.
+    <dd>Set |context|.{{[[powerPreference]]}} to the value of {{MLContextOptions}}'s {{powerPreference}} member.
```
wchao1115, line 520: Is the double square brackets here intentional? Can we remove them? It looks unpolished and makes the text harder to read in the final rendering.

wchao1115 (changes requested):

index.bs
anssiko: I fully agree it does not look pretty, and I'm not a fan of this look personally, but this is the convention so I suggest we align with it. That is, it is intentional.

(Web IDL inherits the square brackets convention for internal slots from ECMA-262: https://webidl.spec.whatwg.org/#ecmascript-binding)

wchao1115 (approved):

Issue #243, opened 2021-12-22
op metadata that helps avoid implementation mistakes


quidity:
Some way to help implementors validate that data is being passed correctly between nodes in the computation graph, has the expected sizes, etc.

Perhaps a little language which describes the required sizes/shapes and can be used to create checks that can be enforced as a graph is constructed, and as a graph is executed.

e.g. fields for ops which indicate:

Does it change input/output tensor contents?
Does it change input/output tensor sizes?
Can the input tensor be the output tensor?
Shape of the input, shape of the output?
Constraints on other input parameters that can be turned into automatic checks.

anssiko:
The group had this issue on its [agenda](https://www.w3.org/2022/01/27-webmachinelearning-minutes.html#t11
), but we did not record anything in the minutes.


All - please share your feedback for the proposal on this issue.

anssiko:
We discussed this issue again on our [call](https://www.w3.org/2022/02/10-webmachinelearning-minutes.html#t01). It was noted it may not be possible to extract very useful metadata for all ops given the complexity of operations.

@quidity, we'd like to hear which of these options would work for you:
1. Would you prefer us to add a new "Metadata" section for each op to accompany "Arguments" and "Returns"? See the relu example below.
2. Or could we perhaps provide all this information combined in an appendix? It could allow us to present this information in a more concise form, for example as a table. We could then link to that from applicable op definitions. See the example table below.
3. Something else?

---
### Option 1: add metadata into each op definition (using [relu](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-relu) as an example):

### 7.7.21. relu
Compute the [rectified linear function](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)) of the input tensor.
```
partial interface MLGraphBuilder {
  MLOperand relu(MLOperand x);
  MLOperator relu();
};
```

**Arguments:**
x: an [MLOperand](https://www.w3.org/TR/webnn/#mloperand). The input tensor.

**Returns:**

an [MLOperand](https://www.w3.org/TR/webnn/#mloperand). The output tensor of the same shape as x.
an [MLOperator](https://www.w3.org/TR/webnn/#mloperator). The operator representing the relu operation.

**Metadata:**
Does it change input/output tensor contents?
Does it change input/output tensor sizes?
Can the input tensor be the output tensor?
Shape of the input, shape of the output?
Constraints on other input parameters that can be turned into automatic checks.

---

[...]

PR #251, opened 2022-02-03
Update Security Considerations per review feedback

### LINKED ISSUE #241: General Security Questions
I'm a security reviewer from Google Chrome and have some general questions - they almost certainly have good answers and it's very possible I missed finding them in the spec. I also have some specific suggestions which I will open as distinct issues shortly.

1. I think I disagree with the answer given in 2.9 in https://github.com/webmachinelearning/webnn/blob/main/security-privacy.md. This is introducing a new scripting language, built up in js then executed in a number of different interpreters (“cpu”, “gpu” etc.). The complexity of the operations is such that the likelihood of errors that ultimately result in out of bounds accesses is high, and malicious sites will have significant control over operations. This could be addressed more directly in the security/privacy explainer.
2. Operations such as split/slice/squeeze that change the shape of tensors mid-calculation may lead to incorrect assumptions in later operations - for instance if eliding bounds checks this could lead to out of bounds accesses. It would be good for their to be operation level metadata that might be consumed by implementors to help prevent such problems.
3. The universe of operations is likely to vary in future - how will consumers discover which operations are available (short of enumerating them through failures to instantiate)? How will operations be deprecated (for instance if they turn out to be badly implemented?)
4. It feels like .build and .compute should be asynchronous in all cases?
5. New side channels will be made available from shared resources (cpu/gpu). Timeable things should be out of process so incur at least some ipc to achieve anything. Probably not a massive worry when compared with already sharing a cpu between processes running renderers.
6. Verify: Sites must delegate permission to host/run models.
7. Verify: No serialization or caching yet - although this is likely in future.
8. Control over how a model is run - (selecting cpu/gpu/tpu say) - is this too much power for the consuming site - it will for instance make it possible to more directly target a flawed implementation. It's not clear why this is required.

Sub issues:
* #242 
* #243
* #244

anssiko:
Fix #241

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/251.html" title="Last updated on Mar 24, 2022, 9:30 AM UTC (a737271)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/251/19a09ab...a737271.html" title="Last updated on Mar 24, 2022, 9:30 AM UTC (a737271)">Diff</a>


anssiko:
@wchao1115 @huningxin I'm marking this PR as "Ready for review".

My expectation is the first PR that lands will be amended in subsequent PRs as we receive further review feedback. It is fine to have inline issues as reminders in the spec too, but if we're able to fix those via PR review process that'd be of course great.

anssiko:
Here's a direct link to the visual diff of the Security Considerations this PR updates: https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/251/19a09ab...69db252.html#security

index.bs
```diff
@@ -383,6 +383,41 @@ This API is disabled by default in all cross-origin frames using the [[#permissi
...
+
+Purpose-built Web APIs for measuring high-resolution time mitigate against timing attacks using techniques such as resolution reduction, adding jitter, detection of abuse and API call throttling [[hr-time-3]]. It is expected a typical multi-process architecture implementation of the WebNN API makes use of inter-process communication (IPC) mechanism that inherently adds additional jitter on top of the built-in high-resolution time mitigations. Regardless of these mitigations, implementers are always advised to consider and test their implementations against timing attacks, and in particular if IPC is not used.
+
+## Guidelines for new operations ## {#security-new-ops}
+
+To ensure operations defined in this specification are shaped in a way they can be implemented securely, this section includes guidelines on how operations are expected to be defined to reduce potential for implementation problems. These guidelines are expected to evolve over time to align with industry best practices:
+
+- Avoid reshaping tensors
```
huningxin, line 406: I am not quite sure about this one. Do we have any references, say ONNX document?

index.bs
```diff
@@ -383,6 +383,41 @@ This API is disabled by default in all cross-origin frames using the [[#permissi
 
 This API allows creation of an {{MLContext}} from a {{GPUDevice}} or {{WebGLRenderingContext}} defined by WebGPU and WebGL specifications respectively. See <a href="https://gpuweb.github.io/gpuweb/#security">WebGPU Security Considerations</a> and <a href="https://www.khronos.org/registry/webgl/specs/latest/1.0/#4">WebGL Security Consideration</a> for more information regarding security characteristics of these contexts.
 
+Some operations such as {{split()}}, {{slice()}} and {{squeeze()}} that change the shape of tensors mid-calculation may lead to incorrect assumptions in later operations and possibly lead to out-of-bounds access attempts. Implementers are adviced to anticipate such out-of-bounds access issues and add appropriate mitigations.
```
huningxin, line 386: As @wchao1115 [mentioned](https://github.com/webmachinelearning/webnn/issues/243) in #243, the shape inference is done during the graph building stage. When compute the graph, the input shape of each op is finalized. Probably we could make it clear about that.

huningxin (commented):
Thanks for making this PR, @anssiko !

index.bs
anssiko:
```suggestion
```

The [ONNX document](https://lists.w3.org/Archives/Public/www-archive/2021Nov/att-0000/W3C_Adding_new_Operators.pdf) does not mention this. I believe this must be my misinterpretation/oversimplification of [our discussion](https://www.w3.org/2022/01/27-webmachinelearning-minutes.html#t03).

I'm suggesting we drop this bullet unless we can reword this to make sense in this context.

index.bs
anssiko:
```suggestion
Some operations such as {{split()}}, {{slice()}} and {{squeeze()}} that change the shape of tensors mid-calculation may lead to incorrect assumptions in later operations and possibly lead to out-of-bounds access attempts. Implementers are advised to anticipate such out-of-bounds access issues and add appropriate mitigations.

This concern is in part mitigated by means of API design. Once the graph is fully constructed and compiled, the input shapes into each of the operations in the graph are inferred and finalized. The bounds checking occurs when the compute method is invoked that executes the graph against the actual data. No actual data is bound to the compiled graph before this stage. It is the implementation's responsibility to make sure proper bounds checking occurs against the shapes of the data already inferred by that time.
```
I added and adapted @wchao1115's explanation (second paragraph). Please review and suggest changes.

anssiko:
@dontcallmedom @wchao1115 PTAL prior to the WG's call this week.

index.bs
```diff
@@ -383,6 +383,41 @@ This API is disabled by default in all cross-origin frames using the [[#permissi
...
+
+Issue: Document operations susceptible to out-of-bounds access as a guidance to implementers.
+
+As a future-proofing, the API design allows certain operation (annotated with "can be generically emulated") defined in this specification to be deprecated for security, performance, or other reasons without breaking compability. This is made possible by high-level functions that are defined in terms of smaller primitive operations defined in this specifications. This enables a native implementation of a high-level function to be replaced with a polyfill implementation.
+
+Issue: Investigate side channel attack feasibility considering the current state where CPU is shared between processes running renderers.
+
+In order to not allow an attacker to target a specific implementation that may contain a flaw, the [[#programming-model-device-selection]] mechanism is a hint only, and the concrete device selection is left to the implementation. As a further mitigation, no device enumeration mechanism is defined.
```
dontcallmedom, line 394:
```suggestion
In order to not allow an attacker to target a specific implementation that may contain a flaw, the [[#programming-model-device-selection]] mechanism is a hint only, and the concrete device selection is left to the implementation - a user agent could for instance choose never to run a model on a device with known vulnerabilities. As a further mitigation, no device enumeration mechanism is defined.
```

index.bs
```diff
@@ -383,6 +383,41 @@ This API is disabled by default in all cross-origin frames using the [[#permissi
...
+
+Issue: Hinting partially mitigates the concern. Investigate additional mitigations.
+
+The API design minimizes the attack surface for the compiled computational graph. The {{MLGraphBuilder}} interface that hosts the various operations is a data definition API and as such doesn't execute anything, only constructs data. What follows, is that the potential for an attack is limited to when binding the data to the graph before executing it by invoking the {{MLGraph/compute()}} method. This enables implementers to focus on hardening the {{MLGraph/compute()}} method. For example, by making sure it honors the boundary of data and fails appropriately when the bounds are not respected.
```
dontcallmedom, line 398: this will need an editorial update if/when #257 lands

index.bs
```diff
@@ -383,6 +383,41 @@ This API is disabled by default in all cross-origin frames using the [[#permissi
...
+
+Purpose-built Web APIs for measuring high-resolution time mitigate against timing attacks using techniques such as resolution reduction, adding jitter, detection of abuse and API call throttling [[hr-time-3]]. It is expected a typical multi-process architecture implementation of the WebNN API makes use of inter-process communication (IPC) mechanism that inherently adds additional jitter on top of the built-in high-resolution time mitigations. Regardless of these mitigations, implementers are always advised to consider and test their implementations against timing attacks, and in particular if IPC is not used.
```
dontcallmedom, line 400:
```suggestion
Purpose-built Web APIs for measuring high-resolution time mitigate against timing attacks using techniques such as resolution reduction, adding jitter, detection of abuse and API call throttling [[hr-time-3]]. The practical deployment of WebNN implementations are likely to bring enough jitter to make timing attacks impractical (e.g. because they would use IPC) but implementers are advised to consider and test their implementations against timing attacks.
```

index.bs
```diff
@@ -383,6 +383,41 @@ This API is disabled by default in all cross-origin frames using the [[#permissi
...
 
+Some operations such as {{split()}}, {{slice()}} and {{squeeze()}} that change the shape of tensors mid-calculation may lead to incorrect assumptions in later operations and possibly lead to out-of-bounds access attempts. Implementers are adviced to anticipate such out-of-bounds access issues and add appropriate mitigations.
+
+Issue: Document operations susceptible to out-of-bounds access as a guidance to implementers.
+
+As a future-proofing, the API design allows certain operation (annotated with "can be generically emulated") defined in this specification to be deprecated for security, performance, or other reasons without breaking compability. This is made possible by high-level functions that are defined in terms of smaller primitive operations defined in this specifications. This enables a native implementation of a high-level function to be replaced with a polyfill implementation.
```
wchao1115, line 390: I'm not sure I agree with this statement. How do emulated operations pose security threat and should be deprecated in the future if it's an implementation choice of the API?

index.bs
wchao1115, line 394: It's actually important for the applications to understand how a device is selected and that the device selection rule, if we  would have one, should be a normative behavior and not merely informative.

index.bs
wchao1115: Reshaping is quite common in an ML graph. I'm not entirely sure why it should be discouraged from a security standpoint.

index.bs
anssiko: Suggested change updated accordingly.

index.bs
anssiko: This refers to the following ops that can be generically emulated (and for which we have informative "can be generically emulated" text in place):

`batchNormalization, clamp, elu, gemm, gru, gruCell, hardSigmoid, hardSwish, instanceNormalization, leakyRelu, linear, relu, sigmoid, softmax, softplus, softsign, split, tanh`

Suggested change updated. I did not enumerate these ops, but refer to them as "certain operations that can be generically emulated".

index.bs
anssiko: It sounds like we'll simply drop this bullet as suggested.

anssiko:
@wchao1115 @dontcallmedom @huningxin thanks for your comments, we're iterating toward a mergeable PR.

For discussion on this week's call.

index.bs
anssiko:
```suggestion
As a future-proofing measure, the API design allows certain operations that can be generically emulated to be deprecated for security, performance, or other reasons without breaking compatibility. This is made possible by high-level functions that are defined in terms of smaller primitive operations defined in this specifications. This enables a native implementation of a high-level function to be replaced with a polyfill implementation.
```

anssiko: (Reposting to the right line 390:)

This refers to the following ops that can be generically emulated (and for which we have informative "can be generically emulated" text in place):

`batchNormalization, clamp, elu, gemm, gru, gruCell, hardSigmoid, hardSwish, instanceNormalization, leakyRelu, linear, relu, sigmoid, softmax, softplus, softsign, split, tanh`

Suggested change updated. I did not enumerate these ops, but refer to them as "certain operations that can be generically emulated".

anssiko:
I incorporated all the suggestions made in this PR to trigger a PR preview update.

See: [Preview](https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/251.html) | [Diff](https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/251/19a09ab...a737271.html)

huningxin (approved):

Issue #230, opened 2021-11-18
Should WebNN support async APIs?


huningxin:
As mentioned in https://github.com/webmachinelearning/webnn/issues/229, the existing WebNN graph building (MLGraphBuilder.build) and execution (MLGraph.compute) are sync APIs. This is required by Wasm (C++) based ML framework's backend implementation. To avoid blocking the main thread, the good practice is to call these synchronous APIs in a worker context.

There are JavaScript based ML frameworks, like TensorFlow.js, that are mainly used in the main thread. Should WebNN support async APIs in the main thread? This would help not only the JS ML frameworks but also broader JS adoption of the API.

/cc @pyu10055

bbernhar:
@huningxin 

I can help fill in the GPU story a bit.

GPUDevice does not require WebNN compute() to be sync. WebGPU schedules work on the GPU/queue timeline which is async by design. WebNN behavior's here is non-normative and should clarify it shares ownership of WebGPU's (implicit) queue so calling compute() goes back on the queue timeline, which never blocks the main thread. This will be important for interop because we cannot have compute() work being scheduled alongside WebGPU work on two different timelines.

huningxin:
Thanks @bbernhar for the clarification of GPU story. I agree this is important for WebNN-WebGPU interop and we need to improve this part.

> GPUDevice does not require WebNN compute() to be sync.

I suppose you mean if WebNN `MLContext` is [created from `GPUDevice`](https://www.w3.org/TR/webnn/#dom-ml-createcontext-gpudevice-gpudevice), the sync `compute()` would not block the main thread because its actual work is scheduled async on the GPU/queue timeline. Correct?

I think this align with what @RafaelCintron shared in the [WebML WG Teleconference – 18 Nov 2021](https://www.w3.org/2021/11/18-webmachinelearning-minutes.html#t05):

> RafaelCintron: WebGPU and WebGL are technically async APIs because you submit it on the CPU which then queues it to the GPU
> … when WebNN is used in the GPU, the commands are executed in parallel from a CPU perspective, so essentially async - we should block on the CPU until the inference is done

This is true if the inputs and outputs are GPU resources, such as `GPUBuffer`, because developers can use [`GPUBuffer.mapAsync`](https://www.w3.org/TR/webgpu/#dom-gpubuffer-mapasync) to read back the results without blocking the main thread.

[...]

Issue #226, opened 2021-11-10
Integration with real-time video processing


dontcallmedom:
The WebRTC Working Group is working on an API to allow fast processing of real-time video, with two proposals under discussion towards convergence:  https://github.com/alvestrand/mediacapture-transform and https://github.com/jan-ivar/mediacapture-transform (see also [relevant issues on convergence](https://github.com/w3c/mediacapture-extensions/issues?q=is%3Aissue+is%3Aopen+label%3Amediacapture-transform)). Chromium has started shipping an implementation based on the first proposal which should allow for initial experimentation with the overall approach.

Since we can expect a lot of this real-time processing to be done based on Machine Learning models, and as suggested by the [Web Machine Learning Working Group charter](https://www.w3.org/2021/04/web-machine-learning-charter.html#coordination), we should ensure that models loaded via WebNN-backed JS frameworks can be used in the context of that API (in particular, of a WHATWG Streams-based API, running in a worker context, with video frames coming from a webcam likely stored in a GPU memory context), and that it delivers actual performance improvements (in particular that any boost from the hardware acceleration provided by WebNN doesn't get overtaken by cost associated with e.g. memory copies).

My sense is that the best way to determine this would be:
* to build a prototype that integrates the mediacapture-transform API (in a worker context) with e.g. a TF.js model that allows for background blur
* measure performance of the said prototype across various TF.js backends, including a WebNN-native one; ideally this would include specific measurements of memory copies, although the raw result on FPS may already give sufficient hints

While the real-time video processing framework in WebRTC is still somewhat in flux, I think we have enough convergence on the overall picture and a good enough basis for experimentation with the Chromium implementation to get started with such a work.  The WebRTC Samples repo has a [few examples of that API in action](https://github.com/webrtc/samples/tree/gh-pages/src/content/insertable-streams) (`video-crop` in particular exercises it in a worker context).

/cc @aboba @alvestrand @jan-ivar

anssiko:
Thanks @dontcallmedom for starting this discussion that clearly benefits from coordination between the WebML and WebRTC WGs.

[...]

PR #257, opened 2022-03-10
Context-based graph execution methods for different threading models.

wchao1115:
Alternative design to #255 for various graph execution methods, with both *immediate* and *async execution* methods folded into the context interface, making it easier for developer to associate execution method to the context created with different options. The *queued execution* method, however, is left to a separate `MLCommandEncoder` interface as it requires multi-step calling pattern that ends with the `finish` method, consistent with the WebGPU GPU command encoder calling pattern.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/257.html" title="Last updated on Apr 22, 2022, 5:34 PM UTC (0ea819d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/257/5b6913b...0ea819d.html" title="Last updated on Apr 22, 2022, 5:34 PM UTC (0ea819d)">Diff</a>


index.bs
```diff
@@ -2166,71 +2428,121 @@ interface MLGraph {
...
+  required MLGPUResource resource;
+  required sequence<long> dimensions;
+};
+
+typedef record<DOMString, (MLGPUResource or MLGPUInput)> MLNamedGPUInputs;
+typedef record<DOMString, MLGPUResource> MLNamedGPUOutputs;
+
+[SecureContext, Exposed=(Window, DedicatedWorker)]
+interface MLCommandEncoder {
+  undefined initializeGraph(MLGraph graph, MLNamedGPUInputs inputs);
```
anssiko, line 2447: We may want to name `inputs` arguments passed to `initializeGraph()` and `dispatch()` differently to make it clearer their semantics differ. I think this would improve developer ergonomics.

wchao1115: I will add more descriptive text into describing these two methods in more detail. I dont think just changing the param name will make any material difference.

wchao1115: BTW, I'll encourage other reviewers to take a look at this PR and given their opinion on it relative to #255. I hope we can come to an agreement on this important technical topic soon. It has been a while in the making.

index.bs
```diff
@@ -432,10 +434,29 @@ the computation graph used to compute one or more specified outputs. The key
...
 
-The {{MLGraph/compute()}} method of the {{MLGraph}} interface is used to execute the
-compiled computation graph (to perform inference). The caller supplies the input
-values using {{MLNamedInputs}}, binding the input {{MLOperand}}s to their values.
-The caller supplies pre-allocated buffers for output {{MLOperand}}s using {{MLNamedOutputs}}.
+Once the {{MLGraph}} is constructed, there are multiple ways by which the graph may be executed. The
+{{MLContext/compute()}} method represents a way the execution of the graph is carried out immediately 
+on the calling thread, which must also be a worker thread. The execution produces the results of the computation 
+from all the inputs bound to the graph. This type of execution is limited only to when the computational device 
+bound to the context is a CPU device.
```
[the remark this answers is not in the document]
wchao1115: @dontcallmedom The GPU execution is inherently async but not on the app's timeline. It's async on the GPU timeline. So having a sync call on a GPU device would mean stalling the CPU until the GPU queue has a chance to be executed. That is undesirable and none of our known use case will do that.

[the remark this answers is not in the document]
wchao1115: The TF.js use case on a GPU device is such that the execution call is done on the main thread and not the worker thread. However, the compute result is deferred until the GPU executes the result and the readback occurs. The reason the sync API is limited to only the worker thread is for a different purpose. It's because even for the CPU execution, the main thread should never block. 

Again, this change assumes that the app knows what type of MLContext it creates. WebNN makes that assumption.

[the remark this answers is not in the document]
dontcallmedom: > the app knows what type of MLContext it creates. WebNN makes that assumption.

To clarify my point - I'm not arguing here (as I did in #255) that the `MLContext` may be coming from a third party library. My assertion is that as a developer, even if I set `MLDdevicePreference` to CPU, the spec gives no guarantee that I'll get back a CPU-based `MLContext`, and doesn't even expose if the `MLContext` I got back is CPU-based or not. So as a developer, I don't know if I can or cannot execute the sync version of the API, except by actually executing it and getting an exception (which I don't think is a practical API to use).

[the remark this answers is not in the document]
wchao1115: > the MLContext may be coming from a third party library

I think we will need to establish if this is a supported use case. As of the current design goal, since WebNN is designed to be a framework backend, it is highly unlikely that the MLContext will be coming from a third party library. Frameworks such as TF.js owns the creation of MLContext and the execution that follows. 

>  if I set MLDdevicePreference to CPU, the spec gives no guarantee that I'll get back a CPU-based MLContext, and doesn't even expose if the MLContext I got back is CPU-based or not.

This is a separate issue and a valid concern. In order to support a default setting, there needs to be a defined default value for `MLDevicePreference` and that it's currently defined as a CPU device. To make it clearer, we could simply remove the `cpu` option from the enum. Should we go ahead and do that?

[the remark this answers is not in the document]
dontcallmedom: > there needs to be a defined default value for MLDevicePreference and that it's currently defined as a CPU device. To make it clearer, we could simply remove the cpu option from the enum. Should we go ahead and do that?

I don't think this helps - the real question is whether there is a guarantee that requesting a default (or a CPU) device gives you a CPU MLContext or not. My reading of the spec is that there isn't such a guarantee.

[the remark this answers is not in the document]
wchao1115: My point is to make that guarantee thru the `default` option, and eliminate the `cpu` option as it would be redundant.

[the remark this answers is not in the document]
dontcallmedom: let's bring this to a separate issue - I think I would keep `"cpu"` as a name and instead make it the WebIDL-default value of `MLContextOptions.devicePreference`; but my sense is that the primary question around such a change is more about the impact of making `devicePreference` no longer a hint but a setting.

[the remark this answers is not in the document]
anssiko: We can use issue https://github.com/webmachinelearning/webnn/issues/169 for device selection discussion.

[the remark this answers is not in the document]
wchao1115: I agree with making "cpu" a default option and remove "default" as @dontcallmedom suggested here. I don't think the word "preference" necessarily implies a hint.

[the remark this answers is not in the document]
huningxin: As I [commented](https://github.com/webmachinelearning/webnn/pull/255#discussion_r816500166) in #255, I think it should support GPU as well. It would allow the caller who requires sync API could also access GPU, e.g., TFLite WebNN delegate. When the compute device is GPU, it would block the calling thread until the result is readback into the output ArrayBuffers. This would not impact the responsiveness of the main (UI) thread as its usage is restricted in the DedicatedWorker.

explainer.md
```diff
@@ -47,7 +47,7 @@ const bufferB = new Float32Array(4).fill(0.8);
 const bufferC = new Float32Array(4);
 const inputs = {'A': bufferA, 'B': bufferB};
 const outputs = {'C': bufferC};
-graph.compute(inputs, outputs);
+context.compute(inputs, outputs);
```
huningxin, line 50: missing `graph` for compute?

explainer.md
```diff
@@ -153,7 +154,7 @@ export class NSNet2 {
       'gru94': gru94Buffer,
       'gru157': gru157Buffer
     };
-    return this.graph.compute(inputs, outputs);
+    return this.context.compute(inputs, outputs);
```
huningxin, line 157: ditto.

explainer.md
wchao1115: Fixed

explainer.md
wchao1115: Fixed.

explainer.md
wchao1115: > As I https://github.com/webmachinelearning/webnn/pull/255#discussion_r816500166 in https://github.com/webmachinelearning/webnn/pull/255, I think it should support GPU as well.

I'll defer this to @RafaelCintron who insisted that GPU sync call is bad for user experience.

index.bs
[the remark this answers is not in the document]
wchao1115: I'll defer this to @RafaelCintron who insisted that GPU sync call is bad for user experience.

index.bs
wchao1115: Please look at the latest commit. I've removed `MLCommandEncoder` and let `computeAsync` call handle both the CPU and GPU async use case per discussion with @RafaelCintron and @bbernhar. Detail please refer to this reply https://github.com/webmachinelearning/webnn/pull/255#issuecomment-1079981083

index.bs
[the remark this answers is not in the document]
wchao1115: > let's bring this to a separate issue - I think I would keep `"cpu"` as a name and instead make it the WebIDL-default value of `MLContextOptions.devicePreference`; but my sense is that the primary question around such a change is more about the impact of making `devicePreference` no longer a hint but a setting.

In the latest commit, I've removed the `default` option and rename `MLDevicePreference` to `MLDeviceType` to make it more clear that it is not merely a hint. @dontcallmedom Please take a look.

index.bs
```diff
@@ -489,18 +508,17 @@ An {{MLContext}} interface represents a global state of neural network execution
...
 <table>
-  <tr><th>Device Type<th>ArrayBufferView<th>GPUBuffer<th>GPUTexture<th>WebGLBuffer<th>WebGLTexture
-  <tr><td>GPUDevice<td>Yes<td>Yes<td>Yes<td>No<td>No
-  <tr><td>WebGLRenderingContext<td>Yes<td>No<td>No<td>Yes<td>Yes
-  <tr><td>default<td>Yes<td>No<td>No<td>No<td>No
-  <tr><td>gpu<td>Yes<td>No<td>No<td>No<td>No
+  <tr><th>Device Option<th>ArrayBufferView<th>GPUBuffer<th>GPUTexture<th>WebGLBuffer<th>WebGLTexture
   <tr><td>cpu<td>Yes<td>No<td>No<td>No<td>No
+  <tr><td>gpu (GPUDevice == null)<td>Yes<td>No<td>No<td>No<td>No
+  <tr><td>gpu (GPUDevice != null)<td>Yes<td>Yes<td>Yes<td>No<td>No
```
huningxin, line 520: If the `MLContext` is created from a `GPUDevice`, I would suggest it only accepts the `GPUBuffer` and `GPUTexture`. This design would help simplify the spec / implementation of WebNN API and leave the `ArrayBufferView` uploading / readback to WebGPU API where it is already well defined.

huningxin:
Summarize my points, there should be three types of `MLContext` that support three usages:

| Usage | Creation Method | Device Type | Execution Method | Buffer Type |
|----|----|----|----|----|
| Default | Create from an `MLContextOptions` | CPU or GPU, <br> device selected by device and power preferences | async `compute` (both main thread and worker) <br> sync `compute` (worker only) | `ArrayBufferView` |
| WebGPU interop | Create from <br> a `GPUDevice` | GPU, <br> device selected by WebGPU API | submit `MLGraph` as a `GPUCommandBuffer` to `GPUQueue` | `GPUBuffer` and `GPUTexture` |
| WebGL interop | Create from a `WebGLRenderingContext` | GPU, <br> device selected by WebGL API | sync `compute` (both main thread and worker) ? | `WebGLBuffer` and `WebGLTexture` |

I have an open for the execution method for WebGL interop usage: should it be sync API and be available in both main thread and worker given GPU execution is async by natural? But how would it happen and be implemented? It seems to me there is lack of investigation on WebGL interop. Should we remove the WebGL interop support and leave it for following PR?

index.bs
```diff
@@ -535,74 +552,64 @@ enum MLPowerPreference {
 };
 
 dictionary MLContextOptions {
-  MLDevicePreference devicePreference = "default";
+  MLDeviceType deviceType = "cpu";
   MLPowerPreference powerPreference = "default";
+  GPUDevice gpuDevice = null;
+  WebGLRenderingContext glContext = null;
```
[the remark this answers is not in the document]
wchao1115: When `deviceType` is set to anything but "gpu", the `gpuDevice` even if given is ignored. I don't think this use case is uncommon considering that every field in the `MLContextOptions` is optional. By doing it this way, we avoid having multiple overload `createContext` methods.

wchao1115:
@huningxin The `default` device option is problematic for our use case because the caller doesn't know what type of device the user agent has selected. That type of design based on behavior hints is fine as long as the user agent retains full control of all the related behaviors when it decides to select a certain type of device. But in our case, the caller must choose the execution method that is best for their situation. This means that the caller must be conscious of their choice of the device type to use and therefore cannot defer that decision to the user agent. @dontcallmedom expressed this concern in his review, which I agreed.

huningxin:
@wchao1115 , my point is the spec should allow developers to specify their intent that "select the best device automatically". For that use case, the ["auto" proposal](https://github.com/webmachinelearning/model-loader/issues/30) from @yuhonglin is probably a good fit.

> On a separate issue, a design that allows synchronous GPU execution should be avoided because the CPU could be left needlessly stalled for a long period of time.

I don't think this is an issue, because the sync `compute` method is restricted to be used only in the worker thread. The stall of worker thread won't impact the responsiveness of UI which is running in the main thread. If developers don't want CPU to wait for GPU execution, they are free to use the async `compute` method.

As my understanding, the sync `compute` method is required to support a framework code base that expects sync API call in a backend, e.g., TFLite WebNN delegate. The spec shouldn't restrict this type of frameworks only to CPU execution.

wchao1115:
Can we keep the 'auto' proposal as a separate issue for now? This is a very hard topic that to my knowledge none of the leading cross-vendor ML platform has fully solved. The only known case when this is supported today is CoreML and that only works on Apple M1 hardware platform.

As for supporting sync call on the GPU device on the worker thread, I would like to understand a bit more of the current use case. Since the GPU execution is async by nature (but on the GPU timeline), it seems odd that a worker thread would want to block for it.

index.bs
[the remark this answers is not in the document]
wchao1115: The way you explain it here is not really "default" but rather "auto" option i.e. automatic device selection. This is a hard subject with potential backward and forward compatibility guarantee involved. Also the definition of "best" varies greatly depending on what the application is trying to do. Generally speaking, this type of policy behavior should not belong to the backend component like the WebNN API. Rather, it should be part of the application logic or in the framework on behalf of the application. 

"default" and "auto" aren't really the same thing. Default is a well-defined behavior that is chosen to be a default behavior when nothing more explicit is specified, while the behavior of `auto` is dynamic implying the notion of "best fit", a more open-ended policy. 

I prefer that we discuss automatic device selection as a separate issue.

huningxin:
> As for supporting sync call on the GPU device on the worker thread, I would like to understand a bit more of the current use case.
 
As I mentioned, the sync `compute` call is required by a framework backend that infers sub-graph synchronously, for example [TFLite WebNN delegate](https://github.com/tensorflow/sig-tfjs/blob/main/rfcs/20211020-webnn-delegate-for-tflite-web.md) needs to implement the sync [`SimpleDelegateKernelInterface::Eval`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/delegates/utils/simple_delegate.h#L61) method for sub-graph inference. The spec should allow these frameworks to use GPU.

index.bs
```diff
@@ -626,14 +655,244 @@ interface MLContext {};
...
+                1. Set |outputSize| to the product of |outputSize| and |dimension|.
+            1. If |outputSize| is greater than the length of |value|, then:
+                1. Throw a {{DataError}} {{DOMException}} and stop.
+            1. Else:
+                1. Set the values of |value| to the values of |outputTensor|.
+    1. Return Promise<{{undefined}}>.
+</div>
+
+<div class="note">
+When the {{[[contextType]]}} of {{MLContext}} is set to [=webgpu-context|webgpu=] with the {{MLContextOptions}}.{{gpuDevice}} also specified, the user agent queues the ML workload needed to compute the result of the operations in the graph in an internal command buffer before submitting it to the {{gpuDevice}}'s default {{GPUQueue}}. The user agent then returns the call with a {{Promise}} that resolves once the queue finishes processing all the work submitted to it.
```
huningxin, line 891: It is unclear to me that who submits the internal command buffer to gpuDevice's default GPUQueue. And when to submit? I am also a bit confused by the usage of the returned promise. Is its usage same as the promise returned by WebGPU's [`GPUQueue::onSubmittedWorkDone`](https://www.w3.org/TR/webgpu/#dom-gpuqueue-onsubmittedworkdone)?

wchao1115:
> As I mentioned, the sync `compute` call is required by a framework backend that infers sub-graph synchronously. For example, [TFLite WebNN delegate](https://github.com/tensorflow/sig-tfjs/blob/main/rfcs/20211020-webnn-delegate-for-tflite-web.md) needs to implement the sync [`SimpleDelegateKernelInterface::Eval`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/delegates/utils/simple_delegate.h#L61) 

Thanks @huningxin. I think it makes sense. Can we assume that the TFLite delegate use case only happen in the worker thread?

huningxin:
@wchao1115 
> Can we assume that the TFLite delegate use case only happen in the worker thread?

I suppose so. People experienced [unresponsiveness issue](https://github.com/tensorflow/tfjs/issues/2723) by inferring large model with TF.js Wasm backend (same applies to TFLite Wasm backend) in main thread and [request](https://github.com/tensorflow/tfjs/issues/5842) to move all heavy processing to worker thread. /cc @pyu10055 

Once this PR landed, we can update the design of TFLite WebNN delegate to reflect this restriction. WDYT?

wchao1115:
Ok. Sounds good. I'll make that change.

index.bs
bbernhar: At some point, I suspect `computeAsync` will probably need to also accept a sequence of `GPUCommandBuffers` (created by WebGPU, not WebNN) if WebNN is to efficiently interleave 3D/Compute and ML work. 

WebNN's implementation of `computeAsync` has no way to de-dup submits. `GPUQueue.Submit` was intended to be called _only once per frame_ - the per-submit call overhead by the WebGPU runtime is quite high due to GC. So long as `computeAsync` is responsible to _execute_ `GPUCommandBuffers`, there is no issue mixing ML and GPU commands in them and batching them together into a single `computeAsync` call will be far more efficient for interop. 

Hopefully that makes the issue a bit clearer.

wchao1115: > At some point, I suspect computeAsync will probably need to also accept a sequence of GPUCommandBuffers 

At that point, I'd much rather reintroduce the `MLCommandEncoder` interface and have all the fine-grained, intra-command buffer interop be done there as opposed to overloading `computeAsync` to become something much more complicated. 

> WebNN's implementation of computeAsync has no way to de-dup submits. 

Given that `computeAsync` will submit to the queue on behalf of the caller the content that is only generated and recorded within its own internal command buffer, I don't see a possibility of `computeAsync` causing a double submission of the same workload if that's what you're referring to. Am I missing anything?

bbernhar: @wchao1115

Consider what happens if a WebGPU buffer, being rendered into by WebGPU is also referenced by the internal command buffer generated by WebNN (eg. shared tensor, if you will). If the WebGPU developer does not call submit, the buffer contents will not be updated. Meanwhile, if `computeAsync` only "submits" it's own internal command buffer, it will see the wrong result. To remedy, the WebNN developer has no choice but to submit twice. Once through WebGPU, then again using WebNN.  Now, what if I want to do a GPU pre-pass AND post-pass with some ML work in-between? That's at-least 2x WebGPU submits per frame or "submit per ml context". If the cost of submission exceeds the benefit of using ML, no WebNN for you.

huningxin: > Now, what if I want to do a GPU pre-pass AND post-pass with some ML work in-between?

The [background blur sample](https://github.com/webmachinelearning/webnn/issues/226#issuecomment-1074968279) requires to execute the following steps for every video frame:

1. Import `VideoFrame` into a WebGPU `GPUTexture`.
2. Blur the imported texture by a WebGPU shader (blur.wgsl).
3. Prepare (normalize) the imported texture to the input tensor (`GPUBuffer`) by a WebGPU compute shader (preprocessing.wgsl).
4. Background segmentation by WebNN graph. Its input is the prepared `GPUBuffer` and its output is another `GPUBuffer` that contains the segmap.
6. Blend the imported texture and the blurred texture based on the segmap (blend.wgsl) into another `GPUTexture` (blend.wgsl). Then draw this texture into an offscreen canvas.
7. Export the offscreen canvas to a VideoFrame.

So with this design, there would be three submits per video frame: 1) `GPUQueue::submit` after step 3; 2) `MLContext::computeAsync` at step 4 submits the ML workload internally; 3) another `GPUQueue::submit` after step 6.

Do I understand correctly?

wchao1115: @huningxin I believe you are correct. 

@bbernhar Correct. The input buffer to an ML graph requires the content to be properly in place before the graph is executed. It is the caller's responsibility to make sure that the content is first rendered before it being processed in the ML graph during an inference call. Technically, this can be done either by having both the render workload and the ML workload be in the same command buffer submitted together, or on separate command buffers submitted individually or together.  When you said "submit twice" I thought you meant having the same workload in a command buffer submitted twice. That is not the case, so thanks for clarifying it. 

Do we really think that the cost of multiple queue submissions can overshadow either the rendering cost or the inference cost in the ML graph? I find that hard to believe. So although having a single submission seems cleaner from the WebGPU point of view. I don't really think multiple submission is a deal breaker, especially if it buys us a clear and clean separation between how the rendering workload and ML workload are executed across WebGPU and WebNN. I think it's just a trade-off.

bbernhar: @wchao1115 

I would argue MLCommandBuffer and GPUCommandBuffer, used by WebNN and WebGPU respectively, equally demonstrate a "clean separation". My point is submit has a _fixed CPU overhead_ and it's recommended (for WebGPU developers) to submit enough work (at once) to hide this latency. Otherwise, smaller submits finish faster then newer ones can be submitted and the GPU will sit idle.

huningxin: According to my experience of adding WebNN (segmentation) into a WebGPU-based (blur) pipeline, if the ML workload could be wrapped into an `MLCommandBuffer`, and be submitted together with pre and post `GPUCommandBuffer`s via a single `GPUQueue::submit`, it would be simple and clean.

I would also recommend to keep `computeAsync` just working with `ArrayBufferView`s. This would help separate the WebNN standalone usage from WebNN/WebGPU interop usage. I hope it would make it simple for developers to select the execution method based on the usage, e.g., something like:

1. If use WebNN API standalone, then:
   1. if need sync API, then:
      1. use `MLContext::compute` in worker thread
   1. else:
      1. use `MLContext::computeAsync`
3. else if use WebNN in WebGPU pipeline, then:
   1. (open to discuss) `MLCommandBuffer` or `MLCommandEncoder` proposal
5. else if use WebNN in WebGL pipeline, then:
   1. (need more investigation)

My points are also reflected in [the table](https://github.com/webmachinelearning/webnn/pull/257#issuecomment-1080120454) of earlier comment if that is more clear.

wchao1115: I believe I've made my point in this PR that there is no perfect design that satisfies all requirements. And I feel we have been circling around what I see as an unavoidable trade-off between the two designs I've been proposing all along. 

1. WebNN returns a `GPUCommandBuffer` filled with ML workloads, then let the caller be responsible for all workload submission. Since the caller owns workload submission, they can fold and order graphics and ML workload in a single submission. This is what @huningxin is arguing for, and it is my original proposal. This design however poses an implementation challenge in that WebNN must also implement the returning command buffer such that the WebGPU implementation can understand and process. Simply because it requires that a WebNN method returns a WebGPU interface.
2. WebNN owns both the generation and the submission of its own ML workload via its own internal command buffer. The only dependency on WebGPU is the incoming `GPUDevice` and its associated default command queue. This is a simpler approach that decouples WebNN from any WebGPU-generated type (e.g. `GPUCommandBuffer` etc) but with a drawback that it may require multiple submissions from the caller in an interop scenario e.g. the blur prototype Ningxin is working on. It is also the design that I believe @bbernhar preferred in our offline conversation with @RafaelCintron on the topic before I removed `MLCommandEncoder` in the latest commit based on that consensus.

I hope I summarize it well enough here. Ultimately, unless someone else wants to propose something different, I think we need to choose between the two and agree on the trade-off that each design presents.

wchao1115: @bbernhar Because `computeAsync` also has to take care of the CPU cases. Your proposal calls for the callers optionally pass in their `GPUCommandBuffer` with rendered content from WebGPU to the WebNN workload submission method. It's just a bit messy to have a method taking WebGPU command buffers while at the same time able to populate CPU ML work on a worker thread.

huningxin: @bbernhar 

> Not sure I understand why it would be more complicated to re-use MLContext::computeAsync for case (2) too, do you mind elaborating a bit more?

Because the `MLContext::computeAsync` also needs to support `ArrayBufferView` for case (1). If the compute method only submits ML workload to the device queue for case (2), it doesn't have to be async call, just like `GPUQueue::submit`. The returned promise seems duplicated and confusing to me. It also requires multiple queue submissions for video processing use case. And as @RafaelCintron [mentioned](https://github.com/webmachinelearning/webnn/issues/230#issuecomment-1031849403), if WebGPU adds multi-queue in the future, this method needs to be changed again since it currently only submits to the default queue. I suppose the complex comes by packing the two usages into one method. 

Another perspective is, if we keep `MLContext::computeAsync` (and `MLContext::compute`) only for case (1), it would give us a "clean separation" of the dependency. I think it would enable us move fast on the spec development for WebNN standalone usage (case (1)).

> Getting GPUQueue::submit to accept MLCommandBuffer (or GPUCommandBuffers filled with ML commands) requires a discussion with the WebGPU WG, at a minimum.

I would hope that conversation could be started as early as possible. There is an open [issue](https://github.com/gpuweb/gpuweb/issues/2500) in WebGPU repo for this. The current [plan](https://www.w3.org/2022/03/24-webmachinelearning-minutes.html#t07) is that we make the consensus first within Web ML WG and then we coordinate with WebGPU WG for broader review. Does the plan make sense?

bbernhar: > If the compute method only submits ML workload to the device queue for case (2), it doesn't have to be async call, just like GPUQueue::submit

`GPUQueue::submit` is always `async`, hence my confusion of having another "async" execution path for WebNN (but ONLY for the CPU). It's not obvious to me it's a change of dispatch (ie. worker thread vs WARP) that distinguishes them, since the returned promise can be ignored.

Your plan SGTM.

wchao1115:
Commit #8 contains the following changes. Hopefully this is the last for this PR. 
1. Allow GPU sync execution (limited to worker thread calling thread). I spoke with @RafaelCintron and he is signed off on that.
2. Both sync and async execution only accepts `ArrayBufferView` -- meaning default context (those created with `MLContextOptions`) always upload/download inputs/outputs if needed by the GPU device. 
3. Device type must be explicit, default to CPU device. We assume that the caller of WebNN always deal with specific device type, as defering that choice to the user agent will only make resource management more complicate and less efficient. 
4. Reintroduce `MLCommandEncoder`. After much back and forth, I think a separate context type that explicitly deals with WebGPU interop is the most practical and clean approach. An implementation that can't support this requirement should fail context creation. Also, after much deliberation, I prefer to stick with the original proposal of `MLCommandEncoder` that produces WebGPU-compatible `GPUCommandBuffer` as opposed to a WebNN-specific command buffer. We need to have a follow-up with the WebGPU CG on how to implement WebNN in a way that produces a compatible `GPUCommandBuffer`. 

This change is unlikely to make everyone happy, that solution probably doesn't exist. But I hope it's a reasonable enough compromise where everyone gets something they want. The net result of completing this change at this stage of the draft spec is certainly going to be far better than nothing.

index.bs
```diff
@@ -626,14 +647,264 @@ interface MLContext {};
...
+            1. For each |dimension| of dimensions of |outputTensor|:
+                1. Set |outputSize| to the product of |outputSize| and |dimension|.
+            1. If |outputSize| is greater than the length of |value|, then:
+                1. Throw a {{DataError}} {{DOMException}} and stop.
+            1. Else:
+                1. Set the values of |value| to the values of |outputTensor|.
+    1. Return Promise<{{undefined}}>.
+</div>
+
+<div class="note">
```
huningxin, line 889: This note section should be removed because the `computeAsync` method is now only allowed in an `MLContext` created with `MLContextOptions`.

index.bs
```diff
@@ -2202,159 +2461,140 @@ interface MLGraph {
...
-<div class="example">
-The following code showcases the computation with dynamic input dimensions.
-<pre highlight="js">
-function sizeOfShape(array) {
-  return array.reduce(
-      (accumulator, currentValue) => accumulator * currentValue);
-}
+<script type=idl>
+partial interface MLCommandEncoder {
+  undefined initializeGraph(MLGraph graph, MLNamedGPUInputs constants);
```
huningxin, line 2499: Does it mean the caller should define the weight operands by `MLGraphBuilder::input` when building the graph?

This may lead to different ways of defining weights:
1.  Using `MLGraphBuilder::constant` for `MLContext` created from `MLContextOptions`
2.  Using `MLGraphBuidler::input` for `MLContext` created from `GPUDevice`

Should we also define `MLContext::initializeGraph` (and probably also with `MLContext::initializeGraphAsync`) that takes `MLNamedArrayInputs`?

Then the `MLGraphBuilder::constant` could just create a named placeholder (or just overload `MLGraphBuilder::input`) that will be bound to different buffer type according to the context type when `initializeGraph` invoked.

huningxin (commented):
Thanks much @wchao1115 for this update.

> Allow GPU sync execution (limited to worker thread calling thread). 

+1

> Both sync and async execution only accepts ArrayBufferView 

+1

> Device type must be explicit, default to CPU device. 

+1

> Reintroduce MLCommandEncoder. 

+1 with an open on constants initialization. Please take a look.

index.bs
wchao1115: Good point. Will do.

index.bs
wchao1115: Or we can just remove the second param to `initializeGraph`. Your comment makes me realize that the second param is entirely redundant to what already in the graph. There is already an overload for `constant` in the graph builder interface that would let the caller pass in a different type of resource e.g. `GPUBuffer`. 

I don't think we need `initializeGraphAsync`. The process of recording the workload to the command buffer typically happens very quickly and entirely on the CPU side. It's hardly a potentially block call and shouldn't have the need to offload it to an async call.

huningxin: > The process of recording the workload to the command buffer typically happens very quickly and entirely on the CPU side. It's hardly a potentially block call and shouldn't have the need to offload it to an async call.

This is true for `MLCommandEncoder`.

However, we may need to define how the "weight preprocessing" happens in a default `MLContext`. In current design, I suppose it would happen upon `MLGraphBuilder::build` being invoked. If the device type is `gpu`, the steps of weight processing might be:
1. record the initialization of the graph
2. submit the command buffer to the queue
3. wait for queue finishes the initialization

As the `build` is a sync method, it may block the calling thread while waiting for GPU queue execution.

> There is already an overload for `constant` in the graph builder interface that would let the caller pass in a different type of resource e.g. `GPUBuffer`.

I think the constant / weight uploading and preprocessing should also be context-based. Correspondingly the overload for `constant` in the graph builder could be removed. Something like:

| Context type | Constant / weight uploading and preprocessing method | Buffer type |
|---|---|---|
| "default" | `MLContext::initializeGraphAsync(graph, constants)` and <br> `MLContext::initializeGraph(graph, constants)` (only for worker) | `ArrayBufferView` |
| "webgpu" | `MLCommandEncoder::initializeGraph(graph, constants)` | `GPUBuffer` |

Hopefully, this would also make the graph builder implementation simpler, because it won't need to hold constant buffers any more.

wchao1115: Weight preprocessing in the "default" context case can be done on the first call to `compute`. I don't think there is a need for a dedicated `initializeGraph` call there. 

The `MLCommandEncoder` case needs an `initializeGraph` call because the purpose of the call isn't to execute the graph, but only to record the workload and return it in a command buffer. By producing the initialize workload in a command buffer, it allows more flexibility for the caller to choose how/when to submit them for execution.

wchao1115: Please take a look at the latest commit 59bbbe9a4821c6e9b6b5c11384bf9c6e50c60533.

anssiko (approved):

index.bs
huningxin: Thanks @wchao1115 , the latest commit looks good.

> Weight preprocessing in the "default" context case can be done implicitly on the first call to `compute`.

It's possible. At minimum, we should define when and where this happens. Otherwise, this would lead to some delay in the first call to `compute` that the caller may not expect. IMO, the flexibility and explicit control of the graph initialization by `MLCommandEncoder::initializeGraph` is useful for a good user experience, because the caller can ensure the graph is ready to compute in an efficient way. We may think about supporting it in the "default" context.

Given this is out of scope of this PR focusing on graph execution, I propose we deal with it with a separate issue. WDYT?

wchao1115: For the default context, I think your suggestion to have weight processing done as the last step of `MLGraphBuilder.build` makes a lot of sense to me, given that all the constant operands are already in place in the graph builder state at that time. I can put a note section to clarify that if it's useful. 

I still think having a separate `initializeGraph` call for the WebGPU context case is useful since it'll allow the app to fold the initialize workload to other graphics workload that they already have in the same queue submission.

huningxin: > For the default context, I think your suggestion to have weight processing done as the last step of `MLGraphBuilder.build` makes a lot of sense to me, given that all the constant operands are already in place in the graph builder state at that time. I can put a note section to clarify that if it's useful.

I think it's useful. But there is still an issue that the sync `build` method may block the calling thread until the graph initialization is done. This may be even worse when the device type is gpu that involves uploading weights to gpu.

> I still think having a separate `initializeGraph` call for the WebGPU context case is useful since it'll allow the app to fold the initialize workload to other graphics workload that they already have in the same queue submission.

I'd agree. Then we should also note that for the webgpu context, the `MLGraphBuilder.build` won't process the weights. And this would make the weights processing a bit complex because the `MLGraph` implementation may need to hold the unprocessed weights until `initializeGraph` being called. Actually I like the `initializeGraph` that takes the constant buffers. It allows the implementation simply pass the caller-supplied constant buffers to native ML API without holding a copy or reference.

wchao1115: > I can put a note section to clarify that if it's useful.

Added in this commit 0ed918fac4da4ada3501fffe2513da429e8e1a11 . Please take a look.

index.bs
```diff
@@ -989,6 +989,10 @@ interface MLGraphBuilder {
 };
 </script>
 
+<div class="note">
+The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph builder state up to the specified output operands into a compiled graph according to the type of {{MLContext}} backing it. When the {{[[contextType]]}} of the {{MLContext}} is set to [=default-context|default=] with the {{MLContextOptions}}.{{deviceType}} set to [=device-type-gpu|gpu=], the compiled graph is initialized right before the {{MLGraphBuilder/build()}} method call returns. This initialization stage is important for optimal performance of the subsequent graph executions. See [[#api-mlcommandencoder-graph-initialization]] for more detail.
```
huningxin, line 993: This initialization stage should also be important for cpu device. According to the existing CPU backends of WebNN-native, including XNNPack, oneDNN or MLAS, all of them will re-layout/re-order the static weights to optimized memory layout (packed/blocked) for optimal performance.

As the `build` method is a sync call, it would block until the graph is initialized. It would impact the responsiveness of UI when calling it on the main thread.

wchao1115: On responsiveness I think you can make the same argument even without considering weight processing. A straight-up compilation process could be expensive regardless. Perhaps, it should be constrained to a worker thread as well?

huningxin: You are right, actually I mentioned it in the original issue report https://github.com/webmachinelearning/webnn/issues/230.  

> Perhaps, it should be constrained to a worker thread as well?

Yes, we probably can do the same for the `build` as the `compute` that restricts the sync `build` method in the worker thread and introduces aync `buildAsync` method for the main thread.

wchao1115: How about 0ea819d7d57b000cce5188aa613fe5571f8de463 now?

wchao1115: How about this 0ea819d7d57b000cce5188aa613fe5571f8de463?

bbernhar:
@wchao1115 


I would make it clear (4) makes non-trivial assumptions (and breaking changes) with WebGPU. 1) It assumes WebGPU will support a completely new mechanism of interop at some point, and 2) WebGPU will treat `GPUCommandBuffer`, containing _non-WebGPU_ commands, equally equivalent to a `GPUCommandBuffer`, containing _only WebGPU_ commands.

Stepping back, what exactly is the problem (4) is trying to solve? To enable more efficient WebGPU interop for WebNN? Or forgo WebNN direct access to the GPU by using WebGPU?

bbernhar:
@wchao1115 

I like the use of `MLCommandBuffer` but I suggest we offer it for WebNN Interop. WebGPU interop wasn't meant to be a means for gate keeping GPU access on the web. Changing of the WebGPU execution model, by allowing ML work to execute under WebGPU, and requiring a new means of WebGPU interop, by sharing command-buffers with WebNN are just what-ifs and far from practical - does WebGPU do this for every API ?!? 

Since WebNN has no actual dependency on WebGPU's execution model, such a justification to WebGPU WG will be very hard and unlikely anyway and WebNN Interop would allow progress...

index.bs
huningxin: Thanks @wchao1115 , https://github.com/webmachinelearning/webnn/commit/0ea819d7d57b000cce5188aa613fe5571f8de463 looks good to me.

Should we support asynchronous graph initialization by introducing `buildAsync` for main thread usage? Because without that, the only way to build an `MLGraph` is calling `build` in a worker. It means, for asynchronous execution in the main thread, the caller has to transfer the `MLGraph` back to the main thread. This may be inconvenient and inefficient. So I propose we define `buildAsync` like:

```js
Promise<MLGraph> buildAsync(MLNamedOperands outputs);
```

Any thoughts?

huningxin:
@bbernhar 

> does WebGPU do this for every API ?!?

I suppose not. I think the use case, such as video background blur on GPU (#249) , would drive the requirement of WebGPU interop. As the investigation of #249 shows, the apps can do this with WebGPU-only pipeline. For better ML compute performance, the apps may also want to plug WebNN into that pipeline. The `MLCommandEncoder` proposal allows this usage. 

> I like the use of `MLCommandBuffer` but I suggest we offer it for WebNN Interop.

WebNN interop is a good perspective. As we discussed in the last WG call, @wchao1115 already did a great job in this PR that gives a clear separation between WebNN standalone usage and WebGPU interop. I believe it would allow more efficient iterations for each usage respectively. So I would suggest to merge this PR first and file a separate issue for WebNN interop proposal.

WDYT?

index.bs
wacky6: Should we change `build()` to return `Promise<MLGraph>`? Having the API default to async gives us lots of flexibility in the future.

Also, it's easy to write async calls in synchronous style with JS async-await. Converting sync / blocking calls to async / non-blocking calls is much harder.

huningxin: Thanks @wacky6 

> Also, it's easy to write async calls in synchronous style with JS async-await.

It's true. 

However, it is hard or inefficient for a C++ inference code of Wasm-based ML framework to integrate. That C++ inference code usually expects a sync method call. The sync methods (build/compute) are designed for that use case.

My suggestion is to keep the sync `build` method but restrict it in worker scope and introduce the async `buildAsync` for main thread / JS usage.

wacky6: Thanks for the explanation!

I think there're libraries to wrap async Web API calls in JavaScript. Like: https://web.dev/asyncify/ .

Are there any constraints limiting such libraries wrapping async WebNN methods to WASM sync method calls?

For Web APIs in general, I don't think "Async" suffic to JavaScript API is used elsewhere. Async functions are first-class citizens in JS, and most web developers expect async APIs nowadays.

If `buildAsync()` can be trivially converted to `buildSync` in WASM, I think just providing the `Promise<MLGraph> build()` method is better in terms of simplicity and expandability.

wchao1115: Thanks @wacky6 and @huningxin for the feedback. I've created issue #263 to track this request separately.

bbernhar:
@huningxin 

Interop can go both ways. Plugging WebGPU command buffers into WebNN achieves the exact same result; except, it doesn't break WebGPU and you can progress immediately. If not, I suggest we define it as `MLExternalCommandBuffer` (read-only) to denote such usage is one-way.

index.bs
huningxin: Thanks @wchao1115 . I have no other comments for this PR.

index.bs
huningxin: Track the async build issue in https://github.com/webmachinelearning/webnn/issues/263.

huningxin:
Thanks @bbernhar for your feedback. I am interested in knowing more. Could you please file a separate issue? So we can focus on it.

huningxin (approved):
Looks good, thanks again for @wchao1115 's great work.

bbernhar:
@huningxin Done.

https://github.com/webmachinelearning/webnn/issues/264

PR #286, opened 2022-08-29
Address PING review feedback

### LINKED ISSUE #280: Items from privacy review
See: https://github.com/w3cping/privacy-request/issues/96

(1) Please keep us posted with the results of your research into the extent of the threat of fingerprinting via execution time analysis and potential mitigations.

(2) Re: The ethics (and privacy implications) of the identified and potential use cases of the API. PING observed that a number of the use cases that the API would enable are highly privacy-invasive and, therefore, should include an analysis of the privacy implications of those use cases as well as mitigations (e.g. transparency, policy and/or user controls). (Note: While the described use cases are probably only a subset of use cases that this API could be used for, the examples provided in the specification should help guide the use (or not use) of the API in other situations.)

(3) What are the software implementations (or examples) that would eliminate or reduce compute unit scheduling from being a fingerprinting risk? (The text says: " Furthermore, software implementations can be used to further eliminate such artifacts.")

(3) Re: "If a future version of this specification introduces support for new a [device type](https://www.w3.org/TR/webnn/#device-type) that can only support a subset of [MLOperandType](https://www.w3.org/TR/webnn/#enumdef-mloperandtype)s, that may introduce a new fingerprint", what mitigations does the WG envisage should this occur?

(4) Re: "In general, implementers of this API are expected to be familiar with the [WebGPU Privacy Considerations](https://gpuweb.github.io/gpuweb/#privacy-considerations)", it would be helpful to state that implementers are expected to apply WebGPU Privacy Considerations to their implementations.

(5) "[Power preference](https://www.w3.org/TR/webnn/#power-preference) indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint." - If power preference is set to default, what might the user agent do, and could that reveal information about the device and/or user? (We discussed this during the PING call, but we were not able to sort out whether the hint is fingerprintable. Could you explain in more detail how the power preference works so we can understand the fingerprinting risk?)

Possible typo in privacy considerations: "worload" => "workload"; "allow the implementation better select" => "allow the implementation to better select"

anssiko:
Fix #280

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/286.html" title="Last updated on Sep 1, 2022, 1:00 PM UTC (0740090)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/286/92edaf7...0740090.html" title="Last updated on Sep 1, 2022, 1:00 PM UTC (0740090)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -222,6 +222,8 @@ This section illustrates application-level use cases for neural network
 inference hardware acceleration. All applications in those use cases can be
 built on top of pre-trained deep neural network (DNN) [[models]].
 
+Note: Please be aware that some of the use cases described here, are by their very nature, privacy-invasive. Developers who are planning to use the API for such use cases should ensure that the API is being used to benefit users, for purposes that users understand, and approve. They should apply the Ethical Principles for Web Machine Learning [[webmachinelearning-ethics]] and implement appropriate privacy risk mitigations such as transparency, data minimisation, and users controls.
+
 ### Person Detection ### {#usecase-person-detection}
 
 A user opens a web-based video conferencing application, but she temporarily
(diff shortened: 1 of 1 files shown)


wchao1115 (approved):

anssiko:
@sandandsnow, this PR is an attempt to address the items for privacy review https://github.com/webmachinelearning/webnn/issues/280 raised by PING.

We'd like to get PING's go-ahead before merge.

anssiko:
Merging per PING confirmation: the privacy review has been completed https://github.com/w3cping/privacy-request/issues/96#issuecomment-1234419529

Thank you @sandandsnow and the whole PING!

PR #301, opened 2022-11-30
WiP: editorials

zolkis:
Content
- A trivial change for fixing some line endings.
- Changed the section titles to include interface/method/dictionary etc.
- Improved the ML.createContext() steps using prose also used in other spec's.
- Removed the unused internal slot [[implementation]].
- Added MLGraphBuilder constructor steps.
- Moved forward the permission section in ML interface.

For other interface using [[implementation]] as an internal slot, it's a bad name, kind of circular self-reference. We need to find another name. Implementation refers to the user agent code that implements the given algorithm(s). 

Next, will improve / define algorithms where missing.

If there are commits that are trivial to merge, but others are more controversial, I can create new PRs with the accepted commits cherry-picked from here and leave this PR open.

Feel free to edit/commit here.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/301.html" title="Last updated on Dec 1, 2022, 4:26 PM UTC (d8b8994)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/301/29c0323...zolkis:d8b8994.html" title="Last updated on Dec 1, 2022, 4:26 PM UTC (d8b8994)">Diff</a>


index.bs
```diff
@@ -583,36 +583,33 @@ interface ML {
 </script>
 
 The {{ML/createContext()}} method steps are:
-1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, then throw a "{{SecurityError!!exception}}" {{DOMException}} and abort these steps.
 1. Let |promise| be [=a new promise=].
+1. Return |promise| and run the following steps [=in parallel=].
```
anssiko, line 587: I'd suggest indent the "in parallel" steps that follow so the in parallel steps start from index 1.

I'm used to seeing the "Return _promise_." on its own line after the "in parallel" substeps. Reading the [in parallel definition](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) I don't think there's any normative impact either way so we can return early too if that is a more common practice.

If we want to return the promise on the same line with "run the following steps in parallel", we may want to update [`compute()`](https://www.w3.org/TR/webnn/#dom-mlcontext-compute) similarly.

Maybe return promise on its own line maps more directly with the code that is written to implement this (also for tests) and thus less error prone?

index.bs
```diff
@@ -583,36 +583,33 @@ interface ML {
...
     <dd>Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
     </dl>
-1. Issue the following steps to a separate timeline:
-    1. If the User Agent can support the |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}, then:
-        1. Set |context|.{{MLContext/[[implementation]]}} to an implementation supporting |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}.
-        1. [=Resolve=] |promise| with |context|.
-    1. Else:
-        1. [=Resolve=] |promise| with a new {{NotSupportedError}}.
-1. Return |promise|.
+1. If the UA cannot support |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}, [=reject=] |promise| with a new {{NotSupportedError}} and abort these steps.
```
anssiko, line 603: The spec uses "user agent" in its lower case form elsewhere except in the context of `[[implementation]]` internal slot. I'd suggest using "user agent" here too and update the uppercase form to match.

anssiko (changes requested):
Thanks @zolkis! A added a few minor suggestions.

index.bs
```diff
@@ -937,7 +931,24 @@ interface MLGraphBuilder {
...
 
-### batchNormalization ### {#api-mlgraphbuilder-batchnorm}
+{{MLGraphBuilder}} has the following internal slots:
+<dl dfn-type=attribute dfn-for="MLGraphBuilder">
+    : <dfn>\[[context]]</dfn> of type {{MLContext}}
+    ::
+        The context of type {{MLContext}} associated with this {{MLGraphBuilder}}.
+</dl>
+
+### The [=new=] {{MLGraphBuilder}} constructor steps are:
```
anssiko, line 941: You should remove `###` otherwise bikeshed thinks this line is a section header: https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/301.html#the-new-mlgraphbuilder-constructor-steps-are

When working with bikeshed I usually build locally to check how the output is rendered. The syntax between bikeshed and ReSpec is quite similar but there are differences such as this one. There's also a bikeshed Web API that works without local bikeshed install: https://github.com/webmachinelearning/webnn/#pull-requests

anssiko (changes requested):

index.bs
zolkis: No problem, I can use the prose used in Web Bluetooth, Compute Pressure etc, or the prose used in WebRTC and others.

> the method when invoked, MUST return [a new promise](https://webidl.spec.whatwg.org/#a-new-promise) promise and run the following steps [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel): 

or,

> - If [condition] is true, return a promise [rejected](https://w3c.github.io/webrtc-pc/#dfn-reject) with a newly created WhateverError.
> - (...) other steps
> - Let p be a new promise.
> - (all the other steps)
> - Return p.

I have mostly used/preferred the former, since it gives a better hunch on [what's happening](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) in the event/task/microtask loop. The method really returns and the rest happen on a new timeline that implements "parallel" on the given underlying platform.

index.bs
zolkis: Yes, that's a bug, nowadays it's simply "user agent".

index.bs
zolkis: Good catch, will fix all these.

index.bs
anssiko: I'll let the editors @huningxin @wchao1115 to share which convention they'd prefer to use.

index.bs
```diff
@@ -582,37 +582,34 @@ interface ML {
...
-1. Let |promise| be [=a new promise=].
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" and abort these steps.
+1. Return [=a new promise=] |promise| and run the following steps [=in parallel=].
 1. Let |context| be a new {{MLContext}} object.
-1. Switch on the method's first argument:
+1. Let |options| be the first argument.
+1. Switch on |option|:
     <dl class=switch>
-    <dt>{{MLContextOptions}}
+    <dt>{{MLContextOptions}} or {{undefined}}
```
anssiko, line 599: Good that you brought up the undefined type.

Should we handle undefined type as a separate case in this switch statement? Or maybe better to handle it as part of an Otherwise case to have a fall back to default for all unknown arguments including undefined?

Switch on _options_:
-> MLContextOptions
-> GPUDevice
-> Otherwise

Switch statement semantics are not specified in Web IDL, but it seems Otherwise is being popularized (out of default, otherwise, or else), notably by HTML LS.

index.bs
```diff
@@ -582,37 +582,34 @@ interface ML {
...
     <dd>Set |context|.{{[[contextType]]}} to [=default-context|default=].
-    <dd>Set |context|.{{[[deviceType]]}} to the value of {{MLContextOptions}}'s {{deviceType}}.
-    <dd>Set |context|.{{[[powerPreference]]}} to the value of {{MLContextOptions}}'s {{powerPreference}}.
+    <dd>Set |context|.{{[[deviceType]]}} to the value of |option|'s {{deviceType}} or "[=device-type-cpu|cpu=]".
```
anssiko, line 601: s/option/options

anssiko (changes requested):

index.bs
zolkis: Let's discuss this on the call. I am open - just expressed the JS convention, meaning since we have a default value, and it falls under the same switch option, I handled them together. I am completely open on it, though.

index.bs
zolkis: Thanks, fixed (will be committed later).

zolkis:
As agreed on the call, spawned #304 and #305 for merging the trivial changes. 
Keeping the other commits here for discussion.

anssiko:
@zolkis can we close this now?

zolkis:
I will close (and likely remove) this one as a cleanup, after #310 is merged.

zolkis:
The content here has been merged in other PRs. Closing.

PR #305, opened 2022-12-01
Editorial: improve titles by including interface/method/dict etc.

zolkis:
This is to align with other W3C specification prose. Trivial merge.
Feel free to edit/add commits.

@huningxin @wchao1115

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/305.html" title="Last updated on Dec 1, 2022, 4:40 PM UTC (982fcd2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/305/29c0323...zolkis:982fcd2.html" title="Last updated on Dec 1, 2022, 4:40 PM UTC (982fcd2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -539,7 +539,7 @@ The following table summarizes the types of resource supported by the context cr
 API {#api}
 =====================
 
-## navigator.ml ## {#api-navigator-ml}
+## The navigator.ml interface ## {#api-navigator-ml}
 
 A {{ML}} object is available in the {{Window}} and {{DedicatedWorkerGlobalScope}} contexts through the {{Navigator}}
 and {{WorkerNavigator}} interfaces respectively and is exposed via `navigator.ml`:
@@ -552,7 +552,7 @@ Navigator includes NavigatorML;
 WorkerNavigator includes NavigatorML;
 </script>
 
-## ML ## {#api-ml}
+## The ML interface ## {#api-ml}
 <script type=idl>
 enum MLDeviceType {
   "cpu",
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1211,7 +1211,7 @@ partial interface MLGraphBuilder {
     *output size = (input size - 1) ** *stride + filter size + (filter size - 1) ** *(dilation - 1) - beginning padding - ending padding + output padding*
 </div>
 
-### element-wise binary operations ### {#api-mlgraphbuilder-binary}
+### Element-wise binary operations ### {#api-mlgraphbuilder-binary}
```
huningxin, line 1214: Should it also mention "method"? such as "The methods for element-wise binary operations"?

index.bs
```diff
@@ -1248,7 +1248,7 @@ partial interface MLGraphBuilder {
         - *pow*: Compute the values of the values of the first input tensor to the power of the values of the second input tensor, element-wise.
 </div>
 
-### element-wise unary operations ### {#api-mlgraphbuilder-unary}
+### Element-wise unary operations ### {#api-mlgraphbuilder-unary}
```
huningxin, line 1251: ditto.

zolkis: Same argument as above. I suggest either keeping the generic section name, or split up to subsections for each method.

index.bs
anssiko: For consistency keep the top-level group as "Element-wise unary operations" and have "The x() method" subsections for each of these ops where you define "The x() method steps are:".

anssiko: Yes, for consistency we should do that unless too much boilerplate makes it actually harder to read.

Keep the current names in this PR and push separate PRs for "element-wise binary operations" and "element-wise unary operations".

Smaller self-contained PRs are easier to review. We can then see how it'll look like and which way to go.

zolkis: I think this would expand and clutter the spec way too much, with little added benefit.
What matters in all these methods is how the inputs and outputs are treated between the script and implementation and then the underlying platform.
These ops might be mapped directly to the underlying platform, meaning basically empty shell algorithms (and sections), not adding any useful information.
While we could do these changes, looks lower priority compared to other work on the spec, so I suggest we leave this to a later stage. When done, will be (easily) done as suggested.

anssiko: Sounds good.

anssiko (approved):
We will handle "element-wise binary operations" and "element-wise unary operations" updates in a separate PR.

PR #309, opened 2022-12-08
Add constructor steps to MLGraphBuilder

zolkis:
Cherry-picked from #301 to help reviewing.
It refers to (missing) steps for validating `MLContext`, which will be added in a separate PR.
It is marked as an issue.

@anssiko @wchao1115 @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/309.html" title="Last updated on Dec 8, 2022, 4:14 PM UTC (7c606e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/309/8ac6b40...zolkis:7c606e2.html" title="Last updated on Dec 8, 2022, 4:14 PM UTC (7c606e2)">Diff</a>

PR #310, opened 2022-12-08
MlContext improvements

### LINKED ISSUE #308: Validate MLContext
We need steps to validate an `MLContext` object.
PR in the works.

zolkis:
Several small commits to improve the createContext() steps, and add the validate MLContext steps.
Fixes #308.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/310.html" title="Last updated on Dec 13, 2022, 5:35 PM UTC (fc8854c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/310/8ac6b40...zolkis:fc8854c.html" title="Last updated on Dec 13, 2022, 5:35 PM UTC (fc8854c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -583,6 +583,12 @@ interface ML {
 };
 </script>
 
+### Permissions Policy Integration ### {#permissions-policy-integration}
+
+This specification defines a <a>policy-controlled feature</a> identified by the
+string "<code><dfn data-lt="webnn-feature">webnn</dfn></code>".
+Its <a>default allowlist</a> is <code>'self'</code>.
+
 The {{ML/createContext()}} method steps are:
 1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, then throw a "{{SecurityError!!exception}}" {{DOMException}} and abort these steps.
 1. Let |promise| be [=a new promise=].
@@ -606,12 +612,6 @@ The {{ML/createContext()}} method steps are:
     1. Else:
         1. [=Resolve=] |promise| with a new {{NotSupportedError}}.
 1. Return |promise|.
-
-### Permissions Policy Integration ### {#permissions-policy-integration}
-
-This specification defines a <a>policy-controlled feature</a> identified by the
-string "<code><dfn data-lt="webnn-feature">webnn</dfn></code>".
-Its <a>default allowlist</a> is <code>'self'</code>.
 
 ## The MLContext interface ## {#api-mlcontext}
 The {{MLContext}} interface represents a global state of neural network compute workload and execution processes. Each {{MLContext}} object has associated [=context type=], [=device type=] and [=power preference=].
--- a/index.bs
+++ b/index.bs
@@ -605,15 +605,15 @@ The {{ML/createContext()}} method steps are:
         1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
         1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device-type-cpu|cpu=]".
         1. If |options|["{{powerPreference}}"] [=map/exists=], then set |context|.{{[[powerPreference]]}} to |options|["{{powerPreference}}"]. Otherwise, set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
-1. If the <a>validate MLContext</a> steps given |context| return `false`, [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
+1. If the <a>validate MLContext</a> steps given |context| return `false`, [=reject=] |promise| with a "{{NotSupportedError}}" and abort these steps.
 1. [=Resolve=] |promise| with |context|.
 
 ### The {{ML/createContextSync()}} method ### {#api-ml-createcontextsync}
 The {{ML/createContextSync()}} method steps are:
-1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a {{SecurityError}} and abort these steps.
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" and abort these steps.
 1. Let |options| be the first argument.
 1. Let |context| be the result of running the <a>create context</a> steps given |options|.
-1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a {{NotSupportedError}} and abort these steps.
+1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a "{{NotSupportedError}}" and abort these steps.
 1. Return |context|.
 
 ## The MLContext interface ## {#api-mlcontext}
--- a/index.bs
+++ b/index.bs
@@ -594,18 +594,27 @@ The {{ML/createContext()}} method steps are:
 1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" and abort these steps.
 1. Let |promise| be [=a new promise=].
 1. Return |promise| and run the following steps [=in parallel=].
-1. Let |context| be a new {{MLContext}} object.
 1. Let |options| be the first argument.
-1. If |options| is a {{GPUDevice}} object,
-    1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
-    1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
-    1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
-1. Otherwise,
-    1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
-    1. If |options|'s {{deviceType}} [=map/exists=], set |context|.{{[[deviceType]]}} to its value, otherwise to "[=device-type-cpu|cpu=]".
-    1. If |options|'s {{powerPreference}} [=map/exists=], set |context|.{{[[powerPreference]]}} to its value, otherwise to "[=power-preference-default|default=]".
+1. Run the <dfn>create context</dfn> steps given |options|:
+    1. Let |context| be a new {{MLContext}} object.
+    1. If |options| is a {{GPUDevice}} object,
+        1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
+        1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
+        1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+    1. Otherwise,
+        1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
+        1. If |options|'s {{deviceType}} [=map/exists=], set |context|.{{[[deviceType]]}} to its value, otherwise to "[=device-type-cpu|cpu=]".
+        1. If |options|'s {{powerPreference}} [=map/exists=], set |context|.{{[[powerPreference]]}} to its value, otherwise to "[=power-preference-default|default=]".
 1. If the <a>validate MLContext</a> steps given |context| return failure, [=reject=] |promise| with a new {{NotSupportedError}} and abort these steps.
 1. [=Resolve=] |promise| with |context|.
+
+### The {{ML/createContextSync()}} method ### {#api-ml-createcontextsync}
+The {{ML/createContextSync()}} method steps are:
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" and abort these steps.
+1. Let |options| be the first argument.
+1. Let |context| be the result of running the <a>create context</a> steps given |options|.
+1. If the <a>validate MLContext</a> steps given |context| return failure, throw a {{NotSupportedError}} and abort these steps.
+1. Return |context|.
 
 ## The MLContext interface ## {#api-mlcontext}
 The {{MLContext}} interface represents a global state of neural network compute workload and execution processes. Each {{MLContext}} object has associated [=context type=], [=device type=] and [=power preference=].
--- a/index.bs
+++ b/index.bs
@@ -589,29 +589,23 @@ This specification defines a <a>policy-controlled feature</a> identified by the
 string "<code><dfn data-lt="webnn-feature">webnn</dfn></code>".
 Its <a>default allowlist</a> is <code>'self'</code>.
 
+### The {{ML/createContext()}} method ### {#api-ml-createcontext}
 The {{ML/createContext()}} method steps are:
-1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, then throw a "{{SecurityError!!exception}}" {{DOMException}} and abort these steps.
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" and abort these steps.
 1. Let |promise| be [=a new promise=].
+1. Return |promise| and run the following steps [=in parallel=].
 1. Let |context| be a new {{MLContext}} object.
-1. Switch on the method's first argument:
-    <dl class=switch>
-    <dt>{{MLContextOptions}}
-    <dd>Set |context|.{{[[contextType]]}} to [=default-context|default=].
-    <dd>Set |context|.{{[[deviceType]]}} to the value of {{MLContextOptions}}'s {{deviceType}}.
-    <dd>Set |context|.{{[[powerPreference]]}} to the value of {{MLContextOptions}}'s {{powerPreference}}.
-
-    <dt>{{GPUDevice}}
-    <dd>Set |context|.{{[[contextType]]}} to [=webgpu-context|webgpu=].
-    <dd>Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
-    <dd>Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
-    </dl>
-1. Issue the following steps to a separate timeline:
-    1. If the User Agent can support the |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}, then:
-        1. Set |context|.{{MLContext/[[implementation]]}} to an implementation supporting |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}.
-        1. [=Resolve=] |promise| with |context|.
-    1. Else:
-        1. [=Resolve=] |promise| with a new {{NotSupportedError}}.
-1. Return |promise|.
+1. Let |options| be the first argument.
+1. If |options| is a {{GPUDevice}},
+    1. Set |context|.{{[[contextType]]}} to [=webgpu-context|webgpu=].
+    1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
+    1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+1. Otherwise,
+    1. Set |context|.{{[[contextType]]}} to [=default-context|default=].
+    1. Set |context|.{{[[deviceType]]}} to the value of |options|'s {{deviceType}} if defined, or to "[=device-type-cpu|cpu=]".
+    1. Set |context|.{{[[powerPreference]]}} to the value of |options|'s {{powerPreference}} if defined, or to "[=power-preference-default|default=]".
+1. If the user agent cannot support |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}, [=reject=] |promise| with a new {{NotSupportedError}} and abort these steps.
+1. [=Resolve=] |promise| with |context|.
 
 ## The MLContext interface ## {#api-mlcontext}
 The {{MLContext}} interface represents a global state of neural network compute workload and execution processes. Each {{MLContext}} object has associated [=context type=], [=device type=] and [=power preference=].
--- a/index.bs
+++ b/index.bs
@@ -591,7 +591,8 @@ Its <a>default allowlist</a> is <code>'self'</code>.
 
 ### The {{ML/createContext()}} method ### {#api-ml-createcontext}
 The {{ML/createContext()}} method steps are:
-1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" and abort these steps.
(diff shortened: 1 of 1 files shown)


zolkis:
@anssiko @huningxin @wchao1115 PTAL - thanks.

index.bs
```diff
@@ -583,36 +583,30 @@ interface ML {
...
+1. Return |promise| and run the following steps [=in parallel=].
+1. Let |context| be a new {{MLContext}} object.
+1. Let |options| be the first argument.
+1. If |options| is a {{GPUDevice}},
+    1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
+    1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
+    1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+1. Otherwise,
+    1. Set |context|.{{[[contextType]]}} to "[=default-context|default=].
+    1. Set |context|.{{[[deviceType]]}} to the value of |options|'s {{deviceType}} if defined, or to "[=device-type-cpu|cpu=]".
```
anssiko, line 605: [Dictionary](https://webidl.spec.whatwg.org/#idl-dictionary) options is an ordered map, so we should use "map[key] [exists](https://infra.spec.whatwg.org/#map-exists)" syntax for this and other similar checks.

You should also check that we use "exists" consistently in the spec. I spotted a few occurrences of "exist" that do not link to the dfn in the MLCommandEncoder interface. Spin a new PR for those.

anssiko (changes requested):
Thanks! One suggestion.

index.bs
```diff
@@ -583,36 +583,30 @@ interface ML {
...
 Its <a>default allowlist</a> is <code>'self'</code>.
 
+### The {{ML/createContext()}} method ### {#api-ml-createcontext}
+The {{ML/createContext()}} method steps are:
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" and abort these steps.
+1. Let |promise| be [=a new promise=].
+1. Return |promise| and run the following steps [=in parallel=].
+1. Let |context| be a new {{MLContext}} object.
+1. Let |options| be the first argument.
+1. If |options| is a {{GPUDevice}},
```
zolkis, line 599: @anssiko could you check on this, too? are we allowed to say "is a <object-defined-here>", or just "is an object" and specify prose? I suspect the latter :)

index.bs
anssiko: I keep seeing the following pattern being used with dictionaries:

    1. If options["deviceType"] exists, then
      1. Set context[[deviceType]] to options["deviceType"]. 
    1. Otherwise, set context[[deviceType]] to "cpu".

See e.g. https://www.w3.org/TR/generic-sensor/#abstract-operations

index.bs
anssiko: I think this would be the most explicit:

`1. If options is a GPUDevice object, then`

zolkis: In that case we don't need the `object`, and other specs are not using it, either. 
We can leave it as is for now, and make a sweep-check later on this topic.

anssiko: “is a Foo object” makes the type explicit without the need to click the link.

See also https://www.w3.org/TR/webgpu/#dom-gpuadapter-requestdevice

index.bs
zolkis: Equivalent to the current text, 

> 1. Set |context|.{{[[powerPreference]]}} to the value of |options|'s {{powerPreference}} if it [=map/exists=], or to "[=power-preference-default|default=]".

But I can follow the other style at the expense of verbosity, no problem.

index.bs
zolkis: I see both patterns in different specs (infra, url, html, etc)
 - if an object  or variable is defined in the spec or algo, just use it by name
 - generic object, list, promise etc are referred as "a list", "an object" etc.
or,
- use the type next to the name.
If you prefer having the type mentioned, we can follow that convention. I used to prefer brevity when clarity is not affected.

index.bs
zolkis: BTW nowadays we are not supposed to use `options["deviceType"]`, we should say `options`'  `deviceType` (link), or 
"if `options` includes `deviceType`". Same applies to |context| etc. 

Need to check more specifications.

zolkis: BTW nowadays we are not supposed to use `options["deviceType"]`, we should say `options`'  `deviceType` (link), or 
"if `options` includes `deviceType`". Same applies to |context| etc. 

Need to check more specifications.

zolkis:
Note: for some reason the last commit (createContextSync) didn't appear in the PR, even though it was present on my branch. github didn't find that commit when clicking on its link, giving a 404 - so something went wrong.
Needed to amend (with same message), and force-push: that worked.

wchao1115:
I'm sorry @zolkis what problem we're trying to address here? The issue #308 doesn't have any detail on it. 

I also forgot about the agreement around `createContextSync`. Is that what we previously agreed to? Adding @huningxin as a reviewer also.

huningxin:
> I also forgot about the agreement around `createContextSync`. Is that what we previously agreed to?

The sync version of create context method was there since it was introduced by https://github.com/webmachinelearning/webnn/pull/149 before. Later https://github.com/webmachinelearning/webnn/pull/274 introduced the async create context method and the sync version was restricted to be exposed to dedicated worker. According to the TAG's feedback, the sync version was appended with the "Sync" postfix as `createContextSync()`.

zolkis:
> I'm sorry @zolkis what problem we're trying to address here? The issue #308 doesn't have any detail on it. 

Thank you for taking a look (looks like I can only mention people in PRs, and could not assign reviewers myself).
The need to validate `MLContext` came up in #309 (which needs to be updated when/if this PR is merged).
In addition, it is used in `createContext()` and its sync version. In principle, we need it in all methods where `context` is an argument. 

These are still just editorial changes, since the steps leave the details to implementations completely. To be checked later if that needs improvement - until then, this is just a formal placeholder, signifying the place implementations should do a check.

zolkis:
Accidentally removed @huningxin from reviewers, but can't add back.
@anssiko have your concerns been addressed?

index.bs
anssiko: Thanks @zolkis for the research, in my experience conventions in the infra spec are safe to use. Given that, I think you want to change this line to something like:

```
1. If options["deviceType"] exists, then
  1. Set context[[deviceType]] to options["deviceType"]. 
1. Otherwise, set context[[deviceType]] to "cpu".
```

And similarly on the following line for powerPreference.

index.bs
anssiko: I think including the type too would be better.

I usually check the mothership HTML LS for conventions because it is well maintained. It looks like "is a Foo object" is the convention in algorithms:

https://html.spec.whatwg.org/#media-element-load-algorithm
https://html.spec.whatwg.org/#convert-to-a-list-of-name-value-pairs
https://html.spec.whatwg.org/#font-style-source-object
https://html.spec.whatwg.org/#dom-path2d
https://html.spec.whatwg.org/#check-the-usability-of-the-image-argument
https://html.spec.whatwg.org/#dom-context-2d-fillstyle
https://html.spec.whatwg.org/#dom-context-2d-strokestyle

Also, I had this type of hypothetical test in mind:
```
async function getGPUDevice() {
  const adapter = await navigator.gpu.requestAdapter();
  if (!adapter) { return false; }
  const device = await adapter.requestDevice();
  return device;
}

function isGPUDevice(device) {
  return device instanceof GPUDevice;
}

function isObject(device) {
  return device instanceof Object;
}

const device = await getGPUDevice();
console.log(isGPUDevice(device));
console.log(isObject(device));
```

Both the assertions should return true, thus "is a GPUDevice object" prose more explicitly maps to these two test assertions.

anssiko:
>Accidentally removed @huningxin from reviewers, but can't add back.
@anssiko have your concerns been addressed?

Ping me here when you've addressed the review comments and I'll add the reviewers back.

index.bs
zolkis: Yes, the current text is 

> If |options| is a {{GPUDevice}} object

anssiko: LGTM

index.bs
zolkis: It's more fluffy and repeating stuff than saying, 

> If |options|'s {{powerPreference}} [=map/exists=], set |context|.{{[[powerPreference]]}} to its value, otherwise to "[=power-preference-default|default=]".

but let it be as you wish.

anssiko: The one line version is better in this context. To align with the convention, you should replace "its value" with "options["deviceType"]":

`1. If options["deviceType"] exists, then set context[[deviceType]] to options["deviceType"]. Otherwise, set context[[deviceType]] to "cpu".`

zolkis: I just checked a couple of other specs, and the convention they use is 2 steps, one for "if" and one for "otherwise", just as the first suggestion in my [previous](https://github.com/webmachinelearning/webnn/pull/310#discussion_r1045580031) comment.

zolkis: Not wishing to over-stretch the issue, but avoiding unnecessary repetitions would make the text more clear IMHO.
What about using

```1. If options["deviceType"] exists, then set context[[deviceType]] to its value. Otherwise, set context[[deviceType]] to "cpu".```

Comment (anssiko) [reply index.bs]: I do also think that reads more like English, but if we want to use the infra spec conventions consistently we should replace `its value` with `options["deviceType"]`.

Note that map get is with quote marks and map set without:

https://infra.spec.whatwg.org/#map-get 
https://infra.spec.whatwg.org/#map-set

It is not over-stretching. It is important to agree on these conventions so we can then apply the conventions in follow ups consistently. My recommendation would be to align with the infra spec conventions whenever available.

Comment (zolkis) [reply index.bs]: No problem, it's already written that way (assumed we will stick to the more verbose standard pattern).

**Review by anssiko — APPROVED** [review]
LGTM. This editorial change looks good to me. Thanks for all the iteration to establish the conventions :-)

**Comment by anssiko** [comment]
@huningxin @wchao1115 this editorial update is now ready for your review.

You may find the visual PR Preview and Diff helpful:

<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/310.html" title="Last updated on Dec 12, 2022, 11:52 AM UTC (eed627b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/310/8ac6b40...zolkis:eed627b.html" title="Last updated on Dec 12, 2022, 11:52 AM UTC (eed627b)">Diff</a>

File: index.bs [hunk 3]
```diff
@@ -661,15 +664,20 @@ interface MLContext {};
...
-    : <dfn>\[[implementation]]</dfn>
-    ::
-        The underlying implementation provided by the User Agent.
 </dl>
 
 <div class="note">
 When the {{[[contextType]]}} is set to [=default-context|default=] with the {{MLContextOptions}}.{{deviceType}} set to [=device-type-gpu|gpu=], the user agent is responsible for creating an internal GPU device that operates within the context and is capable of ML workload submission on behalf of the calling application. In this setting however, only {{ArrayBufferView}} inputs and outputs are allowed in and out of the graph execution since the application has no way to know what type of internal GPU device is being created on their behalf. In this case, the user agent is responsible for automatic uploads and downloads of the inputs and outputs to and from the GPU memory using this said internal device.
 </div>
 
+### The validate {{MLContext}} algorithm ### {#api-mlcontext-validate}
```
wchao1115, line 673: Should the text be read *"The {{MLContext}} validation algorithm"* ?

wchao1115 (changes requested):

index.bs
```diff
@@ -661,15 +664,20 @@ interface MLContext {};
...
+To <dfn lt="validate MLContext">validate {{MLContext}}</dfn>, given |context|, run these steps:
+1. If |context|.{{[[contextType]]}} is not "[=webgpu-context|webgpu=]" or "[=default-context|default=], return failure.
```
huningxin, line 675: For the validation failure, should it return `false`? For validation success, it returns `true`.

WebGPU spec uses "return false" for validation steps, such as [validating GPUBufferDescriptor](https://www.w3.org/TR/webgpu/#abstract-opdef-validating-gpubufferdescriptor).

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
-
 ### Permissions Policy Integration ### {#permissions-policy-integration}
 
 This specification defines a <a>policy-controlled feature</a> identified by the
 string "<code><dfn data-lt="webnn-feature">webnn</dfn></code>".
 Its <a>default allowlist</a> is <code>'self'</code>.
 
+### The {{ML/createContext()}} method ### {#api-ml-createcontext}
+The {{ML/createContext()}} method steps are:
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" and abort these steps.
```
huningxin, line 594: Should it be {{SecurityError}} (without "")? It would be better to keep consistent with other errors, like {{NotSupportedError}} below.

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
+    1. Let |context| be a new {{MLContext}} object.
+    1. If |options| is a {{GPUDevice}} object,
+        1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
+        1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
+        1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+    1. Otherwise,
+        1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
+        1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device-type-cpu|cpu=]".
+        1. If |options|["{{powerPreference}}"] [=map/exists=], then set |context|.{{[[powerPreference]]}} to |options|["{{powerPreference}}"]. Otherwise, set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+1. If the <a>validate MLContext</a> steps given |context| return failure, [=reject=] |promise| with a new {{NotSupportedError}} and abort these steps.
```
huningxin, line 608: Is the "new" needed for {{NotSupportedError}}? At line 616, there is no "new"  word in the context of throw a {{NotSupportedError}}. It would be better to keep consistent.

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
+1. [=Resolve=] |promise| with |context|.
+
+### The {{ML/createContextSync()}} method ### {#api-ml-createcontextsync}
+The {{ML/createContextSync()}} method steps are:
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" and abort these steps.
+1. Let |options| be the first argument.
+1. Let |context| be the result of running the <a>create context</a> steps given |options|.
+1. If the <a>validate MLContext</a> steps given |context| return failure, throw a {{NotSupportedError}} and abort these steps.
```
huningxin, line 616: If we adopt returning false for validation failure, checking the return value of validating MLContext steps needs to be updated accordingly.

index.bs
huningxin: ditto, if we adopt returning false.

index.bs
zolkis: Yes, sounds better.

index.bs
zolkis: Yes,  we can return `false`, and then check for "if returns false", instead of "if returns failure".

index.bs
zolkis: Right, thanks for spotting.

index.bs
zolkis: I will check on this one, IIRC some specs are using it this way. If no strong rule found, will use without "new".

zolkis: Using without "new".

zolkis:
Thank you for the review/comments, they have been addressed, and it's better. :)

index.bs
anssiko: When using the `DOMException` [error names](https://webidl.spec.whatwg.org/#idl-DOMException-error-names), the convention is `ErrorNameHere DOMException`:

For example on this line: "... with a "[`NotSupportedError`](https://webidl.spec.whatwg.org/#notsupportederror)" [`DOMException`](https://webidl.spec.whatwg.org/#dfn-DOMException) and about these steps"

index.bs
anssiko, line 594: "[SecurityError](https://webidl.spec.whatwg.org/#securityerror)" [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException)

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
+    1. Let |context| be a new {{MLContext}} object.
+    1. If |options| is a {{GPUDevice}} object,
+        1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
+        1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
+        1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+    1. Otherwise,
+        1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
+        1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device-type-cpu|cpu=]".
+        1. If |options|["{{powerPreference}}"] [=map/exists=], then set |context|.{{[[powerPreference]]}} to |options|["{{powerPreference}}"]. Otherwise, set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+1. If the <a>validate MLContext</a> steps given |context| return `false`, [=reject=] |promise| with a {{NotSupportedError}} and abort these steps.
```
anssiko, line 608: "[NotSupportedError](https://webidl.spec.whatwg.org/#notsupportederror)" [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException)

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
+1. [=Resolve=] |promise| with |context|.
+
+### The {{ML/createContextSync()}} method ### {#api-ml-createcontextsync}
+The {{ML/createContextSync()}} method steps are:
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a {{SecurityError}} and abort these steps.
+1. Let |options| be the first argument.
+1. Let |context| be the result of running the <a>create context</a> steps given |options|.
+1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a {{NotSupportedError}} and abort these steps.
```
anssiko, line 616: "[NotSupportedError](https://webidl.spec.whatwg.org/#notsupportederror)" [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException)

anssiko:
@zolkis we want to add the missing [`DOMException`](https://webidl.spec.whatwg.org/#dfn-DOMException) to a few places. If you search the spec with "DOMException" you'll find a few occurrences outside the scope of this PR (7.3 and 7.9) where we need to add quote marks to the error names. You can do that in a separate PR, open an issue for that editorial tweak.

index.bs
zolkis: Another convention round.
SecurityError is a DOMException by [definition](https://webidl.spec.whatwg.org/#idl-DOMException-error-names).
So, using the linked form 
```
{{SecurityError}}
```
without the quotes and without DOMException it would unambiguously refer to a "SecurityError" DOMException :)  

However, even if not the shortest, we should stick to the [Web IDL convention](https://webidl.spec.whatwg.org/#idl-exceptions):

> [Simple exceptions](https://webidl.spec.whatwg.org/#dfn-simple-exception) can be created by providing their [error name](https://webidl.spec.whatwg.org/#dfn-exception-error-name). A [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) can be [created](https://webidl.spec.whatwg.org/#dfn-create-exception) by providing its [error name](https://webidl.spec.whatwg.org/#dfn-exception-error-name) followed by [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException). Exceptions can also be thrown, by providing the same details required to [create](https://webidl.spec.whatwg.org/#dfn-create-exception) one.

These rules are not used uniformly throughout the spec. 
The question is whether to fix that in this PR, or should it be a separate one, plus document the reasoning in an issue that links this discussion?

zolkis:
I created #314 to document the rule, but I can fix it in this PR as it's quite short, and overlaps with this PR. 

In the end, the separate commits should be visible, i.e. the commit fixing the issue should not be squashed.

zolkis:
The change is done, should I commit it here, or stash it and use it in another PR after this one is merged (as it is for now)?

anssiko:
I suggest commit changes to the lines touched by this PR and leave the full spec error conventions update for another PR.

zolkis:
Ah, the most complex way. It's too late to do that cleanly, I will have a merge commit on the new PR. But it's done now.

index.bs
anssiko, line 594:
```suggestion
1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" {{DOMException}} and abort these steps.

```

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
+    1. Let |context| be a new {{MLContext}} object.
+    1. If |options| is a {{GPUDevice}} object,
+        1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
+        1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
+        1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+    1. Otherwise,
+        1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
+        1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device-type-cpu|cpu=]".
+        1. If |options|["{{powerPreference}}"] [=map/exists=], then set |context|.{{[[powerPreference]]}} to |options|["{{powerPreference}}"]. Otherwise, set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+1. If the <a>validate MLContext</a> steps given |context| return `false`, [=reject=] |promise| with a "{{NotSupportedError}}" and abort these steps.
```
anssiko, line 608:
```suggestion
1. If the <a>validate MLContext</a> steps given |context| return `false`, [=reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}} and abort these steps.
```

index.bs
```diff
@@ -583,36 +583,39 @@ interface ML {
...
+1. [=Resolve=] |promise| with |context|.
+
+### The {{ML/createContextSync()}} method ### {#api-ml-createcontextsync}
+The {{ML/createContextSync()}} method steps are:
+1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" and abort these steps.
+1. Let |options| be the first argument.
+1. Let |context| be the result of running the <a>create context</a> steps given |options|.
+1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a "{{NotSupportedError}}" and abort these steps.
```
anssiko, line 616:
```suggestion
1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a "{{NotSupportedError}}" {{DOMException}} and abort these steps.
```

anssiko (changes requested):
We're almost done here :-) My proposal is to fix the error usage in `createContext()` and `createContextSync()` in this PR so we can land this.

It looks like other review feedback has been addressed, so I'll try to commit the remaining fixes to this branch and merge this PR so we can save precious Editors' review time for the upcoming PRs.

Thanks for your review @huningxin @wchao1115 and @zolkis for crafting this PR!

anssiko (approved):

zolkis:
Thanks!

zolkis:
In some other projects, before merging, the smaller related commits are squashed together, to result in more meaningful history. However, we can also keep full commit history.

PR #321, opened 2022-12-23
Add LSTM to the operator list

wchao1115:
Finally had a chance to do this. Adding `lstm` and `lstmCell` as additional operations in the spec. Since the addition of `gru` and `gruCell` there have been numerous questions about LSTM. As a widely popular recurrent network, supporting LSTM along with GRU is generally seen as important for API completeness. 

This change also made some editorial cleanups of existing wordings for GRU, and proposes a name change of the existing `MLOperator` to `MLActivation` to reflect its true purpose in the spec as a generic definition of an activation function i.e. not every operation can be an activation function, etc.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/321.html" title="Last updated on Jan 9, 2023, 9:34 PM UTC (8545e28)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/321/5a4e73e...8545e28.html" title="Last updated on Jan 9, 2023, 9:34 PM UTC (8545e28)">Diff</a>

Issue #142, opened 2021-02-04
Isomorphic JS story, worker scope exposure?


anssiko:
> 12. What's the isomorphic JS story for this? Also, given that this is attached to vanilla navigator, is this not expected to work in a worker scope?

via https://github.com/w3ctag/design-reviews/issues/570#issuecomment-768875996

RafaelCintron: 12 is a good issue, should not tie to navigator, good for worker scope as well
… as for the isomorphic part, the degree to which you can use this API, depends on how much Node exposes e.g. WebGL in node

cynthia:
> depends on how much Node exposes e.g. WebGL in node

Alternatively, FFI bindings to an existing framework that already has a compute backend (e.g. CUDA) which is abstracted to expose the same API sounds totally feasible.

huningxin:
Agree, it's possible. And the [webnn-polyfill](https://github.com/webmachinelearning/webnn-polyfill) CPU backend (depends on TF.js CPU backend) already works in Node.js for testing. Further, I believe a node.js binding for the native implementation of webnn ([webnn-native](https://github.com/webmachinelearning/webnn-native), code to be submitted) should also work. However, as the [node_setup.js](https://github.com/webmachinelearning/webnn-polyfill/blob/master/node_setup.js) code shows, we emulate a navigator object in node.js and get the webnn-polyfill and tests run. Probably, we need to bind [`ML`](https://webmachinelearning.github.io/webnn/#api-ml) object according to different JS execution environments. Any suggestions @cynthia ?

anssiko:
Relevant to this issue, [Deno 1.8](https://deno.land/posts/v1.8) adds experimental support for WebGPU API. The project’s goal is to run Tensorflow.js on Deno, with GPU acceleration.

anssiko:
>RESOLUTION: Expose ML interface to Worker scope similarly to WebGPU API

via https://www.w3.org/2021/04/15-webmachinelearning-minutes.html#r02

anssiko:
Noting here #163 makes feasible introducing sync variants of `compile()` and `compute()` in the worker context without blocking the main thread.

@huningxin had some use cases in mind. I think this is worth exploring in a separate issue.

PR #322, opened 2022-12-26
Simplify MLContext creation

wchao1115:
Given the explicit support for the WebGPU `gpuDevice` as a dedicated ML context type in addition to the default type, and the recent removal of the support for the WebGLContext device type, the `MLContext` creation can be further simplified by removing the internal GPU device type that is not explicitly specified as a `gpuDevice` parameter to the ML context creation method i.e. a GPU context can now be created only from a WebGPU device. 

The second change is the removal of the `high-performance` power preference. This option was originally defined to facilitate the creation of a GPU context that is focused on high performance, low latency execution of a discrete GPU (presumably as a tiebreaker in a configuration where both an integrated GPU and a dedicated discrete GPU are present). Following the rationale behind the first change to only support a GPU context thru a WebGPU device, this option is no longer necessary since the specified `gpuDevice` must be created from an adapter that must first be explicitly selected by the user of WebGPU i.e. it is now up to the user to choose which GPU adapter (e.g. integrated vs. discrete) that they want to use as a GPU device backing the `MLContext` object.

Lastly, please note that with this change, the *default context* is now an implementation choice. As long as the implementation takes reasonable consideration to the user-specified power preference option, it can implement the default context type any way they want to as long as they aren't a GPU implementation. For instance, the implementation can be entirely CPU-based or as a hybrid CPU-NPU implementation, as long as it fully supports `ArrayBufferView` as in the API contract, it's all good from the spec point of view. I've taken to remove explicit mentions of CPU as a normative requirement in the relevant wordings to make it sound more like an informative consideration.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/322.html" title="Last updated on Jan 13, 2023, 11:11 PM UTC (fde3783)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/322/d87f5d2...fde3783.html" title="Last updated on Jan 13, 2023, 11:11 PM UTC (fde3783)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -516,19 +516,19 @@ a graph execution method that provides the maximum flexibility to callers that a
 application. It does this by placing the workload required to initialize and compute the results of the
 operations in the graph onto a {{GPUCommandBuffer}}. The callers are responsible for the eventual submission
 of this workload on the {{GPUQueue}} through the WebGPU queue submission mechanism. Once the submitted workload
-is completely executed, the result is avaialble in the bound output buffers.
+is completely executed, the result is available in the bound output buffers.
 
 ## Context and Device Association ## {#programming-model-context-device-association}
 
 An {{MLContext}} interface represents a state of neural network execution. An important function of this state is manage resources used in the compilation 
-and execution of the neural network graph. An implementation in a user agent may implement this state in term of CPU resources when an {{MLContext}} is created 
+and execution of the neural network graph. An implementation in a user agent may implement this state in terms of CPU resources when an {{MLContext}} is created 
 without an explicit association with a hardware device (aka. a *"default context"*). However, when an {{MLContext}} is created from a specific {{GPUDevice}} that 
 is also used by the application of WebGPU (aka. a *"GPU context"*), any GPU resource created from the same {{GPUDevice}} such as the {{GPUBuffer}} or {{GPUTexture}} 
 is considered a native resource that can also be used as a graph constant, input, or output operand.
 
 In a situation when a GPU context executes a graph with a constant or an input or output allocated in the system memory as in an {{ArrayBufferView}}, the content 
 is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of 
-the graph execution. This automatic data upload and download cycles will only occur whenever the executing GPU context determines that the data be copied out of or back 
+the graph execution. This automatic data upload and download cycles will only occur whenever the executing GPU context determines that the data must be copied out of or back 
 into the system memory as part of the execution. Additionally, the eventual result of the execution must also be in a known layout format. While the internal execution 
 technique may be optimized for native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert 
 the content back to a known layout format at the end of the graph in order to maintain interoperability as expected by the caller.
--- a/index.bs
+++ b/index.bs
@@ -520,15 +520,16 @@ is completely executed, the result is available in the bound output buffers.
 
 ## Context and Device Association ## {#programming-model-context-device-association}
 
-An {{MLContext}} interface represents a state of neural network execution. An important function of this state is manage resources used in the compilation 
-and execution of the neural network graph. An implementation in a user agent may implement this state in terms of CPU resources when an {{MLContext}} is created 
-without an explicit association with a hardware device (aka. a *"default context"*). However, when an {{MLContext}} is created from a specific {{GPUDevice}} that 
-is also used by the application of WebGPU (aka. a *"GPU context"*), any GPU resource created from the same {{GPUDevice}} such as the {{GPUBuffer}} or {{GPUTexture}} 
-is considered a native resource that can also be used as a graph constant, input, or output operand.
+An {{MLContext}} interface represents a state of a neural network execution. An important function of this state is to manage resources used in the compilation 
+and execution of the neural network graph. An implementation in a user agent may implement this state in terms of CPU resources and execution when an {{MLContext}} 
+is created without an explicit association with a hardware device (aka. a *"default context"*). However, when an {{MLContext}} is created from a WebGPU {{GPUDevice}} 
+(aka. a *"GPU context"*), the implementation uses the specified GPU device as a resource domain for the subsequent compilation and execution of the graph. 
+Any GPU resource such as the {{GPUBuffer}} or {{GPUTexture}} created from the same {{GPUDevice}} is therefore considered a resource of native resource type that 
+can be used to store a graph constant, input, or output operand.
 
 In a situation when a GPU context executes a graph with a constant or an input or output allocated in the system memory as in an {{ArrayBufferView}}, the content 
 is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of 
-the graph execution. This automatic data upload and download cycles will only occur whenever the executing GPU context determines that the data must be copied out of or back 
+the graph execution. These automatic data upload and download cycles will only occur whenever the executing GPU context determines that the data must be copied out of or back 
 into the system memory as part of the execution. Additionally, the eventual result of the execution must also be in a known layout format. While the internal execution 
 technique may be optimized for native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert 
 the content back to a known layout format at the end of the graph in order to maintain interoperability as expected by the caller.
--- a/index.bs
+++ b/index.bs
@@ -663,7 +663,7 @@ To <dfn lt="validate MLContext">validate {{MLContext}}</dfn>, given |context|, r
 1. Return `true`;
 
 ### Synchronous Execution ### {#api-mlcontext-sync-execution}
-Synchronously carries out the computational workload of a compiled graph {{MLGraph}} on the calling thread, which must be a worker thread, to produce results as defined by the operations in the graph. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
+Synchronously carries out the computational workload of a compiled graph {{MLGraph}} on the calling thread, which must be a worker thread, to produce results as defined by the operations in the graph.
 
 <script type=idl>
 partial interface MLContext {
@@ -754,7 +754,7 @@ console.log(&#96;values: ${bufferE}&#96;);
 </div>
 
 ### Asynchronous Execution ### {#api-mlcontext-async-execution}
-Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a CPU worker thread or on a GPU timeline submitting a GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
+Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a CPU worker thread or on a GPU timeline submitting a GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing.
 
 <script type=idl>
 partial interface MLContext {
--- a/index.bs
+++ b/index.bs
@@ -401,7 +401,7 @@ As a future-proofing measure, the API design allows certain operations that can 
 
 Issue: Investigate side channel attack feasibility considering the current state where CPU is shared between processes running renderers.
 
-In order to not allow an attacker to target a specific implementation that may contain a flaw, the [[#programming-model-device-selection]] mechanism is a hint only, and the concrete device selection is left to the implementation - a user agent could for instance choose never to run a model on a device with known vulnerabilities. As a further mitigation, no device enumeration mechanism is defined.
+In order to not allow an attacker to target a specific implementation that may contain a flaw, the [[#programming-model-context-device-association]] mechanism is a hint only, and the concrete device selection is left to the implementation - a user agent could for instance choose never to run a model on a device with known vulnerabilities. As a further mitigation, no device enumeration mechanism is defined.
 
 Issue: Hinting partially mitigates the concern. Investigate additional mitigations.
 
@@ -442,7 +442,7 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring; 
 
 The WebGPU API identifies <a href="https://gpuweb.github.io/gpuweb/#privacy-machine-artifacts">machine-specific artifacts</a> as a privacy consideration. Given the WebNN API defines means to record an ML workload onto a WebGPU-compatible {{GPUCommandBuffer}}, compute unit scheduling may under certain circumstances introduce a fingerprint. However, similarly to WebGPU, such fingerprints are identical across most or all of the devices of each vendor, mitigating the concern. Furthermore, software implementations can be used to further eliminate such artifacts.
 
-The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Device type=] normatively indicates the kind of device and is either "cpu" or "gpu". If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
+The WebNN API defines developer-settable preferences to help inform [[#programming-model-context-device-association]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
 
 If a future version of this specification introduces support for new a  [=device type=] that can only support a subset of {{MLOperandType}}s, that may introduce a new fingerprint.
 
@@ -491,26 +491,27 @@ that shares the same buffer as the input tensor. (In the case of reshape or sque
 the entire data is shared, while in the case of slice, a part of the input data is shared.)
 The implementation may use views, as above, for intermediate values.
 
-Before the execution, the computation graph that is used to compute one or more specified outputs needs to be compiled and optimized. The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion.
+Before the execution, the computation graph that is used to compute one or more specified outputs needs to be compiled and optimized. 
+The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion.
 
-There are multiple ways by which the graph may be compiled. The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. The {{MLGraphBuilder}}.{{MLGraphBuilder/buildSync()}} method compiles the graph immediately on the calling thread, which must be a worker thread running on CPU or GPU device, and returns an {{MLGraph}}. Both compilation methods produce an {{MLGraph}} that represents a compiled graph for optimal execution.
+There are multiple ways by which the graph may be compiled. The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph 
+in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. The 
+{{MLGraphBuilder}}.{{MLGraphBuilder/buildSync()}} method compiles the graph immediately on the calling thread, which must be a worker 
+thread and returns an {{MLGraph}}. Both compilation methods produce an {{MLGraph}} that represents a compiled graph for optimal execution.
 
 Once the {{MLGraph}} is constructed, there are multiple ways by which the graph may be executed. The
 {{MLContext}}.{{MLContext/computeSync()}} method represents a way the execution of the graph is carried out immediately
-on the calling thread, which must also be a worker thread, either on a CPU or GPU device. The execution
-produces the results of the computation from all the inputs bound to the graph.
+on the calling thread, which must also be a worker thread. The execution produces the results of the computation from all the inputs bound to the graph.
 
-The {{MLContext}}.{{MLContext/compute()}} method represents a way the execution of the graph is performed asynchronously
-either on a parallel timeline in a separate worker thread for the CPU execution or on a GPU timeline in a GPU
-command queue. This method returns immediately without blocking the calling thread while the actual execution is
-offloaded to a different timeline. This type of execution is appropriate when the responsiveness of the calling
-thread is critical to good user experience. The computation results will be placed at the bound outputs at the
-time the operation is successfully completed on the offloaded timeline at which time the calling thread is
-signaled. This type of execution supports both the CPU and GPU device.
+The {{MLContext}}.{{MLContext/compute()}} method represents a way the execution of the graph is performed asynchronously either on a parallel 
+timeline in a CPU worker thread or on a GPU timeline executing a GPU command queue. This method returns immediately without blocking the calling 
+thread while the actual execution is offloaded to a different timeline. This type of execution is appropriate when the responsiveness of the calling 
+thread is critical to good user experience. The computation results will be placed at the bound outputs at the time the operation is successfully 
+completed on the offloaded timeline at which time the calling thread is signaled.
 
 In both the {{MLContext}}.{{MLContext/compute()}} and {{MLContext}}.{{MLContext/computeSync()}} execution methods, the caller supplies
-the input values using {{MLNamedArrayBufferViews}}, binding the input {{MLOperand}}s to their values. The caller
-then supplies pre-allocated buffers for output {{MLOperand}}s using {{MLNamedArrayBufferViews}}.
+the input values using {{MLNamedArrayBufferViews}}, binding the input {{MLOperand}}s to their values. The caller then supplies pre-allocated 
+buffers for output {{MLOperand}}s using {{MLNamedArrayBufferViews}}.
 
 The {{MLCommandEncoder}} interface created by the {{MLContext}}.{{MLContext/createCommandEncoder()}} method supports
 a graph execution method that provides the maximum flexibility to callers that also utilize WebGPU in their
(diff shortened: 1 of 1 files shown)


huningxin:
> i.e. a GPU context can now be created only from a WebGPU device.

I concern that this change would prevent an non-browser implementation, such as [WebNN-native Node.js binding](https://github.com/webmachinelearning/webnn-native/tree/main/node), from supporting GPU device.

wchao1115:
Thanks for the comment @huningxin. As noted in the last paragraph, as noted in the last paragraph, non-browser implementation only supports default context by definition, and therefore it can implement the interface as it sees fit, including as a GPU implementation. I can expand on that a bit more in the wording if that helps. Removing GPU support from non-WebGPU device could really simplify the browser-based implementation and makes it clear how WebNN interop with WebGPU.

huningxin:
Thanks for the explanation, @wchao1115 .

The non-browser implementation is one aspect. Actually, I am trying to understand the impact to the default context developers.

Before this change, to utilize a GPU device, the developers can just set the device type to "gpu" when creating the default context. The "gpu" and "cpu" default contexts can share the same code for graph building and compute (e.g., `MLContext.compute()`).

However, after this change, to utilize a GPU device, the developers would need to use WebGPU API to obtain a `GPUDevice` and create a webgpu context with it. Although the graph building code may be reused with the default context, the graph compute code would be different, because the `compute()` method is not allowed in a webgpu context (the spec says this method throws "OperationError" DOMException if not a default context). In particular, developers need to create `MLCommandEncoder` and use its methods including `initializeGraph()`, `dispatch()` and `finish()` to compute a graph. 

I feel this change would bring some extra overhead (WebGPU interop) to the default context developers for GPU access and reduce the code reusability across different types of devices for some use cases.

wchao1115:
With this change the only requirement of the default context is that it needs to support both the GPU resources and the array buffer view as constants, inputs, and outputs. As long as both types of resources are supported, the implementation can be either on the CPU or GPU side. 

That said I'm not familiar with the importance of supporting non-browser implementation out of the same standard web API spec. Is that a requirement for WebNN? Do we have existing customers that require that webnn-native be supported as a one of the standard implementations? My view is that we can still maintain webnn-native that supports both CPU and GPU with an extension e.g. retaining the device preference enum in the webnn-native API but outside of the WebNN spec, etc. 

As for the OperationError failure cases for `compute` I think it's overkill. I don't recall why we need that. We should remove those restrictions.

huningxin:
> With this change the only requirement of the default context is that it needs to support the array buffer view as constants, inputs, and outputs. As long as CPU resources are supported, the implementation can be either on the CPU or GPU side (i.e. with automatic upload/download for the GPU execution)

It's great that the default context implementation can still support GPU. However, with the removal of the device type and high-performance power preference, the default context developers lose the capability to request a GPU implementation.

> That said I'm not familiar with the importance of supporting non-browser implementation out of the same standard web API spec. Is that a requirement for WebNN?

It is tracked by https://github.com/webmachinelearning/webnn/issues/142. I am not aware any existing customers of webnn-native node.js binding. I agree the non-browser implementation is complementary. My focus is the default context usage in web browsers.

> As for the OperationError failure cases for `compute` I think it's overkill. I don't recall why we need that. We should remove those restrictions.

This was introduced by https://github.com/webmachinelearning/webnn/pull/257. 

I think this makes sense because it makes a clear separation by default and webgpu contexts for the two major usages that are WebNN-standalone (default) and WebNN/WebGPU interop. 

As I summarized in above PR [review](https://github.com/webmachinelearning/webnn/pull/257#issuecomment-1080120454) , the difference can be captured by the following table (revised by dropping the WebGL interop).

| Context Type| Creation Method | Device Type | Graph Execution Method | Resource Type |
|----|----|----|----|----|
| default | Create from an `MLContextOptions` | CPU/GPU/NPU, <br> device selected by device and power preferences | `compute()` (async call for both main thread and worker) <br> `computeSync()` (sync call for worker only) | CPU resources: `ArrayBufferView` |
|  webgpu  | Create from <br> a `GPUDevice` | GPU, <br> device selected by WebGPU API | `MLCommandEncoder` records an `MLGraph` into a `GPUCommandBuffer` that can be submitted to a `GPUQueue` | GPU resources: `GPUBuffer` and `GPUTexture` |

wchao1115:
As I proposed in my earlier comment, are you open to leaving the device preference only for webnn-native and not officially on the WebNN spec? That way the webnn-native continues to function as is today while we simplify the contract for WebNN to benefit the browser implementation?

huningxin:
Thanks @wchao1115 , webnn-native should follow the spec as much as possible. It would be fine.

My concern is more about the impact to the default context developers, they are also targeting web browsers. For example, the TFLite WebNN delegate uses the default context. It allows developers to access GPU device by just [setting device type](https://github.com/huningxin/tensorflow/blob/webnn_delegate_side_module/tensorflow/lite/delegates/webnn/webnn_delegate.h#L31) and without any other code modifications. With this change, it has to create a webgpu context and introduce a different graph compute path with WebGPU interop.

wchao1115:
Thanks. I understand your concern from the framework developers' point of view. However, that use case in practice will encourage multiple devices created in the same process if the app also uses WebGPU, with one internally owned by the default context and the other explicitly by the app thru a WebGPU device. Having multiple GPU devices to deal with creates huge complication around performance, and a lot of complications around resource sharing especially if the system has multiple adapters configured (which is a lot more common than most people think, with both Intel integrated and other discrete adapter plugged in). 

It seems pretty clear that people wants WebGPU to be the way webapp deals with GPU going forward. We even dropped support for WebGL to make it more compelling to frameworks and apps to target WebGPU in the future. With that backdrop, having framework developers targeting WebGPU explicitly in the backend may create some short-term integration work but it's a small price to pay because you end up with a much more streamline GPU stack, with a more robust interop, and more efficient support for the apps down the road, which is a huge advantage that multiplies over time.

The current design will make efficient and robust WebGPU interop very hard to accomplish in practice because we intentionally fragment it at the API level.

huningxin:
@wchao1115 

> However, that use case in practice will encourage multiple devices created in the same process if the app also uses WebGPU, with one internally owned by the default context and the other explicitly by the app thru a WebGPU device.

I suppose this issue could be solved by the implementation that manages and shares the GPU device between the WebNN default context and WebGPU. 

This was prototyped by my previous [WebNN-on-Dawn Chromium PoC](https://github.com/huningxin/chromium-src/tree/webnn_on_dawn) (proof-of-concept) that [integrates the WebNN implementation into WebGPU Dawn project](https://github.com/huningxin/dawn/commit/21cf41a210a511f0feefca1593b85b08048f74d9#diff-70bd03dd210ac8ddc339cabf2c8c4567423717e12af0535098a5d06ca7c13ed8). It allows a WebNN context creates and owns a WebGPU/Dawn device internally. The WebGPU/Dawn device is used by the WebNN implementation to execute the command lists that represents the MLGraph.

Another proof is [WebNN-polyfill](https://github.com/webmachinelearning/webnn-polyfill). The default context polyfill implementation is able to run on top of WebGPU via TensorFlow.js WebGPU backend. 

> It seems pretty clear that people wants WebGPU to be the way webapp deals with GPU going forward.

I agreed. And that's exactly why I think WebGPU interop is important especially for the GPU rendering / [real-time video processing](https://github.com/webmachinelearning/webnn/issues/226) use cases.

On the other hand, I still believe it is very compelling if the default context execution model, in particular `MLContext.compute()` with `ArrayBufferViews` resources, could easily target different hardware devices, including CPU, GPU and NPU, by simply setting a device type. I suppose this is widely adopted execution model by native frameworks / OS ML APIs as well.

index.bs
```diff
@@ -770,7 +753,7 @@ console.log(&#96;values: ${bufferE}&#96;);
 </div>
 
 ### Asynchronous Execution ### {#api-mlcontext-async-execution}
-Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
+Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a CPU worker thread or on a GPU timeline submitting a GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
```
huningxin, line 756: @wchao1115 

> As for the OperationError failure cases for `compute` I think it's overkill. I don't recall why we need that. We should remove those restrictions.

Do you mean it should allow the webgpu context supports the `MLContext.compute()` execution method? However, this is not reflected by this PR because the spec text still enforces a default context. You may want to clarify that. Thanks.

huningxin:
@wchao1115

> However, that use case in practice will encourage multiple devices created in the same process if the app also uses WebGPU, with one internally owned by the default context and the other explicitly by the app thru a WebGPU device.

This issue may not exist for some native ML APIs that run in a separated process of WebGPU, such as the [ChromeOS MLService](https://chromium.googlesource.com/chromiumos/platform2/+/HEAD/ml/README.md). The CG's Model Loader API implementation is based on MLService. The device type of default context is used for MLService to select a device. In particular, when creating a model loader, the [implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/ml_model_loader.cc;l=179) will get the device type from the context, pass the device type to MLService and the MLService will select a device used to do model inference. This infrastructure is also shared by the ongoing [WebNN MLService backend prototype](https://chromium-review.googlesource.com/c/chromium/src/+/4091788). 

/cc @wacky6

wacky6:
> However, that use case in practice will encourage multiple devices created in the same process if the app also uses WebGPU, with one internally owned by the default context and the other explicitly by the app thru a WebGPU device.

Is the concern about having multiple JS objects for a single GPU device, and potentially allocating 2x memory in GPU?.

Something like:
1. Create default context (implementation chooses GPU). Developer wants to use this as fallback when there's GPU resource contention.
2. Create GPU context (explicitly choose GPU)
3. Model is duplicated in GPU memory

I don't see a particular reason to discourage this from API design perspective.

---

Note, ChromeOS MLService manages GPU/NPU resource independently of WebGPU process.

In an ideal world, developer don't need to care about device type, MLService will just pick one intelligently.

Some theoretical examples:
- OS decide website engagement, system load, available resource
- OS evicts (e.g. GPU -> CPU), throttles inference when system is underload
- Decide based on model ops (e.g. TPU for matrix heavy models, GPU for convolution based models)

wchao1115:
> I suppose this issue could be solved by the implementation that manages and shares the GPU device between the WebNN default context and WebGPU.

@huningxin I can see how this can be done on a system with a single GPU adapter. But a WebGPU `gpuDevice` is created from an adapter that is explicitly selected by the developer. So, in a multi-adapter configuration, the WebNN implementation may end up create a device different from what the WebGPU user wants. 

> Is the concern about having multiple JS objects for a single GPU device, and potentially allocating 2x memory in GPU?.

@wacky6 The concern is two-folds. First, the uncoordinated implementations between WebGPU and WebNN could end up creating duplicates of the same GPU device on the same adapter. This isn't just two JS objects pointing to the same underlying GPU system device objects (e.g. a D3D device in the Windows case) but two GPU device objects, each wrapped within its own JS API object. This would be quite wasteful from the OS point of view.

Secondly, if we assume that the implementations of the two APIs are tightly coordinated (which shall we say is currently still questionable, but lets assume that for the sake of this conversation), we still have a multi-adapter case to worry about. These are systems with two physical GPU adapters (e.g. an Intel integrated graphics with a discrete GPU adapter plugged in to one of the IO slots). In this settings, a WebGPU user would be responsible for selecting an adapter they want to use as the basis for the `gpuDevice` they will create (i.e. there is no default device in WebGPU, you create a device from a pre-selected adapter). There is no way for the WebNN implementation to know ahead of time which adapter the user would want their `gpuDevice` to be base on. And, if the default context simply picks the default adapter (almost always an integrated one) while the user picks a discrete adapter, you end up having two actual GPU devices with different resource domains lying around while their resources can't actually be shared across the two adapters. So, it's not only wasteful but also not workable i.e. breaking apps.

huningxin:
@wchao1115 

> There is no way for the WebNN implementation to know ahead of time which adapter the user would want their `gpuDevice` to be base on.

That's true.

> And, if the default context simply picks the default adapter (almost always an integrated one) while the user picks a discrete adapter, you end up having two actual GPU devices with different resource domains lying around while their resources can't actually be shared across.

This case seems not an issue because the default context `compute()` method would only take `ArrayBufferView` resources and won't need to interact with any WebGPU resources created on discrete GPU adapter.

And if developers can set the "high-performance" power preference, the WebNN default context may select the same discrete GPU adapter that is selected by WebGPU call explicitly before.

huningxin:
@wacky6

> In an ideal world, developer don't need to care about device type, MLService will just pick one intelligently.

Do you mean the ["auto"](https://webmachinelearning.github.io/model-loader/#dom-mldevicepreference-auto) device preference proposal of Model Loader API?

wacky6:
> Do you mean the ["auto"](https://webmachinelearning.github.io/model-loader/#dom-mldevicepreference-auto) device preference proposal of Model Loader API?

Yes. And for multi-GPU, using ["gpu"] picks the best available one (e.g. prefers discrete over integrated).

wchao1115:
@wacky6 
> Yes. And for multi-GPU, using ["gpu"] picks the best available one (e.g. prefers discrete over integrated).

The definition of "best" varies depending on what the app wants. Discrete GPU consumes a lot of power. It may not be ideal to run a model with low compute requirements where using an integrated GPU could preserve battery life in a mobile situation.

wchao1115:
@huningxin 

> This case seems not an issue because the default context compute() method would only take ArrayBufferView resources and won't need to interact with any WebGPU resources created on discrete GPU adapter.

My point was that in this situation sharing a piece of content between the two API would then become a pair of copies for data download/upload, which is highly inefficient. 

By making a WebGPU device the only GPU device WebNN operates on, we make the API efficient by default. zero-copy by default. 

The reasoning behind this proposal is to simplify and make things more efficient by default. It isn't because something is functionally broken. My chief concern is that the current API provides too many options that could easily lead to inefficient implementation and interop with WebGPU. It seems far cleaner to coalesce around WebGPU device for GPU support of WebNN provided that efficient interop with WebGPU is highly desirable. I actually believe this proposal will help bridge the implementation gap on the GPU side with the Chromium GPU folks. 

As a concrete sample data point, one strong hint of a redundant design is that there are both the `MLPowerPreference` defined on WebNN and the `GPUPowerPreference` on WebGPU. The two enums are basically doing the exact same thing for GPU adapter selection.

huningxin:
@wchao1115 

> My point was that in this situation sharing a piece of content between the two API would then become a pair of copies for data download/upload, which is highly inefficient.

That's true. This is a WebNN/WebGPU interop scenario. Developers should create a webgpu context and use `MLCommandEncoder` for WebGPU interop.

I feel WebNN spec should allow the implementation to pick a GPU device if developers just want to use GPU but don't care about WebGPU interop or multi-adapters etc.,. 

> As a concrete sample data point, one strong hint of a redundant design is that there are both the `MLPowerPreference` defined on WebNN and the `GPUPowerPreference` on WebGPU.

Modern CPUs may also employ hybrid cores, such as [performance-cores and efficient-cores](https://en.wikipedia.org/wiki/Alder_Lake) or [big.LITTLE cores](https://en.wikipedia.org/wiki/ARM_big.LITTLE), to balance power and performance. The `MLPowerPreference` may also help select the cores to run the workloads on these CPU architectures.

wchao1115:
> I feel WebNN spec should allow the implementation to pick a GPU device if developers just want to use GPU but don't care about WebGPU interop or multi-adapters etc.,.

In practice, the GPU implementation will pick the GPU adapter for the device based on the `MLPowerPreference`. The same logic is used when a WebGPU picks an adapter based on `GPUPowerPreference`. By letting the app developer selects the adapter themselves through WebGPU, we reduce the redundancy and remove the ambiguity of having two adapters both active but known only by one of the two API without efficient interoperability.

Even though it may seem that we limit WebNN from being able to manage its own GPU device, in practice no functional capability is being removed here. The developer can still achieve the same outcome as what they can do now but doing so in a more direct way.  WebNN not managing a GPU device lifetime is a good thing from an implementation standpoint.

index.bs
```diff
@@ -519,21 +518,28 @@ operations in the graph onto a {{GPUCommandBuffer}}. The callers are responsible
 of this workload on the {{GPUQueue}} through the WebGPU queue submission mechanism. Once the submitted workload
 is completely executed, the result is avaialble in the bound output buffers.
```
anssiko, line 519: s/avaialble/available

(Line not touched by this PR but this typo caught my eye while reviewing the changes around it.)

index.bs
```diff
@@ -519,21 +518,28 @@ operations in the graph onto a {{GPUCommandBuffer}}. The callers are responsible
...
 
-In a situation when a GPU context executes a graph with a constant or an input in the system memory as an {{ArrayBufferView}}, the input content is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of the graph execution. This data upload and download cycles will only occur whenever the execution device requires the data to be copied out of and back into the system memory, such as in the case of the GPU. It doesn't occur when the device is a CPU device. Additionally, the result of the graph execution is in a known layout format. While the execution may be optimized for a native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert the content back to a known layout format at the end of the graph in order to maintain the expected behavior from the caller's perspective.
+An {{MLContext}} interface represents a state of neural network execution. An important function of this state is manage resources used in the compilation
```
anssiko, line 523: s/is/is to

index.bs
```diff
@@ -519,21 +518,28 @@ operations in the graph onto a {{GPUCommandBuffer}}. The callers are responsible
...
+An {{MLContext}} interface represents a state of neural network execution. An important function of this state is manage resources used in the compilation 
+and execution of the neural network graph. An implementation in a user agent may implement this state in term of CPU resources when an {{MLContext}} is created
```
anssiko, line 524: s/in term of/in terms of

index.bs
```diff
@@ -519,21 +518,28 @@ operations in the graph onto a {{GPUCommandBuffer}}. The callers are responsible
...
+is also used by the application of WebGPU (aka. a *"GPU context"*), any GPU resource created from the same {{GPUDevice}} such as the {{GPUBuffer}} or {{GPUTexture}} 
+is considered a native resource that can also be used as a graph constant, input, or output operand.
 
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's [=power preference=] and [=device type=] specified in the {{MLPowerPreference}} and {{MLDeviceType}} options.
+In a situation when a GPU context executes a graph with a constant or an input or output allocated in the system memory as in an {{ArrayBufferView}}, the content 
+is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of 
+the graph execution. This automatic data upload and download cycles will only occur whenever the executing GPU context determines that the data be copied out of or back
```
anssiko, line 531: s/This automatic data/The automatic data

Also, if the following is a conformance requirement (in practice something for which we can write a test case for):

s/the data be/the data must be 

If it is informative prose instead, we usually try to avoid [RFC2119 terminology](https://www.w3.org/TR/webnn/#w3c-conventions) and could say:

s/the data be/the data has to be

If the latter, informative prose is expected to be wrapped into a "Note:" block. Alternative convention is to add "This section is non-normative." at the beginning of a section that in its entirety is non-normative (=informative), example: https://html.spec.whatwg.org/multipage/introduction.html#is-this-html5?

anssiko (changes requested):
@wchao1115 thanks! I took a look and made some minor suggestions. I think this PR benefits from a WG discussion so I plan to bring this to our next call agenda to solicit more feedback and review.

wchao1115:
> @wchao1115 thanks! I took a look and made some minor suggestions. I think this PR benefits from a WG discussion so I plan to bring this to our next call agenda to solicit more feedback and review.

Thanks. How should handle occasional typos and misspelled words that are irrelevant to the change? It just makes the merge harder. Ideally, I think we should have a round of "scrub only" changes that does not materially impact the rest of the doc content.

anssiko:
It is fine to fix typos if those typos were introduced by the changes in this PR. For other changes, a separate PR keeps us focused.

anssiko:
We should check whether these proposed changes suggest changes to the [performance adaptation](https://www.w3.org/TR/webnn/#usecase-perf-adapt) use case. It is generally recommended to maintain use cases alongside the API as it evolves. In the spirit of atomic PRs any use case changes motivated by API changes should be incorporated into the same PR. If that’d complicate the concrete work, the editors are free to work on these changes in separate PRs and cross-link the two.

Related, there has been a proposal https://github.com/webmachinelearning/webnn/pull/207 to update the performance adaptation use case in the past. The WG considered that update future work given no accompanying API changes were suggested at that time.

wchao1115:
@anssiko This change doesn't affect that use case as it doesn't remove any current capability i.e. the developer can still use WebNN to direct the execution of her model to the CPU via the default context.

index.bs
wchao1115: This section of the new text doesn't describe new behavior, just repeat what was said earlier about automatic data upload/download to and from the GPU. Considering that the entire section contains the overall description of the programming model, I feel that this text naturally belongs here and not in a separate note section.

index.bs
wchao1115: Fixed

index.bs
wchao1115: Fixed

anssiko:
To ease the review: the proposed IDL changes in this PR are as follows (excluding `MLOperator` -> `MLActivation` rename changes already landed in #321):

```diff
Navigator includes NavigatorML;
 WorkerNavigator includes NavigatorML;
 
-enum MLDeviceType {
-  "cpu",
-  "gpu"
-};
-
 enum MLPowerPreference {
   "default",
-  "high-performance",
   "low-power"
 };
 
 dictionary MLContextOptions {
-  MLDeviceType deviceType = "cpu";
   MLPowerPreference powerPreference = "default";
 };
 
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface ML {
+  // Default context
   Promise<MLContext> createContext(optional MLContextOptions options = {});
-  Promise<MLContext> createContext(GPUDevice gpuDevice);
 
   [Exposed=(DedicatedWorker)]
   MLContext createContextSync(optional MLContextOptions options = {});
+
+  // GPU context
+  Promise<MLContext> createContext(GPUDevice gpuDevice);
+
   [Exposed=(DedicatedWorker)]
   MLContext createContextSync(GPUDevice gpuDevice);
 };
```

huningxin:
As discussed in [1/13 WebML WG call](https://www.w3.org/2023/01/12-webmachinelearning-minutes.html#t03), participants would like to hear the feedback from framework developers. 

@pyu10055, any perspectives? Thanks!

wchao1115:
Rebase on main and remove restrictive text that prevents `compute` to work with a WebGPU context per WG feedback on 1/13.

index.bs

index.bs
wchao1115: Fixed

RafaelCintron (approved):

huningxin:
> Rebase on main and remove restrictive text that prevents `compute` and `computeSync` to work with a WebGPU context per WG feedback on 1/13. @huningxin please take a look.

Thanks for the update. It resolved the corresponding [conversation](https://github.com/webmachinelearning/webnn/pull/322#discussion_r1060251126).

anssiko (approved):
My comments have been addressed, thanks!

We should wait for @huningxin to be back from his well-deserved holiday to confirm his feedback and comments has been addressed before we proceed with this PR.

@pyu10055, given the impact of this proposed change to JS frameworks, feedback from TF.js would be invaluable. Please see my compact diff at https://github.com/webmachinelearning/webnn/pull/322#issuecomment-1379983187 if this PR looks tl;dr. All comments and questions welcome.

(As a reminder, this was discussed on our [last call](https://www.w3.org/2023/01/12-webmachinelearning-minutes.html#t03).)

anssiko:
We discussed this on our call https://www.w3.org/2023/02/02-webmachinelearning-minutes.html today.

Here's a summary and proposed next steps:

This feature introduces a normative dependency on the WebGPU API for the GPU path. We discussed how maturization of this important feature would greatly benefit from tight coordination with WebGPU contributors, implementation experience and feedback from key customers such as JS ML framework. We also discussed pros and cons and logistics of landing this feature in a revised CR vs. in the initial CR. The upcoming WebGPU F2F 2023-02-16/17 in the SF bay area would be a great opportunity to socialise and seek feedback on the WebNN-WebGPU interoperability design overall.

@wchao1115 would you be interested in attending that F2F (virtually) and sharing the WebNN-WebGPU interop design at the meeting? @huningxin @RafaelCintron may want to contribute too.

@pyu10055, given the impact of this proposed change to JS ML frameworks, feedback from TF.js would be invaluable. Please see my compact diff at https://github.com/webmachinelearning/webnn/pull/322#issuecomment-1379983187 if this PR looks tl;dr. All comments and questions welcome.

huningxin:
@anssiko 

> We discussed this on our call https://www.w3.org/2023/02/02-webmachinelearning-minutes.html today.

I think we also would like to hear feedback from cross-platform support perspective.

@wacky6 , do you see any gaps to support this change (compact diff at https://github.com/webmachinelearning/webnn/pull/322#issuecomment-1379983187) on WebNN MLService  implementation. 

And this change may also impact [Model Loader API](https://webmachinelearning.github.io/model-loader/#api) because it is sharing the `MLContext` interface with WebNN. Any thoughts from Model Loader's point of view?

anssiko:
@mattsoulanille from the TensorFlow.js team may be able to provide feedback from the ML JS framework perspective on this proposed change.

pyu10055:
Sorry for being late to this discussion, I like @wchao1115's proposal that enables WebNN expose low level implementation details (WebGPU) to the API to enable high level framework to provide more efficient integration, and better ML pipeline. At the same time, I do agree with @huningxin that less experienced users who would prefer a semantically meaningful API, I think this would improve the usability of the API and attract more adoptions. 

I would selfishly prefer having both APIs to address needs of users from different spectrum.

wchao1115:
@anssiko I have no strong opinion on whether this change should be incorporated as part of the upcoming CR or later. I think that's entirely logistical. We should do what enables us to expedite the CR review process, which I'm not an expert of. 

As for the change itself, this line of summary worries me a bit:

> This feature introduces a normative dependency on the WebGPU API for the GPU path.

I'd like to clarify that the spec already takes this normative dependency on WebGPU today without this change. If anything, this change should simplify the implementation of this dependency in practice, as it consolidates GPU support around just one code path with a WebGPU device. 

@pyu10055 I do not consider the WebGPU dependency a *low-level implementation detail*. Again, this dependency already exists at the API level before this change i.e. WebNN already allows the caller to instantiate an `MLContext` from a specified `GPUDevice`.

Given that this change is about removing the support for an internal GPU device option (i.e. the `MLDeviceType == gpu` use case), one of the core arguments that folks have been making against it is that it'll make it harder for the caller of WebNN (e.g. TF.js) to support GPU execution (as it removes an "easier" way of doing it). This argument is based on a wrong assumption that supporting the GPU would require nothing other than an enum value or a Boolean option from the caller. Consider the case where the configuration of the client machine includes multiple GPU adapters for example, which adapter should be used for WebNN execution? Does the framework like TF.js have enough information to decide on behalf of its caller?  What if that decision is either incorrect or must be changed in the future? 

How much should we care about this multi-adapter systems? i.e. how many client devices out there with multiple adapters in it? It turns out that roughly 1/3 of all the PC out there are equipped with a discrete GPU adapter, and every one of those machines also comes with an Intel integrated graphics. 

Allowing the caller to specify precisely which `GPUDevice` (from which GPU adapter exactly) they would want to run inference on mitigates that dangerous ambiguity that could really lead to a very different outcome for the users. One data point today is with ONNX Runtime i.e. ORT does not assume which GPU adapter the caller wants to run on, it relies on the caller to choose explicitly. I think every ML framework that supports the GPU would very likely need to make a similar design choice.

pyu10055:
@wchao1115 I might have not explained myself clearly, I think it is a good thing to framework like TFJS to be able to specify or pass down the gpudevice to WebNN. The question is whether a default device can/should be specified through semantic API (high-performance/low power/etc). I think this is useful, most likely, the end users or framework might not have the understanding or low level hardware access to make a better decision.

wchao1115:
@pyu10055 If what you mean is through the `MLPowerPreference` options, please also take a look at [GPUPowerPreference](https://gpuweb.github.io/gpuweb/#dom-gpurequestadapteroptions-powerpreference). As you can see, they are essentially identical. To select a GPU adapter, the app only needs to make 1 call to [GPU.RequestAdapter](https://gpuweb.github.io/gpuweb/#dom-gpu-requestadapter) with a specified `GPUPowerPreference` "hint" value. It isn't hard to do, can be done easily by the app. 

But if you think you would rather surface a similar adapter hint to the app of TF.js, that call to *requestAdapter* can also be made by TF.js on behalf of the app who passes in the adapter hint. In either case, it can be done above WebNN. 

WebNN is intentionally designed to be a backend API, and not really as a framework API. Having it accepting the device corresponding to the desired adapter removes ambiguity and the burden of maintaining compatibility in the future.

pyu10055:
@wchao1115 Using WebGPU device makes sense if the API allows frameworks or users to specifically create a WebGPU backend. But if the intention of WebNN GPU backend is transparent GPU acceleration, I think WebNN should have an level of abstraction on GPU, exposing WebGPU objects as part of WebNN API will add burdens to non-WebGPU implementations.

wchao1115:
> I think WebNN should have an level of abstraction on GPU, exposing WebGPU objects as part of WebNN API will add burdens to non-WebGPU implementations.

This is already the case today even with the default GPU option i.e. if you intend to run a model on the GPU with an input bound to a GPU resource, that GPU resource must already be a WebGPU resource. Moreover, it must be a WebGPU resource created from the same WebGPU device that WebNN internally manages. This is the reason why WebNN can't really "hide" the underlying WebGPU device from the caller. It's much more natural and robust to have the WebNN caller specifies both what device it wants to run on, and what resource from the same device that it wants to bind the model's input to.

anssiko:
>I have no strong opinion on whether this change should be incorporated as part of the upcoming CR or later. I think that's entirely logistical. We should do what enables us to expedite the CR review process, which I'm not an expert of.

@wchao1115 here's the proposed "status of this document" text that will enable us to expedite the CR process and solicit wider feedback and review on this important design consideration: https://github.com/webmachinelearning/webnn/pull/340

wchao1115:
@anssiko I've commented on #340. I don't think this PR is related to WebGPU interop. As has been explained in this thread, this PR is about removing the internal GPU option in MLContext. WebGPU interop requirement already exists prior to this proposed change.

wchao1115:
Retract until consensus can be found.

PR #319, opened 2022-12-20
WiP: Improve graph execution steps

zolkis:
Factor out graph resource validation. 
Define generic graph execution steps and use them in the sync and async compute() methods. 
Address #316.

@huningxin @wchao1115 @anssiko this requires not only a review, but possibly _development_.
Prose detailing implementation internals has been thin ice in specifications. We need to be generic and flexible to internal optimizations, yet express the intents clearly enough. Your help is needed, please.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/319.html" title="Last updated on Dec 22, 2022, 9:18 AM UTC (071498b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/319/393cd2a...zolkis:071498b.html" title="Last updated on Dec 22, 2022, 9:18 AM UTC (071498b)">Diff</a>

PR #323, opened 2023-01-06
Transfer the input and output views for asynchronous execution

### LINKED ISSUE #318: The input and output resources race condition issue of asynchronous execution
For WebNN graph [asynchronous execution](https://www.w3.org/TR/webnn/#api-mlcontext-async-execution), the main thread and the worker thread may access the input or output array buffers at the same time that would cause the race condition issue.

This issue was captured during the Chromium [CL](https://www.w3.org/TR/webnn/#api-mlcontext-async-execution) review. Thanks @wacky6 for pointing it out. There are some steps may cause this issue:

1. Main thread calls [`MLContext.compute()`](https://www.w3.org/TR/webnn/#dom-mlcontext-compute) with input and output array buffers.
2. The input and output array buffers are passed to worker thread for execution.
3. Main thread may change the array buffers concurrently, e.g., trigger reallocation, detaches the array buffers or modify the content after the validation
4. The worker thread may read / write to the wrong backing stores.

The root cause is, in the current design, both the main thread and worker thread can access the input and output array buffers at the same time. 

A possible solution could be similar to [ReadableStreamBYOBReader](https://streams.spec.whatwg.org/#readablestreambyobreader) interface. It transfers the user-supplied array buffers to a new ones and post the new ones to the worker thread for execution. The new array buffers would share the same backing store of the user-supplied ones, so there are no extra memory copies. But the main thread is not able to change them during the execution, because they are already detached. After worker thread fills the results, the promise will be resolved with the new array buffers. So the main thread can access the results after that. There are more details of this design in @domenic 's [post](https://blog.domenic.me/reading-from-files/).

This may require a change to WebNN `MLContext.compute()` method to something like:

```
dictionary ComputeResult {
  MLNamedArrayBufferViews inputs;
  MLNamedArrayBufferViews outputs;
};

partial interface MLContext {
  Promise<ComputeResult> compute(
      MLGraph graph, MLNamedArrayBufferViews inputs, MLNamedArrayBufferViews outputs);
};
```

huningxin:
fix #318

@wchao1115 @anssiko @wacky6 @RafaelCintron , please take a look. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/323.html" title="Last updated on Jan 11, 2023, 7:13 AM UTC (a750ea4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/323/5a4e73e...huningxin:a750ea4.html" title="Last updated on Jan 11, 2023, 7:13 AM UTC (a750ea4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -40,7 +40,8 @@ urlPrefix: https://webidl.spec.whatwg.org/; spec: WEBIDL
         text: Promise; url: idl-promise
         text: nullable; url: idl-nullable-type
     type: dfn
-        text: transfer array buffer; url: arraybuffer-transfer
+        text: underlying buffer; url: buffersource-underlying-buffer
+        text: buffer byte length; url: buffersource-byte-length
 </pre>
 <pre class="biblio">
 {
@@ -777,9 +778,9 @@ console.log(&#96;values: ${bufferE}&#96;);
 To <dfn for="MLNamedArrayBufferViews">transfer</dfn> an {{MLNamedArrayBufferViews}} |views|:
 1. Let |transferredViews| be a new {{MLNamedArrayBufferViews}}.
 1. For each |key| -> |value| of |views|:
-    1. Let |transferredBuffer| be the result of [=transfer array buffer|transferring=] {{ArrayBuffer}} |value|.\[[ViewedArrayBuffer]].
+    1. Let |transferredBuffer| be the result of [=ArrayBuffer/transfer|transferring=] the [=underlying buffer=] of |value|.
     1. Let |constructor| be the appropriate [=view constructor=] for the type of {{ArrayBufferView}} |value|.
-    1. Let |elementsNumber| be the result of |value|.\[[ByteLength]] ÷ [=element size=] of |value|.
+    1. Let |elementsNumber| be the result of the [=buffer byte length|byte length=] of |value| ÷ [=element size=] of |value|.
     1. Let |transferredView| be [=Construct=](|constructor|, |transferredBuffer|, |value|.\[[ByteOffset]], |elementsNumber|).
     1. Set |transferredViews|[|key|] to |transferredView|.
 1. Return |transferredViews|.
--- a/index.bs
+++ b/index.bs
@@ -33,10 +33,14 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
     type: dfn
         text: element size; url: table-the-typedarray-constructors
         text: element type; url: table-the-typedarray-constructors
+        text: view constructor; url: table-the-typedarray-constructors
+        text: Construct; url: sec-construct
 urlPrefix: https://webidl.spec.whatwg.org/; spec: WEBIDL
     type: interface
         text: Promise; url: idl-promise
         text: nullable; url: idl-nullable-type
+    type: dfn
+        text: transfer array buffer; url: arraybuffer-transfer
 </pre>
 <pre class="biblio">
 {
@@ -733,7 +737,7 @@ partial interface MLContext {
     1. Return {{undefined}}.
 </div>
 
-#### Examples #### {#compilation-examples}
+#### Examples #### {#api-mlcontext-sync-execution-examples}
 
 <div class="example">
 The following code showcases the synchronous computation with optional outputs in a worker.
@@ -769,12 +773,28 @@ console.log(&#96;values: ${bufferE}&#96;);
 </pre>
 </div>
 
+### The {{MLNamedArrayBufferViews}} transfer algorithm ### {#mlnamedarraybufferviews-transfer}
+To <dfn for="MLNamedArrayBufferViews">transfer</dfn> an {{MLNamedArrayBufferViews}} |views|:
+1. Let |transferredViews| be a new {{MLNamedArrayBufferViews}}.
+1. For each |key| -> |value| of |views|:
+    1. Let |transferredBuffer| be the result of [=transfer array buffer|transferring=] {{ArrayBuffer}} |value|.\[[ViewedArrayBuffer]].
+    1. Let |constructor| be the appropriate [=view constructor=] for the type of {{ArrayBufferView}} |value|.
+    1. Let |elementsNumber| be the result of |value|.\[[ByteLength]] ÷ [=element size=] of |value|.
+    1. Let |transferredView| be [=Construct=](|constructor|, |transferredBuffer|, |value|.\[[ByteOffset]], |elementsNumber|).
+    1. Set |transferredViews|[|key|] to |transferredView|.
+1. Return |transferredViews|.
+
 ### Asynchronous Execution ### {#api-mlcontext-async-execution}
-Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
+Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. To prevent the calling thread from modifying the input and output resources while the computation is ongoing, this method [=MLNamedArrayBufferViews/transfer|transfers=] the input and output {{MLNamedArrayBufferViews}} to new views that share the same backing memory allocations. The transferred views are returned to the caller via the promise fulfillment with the computation result written into the backing memory of the output views. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
 
 <script type=idl>
+dictionary ComputeResult {
+  MLNamedArrayBufferViews inputs;
+  MLNamedArrayBufferViews outputs;
+};
+
 partial interface MLContext {
-  Promise<undefined> compute(
+  Promise<ComputeResult> compute(
       MLGraph graph, MLNamedArrayBufferViews inputs, MLNamedArrayBufferViews outputs);
 };
 </script>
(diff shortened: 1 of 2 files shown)


anssiko:
@domenic, as the author of the [`ReadableStreamBYOBReader`](https://streams.spec.whatwg.org/#readablestreambyobreader) design, you may want to look at this PR inspired by your work.

index.bs
```diff
@@ -769,12 +773,28 @@ console.log(&#96;values: ${bufferE}&#96;);
...
+Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. To prevent the calling thread from modifying the input and output resources while the computation is ongoing, this method [=MLNamedArrayBufferViews/transfer|transfers=] the input and output {{MLNamedArrayBufferViews}} to new views that share the same backing memory allocations. The transferred views are returned to the caller via the promise fulfillment with the computation result written into the backing memory of the output views. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
 
 <script type=idl>
+dictionary ComputeResult {
```
wchao1115, line 791: To be naming-consistent with the rest of the API, should we also prefix this type with `ML` e.g. `MLComputeResult`?

index.bs
```diff
@@ -801,31 +822,67 @@ partial interface MLContext {
...
+// 3. Bind inputs to the graph and execute for the result.
+const bufferA = new Float32Array(4).fill(1.0);
+const bufferB = new Float32Array(4).fill(0.8);
+const bufferC = new Float32Array(4);
+const inputs = {'A': bufferA, 'B': bufferB};
+const outputs = {'C': bufferC};
+const result = await context.compute(graph, inputs, outputs);
+// The computed result of [[1, 1], [1, 1]] is in the buffer associated with
+// the output operand.
+console.log('Output value: ' + result.C);
```
wchao1115, line 880: Should this be `result.outputs.C`?

index.bs
```diff
@@ -2607,9 +2664,9 @@ const inputs = {
   'input2': inputBuffer2,
 };
 const outputs = {'output': outputBuffer};
-await context.compute(graph, inputs, outputs);
+const result = await context.compute(graph, inputs, outputs);
 
-console.log('Output value: ' + outputBuffer);
+console.log('Output value: ' + result.output);
```
wchao1115, line 2669: `result.ouputs.output`?

wchao1115 (commented):
Sorry for delay. I was sick the last few days.

index.bs
huningxin: Yes, we should. Fixed in the latest commit.

index.bs
huningxin: Sorry, my typo. Fixed in the latest commit.

index.bs
huningxin: Thanks for catching this bug! fixed in the latest commit.

huningxin:
@wchao1115 

> Sorry for delay. I was sick the last few days.

No worries. Hope you are feeling better now. Thanks for your review and comments. I uploaded a new commit to address your comments. Please take another look.

index.bs
```diff
@@ -769,12 +773,28 @@ console.log(&#96;values: ${bufferE}&#96;);
 </pre>
 </div>
 
+### The {{MLNamedArrayBufferViews}} transfer algorithm ### {#mlnamedarraybufferviews-transfer}
+To <dfn for="MLNamedArrayBufferViews">transfer</dfn> an {{MLNamedArrayBufferViews}} |views|:
+1. Let |transferredViews| be a new {{MLNamedArrayBufferViews}}.
+1. For each |key| -> |value| of |views|:
+    1. Let |transferredBuffer| be the result of [=transfer array buffer|transferring=] {{ArrayBuffer}} |value|.\[[ViewedArrayBuffer]].
```
domenic, line 780: Use https://webidl.spec.whatwg.org/#buffersource-underlying-buffer instead of consulting [[ViewedArrayBuffer]]

index.bs
```diff
@@ -33,10 +33,14 @@ urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
...
         text: element size; url: table-the-typedarray-constructors
         text: element type; url: table-the-typedarray-constructors
+        text: view constructor; url: table-the-typedarray-constructors
+        text: Construct; url: sec-construct
 urlPrefix: https://webidl.spec.whatwg.org/; spec: WEBIDL
     type: interface
         text: Promise; url: idl-promise
         text: nullable; url: idl-nullable-type
+    type: dfn
+        text: transfer array buffer; url: arraybuffer-transfer
```
domenic, line 43: This should not be necessary if you use `[=ArrayBuffer/transfer=]` syntax.

index.bs
```diff
@@ -769,12 +773,28 @@ console.log(&#96;values: ${bufferE}&#96;);
 </pre>
 </div>
 
+### The {{MLNamedArrayBufferViews}} transfer algorithm ### {#mlnamedarraybufferviews-transfer}
+To <dfn for="MLNamedArrayBufferViews">transfer</dfn> an {{MLNamedArrayBufferViews}} |views|:
+1. Let |transferredViews| be a new {{MLNamedArrayBufferViews}}.
+1. For each |key| -> |value| of |views|:
+    1. Let |transferredBuffer| be the result of [=transfer array buffer|transferring=] {{ArrayBuffer}} |value|.\[[ViewedArrayBuffer]].
+    1. Let |constructor| be the appropriate [=view constructor=] for the type of {{ArrayBufferView}} |value|.
+    1. Let |elementsNumber| be the result of |value|.\[[ByteLength]] ÷ [=element size=] of |value|.
```
domenic, line 782: Use https://webidl.spec.whatwg.org/#buffersource-byte-length instead of accessing [[ByteLength]] directly.

domenic (commented):
Some minor suggestions on how to better use Web IDL wrapper algorithms, which are in general preferable to directly using JS spec primitives.

However, it appears we haven't added sufficient wrapper algorithms to Web IDL to replace the Construct() step. So directly using JS spec primitives there is fine, if a bit unfortunate.

There is a more dramatic change I would suggest, if possible. It would be better if your spec algorithms operated on [byte sequences](https://infra.spec.whatwg.org/#byte-sequences) instead of ArrayBufferViews. That is, you would:

- Accept `MLNamedArrayBufferViews`s.
- Transfer the underlying buffers.
- [Get a copy of the bytes](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy) of that buffer. (In a real implementation, this would not make a copy, but in the spec, it's a convenient way to extract the byte sequences and get away from the IDL types. Since the copy is unobservable, this is fine.)
- Modify those byte sequences.
- Eventually, [create](https://webidl.spec.whatwg.org/#arraybufferview-create) ArrayBufferViews from those byte sequences. (Again, the algorithm here is a copy, but an unobservable one that implementations will thus elide.)
- Then stuff those into new `MLNamedArrayBufferViews`s to resolve your promise with.

However, I'm not sure how well this would work for your case exactly. It seems like you care about the view types and such, and tracking that separately might be a pain.

Note that pretty much all of this advice is about how to do things better than the Streams Standard currently does them. Streams was written before we added appropriate helpers to Web IDL, and we haven't taken the time to update Streams yet. That also means people haven't really battle-tested the Web IDL helpers, so they might not work perfectly :). But the above is how I think it should work...

anssiko:
@domenic much thanks for your detailed advice on behalf of the whole WG!

index.bs
huningxin: Fixed by the latest commit.

index.bs
huningxin: Sounds good. Fixed it in the latest commit.

index.bs
huningxin: Thanks for the pointer. Fixed it in the latest commit.

huningxin:
@domenic , I'd like to echo @anssiko and thank you for the detailed feedback! It is very helpful.

> Some minor suggestions on how to better use Web IDL wrapper algorithms, which are in general preferable to directly using JS spec primitives.

I pushed a new [commit](https://github.com/webmachinelearning/webnn/pull/323/commits/a750ea47366e11ba7c96604926a6e7a415353e4a) that uses the Web IDL wrapper algorithms as you suggested.

> There is a more dramatic change I would suggest, if possible. It would be better if your spec algorithms operated on [byte sequences](https://infra.spec.whatwg.org/#byte-sequences) instead of ArrayBufferViews.

Thanks for this proposal! I feel the [byte sequences](https://infra.spec.whatwg.org/#byte-sequences) algorithm would help improve the graph execution steps that @zolkis is working on by https://github.com/webmachinelearning/webnn/pull/319.

Because this PR mainly focuses on fixing #318 by transferring the input/output buffers, we probably can file a separate issue to track this proposal and fix it by a follow-up PR or Zoltan's PR #319.

WDYT?

zolkis:
> Because this PR mainly focuses on fixing https://github.com/webmachinelearning/webnn/issues/318 by transferring the input/output buffers, we probably can file a separate issue to track this proposal and fix it by a follow-up PR or Zoltan's PR https://github.com/webmachinelearning/webnn/pull/319.

From my part I am fine with that. Please finish what you wanted in this PR and I will rework (not only rebase) my PR based on that.

index.bs
```diff
@@ -769,12 +774,28 @@ console.log(&#96;values: ${bufferE}&#96;);
...
+Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. To prevent the calling thread from modifying the input and output resources while the computation is ongoing, this method [=MLNamedArrayBufferViews/transfer|transfers=] the input and output {{MLNamedArrayBufferViews}} to new views that share the same backing memory allocations. The transferred views are returned to the caller via the promise fulfillment with the computation result written into the backing memory of the output views. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it throws an "{{OperationError}}" {{DOMException}}.
```
anssiko, line 789: We could wrap this newly added informative text that documents the motivation for the design into a note block and also cross-reference the related Web IDL warning* in this note, for example:

>Note: In accordance with the [Web IDL warning](https://webidl.spec.whatwg.org/#arraybufferview-write), to prevent the calling thread from modifying the input and output resources while the computation is ongoing, this method [=MLNamedArrayBufferViews/transfer|transfers=] the input and output {{MLNamedArrayBufferViews}} to new views that share the same backing memory allocations. The transferred views are returned to the caller via the promise fulfillment with the computation result written into the backing memory of the output views.

Maybe also move "This method of execution requires ..." above this box.

(*It seems we can't anchor on the warning box directly.)

anssiko (approved):
@huningxin thanks! Approved with one suggestion and with an assumption @zolkis' upcoming update address the [byte sequences](https://infra.spec.whatwg.org/#byte-sequences) algorithm bits.

index.bs
huningxin: > We could wrap this newly added informative text that documents the motivation for the design into a note block and also cross-reference the related Web IDL warning* in this note

Sounds good. Thanks @anssiko . I pushed a new [commit](https://github.com/webmachinelearning/webnn/pull/323/commits/715795c2277e972fe4a36d4a3a693110d44afa72) to fix it. Please take another look.

> Maybe also move "This method of execution requires ..." above this box.

Because there is ongoing [discussion](https://github.com/webmachinelearning/webnn/pull/322#discussion_r1060251126) about this text in PR #322 review. I'd prefer to leaving this out of this PR.

WDYT?

anssiko: LGTM, thanks!

anssiko:
@wchao1115 PTAL

As discussed yesterday, we'd like to get this PR landed soon so @zkis can rework his PR #319 on top of this. Thanks!

wchao1115 (approved):

anssiko:
Thank you everyone! This PR has received adequate review and we're now ready to merge to unblock related work.

I'm especially pleased to see how the cross-group collaboration helped improve the design.

PR #329, opened 2023-01-18
Rework the sync async algorithms based on PR 323

zolkis:
Factor out graph input/output validation.
Factor out graph execution algorithm.
Used them in the sync and async algorithm steps.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/329.html" title="Last updated on Jun 29, 2023, 1:10 PM UTC (ac2c8e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/329/cc34577...zolkis:ac2c8e9.html" title="Last updated on Jun 29, 2023, 1:10 PM UTC (ac2c8e9)">Diff</a>


zolkis:
Looks like there are some changes due to auto-save: in some PRs before me, there were some trailing spaces. Please set your editors to remove trailing spaces.

zolkis:
@huningxin @wchao1115 @anssiko PTAL

zolkis:
Still work in progress... I want to find a better way to express underlying graph execution / compute requests, it also overlaps with other algorithms (graph internal slots, operand internal slots etc.)

zolkis:
@huningxin could you please check for correctness?

index.bs
```diff
@@ -796,38 +796,46 @@ partial interface MLContext {
...
+        1. Let |inputDescriptor| be |graph|.{{MLGraph/[[inputDescriptors]]}}[|key|].
+        1. Let |inputTensor| be a new tensor for |graph|.{{MLGraph/[[implementation]]}} as follows:
+            1. Set the data type of |inputTensor| to the one that matches the [=element type=] of |inputValue|.
+            1. Set the dimensions of |inputTensor| to |inputDescriptor|.{{MLOperandDescriptor/dimensions}}.
+            1. Set the values of elements in |inputTensor| to the values of elements in |inputValue|.
+        1. Request the underlying implementation of |graph| to bind |inputResources|[|key|] to |inputTensor|.
+    1. [=Assert=]: the type of |outputs| is {{MLNamedArrayBufferViews}}.
+    1. Let |results| be the result of invoking |graph|.{{MLGraph/[[implementation]]}} to compute the output of |graph| given |inputResources|. This may involve multiple steps and compute requests.
+        1. If that returns an error, throw an "{{OperationError}}" {{DOMException}} and stop.
+    1. [=Assert=]: the type of |results| is {{MLNamedArrayBufferViews}}.
```
huningxin, line 831: Should `|results|` be `|outputs|`?

index.bs
```diff
@@ -796,38 +796,46 @@ partial interface MLContext {
...
+    1. For each <|key|, |outputValue|> of |outputs|:
+        1. Let |outputDesc| be |graph|.{{MLGraph/[[outputDescriptors]]}}[|key|].
+        1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} to provide the output associated with |key| given |inputResources|[|key|] and stored in |resultTensor|.
```
huningxin, line 834: Should `|inputResources|[|key|]` be `|inputResources|`? The `|key|` is for output.

index.bs
```diff
@@ -796,38 +796,46 @@ partial interface MLContext {
...
+    1. Let |inputResources| denote the input resources of |graph|.{{MLGraph/[[implementation]]}}.
+    1. For each <|key|, |inputValue|> of |inputs|:
+        1. Let |inputDescriptor| be |graph|.{{MLGraph/[[inputDescriptors]]}}[|key|].
+        1. Let |inputTensor| be a new tensor for |graph|.{{MLGraph/[[implementation]]}} as follows:
+            1. Set the data type of |inputTensor| to the one that matches the [=element type=] of |inputValue|.
+            1. Set the dimensions of |inputTensor| to |inputDescriptor|.{{MLOperandDescriptor/dimensions}}.
+            1. Set the values of elements in |inputTensor| to the values of elements in |inputValue|.
+        1. Request the underlying implementation of |graph| to bind |inputResources|[|key|] to |inputTensor|.
+    1. [=Assert=]: the type of |outputs| is {{MLNamedArrayBufferViews}}.
+    1. Let |results| be the result of invoking |graph|.{{MLGraph/[[implementation]]}} to compute the output of |graph| given |inputResources|. This may involve multiple steps and compute requests.
```
huningxin, line 829: What's the usage of `|results|`?

zolkis: Intermediary value, private to the steps. I thought it would be more clear that way, but it seems it's not :), so I will rework this part.

huningxin (approved):
LGTM, thanks @zolkis !

index.bs
```diff
@@ -898,57 +910,32 @@ partial interface MLContext {
...
+        1. If that throws an error, [=reject=] |promise| with the error and stop.
+        1. Otherwise, when <a>execute graph</a> has completed:
+            1. Let |result| be a new {{MLComputeResult}}.
+            1. Set |result|.{{MLComputeResult/inputs}} to |transferredInputs|.
+            1. Set |result|.{{MLComputeResult/outputs}} to |transferredOutputs|.
+            1. [=Resolve=] |promise| with |result| and stop.
+
+    <div class="note">
+        During asynchronous execution, the compute requests in the <a>execute graph</a> algorithm MAY be executed [=in parallel=], or otherwise optimized for the underlying platform.
+    </div>
```
wchao1115, line 938: I think this paragraph may add more confusion to the implementation than it helps. In practice, multiple asynchronous executions on the CPU may be handled independently by multiple worker threads. But for the GPU, although the recording of multiple command lists can be done concurrently by multiple CPU worker threads, the execution of the command lists still happen sequentially (but on the GPU timeline) once they are submitted to the command queue.

zolkis: Right, that's an important distinction, from the perspective of the underlying platform. The (informative) note was written in the web platform terminology and from web spec viewpoint. I wonder should we extend the note with your remark? Then it would be informative AND exact. But I am also fine removing it.

@anssiko, @dontcallmedom do you have an input from the web perspective, is this kind of note useful?
@wchao1115 and @huningxin,  what would you suggest: remove, extend, or rewrite (please suggest text)?

huningxin: > In practice, multiple asynchronous executions on the CPU may be handled independently by multiple worker threads.

For CPU execution, if multiple worker threads execute the graph "in parallel", it would require synchronization between them (e.g. a lock) that protects the graph data structure and also ensures the execution order. It increases the implementation complexity and probably would be easier to introduce bugs. 

Because Chromium also prefers to [Using Sequences Instead of Locks](https://chromium.googlesource.com/chromium/src/+/main/docs/threading_and_tasks.md#Using-Sequences-Instead-of-Locks), this is a proposal [CL](https://chromium-review.googlesource.com/c/chromium/src/+/4382891) that uses `SequencedTaskRunner` of worker pool for XNNPACK CPU backend. With that, although there may be multiple worker pool threads are scheduled to compute the graph, `SequencedTaskRunner` would ensure there is one compute at a time in the posting order. 

/cc @wacky6.

wacky6: > In practice, multiple asynchronous executions on the CPU may be handled independently by multiple worker threads.

I think this is unlikely to be implemented in practice:

- Backends (most likely) aren't re-entrant safe. The implementation need to clone them (e.g. XNN runtime and static values) for each thread for safety. Resource usage is `O(N_threads)`, which might be too much for large models.
- Browsers likely wants to throttle execution (e.g. for background pages) or limit resource usage or preempt under pressure (e.g. preventing an website to consume all available RAM). Single executor (task runner) can achieve this much easier.
- Single executor likely incurs less overhead (e.g. less resource contention and context switches when n_threads > n_cores).
- Accelerators (GPU, TPU, NPU) probably follows the sequenced execution order.

zolkis: Note the expression ["in parallel"](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) in web specs mainly means it should not block the event loop. Details on how that "parallelism" is interpreted are not needed to be explicit. Quoting from the HTML spec,

> To run steps in parallel means those steps are to be run, one after another, at the same time as other logic in the standard (e.g., at the same time as the [event loop](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop)). This standard does not define the precise mechanism by which this is achieved, be it time-sharing cooperative multitasking, fibers, threads, processes, using different hyperthreads, cores, CPUs, machines, etc. By contrast, an operation that is to run immediately must interrupt the currently running task, run itself, and then resume the previously running task.

The [guidance ](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop-for-spec-authors) for spec authors is:

> the overriding guideline is that any work a specification needs to perform that would otherwise block the [event loop](https://html.spec.whatwg.org/multipage/webappapis.html#event-loop) must instead be performed [in parallel](https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel) with it.  This includes (but is not limited to): performing heavy computation; displaying a user-facing prompt; performing operations which could require involving outside systems (i.e. "going out of process").

zolkis: FWIW I think we can delete this note, since the normative text already contains how to run the steps.

zolkis: @wchao1115 , @huningxin : based on the content of this discussion, are you fine keeping this note, or should it be removed?

anssiko: Per https://www.w3.org/TR/webnn/#w3c-conventions RFC 2119 terminology is not to be used in an informative note.

My recommendation would be to remove this note.

zolkis:
@wchao1115 PTAL: this is editorial, but reformulates wording and factors out sub-steps. 
Cleaned up to one squashed commit.

zolkis:
This would need to be merged, as needed in zk-conventions-integration, and only adds editorial changes to a previous PR.

wchao1115:
@zolkis Should we discard this PR in favor of merging the work into the staging branch? This is one of the oldests in the queue.

zolkis:
This is quite different from that staging set of changes. Mainly editorial, factoring out some parts. This still belongs to the main branch IMHO. Technically it is possible to move there, but its value is in the main branch.

wchao1115:
@zolkis Can you help me understand why can't this change be staged in the staging branch like others? Given that it's been many months in the PR queue and still requires 1 more approval with several comments still outstanding, can we please move this out to the staging branch?

zolkis:
> @zolkis Can you help me understand why can't this change be staged in the staging branch like others? Given that it's been many months in the PR queue and still requires 1 more approval with several comments still outstanding, can we please move this out to the staging branch?

I didn't say it can't be staged. I said it does not belong there.
This change has been reviewed and approved on March 17 by one editor.
It has only been missing your review.
I don't see _any_ comments that are outstanding. Could you point to those?

anssiko (approved):

anssiko:
I marked the last remaining discussion as resolved.

@wchao1115 would merging this to main now complicate your expected future changes?

If it does, we can merge to the integration branch instead.

This enhancement PR is a by-product of discussion in PR #323. I understand this context may have been lost because there were multiple now closed WIP PRs before this one.

zolkis:
Right, if merging this conflicts too much with your upcoming PRs @wchao1115 , we can move this to staging, but let's try to merge it here.

anssiko:
CI build fails with:

```
WARNING: Multiple elements have the same ID 'validate-buffer-with-descriptor'.
Deduping, but this ID may not be stable across revisions.
FATAL ERROR: Multiple local 'dfn' <dfn>s have the same linking text 'validate buffer with descriptor'.
```

@zolkis please check and fix.

zolkis:
@wchao1115 could you please check if you'd be fine merging this? It concerns @huningxin 's changes, has 2 approvals and it's editorial, factoring out common parts and making the algorithms more compact. It doesn't strictly belong to the kind of changes in conventions integration, and it's been discussed several times in calls. 

Is there something left undone? Or do you sustain this should be discarded and include the changes in the conventions integration?

wchao1115 (approved):

wchao1115:
> @wchao1115 could you please check if you'd be fine merging this? 

Yes, I approved.

anssiko:
Thanks @wchao1115 for your review. I'll squash and merge so that @zolkis can pull these changes into his integration branch.

Issue #347, opened 2023-02-15
Add the clamp() algorithm


zolkis:
This is to track the algorithm specification for the clamp() operation.
Depends on #336.

inexorabletash:
I believe this is complete. Marking closed, if anything is still outstanding please re-open or file a new issue with the details.

Issue #336, opened 2023-02-01
Add internal slots to MLOperand, MLActivation and basic algorithms


zolkis:
In the validation steps for various operations (for instance Batch Norm), the following dependencies/references/algorithms need to be defined (this entry will be kept updated):
- **Internal slots** for MLOperand: 
    - [[type]] as MLOperandType
    - [[dimensions]] as `sequence<unsigned long>`
    - [[name]] as string (only for input operands)
    - [[data]] as ArrayBufferView (tensor of given dimensions)
    - [[builder]] the associated MLGraphBuilder
- Algorithm (or internal slot) to get the **_rank_** of the operand (for checking range for axis)
- Algorithm to internally **create** an MLOperand given |builder| and |descriptor|, used by op algorithms.
- Algorithm to internally **copy** an MLOperand, used by op algorithms.
- Algorithm to **check dimensions** given |dimensions| and |type|.
- **Internal slots** for MLActivation:
    - [[name]] as string, e.g. "relu", "sigmoid", "softmax", etc.
    - ? [[input]] as MLOperand (needed?)
    - ? [[output]] as MLOperand (intermediate output?)
- Algorithm to **create** an MLActivation, used by the op algorithms.

Related to the clarification of terms requested in #335.

anssiko:
@zolkis given PR https://github.com/webmachinelearning/webnn/pull/400#issuecomment-1602228662 was closed please provide a record here of commit(s) and/or PR(s) that address this issue. When done, close this issue.

My recommendation is to link each PR with an issue using the [Fix keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) either in the PR description of the commit message to enable the associated issue(s) to be automatically closed upon merge.

zolkis:
The content of #400, and previously, #337, has been agreed with @huningxin to be split among the PRs that will have dependencies to this content: 
#347, #348, #398 (#366), #365

PR #348, opened 2023-02-15
[WiP] Add the clamp() algorithm

zolkis:
Add the clamp algorithm. Tracked in #347 (discussion there).
Depends on #336.


zolkis:
During the call today, @huningxin you made a remark that an MLActivation is not needed when an MLOperator is returned. To me it seems it was needed in the [impl](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/webnn/ml_graph_builder.cc;l=551), also for the step of connecting the op input/output. But welcome to suggest better steps here, thanks!

huningxin:
@zolkis 

> To me it seems it was needed in the [impl](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/webnn/ml_graph_builder.cc;l=551), also for the step of connecting the op input/output. But welcome to suggest better steps here, thanks!

The impl you referred to returns an `MLOperand`.

This [impl](https://chromium-review.googlesource.com/c/chromium/src/+/4237384) returns an `MLOperator` (and should be updated to `MLActivation`)

zolkis:
> @zolkis
> 
> > To me it seems it was needed in the [impl](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/webnn/ml_graph_builder.cc;l=551), also for the step of connecting the op input/output. But welcome to suggest better steps here, thanks!
> 
> The impl you referred to returns an `MLOperand`.
> 
> This [impl](https://chromium-review.googlesource.com/c/chromium/src/+/4237384) returns an `MLOperator` (and should be updated to `MLActivation`)

We have a single algorithm for clamp() in JS, and we distinguish for the return values based on the arguments provided to the function. So the two impl need to be combined there. Of course in cpp the implementation can be polymorphic. AFAIK we cannot have 2 separate algorithms mirroring a cpp implementation. We need to figure out what we want in JS, and _then_ implement in cpp. As this is the same case with almost every other function in the API, we need to get this right.
I added a note to clarify the algorithm.

zolkis:
Also, I will need to update #337 with internal constructors for MLOperand and MLActivation as needed in the steps of the clamp() algorithm. I think it's better to add those commits to that PR, not this one.

zolkis:
I modified the algorithm like they would be when #337 is merged. This shows how the `create MLOperand` and `create MLActivation` steps would be used.

zolkis:
I am preparing the PR for input() and constant(), and that changes this PR, too. Moving it back to WiP.

index.bs
```diff
@@ -1212,20 +1212,51 @@ dictionary MLClampOptions {
...
     **Returns:**
-        - an {{MLOperand}}. The output tensor of the same shape as *x*.
-        - an {{MLActivation}}. The activation function representing the clamp operation.
+        - an {{MLOperand}}. The output tensor of the same shape as *input*.
+        - an {{MLActivation}}. The operator representing the clamp operation.
+
+    The {{MLGraphBuilder/clamp()}} method steps are:
+    1. Let |arg1| be the first argument.
+    1. If |arg1| is an instance of {{MLOperand}}, run the {{MLGraphBuilder/clamp(operand, options)}} steps with all the received arguments.
+    1. Otherwise, run the {{MLGraphBuilder/clamp(options)}} steps with all received arguments.
```
zolkis, line 1233: @huningxin : in general, for polymorphic functions, do we prefer adding these boilerplate lines, or just define the two forms separately?

index.bs
```diff
@@ -1249,7 +1239,57 @@ partial interface MLGraphBuilder {
...
+<div algorithm=clamp-options>
+    The {{MLGraphBuilder/clamp(options)}} method steps are:
+    1. Let |options| be the first argument.
+        1. If running the <a>check clamp options</a> steps with |options| returns `false`, then throw a "{{TypeError}}" {{DOMException}} and abort these steps.
+    1. Let |op| be the result of invoking the <a>create MLActivation</a> steps with `"clamp"` and |options|.
+        1. If that throws an error, re-throw the error and abort these steps.
+    1. Make a request to the underlying platform to connect |op| with the [=implementation-defined=] platform operator for clamp, and store a reference to the resulting [=implementation-defined=] platform operator object in |op|.{{MLActivation/[[operator]]}}.
+        <div class="note">
+            This step does not execute the clamp operation yet, it just prepares the operator.
+        </div>
```
wchao1115, line 1290: I don't think this note is needed because the same statement can also be made for all the other graph builder methods that add a given operation into the graph i.e. nothing is executed yet.

wchao1115 (changes requested):

index.bs
zolkis: Indeed it would be better to state that once for all the ops and remove the repetitive notes.

zolkis: Removed.

zolkis:
@wchao1115 concern addressed, PTAL.

zolkis:
@huningxin @wchao1115 PTAL: squashed to one commit.

index.bs
```diff
@@ -1408,7 +1445,54 @@ partial interface MLGraphBuilder {
...
+    **Returns:**
+        - an {{MLOperand}}. The output tensor of the same shape as *operand*.
+</div>
+<div algorithm=clamp>
+    The {{MLGraphBuilder/clamp(operand, options)}} method steps are:
+    1. Let |operand| be the first argument.
+    1. Let |options| be the second argument.
+        1. If running the <a>check clamp options</a> steps with |options| returns `false`, then throw a "{{TypeError}}" {{DOMException}} and abort these steps.
+    1. Let |result| be the result of invoking the <a>copy MLOperand</a> steps given |operand|.
+            1. If that throws an error, re-throw the error and abort these steps.
```
huningxin, line 1472: We may need to create the platform operand for |result|:
1. Make a request to the underlying platform to create an [=implementation-defined=] platform operand with |result|.{{MLOperand/[[descriptor]]}} and store its reference in |result|.{{MLOperand/[[operand]]}}.

index.bs
```diff
@@ -1408,7 +1445,54 @@ partial interface MLGraphBuilder {
...
+    **Returns:**
+        - an {{MLActivation}}. The operator representing the clamp operation.
+</div>
+<div algorithm=clamp-options>
+    The {{MLGraphBuilder/clamp(options)}} method steps are:
+    1. Let |options| be the first argument.
+        1. If running the <a>check clamp options</a> steps with |options| returns `false`, then throw a "{{TypeError}}" {{DOMException}} and abort these steps.
+    1. Let |op| be the result of invoking the <a>create MLActivation</a> steps with `"clamp"` and |options|.
+        1. If that throws an error, re-throw the error and abort these steps.
+    1. Make a request to the underlying platform to connect |op| with the [=implementation-defined=] platform operator for clamp, and store a reference to the resulting [=implementation-defined=] platform operator object in |op|.{{MLActivation/[[operator]]}}.
```
huningxin, line 1493: I suppose we can leverage the steps of {{MLGraphBuilder/clamp(operand, options)}} method. The difference is here it doesn't need register input and output operand.

index.bs
```diff
@@ -754,6 +754,16 @@ To <dfn>create MLOperand</dfn> given |builder| and |desc|, run the following ste
     1. Return |operand|.
 </div>
 
+To <dfn>copy MLOperand</dfn> given |operand|, run the following steps:
+<div class=algorithm-steps>
+    1. If |operand| is not an instance of {{MLOperand}}, then throw a "{{TypeError}}" and stop.
+    1. Let |result| be a new [=object=].
+    1. Set |result|.{{MLOperand/[[builder]]}} to |operand|.{{MLOperand/[[builder]]}}.
+    1. Set |result|.{{MLOperand/[[descriptor]]}} to |operand|.{{MLOperand/[[descriptor]]}}.
+    1. Set |result|.{{MLOperand/[[name]]}} to |operand|.{{MLOperand/[[name]]}}.
```
huningxin, line 763: {{MLOperand/[[name]]}} is only valid for input operand, it should not be copied if the destination operand is not an input.

index.bs
```diff
@@ -1408,7 +1445,54 @@ partial interface MLGraphBuilder {
...
+    **Arguments:**
+        - *operand*: an {{MLOperand}}. The input tensor.
+        - *options*: an optional {{MLClampOptions}}. The optional parameters of the operation.
+            - *minValue*: a {{float}} scalar. Specifies the minimum value of the range. When it is not specified, the clamping is not performed on the lower limit of the range.
+            - *maxValue*: a {{float}} scalar. Specifies the maximum value of the range. When it is not specified, the clamping is not performed on the upper limit of the range.
+    **Returns:**
+        - an {{MLOperand}}. The output tensor of the same shape as *operand*.
+</div>
+<div algorithm=clamp>
+    The {{MLGraphBuilder/clamp(operand, options)}} method steps are:
```
huningxin, line 1467: `operand` is commonly used, e.g. for MLOperand or platform operand etc. Should the variable name be more specific, like `input` or `x`?

index.bs
```diff
@@ -769,16 +779,53 @@ Objects implementing the {{MLActivation}} interface represent activation functio
...
+    To <dfn>create MLActivation</dfn> given |builder|, |name| and |options|, run the following steps:
+  </summary>
+  <div class=algorithm-steps>
+    1. If |builder| is not an instance of {{MLGraphBuilder}}, throw a "{{TypeError}}" and abort these steps.
+    1. If |name| is `undefined` or `null`,  throw a "{{TypeError}}" and abort these steps.
+    1. Let |activation| be a new [=object=].
+    1. Set |activation|.{{MLActivation/[[builder]]}} to |builder|.
+    1. Set |activation|.{{MLActivation/[[name]]}} to |name|.
+    1. If |options| is an [=object=], set |activation|.{{MLActivation/[[options]]}} to |options|.
+    1. Make a request to the underlying platform to bind the [=implementation-defined=] platform operator for |name| to |activation|.{{MLActivation/[[operator]]}}.
```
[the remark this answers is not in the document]
zolkis: I agree. The step says the platform operator is implementation defined. Should I change the prose?

index.bs
```diff
@@ -754,6 +754,16 @@ To <dfn>create MLOperand</dfn> given |builder| and |desc|, run the following ste
     1. Return |operand|.
 </div>
 
+To <dfn>copy MLOperand</dfn> given |operand|, run the following steps:
+<div class=algorithm-steps>
+    1. If |operand| is not an instance of {{MLOperand}}, then throw a "{{TypeError}}" and stop.
+    1. Let |result| be a new [=object=].
+    1. Set |result|.{{MLOperand/[[builder]]}} to |operand|.{{MLOperand/[[builder]]}}.
+    1. Set |result|.{{MLOperand/[[descriptor]]}} to |operand|.{{MLOperand/[[descriptor]]}}.
```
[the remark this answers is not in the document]
zolkis: All internal slots should be immutable, i.e. only changeable with API calls, so assignments to internal slots should involve deep copies by default. I will check this.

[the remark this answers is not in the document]
zolkis: [ECMAScript](https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots) doesn't really shed light on this. Internal slots are supposed to be created and then not modified unless specified differently.

Now we seem to have 2 options:
1. change every line that sets an internal slot to replace `the argument` with "a copy of `the argument`"
2. or, create "set the internal slot `[[name]]` to `<arg>`" steps, where we say "set `[[name]]` to a copy of `<arg>`".

@anssiko @dontcallmedom  any input here?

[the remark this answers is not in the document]
zolkis: @huningxin 
Since this is a general issue for the spec, could we create an issue and fix it outside this PR, in all affected places?

index.bs
zolkis: To me that seems unnecessary to specify, since it seems to be an internal choice of the implementation. But let's try to include it as a start.

index.bs
zolkis: This is for activation and let's try its own steps. Let's keep track if we could reuse these operand and activation platform-related steps in other methods and how to parametrize them. If there is an opportunity, we can factor them out later.

index.bs
zolkis: Oh, so that was the reason for using `x`. Well, `input` is also a loaded name :). 
Since it is an MLOperand, I think using `operand` is just fine.

index.bs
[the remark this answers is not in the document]
huningxin: I am fine to fix it in a separate issue. Please create that issue and add link into the spec. Thanks.

huningxin (approved):
LGTM with a follow-up for "deep-copy" issue.

zolkis:
Moved to #399 with new [target branch](https://github.com/webmachinelearning/webnn/tree/zk-conventions-integration).

PR #355, opened 2023-02-28
Use static padding values for `pad` operation

### LINKED ISSUE #354: Use static padding values for `pad` operation
The current WebNN [`pad`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-pad) operation declares `padding` parameter as an `MLOperand` whose values may be dynamic at runtime, e.g., the compute result of another operation.

```js
  MLOperand pad(MLOperand input, MLOperand padding, optional MLPadOptions options = {});
};
```

Although this definition is close to framework, such as ONNX [`Pad`](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Pad) op, it is not widely available on native ML APIs. This may lead to complex implementation if the native ML APIs only support static padding values. 

For example, DirectML [`DML_PADDING1_OPERATOR_DESC`](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_padding1_operator_desc) defines `startPadding` and `endPadding` as two unsigned integer arrays.
```c++
struct DML_PADDING1_OPERATOR_DESC {
  const DML_TENSOR_DESC *InputTensor;
  const DML_TENSOR_DESC *OutputTensor;
  DML_PADDING_MODE      PaddingMode;
  DML_TENSOR_DATA_TYPE  PaddingValueDataType;
  DML_SCALAR_UNION      PaddingValue;
  UINT                  DimensionCount;
  const UINT            *StartPadding;
  const UINT            *EndPadding;
};
```

XNNPACK [`xnn_define_static_constant_pad()`](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#LL924C17-L924C47) defines `pre_paddings` and `post_paddings` as two `size_t` arrays.
```c++
enum xnn_status xnn_define_static_constant_pad(
  xnn_subgraph_t subgraph,
  const size_t* pre_paddings,
  const size_t* post_paddings,
  float padding_value,
  uint32_t input_id,
  uint32_t output_id,
  uint32_t flags);
```

MPSGraph [`padTensor`](https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/3580502-padtensor?language=objc) also defines `lefPadding`NSArray<NSNumber *>`
```c++
(MPSGraphTensor *)padTensor:(MPSGraphTensor *)tensor 
              withPaddingMode:(MPSGraphPaddingMode)paddingMode 
                  leftPadding:(MPSShape *)leftPadding 
                 rightPadding:(MPSShape *)rightPadding 
                constantValue:(double)constantValue 
                         name:(NSString *)name;
```

The proposal is to use static padding values for WebNN `pad` operation that would simplify the implementation on native ML APIs, such as
```js
  MLOperand pad(MLOperand input,
                sequence<unsigned long> beginningPadding,
                sequence<unsigned long> endingPadding,
                optional MLPadOptions options = {});
};
```

The ML framework could map to WebNN static padding values if the padding operand is a constant and still support dynamic padding values by its own kernels.

Any thoughts?

huningxin:
fix #354

@wchao1115 @wacky6 @anssiko @pyu10055 @RafaelCintron , PTAL. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/355.html" title="Last updated on Mar 1, 2023, 2:00 AM UTC (91547d4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/355/14c6670...huningxin:91547d4.html" title="Last updated on Mar 1, 2023, 2:00 AM UTC (91547d4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2353,55 +2353,61 @@ dictionary MLPadOptions {
 };
 
 partial interface MLGraphBuilder {
-  MLOperand pad(MLOperand input, MLOperand padding, optional MLPadOptions options = {});
+  MLOperand pad(MLOperand input,
+                sequence<unsigned long> beginningPadding,
+                sequence<unsigned long> endPadding,
+                optional MLPadOptions options = {});
 };
 </script>
 <div algorithm=pad>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input tensor.
-        - *padding*: an {{MLOperand}}. The 2-D Tensor of integer values indicating the number of padding values to add at the beginning and end of each input dimensions. The tensor has shape [*n*, 2] where *n* is the rank of the input tensor. For each dimension *D* of *input*, *padding[D, 0]* indicates how many values to add before the content in that dimension, and *padding[D, 1]* indicates how many values to add after the content in that dimension.
+        - *beginningPadding*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the number of padding values to add at the beginning of each input dimension, of length *N* where *N* is the rank of the input tensor. For each dimension *d* of *input*, *beginningPadding[d]* indicates how many values to add before the content in that dimension.
+        - *endPadding*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the number of padding values to add at the end of each input dimension, of length *N* where *N* is the rank of the input tensor. For each dimension *d* of *input*, *endPadding[d]* indicates how many values to add after the content in that dimension.
         - *options*: an optional {{MLPadOptions}}. The optional parameters of the operation.
             - *mode*: an {{MLPaddingMode}}. The different ways to pad the tensor. When not set, it's assumed to be "constant".
             - *value*: a {{float}}. The pad value when the *options.mode* is set to *"constant"*. When not set, it's assumed to be 0.
 
-    **Returns:** an {{MLOperand}}. The padded output tensor.
+    **Returns:** an {{MLOperand}}. The padded output tensor. Each dimension of the output tensor can be calculated as follow:
+
+    *output size = beginning padding + input size + ending padding*
+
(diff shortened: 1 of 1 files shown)


anssiko (approved):
Thanks!

FYI @zolkis can also take a look considering his editorial improvement work in flight. We should land this using the existing arguments/returns convention and @zolkis will make any editorial updates on top of this in a separate PR when that is timely.

zolkis:
Left a non-substantial comment in #354.

wchao1115:
@huningxin I think we defined `pad` the way it is b/c ONNX supports tensor padding (see [Pad](https://onnx.ai/onnx/operators/onnx__Pad.html#pad)). Do you have the use case that requires it to work on static array? If so, should we allow for both in the definition instead of removing the tensor padding option?

huningxin:
@wchao1115 

> Do you have the use case that requires it to work on static array?

Yes. ONNX [Fast Neural Style Transfer models](https://github.com/onnx/models/tree/main/vision/style_transfer/fast_neural_style#fast-neural-style-transfer) and TFLite [Pose Landmarks Detection models](https://google.github.io/mediapipe/solutions/models.html#pose) use static padding values.

> If so, should we allow for both in the definition instead of removing the tensor padding option?

The issue is, as my [survey](https://github.com/webmachinelearning/webnn/issues/354#issue-1601135802), the tensor padding is not widely available on native ML APIs currently. The pad of DirectML, XNNPACK and MPSGraph only support static padding values in arrays.

So, we probably should start with static padding values and add tensor padding options back later.

WDYT?

wchao1115 (approved):

huningxin:
Thanks for the review and approvals. I am going to merge this PR.

PR #357, opened 2023-02-28
Add prelu (Parametric ReLU) into the operation list

### LINKED ISSUE #356: Propose to add Parametric ReLU into the operation list
### Overview

[Parametric ReLU](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)#Parametric_ReLU) (PRelu) is a type of leaky ReLU that, instead of having a scalar slope like 0.01, making the slope (coefficient of leakage) into a parameter that is learned along with the other neural-network parameters:
```
f(x, slope) = x,         if x >= 0
              slope * x, otherwise
```

WebNN already supports [`leakyRelu`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-leakyrelu).

### Use case
PRelu is a widely used activation function, in particular it is useful for WebNN [Facial Landmark Detection](https://www.w3.org/TR/webnn/#usecase-facial-landmarks) use case. For example, MediaPipe [Face Mesh](https://google.github.io/mediapipe/solutions/models.html#face-mesh) models use PRelu.

### Performance consideration
Regarding to WebNN [Guidelines for new operations](https://www.w3.org/TR/webnn/#security-new-ops), 
> If an operation can be decomposed to low level primitives:

PRelu can be decomposed to low level primitives

> Add an informative emulation path

The emulation sample code may be
```js
return builder.add(builder.max(builder.constant(0), x),
                   builder.mul(slope, builder.min(builder.constant(0), x)));
```

> Prefer primitives over new high level operations but consider performance consequences

PRelu can be implemented by more efficient implementation of native ML APIs, such as 

* XNNPACK [`xnn_define_prelu()`](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#LL1164C17-L1164C33) 
* DirectML [`DML_ACTIVATION_PARAMETERIZED_RELU_OPERATOR_DESC`](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_activation_parameterized_relu_operator_desc)
* MPSGraph [`leakyReLUWithTensor`](https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/3867066-leakyreluwithtensor?language=objc) (defined as a variant of leakyRelu)

therefore its usage is encouraged from the performance standpoint.

huningxin:
fix #356

@wchao1115 @pyu10055 @anssiko @wacky6 @RafaelCintron , PTAL. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/357.html" title="Last updated on Feb 28, 2023, 6:09 AM UTC (d0686bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/357/14c6670...huningxin:d0686bc.html" title="Last updated on Feb 28, 2023, 6:09 AM UTC (d0686bc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2479,6 +2479,34 @@ partial interface MLGraphBuilder {
     </div>
 </div>
 
+### The prelu() method ### {#api-mlgraphbuilder-prelu}
+Calculate the <a href="https://en.wikipedia.org/wiki/Rectifier_(neural_networks)#Parametric_ReLU">parametric version of rectified linear function (Parametric Relu)</a> on the input tensor element-wise. Parametric Relu is a type of leaky ReLU that, instead of having a scalar slope like 0.01, making the slope (coefficient of leakage) into a parameter that is learned during the model training phase of this operation. The calculation follows the expression `max(0, x) + slope ∗ min(0, x)`.
+<script type=idl>
+partial interface MLGraphBuilder {
+  MLOperand prelu(MLOperand x, MLOperand slope);
+};
+</script>
+<div algorithm=prelu>
+    **Arguments:**
+        - *x*: an {{MLOperand}}. The input tensor.
+        - *slope*: an {{MLOperand}}. The slope tensor. Its shape is either the same as, or unidirectionally broadcastable to the shape of input tensor *x* according to [[!numpy-broadcasting-rule]].
+
+    **Returns:**
+        - an {{MLOperand}}. The output tensor of the same shape as *x*.
+
+    <div class="note">
+    The behavior of this operation can be generically emulated from the usage of
+    other operations as follow. However, user agents typically have a more
+    efficient implementation for it, therefore its usage is encouraged from the
+    performance standpoint.
+    <pre highlight="js">
+    return builder.add(builder.max(builder.constant(0), x),
+                       builder.mul(slope, builder.min(builder.constant(0), x)));
+    </pre>
+    </div>
+</div>
+
+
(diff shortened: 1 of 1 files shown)


anssiko (approved):
Given good use case and adequate native ML API support this LGTM. There's potential to fit this into the initial CR if we get the required reviews in time.

wchao1115:
Can prelu be one of the fused activation function? If so, you might want to add the `MLActivation` ctor for it as well.

huningxin:
@wchao1115 

> Can prelu be one of the fused activation functions? If so, you might want to add the `MLActivation` ctor for it as well.

I am not sure, so it is not in the PR. 

It seems to me that the current DirectML [activations that are supported for fusion](https://learn.microsoft.com/en-us/windows/ai/directml/dml-fused-activations#activations-that-are-supported-for-fusion) and XNNPACK [xnn_post_operation ](https://github.com/google/XNNPACK/blob/master/src/xnnpack/post-operation.h) don't support `prelu` as fused activation function.

If `prelu` can be fused activation function, the `scope` MLOperand needs to be constant, correct?

Any recommendations?

wchao1115:
Thanks for checking. I think it is fine to leave it out for now.

wchao1115 (approved):

huningxin:
Thanks for review and approvals. I am going to merge this PR.

Issue #317, opened 2022-12-20
The type of MLTransposeOptions.permutation should be unsigned long


lisa0314:
As the spec said, the values in the permutation sequence must be within the range from 0 to N-1. I suppose that the type of MLTransposeOptions.permutation should be unsigned long.

inexorabletash:
Fixed by #359

PR #359, opened 2023-03-03
Use unsigned integer axis

### LINKED ISSUE #345: Use unsigned long for axis of concat operation
```
  MLOperand concat(sequence<MLOperand> inputs, long axis);
};
```
We now limit the value of [axis](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-concat-inputs-axis-axis) in the interval [0, N) where N is the rank of all the inputs. It should be a positive value and we may use `unsigned long` for axis.

huningxin:
fix #345, #317

This PR makes the axis/axes definitions consistent by using the unsigned integer type and valid value range [0, N-1] where N is the tensor rank across the spec.

@wchao1115 @wacky6 @fdwr @anssiko @pyu10055 @RafaelCintron , PTAL. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/359.html" title="Last updated on Mar 4, 2023, 12:35 AM UTC (9781939)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/359/14c6670...huningxin:9781939.html" title="Last updated on Mar 4, 2023, 12:35 AM UTC (9781939)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1158,7 +1158,7 @@ Normalize the tensor values of input features across the batch dimension using [
 dictionary MLBatchNormalizationOptions {
   MLOperand scale;
   MLOperand bias;
-  unsigned long axis = 1;
+  unsigned long axis = 0;
   float epsilon = 1e-5;
   MLActivation activation;
 };
@@ -1176,7 +1176,7 @@ partial interface MLGraphBuilder {
         - *options*: an optional {{MLBatchNormalizationOptions}}. The optional parameters of the operation.
               - *scale*: an {{MLOperand}}. The 1-D tensor of the scaling values whose length is equal to the size of the input dimension denoted by *options.axis*.
               - *bias*: an {{MLOperand}}. The 1-D tensor of the bias values whose length is equal to the size of the input dimension denoted by *options.axis*.
-              - *axis*: an {{unsigned long}} scalar. The index to the feature count dimension of the input shape for which the mean and variance values are. Its value must be in the range [0, N) where N is the rank of input tensor. When it's not specified, the default value is 1.
+              - *axis*: an {{unsigned long}} scalar. The index to the feature count dimension of the input shape for which the mean and variance values are. Its value must be in the range [0, N-1] where N is the rank of input tensor. When it's not specified, the default value is 0.
               - *epsilon*: a {{float}} scalar. A small value to prevent computational error due to divide-by-zero. The default value is 0.00001 when not specified.
               - *activation*: an {{MLActivation}}. The optional activation function that immediately follows the normalization operation.
 
@@ -1263,7 +1263,7 @@ partial interface MLGraphBuilder {
     **Arguments:**
         - *inputs*: a sequence of {{MLOperand}}. All input tensors must have the
             same shape, except for the size of the dimension to concatenate on.
-        - *axis*: an {{unsigned long}} scalar. The axis that the inputs concatenate along. Its value must be in the range [0, N) where N is the rank of input tensors.
+        - *axis*: an {{unsigned long}} scalar. The axis that the inputs concatenate along. Its value must be in the range [0, N-1] where N is the rank of input tensors.
 
     **Returns:** an {{MLOperand}}. The concatenated tensor of all the inputs along
     the *axis*. The output tensor has the same shape except on the dimension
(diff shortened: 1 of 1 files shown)


anssiko (approved):
🚀

fdwr:
@huningxin : Should we split the `MLBatchNormalizationOptions` default into a separate PR for clarity from the signed -> unsigned aspect?

index.bs
```diff
@@ -2862,7 +2860,7 @@ partial interface MLGraphBuilder {
...
         - *input*: an {{MLOperand}}. The input N-D tensor.
         - *options*: an optional {{MLTransposeOptions}}. The optional parameters of the operation.
-            - *permutation*: a sequence of {{long}} values. The values used to permute the output shape. When it's not specified, it's set to `[N-1...0]`, where `N` is the rank of the input tensor. These default values cause the output to become a transposed tensor of the input. When specified, the number of values in the sequence must be the same as the rank of the input tensor, and the values in the sequence must be within the range from 0 to N-1 with no two or more same values found in the sequence.
+            - *permutation*: a sequence of {{unsigned long}} values. The values used to permute the output shape. When it's not specified, it's set to `[N-1...0]`, where `N` is the rank of the input tensor. These default values cause the output to become a transposed tensor of the input. When specified, the number of values in the sequence must be the same as the rank of the input tensor, and the values in the sequence must be within the range from 0 to N-1 with no two or more same values found in the sequence.
```
fdwr, line 2863: How about adding commas here `[N-1, ..., 0]` and/or adding a tiny example? "When it's not specified, it's set to `[N-1...0]`, where `N` is the rank of the input tensor (e.g. [2,1,0] for a 3D tensor)."

fdwr (approved):
Thanks. Minor proposals, but I'm also content as-is. Rescanning the spec, it appears you got all the "long axis" and "sequence<long>" occurrences. 👍

huningxin:
@fdwr 

> @huningxin : Should we split the `MLBatchNormalizationOptions` default into a separate PR for clarity from the signed -> unsigned aspect?

sgtm. I removed the change of `MLBatchNormalizationOptions.axis` default value in the latest commit https://github.com/webmachinelearning/webnn/pull/359/commits/ef94c4e4c74d0f9087e59b567300c925b616f253. Please take another look. Thanks!

index.bs
huningxin: sgtm. I applied this proposal to `MLTransposeOptions.permutation` and also made `MLSliceOptions.axes` aligned in latest commit https://github.com/webmachinelearning/webnn/pull/359/commits/ef94c4e4c74d0f9087e59b567300c925b616f253. Please take another look. Thanks!

fdwr (approved):
LGTM. TY.

wacky6 (approved):

wchao1115 (approved):

huningxin:
Thanks for the review and approvals. With that, I am going to merge this PR.

fdwr:
@huningxin : I should have noticed earlier while reviewing, but I just noticed `slice`s parameter `starts` still has the negative number policy and accepts axis coordinates as `long` rather than `unsigned long`; and its `sizes` parameter has an unusual policy that -1 is special, whereas sizes in resample2D is `sequence<unsigned long>`. Do we want to rationalize these for consistency too, having higher level policy like this determined by the calling framework?

huningxin:
Thanks @fdwr . This PR focused on the axis/axes, so it just changed [`MLSliceOptions.axes`](https://www.w3.org/TR/webnn/#dom-mlsliceoptions-axes) to `sequence<unsigned long>`. For `starts`, because it is actually the offset along each dimension, it was left as-is. There was similar reason for unchanged `sizes`.

However, I agree with you that using `unsigned long` for `starts` and `sizes` may help the API consistency and simplify the implementation. The `slice` op of native ML APIs, like DirectML and XNNPACK, seem to only support unsigned starts and sizes. Another consideration would be whether we should simplify the `pad` by removing `MLSliceOptions.axes`. It can just require the length of `starts` and `sizes` is same as input tensor's rank.

I propose to open a separate issue to track it, WDYT?

fdwr:
@huningxin : 👍 Separate issue - thanks.

>may help the API consistency and simplify the implementation ...

Yeah, we see that every other `sizes` parameter is `unsigned long` (see below), leaving only `slice` as `sequence<long> sizes`.

```
dictionary MLResample2dOptions {
...
  sequence<unsigned long> sizes;
...
};

dictionary MLPool2dOptions {
  ...
  sequence<unsigned long> outputSizes;
};



dictionary MLPool2dOptions {
...
  sequence<unsigned long> outputSizes;
};
```

> Another consideration would be whether we should simplify the pad by removing MLSliceOptions.axes. It can just require the length of starts and sizes is same as input tensor's rank.

🤔 Not sure, as I want to know what XNNPack does first too (DML_SLICE1_OPERATOR_DESC expects all dimensions to be listed explicitly, and unchanged dimensions just use the full range), but I generally lean the direction of being more explicit for WebNN.

huningxin:
@fdwr 

> as I want to know what XNNPack does first too

[`xnn_define_static_slice`](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#L1417) also expects all dimensions to be listed.

> but I generally lean the direction of being more explicit for WebNN.

+1

PR #361, opened 2023-03-09
update pad example

BruceDai:
@huningxin @anssiko PTAL, thanks.


huningxin (approved):
Good catch! Thanks for the fix @BruceDai !

anssiko (approved):
👍

PR #362, opened 2023-03-20
Update arch diagram to remove WebGL

dontcallmedom:
also provide SVG source


anssiko (approved):
Nice work, thanks!

TIL: [Inkscape can convert PNG to SVG](https://linuxhint.com/convert-png-to-svg-inkscape/).

anssiko:
One detail to note: some frameworks such as TF.js still use a [WebGL backend](https://www.tensorflow.org/js/guide/platform_environment#webgl_backend). I think this is illustrated by the arrow pointing downwards toward the WebGL/WebGPU box. For conciseness and future-proofing reasons I'm supportive of removing WebGL from this diagram given WebNN dropped support and is the context here.

Given this is now in an SVG format we can easily maintain this diagram as our architecture evolves.

huningxin:
For the change on macOS/iOS, the platform API is changed from MLCompute to BNNS/MPS, is that suggested by Apple folks?

dontcallmedom:
> For the change on macOS/iOS, the platform API is changed from MLCompute to BNNS/MPS, is that suggested by Apple folks?

Not at all - I'm not sure how I came to have that change in my SVG version of the diagram; I corrected that to MLCompute.

content/webnn_arch.svg
huningxin, line 1: webnn_arch.svg LGTM.

content/webnn_arch.png
huningxin, line 1: You may miss updating content/webnn_arch.png. It still changes ML Compute to BNNS/MPS.

huningxin (approved):
LGTM, thanks @dontcallmedom !

PR #364, opened 2023-03-22
Add the builder.input() method steps

zolkis:
Add the input() method algorithm.
Depends on #337

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/364.html" title="Last updated on May 15, 2023, 6:32 AM UTC (25f8982)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/364/a115073...zolkis:25f8982.html" title="Last updated on May 15, 2023, 6:32 AM UTC (25f8982)">Diff</a>


wchao1115 (approved):

zolkis:
@huningxin PTAL :)

index.bs
```diff
@@ -1149,9 +1149,35 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+<div algorithm=input>
+    The {{MLGraphBuilder/input(name, descriptor)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |name| be the first argument.
+        1. If |name| is `undefined` or an empty [=string=], then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |descriptor| be the second argument.
+        1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1.  If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1171: nit: one extra space before "If"

index.bs
```diff
@@ -1149,9 +1149,35 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1172: What would happen if user code passes a descriptor for scalar operand (`dimensions` is undefined)? This `input()` algorithm seems only to support tensor operand now.

huningxin: Actually, the `constant(descriptor, bufferView)` algorithm in PR #365 should also address this issue.

zolkis: Right. So this is where conventions about `descriptor` become important.
I am thinking if it's enough to distinguish between vector and scalar solely on the existence of a `dimensions` member. But for now let's continue with that working assumption and keep it in mind.

About `constant(descriptor, bufferView)`: the `descriptor.dimensions` must be defined, otherwise it's an error. There is a polymorphic variant for specifying scalar constant. Should we change that?

zolkis: Done, please check if correct. Using an assert seemed better choice than just making a note.
Forgot to fix the white space error, such a small change that I amended the same commit and force-pushed.

index.bs
```diff
@@ -1168,8 +1168,10 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
         1. If |name| is `undefined` or an empty [=string=], then throw a "{{TypeError}}" {{DOMException}} and stop.
     1. Let |descriptor| be the second argument.
         1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
-        1.  If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
-        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1171: "byte length" algorithm also need to handle scalar operand descriptor. Its current definition would break if `descriptor.dimensions` is undefined.

index.bs
```diff
@@ -1168,8 +1168,10 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
...
+        1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input.
```
huningxin, line 1172: Move this assertion before byte length check?

index.bs
huningxin: > About constant(descriptor, bufferView): the descriptor.dimensions must be defined, otherwise it's an error.

That's fine. No change needed.

index.bs
zolkis: Actually I moved the byte length check to the vector section.
If we need to check the scalar, that would be an additional step.

huningxin: Sounds good.

huningxin:
@zolkis , you may also want to solve the conflicts, squash the commits and edit the commit log before merging. Thanks!

zolkis:
@huningxin 
I rebased, squashed, fixed conflicts, verified with build, then force pushed. 
Since there were a lot of other commits in between filing the PR and now, those commits appear here now.
Needed to add one more commit to merge with the main branch.

PR #365, opened 2023-03-22
Add the constant() method steps.

zolkis:
Add the constant() algorithms.
This another take on how to specify polymorphic functions (compare that with clamp() in #348 
I like this approach more.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/365.html" title="Last updated on May 14, 2023, 7:26 PM UTC (dad23d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/365/b0fb131...zolkis:dad23d1.html" title="Last updated on May 14, 2023, 7:26 PM UTC (dad23d1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -695,7 +695,54 @@ For instance, an {{MLOperand}} may represent a constant feeding to an operation 
 interface MLOperand {};
 </script>
 
-See also [[#security-new-ops]]
+{{MLOperand}} has the following internal slots:
+<div class=internal-slots>
+<dl dfn-type=attribute dfn-for="MLOperand">
+    : <dfn>\[[builder]]</dfn> of type {{MLGraphBuilder}}
+    ::
+        The {{MLOperand}}'s associated builder object.
+
+    : <dfn>\[[descriptor]]</dfn> of type {{MLOperandDescriptor}}
+    ::
+        The {{MLOperand}}'s descriptor.
+
+    : <dfn>\[[name]]</dfn> of type [=string=]
+    ::
+        The {{MLOperand}}'s name (only for input operands).
+
+    : <dfn>\[[operand]]</dfn> of type [=object=]
+    ::
+        Reference to {{MLOperand}}'s corresponding [=implementation-defined=] platform operand object.
+
+    : <dfn>\[[operator]]</dfn> of type [=object=]
+    ::
+        Reference to {{MLOperand}}'s corresponding [=implementation-defined=] platform operator object.
+</dl>
+</div>
+
+Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/constructor()}} constructor to an {{MLContext}} object, an {{MLOperand}} is also always bound to the same {{MLContext}} object.
+
+#### Creating {{MLOperand}} #### {#api-mloperand-create}
+The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, internally using the following algorithms.
+
+To <dfn>create MLOperand</dfn> given |builder| and |desc|, run the following steps:
+<div class=algorithm-steps>
+    1. If |builder| is not an instance of {{MLGraphBuilder}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. If |desc| is not an [=object=] that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |operand| be a new [=object=].
+    1. Set |operand|.{{MLOperand/[[builder]]}} to |builder|.
+    1. Set |operand|.{{MLOperand/[[descriptor]]}} to |desc|.
+    1. Return |operand|.
+</div>
+
+To <dfn>check dimensions</dfn> given |dimensions| and |type|, run the following steps:
+<div class=algorithm-steps>
+    1. If |dimensions| is not an array of positive numbers, return `false`;
+    1. If |dimensions|.length is 0, return `false`.
+    1. If |dimensions|.length is too large to be supported by the implementation, return `false`.
+    1. If any element of |dimensions| is not a positive number, or it is too large to be supported by the implementation given |type|, return `false`.
+    1. Return `true`.
+</div>
 
 ### The MLActivation interface ### {#api-mlactivation}
 
@@ -829,6 +876,13 @@ partial interface MLContext {
             1. Else:
                 1. Set the values of |value| to the values of |outputTensor|.
     1. Return {{undefined}}.
+</div>
+
+<div class=algorithm-steps>
+To <dfn>validate buffer with descriptor</dfn> given |bufferView| and |descriptor|, run the following steps:
+    1. If |bufferView| is not an {{MLBufferView}}, return `false`.
+    1. If |bufferView|'s [=element type=] does not match to |descriptor|.{{MLOperandDescriptor/type}}  according to [this table](#appendices-mloperandtype-arraybufferview-compatibility), return `false`.
+    1. If |bufferView|.\[[ByteLength]] is not equal to the [=byte length=] of |descriptor|, return `false`.
 </div>
 
 #### Examples #### {#api-mlcontext-sync-execution-examples}
@@ -1035,7 +1089,7 @@ partial interface MLCommandEncoder {
 </div>
 
 <div class="note">
-Graph initialization stage typically involves a process known as "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. The initializing inputs are typically the constant weight data specified through the {{MLGraphBuilder}}.{{MLGraphBuilder/constant(desc, bufferView)}} method as constant operands during graph construction time.
+Graph initialization stage typically involves a process known as "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. The initializing inputs are typically the constant weight data specified through the {{MLGraphBuilder/constant(descriptor, bufferView)|MLGraphBuilder/constant()}} method as constant operands during graph construction time.
 </div>
 
 ### Dispatch Execution Commands ### {#api-mlcommandencoder-dispatch-commands}
@@ -1119,7 +1173,7 @@ interface MLGraphBuilder {
   MLOperand input(DOMString name, MLOperandDescriptor desc);
 
   // Create an operand for a graph constant.
-  MLOperand constant(MLOperandDescriptor desc, MLBufferView bufferView);
+  MLOperand constant(MLOperandDescriptor descriptor, MLBufferView bufferView);
 
   // Create a single-value operand from the specified number of the specified type.
   MLOperand constant(double value, optional MLOperandType type = "float32");
@@ -1149,8 +1203,67 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
 1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" {{DOMException}} and abort these steps.
 1. Let |context| be the first argument.
 1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a "{{TypeError}}" and abort these steps.
-
 1. Set {{MLGraphBuilder/[[context]]}} to |context|.
+
+### The constant() method  ### {#api-mlgraphbuilder-constant-method}
+Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
+
+#### The {{MLGraphBuilder/constant(descriptor, bufferView)}} method  #### {#api-mlgraphbuilder-constant}
+<div class="note">
+    **Arguments:**
+        - *descriptor*: an {{MLOperandDescriptor}} object
+        - *bufferView*: an {{MLBufferView}}
+    **Returns:**: an {{MLOperand}} object.
+</div>
+
+<div algorithm=constant class=algorithm-steps>
+The {{MLGraphBuilder/constant(descriptor, bufferView)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |descriptor| be the first argument.
+    1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1. If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |bufferView| be the second argument.
+        1. If invoking <a>validate buffer with descriptor</a> given |bufferView| and |descriptor| return `false`, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=] and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Let |bytes| be the result of invoking [[=get a copy of the bytes held by the buffer source=]] given |bufferView|.
+    1. Make a request to the underlying platform to register |operand| as a tensor constant with |bytes| as value and store a reference to the corresponding [=implementation-defined=] object to |operand|.{{MLOperand/[[operand]]}}.
+        1. If that fails, throw an "{{OperationError}}" {{DOMException}} and stop.
+    1. Return |operand|.
+</div>
+
+#### The {{MLGraphBuilder/constant(value, type)}} method  #### {#api-mlgraphbuilder-constant-value-type}
+
+<div class="note">
+    **Arguments:**
+        - *value*: a number
+        - *type*: an optional {{MLOperandType}}, by default *"float32"*.
+    **Returns:**: an {{MLOperand}} object.
+</div>
+
+<div algorithm=constant-value-type class=algorithm-steps>
+The {{MLGraphBuilder/constant(value, type)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. If |value| is not a [=number=], then then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. If |type| is `undefined`, let |type| be `"float32"`.
+    1. Otherwise, if |type| is not one of {{MLOperandType}}, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |descriptor| be a new {{MLOperandDescriptor}}.
+        1. Set |descriptor|.{{MLOperandDescriptor/type}} to |type|.
+        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `undefined`.
+            <div class="note">
+                In the case of a scalar constant, |descriptor|.{{MLOperandDescriptor/dimensions}} is ignored.
+            </div>
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=] and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Make a request to the underlying platform to register |operand| as a scalar constant with |value| as value and store a reference of the [=implementation-defined=] platform object for the corresponding (scalar or tensor constant) operand to |operand|.{{MLOperand/[[operand]]}}.
+        1. If that throws, re-throw the error and stop.
+    1. Return |operand|.
+</div>
 
 ### The batchNormalization() method ### {#api-mlgraphbuilder-batchnorm}
 Normalize the tensor values of input features across the batch dimension using [[Batch-Normalization]]. For each input feature, the mean and variance values of that feature supplied in this calculation as parameters are previously computed across the batch dimension of the input during the model training phase of this operation.


index.bs
```diff
@@ -1149,9 +1149,62 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
 1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" {{DOMException}} and abort these steps.
 1. Let |context| be the first argument.
 1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a "{{TypeError}}" and abort these steps.
-
 1. Set {{MLGraphBuilder/[[context]]}} to |context|.
 
+### The constant() method  ### {#api-mlgraphbuilder-constant-method}
```
dontcallmedom, line 1154:
```suggestion
### The <dfn method for='MLGraphBuilder'>constant()</dfn> method
```

index.bs
```diff
@@ -1035,7 +1035,7 @@ partial interface MLCommandEncoder {
 </div>
 
 <div class="note">
-Graph initialization stage typically involves a process known as "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. The initializing inputs are typically the constant weight data specified through the {{MLGraphBuilder}}.{{MLGraphBuilder/constant(desc, bufferView)}} method as constant operands during graph construction time.
+Graph initialization stage typically involves a process known as "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. The initializing inputs are typically the constant weight data specified through the <a href="#api-mlgraphbuilder-constant-method">constant()</a> method as constant operands during graph construction time.
```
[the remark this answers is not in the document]
zolkis: Thanks Dom. In this case I used this form because the function is polymorphic and we have multiple versions. Will need to experiment a bit - and the same problem will be there will all the polymorphic functions.

index.bs
```diff
@@ -1149,9 +1149,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1.  If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |bufferView| be the second argument.
+        1. If |bufferView| is not an {{MLBufferView}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1. If |descriptor|.{{MLOperandDescriptor/type}} is not equal to |bufferView|'s [=element type=], throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=byte length=] of |descriptor| is not equal with |bufferView|.\[[ByteLength]], throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=], `"constant"` and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Make a request to the underlying platform to register |operand| as a constant with |bufferView| as value.
```
huningxin, line 1179: I suppose the bytes of `|bufferView|` should be used when request to the underlying platform. Should we use [get a copy of the bytes](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy) algorithm of WebIDL?

index.bs
```diff
@@ -1149,9 +1149,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        - *bufferView*: an {{MLBufferView}}
+    **Returns:**: an {{MLOperand}} object.
+</div>
+<div algorithm=constant>
+    The {{MLGraphBuilder/constant(descriptor, bufferView)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |descriptor| be the first argument.
+    1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
```
huningxin, line 1170: Should the validation of `MLOperandDescriptor` be extracted as an algorithm? I suppose it would be reused by `MLGraphBuilder.input()` as well.

index.bs
```diff
@@ -1149,9 +1149,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1.  If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |bufferView| be the second argument.
```
huningxin, line 1173: Should the validation of `|bufferView|` be extracted as an algorithm? I suppose it would be reused by graph compute steps "validate graph resources" defined in https://github.com/webmachinelearning/webnn/pull/329.

index.bs
```diff
@@ -1149,9 +1149,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1. If |bufferView| is not an {{MLBufferView}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1. If |descriptor|.{{MLOperandDescriptor/type}} is not equal to |bufferView|'s [=element type=], throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=byte length=] of |descriptor| is not equal with |bufferView|.\[[ByteLength]], throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=], `"constant"` and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Make a request to the underlying platform to register |operand| as a constant with |bufferView| as value.
+        1. If that fails, throw an "{{OperationError}}" {{DOMException}} and stop.
+    1. Return |operand|.
```
huningxin, line 1181: I suppose there should be an internal slot of `MLOperand` that represents the platform operand object. Otherwise, when this `MLOperand` be passed to an operator build method, e.g. `conv2d(input, filter)` where filter is a constant `MLOperand`, the algorithm of `conv2d` would refer to the corresponding platform operand object of `filter` to create a `conv2d` platform operator object.

index.bs
```diff
@@ -1149,9 +1149,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |value| be the first argument.
+        1. If |value| is not a number, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |type| be the second argument.
+        1. If |type| is `undefined`, let |type| be `"float32"`.
+        1. Otherwise, if |type| is not one of {{MLOperandType}}, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |descriptor| be a new {{MLOperandDescriptor}}.
+        1. Set |descriptor|.{{MLOperandDescriptor/type}} to |type|.
+        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `{ 1 }`.
```
huningxin, line 1203: Set dimensions to `[ 1 ]`?

index.bs
huningxin: Cross refer to https://github.com/webmachinelearning/webnn/pull/337/files#r1139520188

huningxin:
Does it make sense to pick the dependent definitions of https://github.com/webmachinelearning/webnn/pull/337, such as `MLOperand` internal slots and `create MLOperand` algorithm definitions, and merge them into this PR?

That would share a clear usage of the internal slots and algorithm definitions for the interface method steps. And that hopefully would make the PR easier for review.

WDYT?

zolkis:
Well, all of the other PRs depend on #337, so I suggest we review and merge #337 first. Then I can rebase. 
It has been "tested" by so many PRs, including polymorphic functions / activations, so it's quite stable from my p.o.v.
It is also very hard to add new algorithms with these dependencies still hanging out. I need to test everything in a separate integration repository.
For me it doesn't make much sense to pull #337 in every other PR and leave it open, since it will create a merge hell if I need to touch that part elsewhere.

What I suggest is to clean up and merge the open PRs from bottom up. Let's try to clean up the PRs that are open now, sync-async exec, operand-activation-slots, input, constant, clamp, concat, batch norm.

I will update them today with the suggestions above (factor out more algorithms), but then we should clean up and merge these.

In a week I need to publish PRs for the rest of the algorithms, but will hold them until we could merge the ones currently open.

index.bs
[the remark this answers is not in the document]
dontcallmedom: unless I'm mistaken, bikeshed allows to refer to a specific version of a polymorphic function by specifying the parameters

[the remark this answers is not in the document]
zolkis: But what if I want to refer to the section, not one of the versions? Since here we mean either version.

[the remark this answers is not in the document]
dontcallmedom: ah, I understand now - then your existing approach is indeed probably the best, sorry for the noise :)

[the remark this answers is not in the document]
zolkis: With the `{{ x | y }}` notation, I can put in `x` _one_ of the versions and `y` will be shown. But it's not what I want: I'd like to link the section instead. 

So far I figured out 2 choices: the one above using a `<a href=#section-name>` tag, and the one below:
- make a dfn tag in the section title
  `### The <dfn>MLGraphBuilder/constant()</dfn> method  ### {#api-mlgraphbuilder-constant-method}`
- and use `<a>MLGraphBuilder/constant()</a>`

Both are manual linking, and what I used is more natural and versatile.

But if I must use autolinking, I can use this notation, linking to the version first described in the spec:
`{{MLGraphBuilder/constant(descriptor, bufferView)|MLGraphBuilder/constant()}}`

For lack of better, and for making peace, I am going to use that for now.

index.bs

index.bs
zolkis: OK. I will define that in #329 and use it here (with broken dependency until that one is merged).

zolkis: Done.

index.bs
zolkis: Under consideration how to do it correctly... I will come up with something today.

index.bs

index.bs
zolkis: OK, pushed one version, let's see how do you like it.

index.bs
```diff
@@ -1149,9 +1149,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+    **Returns:**: an {{MLOperand}} object.
+</div>
+<div algorithm=constant>
+    The {{MLGraphBuilder/constant(descriptor, bufferView)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |descriptor| be the first argument.
+    1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1.  If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1171: nit: one more space between "1." and "If".

index.bs
```diff
@@ -1149,9 +1149,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1172: If `check dimensions` definition in #337 can be pulled into this PR, it would be much helpful.

index.bs
```diff
@@ -1149,9 +1149,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+    1. Let |bufferView| be the second argument.
+        1. If invoking <a>validate buffer with descriptor</a> given |bufferView| and |descriptor| return `false`, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=], `"constant"` and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Let |buffer| be the result of invoking [[=get a copy of the bytes held by the buffer source=]] given |bufferView|.
```
huningxin, line 1177: Would `|bytes|` be a better variable name? It is used by WebIDL [get a copy of the bytes held by the buffer source](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy) algorithm and it is distinguished from the underlying array buffer of the `|bufferView|`.

index.bs
```diff
@@ -1149,9 +1149,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |descriptor| be the first argument.
+    1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1.  If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |bufferView| be the second argument.
+        1. If invoking <a>validate buffer with descriptor</a> given |bufferView| and |descriptor| return `false`, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=], `"constant"` and |descriptor|.
```
huningxin, line 1175: If `create MLOperand` definition in #337 can be pulled into this PR, it would be very helpful.

index.bs
```diff
@@ -1149,9 +1149,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1. If that throws, re-throw the exception and stop.
+    1. Let |buffer| be the result of invoking [[=get a copy of the bytes held by the buffer source=]] given |bufferView|.
+    1. Make a request to the underlying platform to register |operand| as a constant with |buffer| as value and store a reference to the corresponding [=implementation-defined=] object to |operand|.{{MLOperand/[[operand]]}}.
```
huningxin, line 1178: If `MLOperand` internal slots definition in #337 can be pulled into this PR, it would be helpful.

index.bs
huningxin: We can do it later when adding `MLGraphBuilder.input()`.

huningxin:
@zolkis 

> Well, all of the other PRs depend on https://github.com/webmachinelearning/webnn/pull/337, so I suggest we review and merge https://github.com/webmachinelearning/webnn/pull/337 first.

That's actually the issue, at least to me. It means that I cannot ensure #337 is correct before I review all the other PRs depending on it.

> What I suggest is to clean up and merge the open PRs from bottom up

Instead, I feel a top-down approach might be helpful. 

For example, in this PR, `MLGraphBuilder.constatnt()` steps clearly need to validate `MLOperandDescriptor`, create `MLOperand` and set platform operand to its internal slot. So, if these internal slot and algorithm definitions are included in one PR, it would make the review to be self-contained. Once approved, it can be merged without any dependencies. That would give you a good base for other `MLOperand` interfaces, like `MLGraphBuilder.input()`, instead of chaining all these PRs together.

Of course, I am open to other reviewers' opinion.

zolkis:
I can do that. But all other PR's are also dependent on #337. Then I'd need to pull #337 in each and every one of them? Actually not all of #337, since only parts are needed, so that would take apart #337, which is now integral (shows in one piece the needed changes on MLOperand and MLActivation) and easy to review. 

I think the morale is that people prefer self-contained PRs, even if they also touch other parts of the spec (which I wanted to avoid, also based on feedback on some earlier PRs).

huningxin:
> Then I'd need to pull #337 in each and every one of them?

I don't think so. I suppose with initial several method steps PRs merged, most of the #337 should be absorbed and merged. For example, once this `MLGraphBuidler.constant()` PR with its deps merged, I don't expect you need #337 for `MLGraphBuilder.input()` any more. I expect similar situation for operators.

IMO, the internal slots and algorithms should be defined only if they have a clear use case, e.g. the interface method steps use them. The "top-down" approach would hopefully make this very clear.

zolkis:
> IMO, the internal slots and algorithms should be defined only if they have a clear use case, e.g. the interface method steps use them. 

Since I have prepared a lot of PRs in the background, I know for sure they have a clear use, and I wanted to group them in a clear way into a single PR, especially when that has earlier been encouraged.

But I do understand that for a reviewer the broken dependencies in PRs is not acceptable, and also that merging a claimed dependency just based on trust that all of it will be needed is not necessarily a good practice :). 

Anyway, everything doesn't have to be perfect now and we might touch those internal slots later, too.

index.bs
zolkis: Done

index.bs
zolkis: Done

index.bs
zolkis: Done

zolkis:
@huningxin I pulled the dependencies and removed the unneeded ones. Now the bs builds clean, but I expect merge hell when #337 and especially #329 is merged. We should preferably merge those first.

index.bs
```diff
@@ -1149,9 +1197,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |value| be the first argument.
+        1. If |value| is not a number, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |type| be the second argument.
+        1. If |type| is `undefined`, let |type| be `"float32"`.
+        1. Otherwise, if |type| is not one of {{MLOperandType}}, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |descriptor| be a new {{MLOperandDescriptor}}.
+        1. Set |descriptor|.{{MLOperandDescriptor/type}} to |type|.
+        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `[ 1 ]`.
```
huningxin, line 1250: Should we leave `dimensions` undefined? Because the [spec of `MLOperandDescriptor`](https://www.w3.org/TR/webnn/#api-mloperanddescriptor) mentions 

> The dimensions field is only required for tensor operands.

This method actually creates a scalar operand.

index.bs
```diff
@@ -1149,9 +1197,63 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+    1. Let |bufferView| be an {{ArrayBuffer}} of type corresponding to |type| according to [this table](#appendices-mloperandtype-arraybufferview-compatibility), containing |value| as its single element.
+    1. Let |operand| be the result of invoking the {{MLGraphBuilder/constant(descriptor, bufferView)}} steps given |descriptor| and |bufferView|.
```
huningxin, line 1252: Should it request the underlying platform to register a corresponding platform scalar operand with `|type|` and `|value|`? 

Some native ML APIs support to create scalar operand, e.g. [NNAPI](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1gaf06d1affd33f3bc698d0c04eceb23298). `MLGraphBuilder.constant(value, type)` should not just be polyfilled by `MLGraphBuilder.constant(descriptor, bufferView)`.

huningxin (commented):
@zolkis , thanks for adding the dependent internal slots and algorithms.

`MLGraphBuilder.constant(descriptor, bufferView)` looks good to me.

Left two questions for `MLGraphBuilder.constant(value, type)`, please take a look.

index.bs
```diff
@@ -1149,9 +1197,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+</div>
+<div algorithm=constant>
+    The {{MLGraphBuilder/constant(descriptor, bufferView)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |descriptor| be the first argument.
+    1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1. If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1220: Should it only run "check dimensions" steps only when `|descriptor|.{{MLOperandDescriptor/dimensions}}` is present?

index.bs
```diff
@@ -1149,9 +1197,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+    **Returns:**: an {{MLOperand}} object.
+</div>
+<div algorithm=constant>
+    The {{MLGraphBuilder/constant(descriptor, bufferView)}} steps are:
+    <div class="note">
+        The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
+    </div>
+    1. Let |descriptor| be the first argument.
+    1. If |descriptor| is not an an object that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+        1. If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
```
huningxin, line 1219: "byte length" algorithm may also need to be fixed for the case that `dimensions` is not present.

index.bs
```diff
@@ -1149,9 +1197,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1. If |value| is not a number, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |type| be the second argument.
+        1. If |type| is `undefined`, let |type| be `"float32"`.
+        1. Otherwise, if |type| is not one of {{MLOperandType}}, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |descriptor| be a new {{MLOperandDescriptor}}.
+        1. Set |descriptor|.{{MLOperandDescriptor/type}} to |type|.
+        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `[ 1 ]`.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=], `"constant"` and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Make a request to the underlying platform to register |operand| as a constant with |bytes| as value and store a reference of the corresponding [=implementation-defined=]  platform object for the scalar operand to |operand|.{{MLOperand/[[operand]]}}.
```
huningxin, line 1253: `|bytes|` is not defined here, should it be `|value|`?

index.bs
```diff
@@ -1149,9 +1197,64 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
...
+        1. If the [=byte length=] of |descriptor| is not supported by the underlying platform, then throw a "{{DataError}}" {{DOMException}} and stop.
+        1. If the [=check dimensions=] steps given |descriptor|.{{MLOperandDescriptor/type}} and |descriptor|.{{MLOperandDescriptor/dimensions}} return `false`, throw a "{{DataError}}" {{DOMException}} and stop.
+    1. Let |bufferView| be the second argument.
+        1. If invoking <a>validate buffer with descriptor</a> given |bufferView| and |descriptor| return `false`, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |operand| be the result of invoking the <a>create MLOperand</a> steps with [=this=], `"constant"` and |descriptor|.
+        1. If that throws, re-throw the exception and stop.
+    1. Let |bytes| be the result of invoking [[=get a copy of the bytes held by the buffer source=]] given |bufferView|.
+    1. Make a request to the underlying platform to register |operand| as a constant with |bytes| as value and store a reference to the corresponding [=implementation-defined=] object to |operand|.{{MLOperand/[[operand]]}}.
```
huningxin, line 1226: The discussion for `MLGraphBuilder.constant(value, type)` reminds me that this step may need to distinguish **tensor** and **scalar** constant. Let's say if `descriptor.dimensions` is present and valid, request the underlying platform to register a tensor constant. If `diescriptor.dimensions` is not present, register a scalar constant.

index.bs
huningxin: This step should be removed.

index.bs
zolkis: Sounds good.

index.bs
zolkis: Yes, now I got how the alternative plays out. Thanks!

index.bs
zolkis: Did we define somewhere and algorithm to get dimensions of a tensor?

index.bs
```diff
@@ -1241,16 +1261,22 @@ Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
     1. Let |value| be the first argument.
-        1. If |value| is not a number, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+        1. If |value| is an array, then let |shape| be the result of invoking [=MLOperand/shape=] given |value|.
```
huningxin, line 1264: Should it just throw error if `|value|` is not a number, according to the current [IDL definition](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-constant-value-type) of `MLGraphBuilder.constant(value, type)`? If this is a new proposal, it probably should be in a separate issue.

index.bs
huningxin: What's the use case of the "algorithm to get dimensions of a tensor"? In the latest commit, it seems to be used to infer the shape of a multi-dimensional JS array. But current `MLGraphBuilder.constant(value, type)` method only takes a number as input to define a scalar constant operand.

The `dimensions` of internal slot `MLOperand.descriptor` defined in this PR is the shape of a tensor operand.

index.bs
zolkis: I thought a constant could be a scalar or a tensor? So here it's just checking if it's an array, and infers the dimensions/shape of the array as a basis to create the descriptor for the new MLOperand.

index.bs
zolkis: I need to get the shape of a tensor when a constant tensor is given as a value to the constant() method. Its shape is used for creating the descriptor/dimensions for the constant operand.

index.bs
huningxin: > I thought a constant could be a scalar or a tensor?

Yes. [`constant(value, type)`](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-constant-value-type) method builds a scalar constant. [`constant(descriptor, bufferView)`](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-constant) method builds a tensor (`descriptor.dimensions` is present) or scalar constant (`descriptor.dimensions` is not present) .

> So here it's just checking if it's an array, and infers the dimensions/shape of the array as a basis to create the descriptor for the new MLOperand.

The current IDL [defines](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-constant-value-type) the `value` parameter as a double. If it is an array, it should report error.

index.bs
```diff
@@ -1243,40 +1328,37 @@ Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
...
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
-    1. Let |value| be the first argument.
-        1. If |value| is an array, then let |shape| be the result of invoking [=MLOperand/shape=] given |value|.
-            1. If that throws, re-throw the exception and stop.
-        1. Else if |value| is not a number, then throw a "{{TypeError}}" {{DOMException}} and stop.
-    1. Let |type| be the second argument.
-        1. If |type| is `undefined`, let |type| be `"float32"`.
-        1. Otherwise, if |type| is not one of {{MLOperandType}}, then  throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. If |value| is a [=number=] array, then then throw a "{{TypeError}}" {{DOMException}} and stop.
```
huningxin, line 1350: My last [comment](https://github.com/webmachinelearning/webnn/pull/365#discussion_r1152847950) may confuse. But according to the current IDL, the check should be if `|value|` is not a number, it throw an error.

index.bs
zolkis: OK, it seems that I missed if there was a convention that dimensions can be missing from descriptor.
Right now the Web IDL mandates the `descriptor` argument, and I inferred it needs both `type` and `dimensions`.
How should the algorithm behave if one or both are missing?

index.bs
zolkis: Sorry, that was a typo, or merge hickup. Fixed :)

index.bs
zolkis: They require the `descriptor`, but the steps do have the gap of checking the descriptor. Before fixing either of these, we should agree (or I should understand) how they are supposed to work when `dimensions` and/or `type` is missing.

huningxin (approved):
LGTM, thanks @zolkis !

index.bs
```diff
@@ -686,16 +686,71 @@ dictionary MLOperandDescriptor {
 
 ### The MLOperand interface ### {#api-mloperand}
 
-An {{MLOperand}} represents an intermediary graph being constructed as a result of compositing parts of an operation into a fully composed operation.
+An {{MLOperand}} represents an intermediary graph edge being constructed as a result of compositing parts of an operation into a fully composed operation.
```
wchao1115, line 689: I think the existing description is already accurate as an operand can indeed be a whole sub-graph and not just a specific graph edge.

index.bs
```diff
@@ -686,16 +686,71 @@ dictionary MLOperandDescriptor {
...
 
 For instance, an {{MLOperand}} may represent a constant feeding to an operation or the result from combining multiple constants together into an operation. See also [[#programming-model]].
 
 <script type=idl>
+enum MLOperandKind {"constant", "input", "output"};
```
wchao1115, line 694: I'm not sure I understand the reason to add the new `MLOperandKind` here. Has there been an issue opened for this or a discussion in a WG meeting for the need of it?

index.bs
```diff
@@ -686,16 +686,71 @@ dictionary MLOperandDescriptor {
...
+{{MLOperand}} has the following internal slots:
+<div class=internal-slots>
+<dl dfn-type=attribute dfn-for="MLOperand">
+    : <dfn>\[[builder]]</dfn> of type {{MLGraphBuilder}}
+    ::
+        The {{MLOperand}}'s associated builder object.
+
+    : <dfn>\[[kind]]</dfn> of type {{MLOperandKind}}
+    ::
+        The {{MLOperand}}'s kind.
```
wchao1115, line 708: Why do we need `MLOperandKind`?

index.bs
```diff
@@ -686,16 +686,71 @@ dictionary MLOperandDescriptor {
...
+    ::
+        Reference to {{MLOperand}}'s corresponding [=implementation-defined=] platform operator object.
+</dl>
+</div>
+
+Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/constructor()}} constructor to an {{MLContext}} object, an {{MLOperand}} is also always bound to the same {{MLContext}} object.
+
+#### Creating {{MLOperand}} #### {#api-mloperand-create}
+<div class="note">
+    The {{MLOperand}} objects are created internally by the methods of {{MLGraphBuilder}}. Based on these steps, exposing a constructor would be possible as well, but needs an explicit use case.
```
wchao1115, line 732: Why? `MLOperand` is an interface that represents either an individual operation or the whole composition of operations being constructed by the graph builder. Besides the methods to create the operands for graph inputs, those methods off the graph builder should already represent every operation as `MLOperand`. There should be no need for additional explicit constructor of `MLOperand` if that's what you mean.

index.bs
```diff
@@ -686,16 +686,71 @@ dictionary MLOperandDescriptor {
...
+</div>
+To <dfn>create MLOperand</dfn> given |builder|, |kind| and |desc|, run the following steps:
+<div class=algorithm-steps>
+    1. If |builder| is not an instance of {{MLGraphBuilder}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. If |kind| is not an {{MLOperandKind}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. If |desc| is not an [=object=] that [=implements=] {{MLOperandDescriptor}}, then throw a "{{TypeError}}" {{DOMException}} and stop.
+    1. Let |operand| be a new [=object=].
+    1. Set |operand|.{{MLOperand/[[builder]]}} to |builder|.
+    1. Set |operand|.{{MLOperand/[[kind]]}} to |kind|.
```
wchao1115, line 741: No need for Kind.

wchao1115 (changes requested):

index.bs
zolkis: Thanks Chai for reviewing. An issue is indeed missing ATM, because the dependency of `kind` is future work.

The `kind` is an internal slot of `MLOperand` used in the build steps (currently missing from the spec, but I have a draft for it) to check if the current `operand` in a `graph` is a "constant" or an "input" kind. If it's an input, it's added to 
`|graph|.{{MLGraph/[[inputDescriptors]]}}`, otherwise if it's constant, it's preprocessed/optimized.

It is only an internal slot, hence we don't need IDL definitions for it indeed. We can just use "constant" and "input" as strings in the spec prose.
But we'd need to keep the internal slot - or otherwise please suggest how to discern in algorithmic steps the operands in a graph to be constant or input.
It's also used in the implementation.

We can discuss this in the next call.

index.bs
zolkis: OK, will remove the note.

zolkis: The reasoning behind constructors usually is that APIs should be designed in a way that it's easy to create JS objects from console/scripts that can be connected via their interfaces to the underlying representations.

If we want to keep a tight connection between the JS objects and what they represent, then we can stick with the factory-only design, and use default object constructor steps in their algorithms.

index.bs
zolkis: OK. Should we use mention this explicitly for clarity? (there is only this one occurrence of "intermediary graph" the spec).

> An {{MLOperand}} represents an intermediary graph (sub-graph or edge) being constructed as a result of compositing parts of an operation into a fully composed operation.

Or should we just leave it as it was?

wchao1115: Please leave it as it was.

index.bs
wchao1115: Thanks for clarifying. If it's meant for internal slot only, please remove it from the IDL.

index.bs
wchao1115: We consciously chose the builder method of construction when we define the graph builder interface. More specifically for the work to construct the ML graph, the builder pattern works much better as it allows the implementation to keep transient graph building state in the graph builder interface as the construction progresses. In short, `MLOperand` isn't designed with a constructor pattern in mind.

wchao1115 (changes requested):

index.bs
```diff
@@ -695,7 +695,60 @@ For instance, an {{MLOperand}} may represent a constant feeding to an operation
...
+{{MLOperand}} has the following internal slots:
+<div class=internal-slots>
+<dl dfn-type=attribute dfn-for="MLOperand">
+    : <dfn>\[[builder]]</dfn> of type {{MLGraphBuilder}}
+    ::
+        The {{MLOperand}}'s associated builder object.
+
+    : <dfn>\[[kind]]</dfn> of type {{string}}
+    ::
+        The {{MLOperand}}'s kind, i.e. "input", "constant" or "output".
```
wchao1115, line 707: I'm still a bit confused with how the MLOperand kind is determined (e.g. "input" vs. "output"). For example, when an operand is created, what is the internal value set to the kind slot? Then when the same operand is immediately passed into the construction of the next operand in the ordering of the graph, when and what value should now be set on the kind slot? I'm still questioning if we need it, even as an internal slot. What value do we want to get out of defining it in the implementation of `MLOperand`?

zolkis: I am not against testing if we need it at all, but the implementation certainly uses the "kind". 

As explained above, at least  the "constant" vs "input" differentiation is need in the build steps (currently missing).

I see no ham in keeping this slot, or even the IDL definitions, until we get the algorithms (at least the build algorithm) sorted out. Of course we can remove anything that is not strictly needed.

@huningxin what is your take on this?

huningxin: The platform operand object ([[operand]] internal slot) should already take the kind info, e.g. "constant", when it is being created. The algorithm step is

> Make a request to the underlying platform to register |operand| as a tensor constant with |bytes| as value and store a reference to the corresponding [=implementation-defined=] object to |operand|.{{MLOperand/[[operand]]}}.

So the internal slot [[kind]] probably can be removed unless it is useful for any other algorithm steps.

zolkis: Currently, in the scope of this PR the [[kind]] slot is not needed.
It was added for future dependency, as it turns out, mainly for the build steps.

I guess the simplest is to remove it from the scope of this PR, and eventually re-add later when the build steps PR is filed. Then we can discuss there if/how we can do without it.

index.bs
```diff
@@ -1149,9 +1203,68 @@ The [=new=] {{MLGraphBuilder}} constructor steps are:
 1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, throw a "{{SecurityError}}" {{DOMException}} and abort these steps.
 1. Let |context| be the first argument.
 1. If the <a>validate MLContext</a> steps given |context| return `false`, throw a "{{TypeError}}" and abort these steps.
-
 1. Set {{MLGraphBuilder/[[context]]}} to |context|.
 
+### The `constant()` method  ### {#api-mlgraphbuilder-constant-method}
```
wchao1115, line 1208: I'm a bit confused by the styling of the section name. For example, here you have the `constant()` method name of section 7.6.2 as a quoted string, which produces a monospaced display text in the content, while all the existing sections e.g. the subsequent one 7.6.3 having the batchNormalization name outside of the quote. But then the sections before it have a mixture of quoted and unquoted method names, etc. 

This is the reason behind calling out making a stylistic change a separate PR in our [Contribution Guidelines](https://github.com/webmachinelearning/webnn/blob/main/CONTRIBUTING.md#type-of-change). We have a lot of content in the spec, if a PR comes in with both content and styling changes, we may have missed something during the reviews and caused the document to become inconsistent styling-wise. 

My recommendation for this PR is to separate out styling changes into a separate PR as described in the guidelines.

wchao1115 (changes requested):

index.bs
zolkis: Thanks for taking a look.

If you check this thread, earlier we discussed why the back quotes are used. The `constant()` method is polymorphic. It has 2 forms like `{{MLGraphBuilder/constant(descriptor, bufferView)}}` and `{{MLGraphBuilder/constant(value, type)}}`.
Since neither is used in the title of the main section, the simple back quoted `constant` is used there. This is consistent over my subsequent changes.

For other methods the autolinking definition should be used. 

To apply first the stylistic changes, means a lot of unwelcome manual merge for me. In addition, some of the sections do not even exist yet to which the stylistic changes will be applied. So we'd apply first those stylistic changes that apply to the current version of the document, then those which apply to the additional sections. So anyway we can't avoid doing later stylistic changes, but we can make it extremely complicated for the rules' sake.

zolkis: The reason we agreed to do a few PRs first was exactly to figure out consistent stylistic changes across these selected features, which will then be applied to all subsequent PRs. I mentioned in the last call that this makes the currently open PRs a bit of an exception. We need them to figure out the rules for the single-shot stylistic PR you've been asking for.

Otherwise we can't avoid going back and forth - which is okay for a few exceptions but we want to get it right before scaling up. 

So if the batchNorm PR differs, that is its bug, and I will correct that. By the time we merge these PRs we will have the rules pretty much agreed upon. There are not so many of them, anyway.

index.bs
zolkis: @huningxin I think this is still open.

A spec should have clear instructions how implementations should handle the case `dimensions` (or `type`) is missing from a required `descriptor`.

index.bs
zolkis: I have removed the back quotes from the title and made a note to self to record that preference.

wchao1115 (approved):

index.bs
huningxin: According to the current spec, `type` is required and `dimensions` is optional. If `dimensions` is not present, the `descriptor` represents a scalar value. It's fine to file an issue and fix it in the following PR.

zolkis: Would it make sense to assign a default value `0`, meaning scalar? Or is it better to handle with [=Map/exist=] and mention it means scalar?

huningxin: > is it better to handle with [=Map/exist=] and mention it means scalar?

This sounds good to me. Should we define a separate algorithm for this scalar check?

zolkis: OK, then it should be fine for now as we have separate algorithms. The `constant(descriptor, bufferView)` algorithm checks the `descriptor.dimensions`, but for the scalar case `constant(value, type)` it defines a new descriptor, currently saying the following:
```
        1. Set |descriptor|.{{MLOperandDescriptor/type}} to |type|.
        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `undefined`.
            <div class="note">
                In the case of a scalar constant, |descriptor|.{{MLOperandDescriptor/dimensions}} is ignored.
            </div>
```

That looks concise enough for me, but if you have suggestions to make it better, please tell.

huningxin: Yes, there are separate algorithms, we are fine now.

index.bs
huningxin: There is a separate algorithm for scalar, it is fine now.

huningxin (approved):

huningxin:
@zolkis , could you please squash the 18 commits and edit the commit message before merging? Thanks!

huningxin:
Looks good. Thanks much @zolkis . I am going to merge it.

PR #376, opened 2023-04-10
Simplify `slice` op

### LINKED ISSUE #369: Make the `starts` and `sizes` definitions of `slice` op consistent and simplified
Thanks @fdwr for pointing out the inconsistency of `starts` and `sizes` definitions of [`slice`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-slice) op when revisiting the PR https://github.com/webmachinelearning/webnn/pull/359#issuecomment-1484393520, where Dwayne mentioned: 

>  I just noticed slices parameter starts still has the negative number policy and accepts axis coordinates as long rather than unsigned long; and its sizes parameter has an unusual policy that -1 is special, whereas sizes in resample2D is sequence<unsigned long>. Do we want to rationalize these for consistency too, having higher level policy like this determined by the calling framework?

In a followed [comment](https://github.com/webmachinelearning/webnn/pull/359#issuecomment-1484636388), Dwayne also pointed out that 

> we see that every other sizes parameter is unsigned long (see below), leaving only slice as sequence<long> sizes.

From my point of view, using unsigned integer for `starts` and `sizes` makes sense. It not only would help ensure the consistency across the spec but also would simplify the implementation. The native ML APIs, such as DirectML's [`DML_SLICE1_OPERATOR_DESC`](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_slice1_operator_desc) and XNNPACK's [`xnn_define_static_slice`](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#L1417), usually use unsigned integer starts/offsets and sizes.

Another simplification idea is whether the `MLSliceOptions` should be removed. `MLSliceOptions.axes` is used to indicate the dimensions of the input shape to which starts and sizes apply. However, native ML APIs, e.g. DirectML's [`DML_SLICE1_OPERATOR_DESC`](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_slice1_operator_desc), XNNPACK's [`xnn_define_static_slice`](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#L1417) and MLCompute's [`MLCSliceLayer`](https://developer.apple.com/documentation/mlcompute/mlcslicelayer?language=objc), usually expect the `starts` and `offsets` for all dimensions to be provided. 

With that, a possible simplified `slice` op definition would be
```js
  // The length of `starts` and `sizes` sequences should equal to the rank of input tensor.
  MLOperand slice(MLOperand input, sequence<unsigned long> starts, sequence<unsigned long> sizes);
};
```

Thoughts?

huningxin:
This PR is based on the discussion of #369 as the first step. It includes the changes:

1. Change `starts` and `sizes` parameters to be of a sequence of unsigned long.
2. Remove `MLSliceOptions.axes`.

The `MLSliceOptions.steps` may need more discussion and will be left to a separate PR.

Fix #369

@wchao1115 @anssiko @pyu10055 @wacky6 @fdwr, PTAL!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/376.html" title="Last updated on Apr 13, 2023, 3:11 AM UTC (8efc568)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/376/47d64fc...huningxin:8efc568.html" title="Last updated on Apr 13, 2023, 3:11 AM UTC (8efc568)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2697,7 +2697,7 @@ partial interface MLGraphBuilder {
 <div algorithm=slice>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input tensor.
-        - *starts*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the starting index to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *starts[d]* indicates the starting index to slice in that dimension.
+        - *starts*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the starting index to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *starts[d]* indicates the starting index to slice in that dimension. The starting index must be in the range [0, input size - 1] in that dimension.
         - *sizes*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the number of elements to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *sizes[d]* indicates the number of elements to slice in that dimension. The size must not be 0 and must satisfy the constraint *starting index + size <= input size* in that dimension.
 
     **Returns:** an {{MLOperand}}. The output tensor of the same rank as the input tensor with tensor values stripped to the specified starting and ending indices in each dimension.
--- a/index.bs
+++ b/index.bs
@@ -2698,7 +2698,7 @@ partial interface MLGraphBuilder {
     **Arguments:**
         - *input*: an {{MLOperand}}. The input tensor.
         - *starts*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the starting index to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *starts[d]* indicates the starting index to slice in that dimension.
-        - *sizes*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the size to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *sizes[d]* indicates the size to slice in that dimension. The size must not be 0 and must satisfy the constraint *starting index + size <= input size* in that dimension.
+        - *sizes*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the number of elements to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *sizes[d]* indicates the number of elements to slice in that dimension. The size must not be 0 and must satisfy the constraint *starting index + size <= input size* in that dimension.
 
     **Returns:** an {{MLOperand}}. The output tensor of the same rank as the input tensor with tensor values stripped to the specified starting and ending indices in each dimension.
 </div>
--- a/index.bs
+++ b/index.bs
@@ -1666,11 +1666,11 @@ partial interface MLGraphBuilder {
     let currentRecurrentBias = [];
 
     for (let dir = 0; dir < numDirections; ++dir) {
-      currentWeight.push(builder.squeeze(builder.slice(weight, [dir, 0, 0], [1, -1, -1]), { axes: [0] }));
-      currentRecurrentWeight.push(builder.squeeze(builder.slice(recurrentWeight, [dir, 0, 0], [1, -1, -1]), { axes: [0] }));
-      currentBias.push(options.bias ? (builder.squeeze(builder.slice(options.bias, [dir, 0], [1, -1]), { axes: [0] })) : null);
+      currentWeight.push(builder.squeeze(builder.slice(weight, [dir, 0, 0], [1, 3 * hidden_size, input_size]), { axes: [0] }));
+      currentRecurrentWeight.push(builder.squeeze(builder.slice(recurrentWeight, [dir, 0, 0], [1, 3 * hidden_size, hidden_size]), { axes: [0] }));
+      currentBias.push(options.bias ? (builder.squeeze(builder.slice(options.bias, [dir, 0], [1, 3 * hidden_size]), { axes: [0] })) : null);
       currentRecurrentBias.push(options.recurrentBias ?
-        (builder.squeeze(builder.slice(options.recurrentBias, [dir, 0], [1, -1]), { axes: [0] })) : null);
+        (builder.squeeze(builder.slice(options.recurrentBias, [dir, 0], [1, 3 * hidden_size]), { axes: [0] })) : null);
     }
 
     for (let step = 0; step < steps; ++step) {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2690,23 +2690,15 @@ partial interface MLGraphBuilder {
...
-        - *options*: an optional {{MLSliceOptions}}. The optional parameters of the operation.
-            - *axes*: a sequence of {{unsigned long}}. The dimensions of the input shape to which *starts* and *sizes* apply. The values in the sequence must be in the range [0, N-1] where N is the rank of input tensor. When not specified, the sequence is assumed to be [0, ..., N-1], e.g. [0,1,2] for a 3-D tensor.
+        - *starts*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the starting index to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *starts[d]* indicates the starting index to slice in that dimension.
+        - *sizes*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the size to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *sizes[d]* indicates the size to slice in that dimension. The size must not be 0 and must satisfy the constraint *starting index + size <= input size* in that dimension.
```
fdwr, line 2701: >The size must not be 0

This may need to be allowed, given some models I've seen (depending on their inputs) might temporarily slice all their content to an empty tensor and then readd padding later. Additionally some object detection models can legitimately return 0 bounding rects as a valid result. So, WebNN in general shouldn't disallow them imo. The WebNN backends can always just skip those operators at a key choke point when seeing 0 sizes, since they are generally nops anyway (that's what ORT's DML EP does when it encounters empty tensors). Though, since supporting empty tensors affects more than just `slice`, that's probably better as a separate issue - no action expected right now. 👍

fdwr (approved):
👍 Thanks Ningxin for simplifying the backend implementations, reducing test complexity, and increasing explicit clarity.

wchao1115 (approved):

index.bs
wacky6, line 2701: nit: instead of "the size to slice", perhaps "the number of elements to slice". 

"the number of elements" could xref to its definition in the spec (if it's defined).

huningxin: > perhaps "the number of elements to slice".

sgtm. Fixed in the new commit, please take another look.

huningxin: @fdwr 

> given some models I've seen (depending on their inputs) might temporarily slice all their content to an empty tensor and then readd padding later. Additionally some object detection models can legitimately return 0 bounding rects as a valid result. So, WebNN in general shouldn't disallow them imo. 

Thanks for your inputs. Those sound like good use cases. It probably deserves a separate issue for WG discussion. 

WDYT?

wacky6 (approved):

index.bs
fdwr: Yep, separate discussion.

anssiko (approved):
(Publication to https://www.w3.org/TR/webnn/ upon PR merge will resume once https://github.com/speced/bikeshed-boilerplate/pull/38 lands. I expect that to happen soon. I suggest hold on merging until that happens so we can validate the auto-publish job works as planned.)

index.bs
```diff
@@ -2690,23 +2690,15 @@ partial interface MLGraphBuilder {
...
-        - *starts*: a sequence of {{long}}. The starting indices to slice of the corresponding axes of the input shape. A negative index value is interpreted as counting back from the end. For example, the value -1
-        - *sizes*: a sequence of {{long}}. The lengths to slice of the corresponding axes of the input shape.
-            The length value of -1 selects all the remaining elements from the starting index of the given axis.
-        - *options*: an optional {{MLSliceOptions}}. The optional parameters of the operation.
-            - *axes*: a sequence of {{unsigned long}}. The dimensions of the input shape to which *starts* and *sizes* apply. The values in the sequence must be in the range [0, N-1] where N is the rank of input tensor. When not specified, the sequence is assumed to be [0, ..., N-1], e.g. [0,1,2] for a 3-D tensor.
+        - *starts*: a sequence of {{unsigned long}}. The sequence of unsigned integer values indicating the starting index to slice of each input dimension, of length N where N is the rank of the input tensor. For each dimension *d* of *input*, *starts[d]* indicates the starting index to slice in that dimension.
```
BruceDai, line 2700: Does it need to describe explicitly that **starts index** for each dimension d must be in the range **[0, input size - 1]**?
@huningxin @fdwr WDYT? Thanks.

huningxin: The *starting index* may not be *input size - 1*, because there would be 0 elements to slice which is not allowed in the current definition.

The *starting index* should also satisfy the constraint *starting index + size <= input size* where *size* must not be 0 for each dimension.

Later, as @fdwr [recommended](https://github.com/webmachinelearning/webnn/pull/376#discussion_r1162180789), if 0 slice size is allowed, this constraint should also work.

@BruceDai , do you see any challenges to implement it?

BruceDai: > do you see any challenges to implement it?

No. My concern is only for adding explicitly description for starting index.

BruceDai: > The starting index may not be input size - 1

The starting index could be input size - 1  when size = 1.
For example producing a slice of the 1D input tensor( it's value is [0, 1, 2, 3] ) by starts = [3] and sizes = [1],  result is expected as [3], and now starting index + size = 4 = input size satisfy **the constraint starting index + size <= input size**. Am I understanding this correctly? PTAL, thanks.

huningxin: Thanks @BruceDai , your example is correct. My previous thought was wrong (comment updated). I'd agree it makes sense to add [0, input size - 1] constraint to starting index.

huningxin: Fixed in the latest commit https://github.com/webmachinelearning/webnn/pull/376/commits/8efc5685cc9adea78debe3cb3d2e829afda7331a, please take a look. Thanks!

BruceDai: LGTM, thanks.

huningxin: > The starting index may not be input size - 1

What I thought was if we allow 0 size dimension for the use cases that @fdwr [mentioned ](https://github.com/webmachinelearning/webnn/pull/376#discussion_r1162180789), the range would become [0, -1] which won't make sense. 

Because we will discuss that in a separate issue, it would be fine to add starting index range [0, input size - 1] into this PR. Any thoughts @fdwr ?

fdwr: @huningxin : Yes, given the current minimum requirement of size > 0, then `[0, inputSize - 1]` is the correct range.

anssiko:
@huningxin @wchao1115 feel free to merge when you’re settled with the remaining discussion.

https://github.com/speced/bikeshed-boilerplate/pull/38 was merged unblocking publications to https://www.w3.org/TR/webnn/

This is the first time we publish a Candidate Recommendation Draft (CRD) with this automated workflow so if any issues occur @dontcallmedom will be the expert to assist.

huningxin:
Thanks @anssiko . I think we've resolved all conversations. I am going to merge it. Thanks all for the reviews and approvals. 

Let's see whether the automated workflow works.

anssiko:
Works as expected: https://www.w3.org/TR/webnn/ 🚀

Thanks @dontcallmedom for reconfiguring this worflow.

PR #387, opened 2023-05-09
Fix first-wave models table formatting

anssiko:
It looks like the extra whitespaces in https://github.com/webmachinelearning/webnn/commit/c82fd2df44e22b1394d29e6025b6f06f987f3df7 confused the markdown renderer even if they made the source much prettier :)

This PR simply takes the previous commit https://github.com/webmachinelearning/webnn/commit/332fbbd85ec2238f45fb4789f21a3fc50f310e53 that renders properly and replaces all occurrences of:
http://webmachinelearning.github.io/webnn/
https://webmachinelearning.github.io/webnn/

With:
https://www.w3.org/TR/webnn/

To bring it on par with the latest while rendering properly.

PTAL @ibelem


ibelem:
LGTM, thanks @anssiko!

anssiko:
Thanks @ibelem!

PR #445, opened 2023-08-09
Fix missing <details open> tags

wchao1115:
Fix a few missing <details open> tags for the relevant dev notes to stay consistent with the new style convention.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1814,7 +1814,10 @@ partial interface MLGraphBuilder {
 </details>
 
 <div class="note">
+  <details open>
+    <summary>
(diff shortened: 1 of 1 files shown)


zolkis:
I am fine with that, if you think it makes sense to collapse the notes.

zolkis (approved):
Thanks

anssiko (approved):

wchao1115:
> I am fine with that, if you think it makes sense to collapse the notes.

Not every note. Just the ones with pseudo code that shows how an operator can be decomposed. The reason is this is one of the design principles of webnn when it comes to its ability to support fusions. We agreed at the time to highlight these sections in the doc with `<div class="note">`. I simply searched all the occurrences of the word "emulated" in the note section in the doc and make them consistent style-wise. I'm pushing a change here to the `zk-convention-integration` branch as part of my review so that when all is clear, we can merge the branch wholesale to `main` then delete it.

wchao1115:
> Thanks

If you're ok with the change, please approve the PR. Thanks!

zolkis (approved):

Issue #454, opened 2023-08-23
(bikeshed naming) MLOperandDataType


fdwr:
Various ML libraries accept different kinds of operands to their operators (e.g. TensorFlow might accept [tensors](https://www.tensorflow.org/api_docs/python/tf/Tensor) or [variables](https://www.tensorflow.org/api_docs/python/tf/Variable); ONNX accepts the operand types [dense tensors](https://onnx.ai/onnx/repo-docs/IR.html), sparse tensors, maps, sequences, optionals; ...).

*Inside* that operand type (e.g. tensor), there may be a subtype for the element data (e.g. float32, bfloat16, int8, int32...), or `dtype` for short, as found in [NumPy data type `numpy.dtype`](https://numpy.org/doc/stable/reference/arrays.dtypes.html), [TensorFlow `tf.dtype`](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType), and [PyTorch `torch.dtype`](https://pytorch.org/docs/stable/tensor_attributes.html#torch.dtype); or "data type" as found in [XNNPack xnn_datatype](https://github.com/google/XNNPACK/blob/ae2875cb2538b00f11564c5cbf649249caf510f1/include/xnnpack.h#L211) and [DirectML DML_TENSOR_DATA_TYPE](https://learn.microsoft.com/en-us/windows/win32/api/directml/ne-directml-dml_tensor_data_type).

Sometimes it makes sense in a standard to name something different from general precedent when it is more informative/clearer, but the current naming [`MLOperandType`](https://www.w3.org/TR/webnn/#enumdef-mloperandtype) actually means the data type (float32, int8...), which is both inconsistent with the community (harder to map this term to everything else) *and* adds no clarity. Additionally, if future WebNN ever supports other kinds of operands besides dense tensors (e.g. sparse tensors), it will be rather awkward to speak of this new type because there will be no vocabulary left for it, since the term `MLOperandType` was already consumed to mean the *subtype* within the operand type. Renaming `MLOperandType` -> `MLOperandDataType` will enable future expansion (if needed) where `MLOperandType` can be reintroduced to mean the actual operand type. I liken the current situation to having an enum named `Animal` which contains dog breeds {`Poodle`, `Bulldog`, `Boxer`}, which makes it harder to speak of actual animal types if later we want to introduce cats.

zolkis:
[...]

PR #464, opened 2023-09-08
MLOperandType -> MLOperandDataType

fdwr:
Motivation from https://github.com/webmachinelearning/webnn/issues/454 :

Various ML libraries accept different kinds of operands to their operators (e.g. TensorFlow accepts operand types like [tensors](https://www.tensorflow.org/api_docs/python/tf/Tensor) or [variables](https://www.tensorflow.org/api_docs/python/tf/Variable); ONNX accepts the operand types [dense tensors](https://onnx.ai/onnx/repo-docs/IR.html), sparse tensors, maps, sequences, optionals; ...).

*Inside* that operand type (e.g. tensor), there may be a subtype for the element data (e.g. float32, bfloat16, int8, int32...), or `dtype` for short, as found in [NumPy data type `numpy.dtype`](https://numpy.org/doc/stable/reference/arrays.dtypes.html), [TensorFlow `tf.dtype`](https://www.tensorflow.org/api_docs/python/tf/dtypes/DType), and [PyTorch `torch.dtype`](https://pytorch.org/docs/stable/tensor_attributes.html#torch.dtype); or "data type" as found in [XNNPack xnn_datatype](https://github.com/google/XNNPACK/blob/ae2875cb2538b00f11564c5cbf649249caf510f1/include/xnnpack.h#L211) and [DirectML DML_TENSOR_DATA_TYPE](https://learn.microsoft.com/en-us/windows/win32/api/directml/ne-directml-dml_tensor_data_type).

Sometimes it makes sense in a standard to name something different from general precedent when it is more informative/clearer, but the current naming [`MLOperandType`](https://www.w3.org/TR/webnn/#enumdef-mloperandtype) actually means the data type (float32, int8...), which is both inconsistent with the community (harder to map this term to everything else) *and* adds no clarity. Additionally, if future WebNN ever supports other kinds of operands besides dense tensors (e.g. sparse tensors), it will be rather awkward to speak of this new type because there will be no vocabulary left for it, since the term `MLOperandType` was already consumed to mean the *subtype* within the operand type. Renaming `MLOperandType` -> `MLOperandDataType` will enable future expansion (if needed) where `MLOperandType` can be reintroduced to mean the actual operand type. I liken the current situation to having an enum named `Animal` which contains dog breeds {`Poodle`, `Bulldog`, `Boxer`}, which makes it harder to speak of actual animal types if later we want to introduce cats.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -629,7 +629,7 @@ The WebGPU API identifies <a href="https://gpuweb.github.io/gpuweb/#privacy-mach
 
 The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Device type=] normatively indicates the kind of device and is either "cpu" or "gpu". If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
 
-If a future version of this specification introduces support for new a  [=device type=] that can only support a subset of {{MLOperandType}}s, that may introduce a new fingerprint.
+If a future version of this specification introduces support for new a  [=device type=] that can only support a subset of {{MLOperandDataType}}s, that may introduce a new fingerprint.
 
 In general, implementers of this API are expected to apply <a href="https://gpuweb.github.io/gpuweb/#privacy-considerations">WebGPU Privacy Considerations</a> to their implementations where applicable.
 
@@ -884,7 +884,7 @@ enum MLInputOperandLayout {
   "nhwc"
 };
 
-enum MLOperandType {
+enum MLOperandDataType {
   "float32",
   "float16",
   "int32",
@@ -895,7 +895,7 @@ enum MLOperandType {
 
 dictionary MLOperandDescriptor {
   // The operand type.
-  required MLOperandType type;
+  required MLOperandDataType type;
 
   // The dimensions field is only required for tensor operands.
   sequence<unsigned long> dimensions;
@@ -910,7 +910,7 @@ dictionary MLOperandDescriptor {
     1. Let |elementLength| be 1.
     1. [=map/For each=] |dimension| of |desc|.{{MLOperandDescriptor/dimensions}}:
         1. Set |elementLength| to |elementLength| × |dimension|.
-    1. Let |elementSize| be the [=element size=] of one of the {{ArrayBufferView}} types that matches |desc|.{{MLOperandDescriptor/type}} according to [this table](#appendices-mloperandtype-arraybufferview-compatibility).
+    1. Let |elementSize| be the [=element size=] of one of the {{ArrayBufferView}} types that matches |desc|.{{MLOperandDescriptor/type}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility).
     1. Return |elementLength| × |elementSize|.
   </div>
 </details>
@@ -1198,7 +1198,7 @@ partial interface MLContext {
   </summary>
   <div class=algorithm-steps>
     1. If |bufferView| is not an {{MLBufferView}}, return false.
-    1. If |bufferView|'s [=element type=] does not match to |descriptor|.{{MLOperandDescriptor/type}}  according to [this table](#appendices-mloperandtype-arraybufferview-compatibility), return false.
+    1. If |bufferView|'s [=element type=] does not match to |descriptor|.{{MLOperandDescriptor/type}}  according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility), return false.
     1. If |bufferView|.\[[ByteLength]] is not equal to the [=byte length=] of |descriptor|, return false.
   </div>
 </details>
@@ -1539,7 +1539,7 @@ interface MLGraphBuilder {
   MLOperand constant(MLOperandDescriptor descriptor, MLBufferView bufferView);
 
   // Create a single-value operand from the specified number of the specified type.
-  MLOperand constant(double value, optional MLOperandType type = "float32");
+  MLOperand constant(double value, optional MLOperandDataType type = "float32");
 
   // Compile the graph up to the specified output operands asynchronously.
   Promise<MLGraph> build(MLNamedOperands outputs);
@@ -1721,7 +1721,7 @@ Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
 <div>
     **Arguments:**
         - *value*: a number
-        - *type*: an optional {{MLOperandType}}, by default *"float32"*.
+        - *type*: an optional {{MLOperandDataType}}, by default *"float32"*.
     **Returns:**: an {{MLOperand}} object.
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -895,7 +895,7 @@ enum MLOperandType {
 
 dictionary MLOperandDescriptor {
   // The operand type.
```
huningxin, line 897: nit: Should it also be updated to "// The operand data type."?

index.bs
```diff
@@ -895,7 +895,7 @@ enum MLOperandType {
...
-  required MLOperandType type;
+  required MLOperandDataType type;
```
huningxin, line 898: Should this member be renamed to `dataType` accordingly?

huningxin (commented):
Thanks @fdwr for this PR!

huningxin (approved):
Looks good, thanks!

wchao1115 (approved):

huningxin:
@fdwr , there are code samples in explainer.md (type -> dataType) need to be updated. Would you like to do that in this PR or in a separate one?

@Honry @BruceDai , this is a API breaking change. Once it is merged, the WPT, samples and framework integration need to be updated accordingly.

BruceDai:
> this is a API breaking change. Once it is merged, the WPT, samples and framework integration need to be updated accordingly.

@huningxin Thanks for your kind reminder. Sure, I will update them.

fdwr:
> @fdwr , there are code samples in explainer.md (type -> dataType) need to be updated. Would you like to do that in this PR or in a separate one?

@huningxin: Thanks. Cleaner to do together one. Ctrl+Shft+F'd the repo - shouldn't be anymore 🤞.

huningxin:
Still LGTM, thanks much @fdwr !

index.bs
```diff
@@ -910,7 +910,7 @@ dictionary MLOperandDescriptor {
     1. Let |elementLength| be 1.
     1. [=map/For each=] |dimension| of |desc|.{{MLOperandDescriptor/dimensions}}:
         1. Set |elementLength| to |elementLength| × |dimension|.
-    1. Let |elementSize| be the [=element size=] of one of the {{ArrayBufferView}} types that matches |desc|.{{MLOperandDescriptor/type}} according to [this table](#appendices-mloperandtype-arraybufferview-compatibility).
+    1. Let |elementSize| be the [=element size=] of one of the {{ArrayBufferView}} types that matches |desc|.{{MLOperandDescriptor/type}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility).
```
BruceDai, line 913: @fdwr Here `MLOperandDescriptor/type` should be `MLOperandDescriptor/dataType`, and please update the others below, thanks.

BruceDai (commented):
@fdwr Thanks for this PR. 
Please also update `MLOperandDescriptor/type` to `MLOperandDescriptor/dataType`, thanks.

index.bs
huningxin: Good catch!

fdwr: Ooh sneaky, using `/` instead of .type, :type, type:: datatype- ...

index.bs
```diff
@@ -894,8 +894,8 @@ enum MLOperandType {
 };
 
 dictionary MLOperandDescriptor {
-  // The operand type.
-  required MLOperandType type;
+  // The operand data type.
+  required MLOperandDataType dataType;
```
Honry, line 898: This change would break trial for WebNN samples/WebNN EP on Chrome/Edge stable for a period of time (at lease one round of Stable release cycle), we can do workaround by adding a duplicated entry to the JS code, e.g. 
`const aTensorDesc = {type: 'float32', dataType: 'float32', dimensions: [1, 3, 3, 3]};`

And we may encounter more such kind of API breaking before the spec becoming mature. It's painful for developers who want to try WebNN early.

fdwr: Yeah, that's why I proposed this right after the recent big spec refactor 446, but before v2 ops were added, because up until this time, we've liberally updated the spec (e.g. changing signed dimension to unsigned, MLOperator -> MLActivation...), but it will become harder over time. 🕗...

wchao1115: I think this is a good discussion about how much we should tolerate a breaking change at this stage in the spec life cycle, now that folks started to implement it. And although this probably won't be the last breaking change before the spec enters a more stable stage (still to be defined when), the bar for a breaking change to go in should continue to rise overtime. 

So, name change is a breaking change, but if it's early enough, we can still look away. But I don't think this period should last long.

fdwr:
"1 workflow awaiting approval"
Who do I ask about this aspect? It's been a month since creation. Thanks.

anssiko:
@fdwr thanks for the ping, I approved this run now. (The editors should be able to approve workflow runs from public forks. If it is not so, let me tweak the GH knobs to make it so.)

There remains one Bikeshed build error to be fixed:
<details><summary>Error details</summary>

```
  [INFO] Build bikeshed document "index.bs" (gh)…
    $ bikeshed --die-on=link-error spec "index.bs" "index.bs.built.html" 
      LINK ERROR: Multiple possible 'idl' local refs for 'constant(descriptor, bufferView)'.
      Randomly chose one of them; other instances might get a different random choice.
      {{MLGraphBuilder/constant(descriptor, bufferView)|MLGraphBuilder/constant(value, type)}}
```

</details>

This may be related to https://github.com/speced/bikeshed/issues/1148 which is still open...

You can [build locally](https://github.com/webmachinelearning/webnn/blob/main/README.md) to test possible workarounds. Maybe @inexorabletash has run into this overloading issue with Bikeshed and has a workaround or fix?

Once the build is green, @wchao1115 @huningxin my expectation is you consider @Honry's feedback and coordinate with him on samples/WebNN EP changes to minimize disruption to early adopters.

inexorabletash:
You need to make these changes as well:

```diff
diff --git a/index.bs b/index.bs
index 9cf75a8..c2618b4 100644
--- a/index.bs
+++ b/index.bs
@@ -1430,7 +1430,7 @@ partial interface MLCommandEncoder {
   </summary>
   <div>
     <div class="note">
-        Graph initialization stage typically involves a process known as "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. The initializing inputs are typically the constant weight data specified through the {{MLGraphBuilder/constant(descriptor, bufferView)|MLGraphBuilder/constant(value, type)}} method as constant operands during graph construction time.
+        Graph initialization stage typically involves a process known as "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. The initializing inputs are typically the constant weight data specified through the {{MLGraphBuilder/constant(descriptor, bufferView)|MLGraphBuilder/constant(value, dataType)}} method as constant operands during graph construction time.
     </div>
   </div>
 </details>
@@ -1716,7 +1716,7 @@ Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
   </div>
 </details>
 
-#### The {{MLGraphBuilder/constant(value, type)}} method  #### {#api-mlgraphbuilder-constant-value-type}
+#### The {{MLGraphBuilder/constant(value, dataType)}} method  #### {#api-mlgraphbuilder-constant-value-type}
 
 <div>
     **Arguments:**
@@ -1727,14 +1727,14 @@ Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLGraphBuilder>constant(|value|, |type|)</dfn> method steps are:
+    The <dfn method for=MLGraphBuilder>constant(|value|, |dataType|)</dfn> method steps are:
   </summary>
   <div class=algorithm-steps>
     <div class="note">
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
-        1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |type|.
+        1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |dataType|.
         1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `undefined`.
             <div class="note">
                 In the case of a scalar constant, |descriptor|.{{MLOperandDescriptor/dimensions}} is ignored.
```

After that, the spec should build without errors.

inexorabletash:
And to be clear, this isn't a _workaround_; the change updated the IDL but not the actual definition (`<dfn>`) and various references to it. Bikeshed's LINK ERROR is telling you that it's trying to link things to a definition that doesn't exist. That said, the error message is either misleading or downright wrong, because it's saying:

```
LINK ERROR: Multiple possible 'idl' local refs for 'constant(descriptor, bufferView)'.
Randomly chose one of them; other instances might get a different random choice.
{{MLGraphBuilder/constant(descriptor, bufferView)|MLGraphBuilder/constant(value, dataType)}}
```

... but actual the problem is that links to `constant(value, dataType)` are failing because the overload definition was still for for `constant(value, type)`. And it's suggesting using `constant(value, dataType)` but that's the problem! I think it's trying to be too clever and confusing itself. I'll file a bug.

anssiko:
I triggered a re-run and the build is now green. Thanks @fdwr @inexorabletash!

The editors @huningxin @wchao1115 are now free to merge considering @Honry's feedback for coordination on samples/WebNN EP changes.

anssiko:
@Honry it looks like the WebNN EP workaround is integrated. Any other changes you want to deploy prior to this spec PR merge?

Honry:
> @Honry it looks like the WebNN EP workaround is integrated. Any other changes you want to deploy prior to this spec PR merge?

That's all, thanks @anssiko, please move forward.

anssiko:
@wchao1115 @huningxin I'll squash and merge this PR now. We have mitigated the compatibility impact to early adopters with an ONNX Runtime patch https://github.com/microsoft/onnxruntime/commit/b3cab55d68cd887c1a43d44f951c93189cf91af6 developed by @Honry (thanks!).

Thanks @fdwr submitting this PR that addresses #454!

Issue #390, opened 2023-05-18
0D scalars


fdwr:
## Context

This arose during prototyping additional models and during Chromium code review and ORT [review](https://github.com/Honry/onnxruntime/pull/9).

## Overview

`[1,2,3]` - 3D tensor shape
`[1,2]` - 2D tensor shape
`[1]` - 1D tensor shape
`[]` - 0D tensor shape (scalar)

A 0D tensor shape vs a 1D tensor with a single element are semantically and functionally distinct. For simple operators (e.g. Add/Sin), treating a scalar as *if* it was a 1D array with a single element yields the same functional result, but for more complex operators where the rank matters (e.g. ReduceSum with all axes and keepDimensions = false, or Gather where the output rank is computed as input.rank + indices.rank - 1...), treating the two as synonymous is problematic and has been causing a multitude of problems with mapping certain ONNX models to WebNN.

## References

Every known ML library represents 0D scalars via the shape `[]` (or `()`):

### Numpy
```python
import numpy

x = numpy.array(42)
y = numpy.add(x, x)
print("value:", y)
print("shape:", y.shape)

# Prints:
# value: 84
# shape: ()
```

### TensorFlow
```python
import tensorflow as tf

x = tf.constant(42, dtype=tf.float32)
y = tf.add(x, x);
print("value:", y)
print("shape:", y.shape)

# Prints:
# value: tf.Tensor(84.0, shape=(), dtype=float32)
# shape: ()
```

### PyTorch
```python
import torch

x = torch.tensor(42, dtype=torch.float)
y = torch.add(x, x)
print("value:", y)
print("shape:", y.shape)
# Prints:
# value: tensor(84.)
# shape: torch.Size([])
```

### ONNX
```python
import onnx

# Scalar via [].
x = onnx.helper.make_tensor(
    "value", onnx.TensorProto.FLOAT, [], [42]
)
```

### SafeTensorss

The [SafeTensors file format](https://github.com/huggingface/safetensors#format) (commonly used with Stable Diffusion models for custom weights) explicitly allows 0D scalars and 0-size tensors.

### XNNPack

Bin Miao says below that XNNPack works.

## Implementation experience

In my Chromium fork, I just needed to change 2 lines, deleting 1 errant validation statement in the graph builder, and adding 1 in the DML backend (TensorDesc constructors sets a minimum rank of 1) for the rest of it to work. Then scalar tests passed:

```javascript
    // 1. Create a computational graph
[...]

Issue #455, opened 2023-08-23
Reserve asserts only for non-obvious things


zolkis:
_Originally posted by @inexorabletash in https://github.com/webmachinelearning/webnn/pull/446#discussion_r1298735835_
            
> FWIW, I'd reserve asserts only for non-obvious things, e.g. where a much earlier step in an algorithm has ensured that a number is non-zero, so a division by it is safe.
> Asserts like these that repeat the concise Web IDL definition in longer prose are just noise and IMHO not useful for implementers.

zolkis:
The difficulty is that e.g. the "create..." and "validate..." type of internal steps have no WebIDL, so asserts make sense there IMHO, but indeed in the algorithms with WebIDL definitions they can be removed.
@anssiko suggested we do this after #446 is merged.

inexorabletash:
FYI I have a branch ready for this, but #514 needs to merge first.

anssiko:
#514 is now merged.

inexorabletash:
The remaining asserts are either:

- Asserting that the _op_ parameter to an internal algorithm is from a set. They seem fine, pending a better suggestion.
- [concat()](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-concat) args, tracked in #503 
- [broadcast-shapes](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-broadcast-shapes), tackled in #534
- [input()](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-input) args, tracked in #502 

... so I'm going to go ahead and close this.

Issue #395, opened 2023-06-05
Use deep copy for setting internal slots?


zolkis:
Should we do deep copy of the descriptor? Would the change of original operand's descriptor impact the copied one?

_Originally posted by @huningxin in https://github.com/webmachinelearning/webnn/pull/348#discussion_r1213183583_

zolkis:
[ECMAScript](https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots) doesn't really shed light on this. Internal slots are supposed to be created and then not modified unless specified differently.

Now we seem to have 2 options:
1. change every line that sets an internal slot to replace `the argument` with "a copy of `the argument`"
2. or, create steps to "set the internal slot `[[name]]` to `<arg>`" steps, where we say "set `[[name]]` to a copy of `<arg>`".

@anssiko @dontcallmedom  any input here?

anssiko:
If we need to do this in many places maybe it makes sense to define an abstract operation for doing a deep copy? Did you check whether any of the infra specs incl. HTML have reusable helpers for this?

For some inspiration, for structured data there's a bunch of abstract operations in HTML https://html.spec.whatwg.org/multipage/structured-data.html referenced by other specs such as IndexedDB, Web Workers.

zolkis:
Yes, we can define separate "set internal slot" steps for that and when it becomes specified in INFRA or ECMA, we can switch.

What we need is to prevent changes from client after setting, i.e. to pass ownership of the init dict to the internal slot's object. An emulation of that is [deep] copy / pass by value, usually referenced as "copy of" in WebIDL, as suggested above.
Is that (option 2) good to go?

anssiko:
Yes, I think a PR for that option would help.

zolkis:
Since it affects all internal slots and the whole spec, I will do it after the current batch of PRs is merged.
Maybe after the stylistic changes.

zolkis:
Setting internal slots has been explained [here](https://github.com/webmachinelearning/webnn/issues/450#issuecomment-1683052506).

inexorabletash:
I've been going through the spec and I don't see anywhere where an internal slot holds a non-primitive data type, and the properties are mutated after it is set. Most specifically, once a MLOperand's [[descriptor]] is set I never see it changing.

The closest I've spotted is in [resample output sizes](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-resample-output-sizes) where the input descriptor is cloned then mutated with:

[...]

PR #476, opened 2023-11-02
Fix dimensions for 0D scalars

fdwr:
From issue: https://github.com/webmachinelearning/webnn/issues/390

A 0D tensor shape vs a 1D tensor with a single element are semantically and functionally distinct, but the current spec mistreats 0D scalars and 1D tensors like they are the same thing, and this messes with the shape inference for operators like `reduceSum` (with all axes and `keepDimensions = false`) or `gather` where the output rank depends on the correct rank for input and indices tensors. This caused models like Segment Anything and Stable Diffusion to fail while prototyping, and the issue arose again during Chromium and ORT WebNN EP code review.

`[1,2,3]` - 3D tensor shape
`[1,2]` - 2D tensor shape
`[1]` - 1D tensor shape
`[]` - 0D tensor shape (scalar) <--- (every known ML library represents 0D scalars via the shape `[]`)

Fixing this in the Chromium fork needed changing just two lines, and we want to address this sooner than later before changing just a few lines becomes *multiple* lines.

***
<a href="https://pr-preview.s3.amazonaws.com/fdwr/webnn/pull/476.html" title="Last updated on Feb 13, 2024, 12:53 AM UTC (297396f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/476/55183f7...fdwr:297396f.html" title="Last updated on Feb 13, 2024, 12:53 AM UTC (297396f)">Diff</a>


zolkis:
Thanks for spotting / correcting that.

index.bs
```diff
@@ -4863,9 +4862,9 @@ partial interface MLGraphBuilder {
   <div class=algorithm-steps>
     1. [=Assert=]: the type of |input| is {{MLOperand}}.
     1. Let |outputShape| be an empty array of {{unsigned long}}.
-    1. If |newShape| is a scalar [=number=], set |outputShape| to `«  1  »`.
+    1. If |newShape| is a scalar [=number=], set |outputShape| to `«»`.
```
huningxin, line 4865: |newShape| should not be a number. We should assert that |newShape| is an array of unsigned long.

fdwr: Does `[=number=]` in this weird syntax refer to a JS `Number`? I presume that be changed to `{{unsigned long long}}` then (I don't get the significance of of `[[` vs `[=`, but `{{unsigned long long}}` is what I see being used elsewhere).

index.bs
```diff
@@ -1611,7 +1610,7 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
     </div>
     1. If |name| is empty, then [=exception/throw=] a {{TypeError}}.
     1. [=Assert=]: the type of |descriptor| is {{MLOperandDescriptor}}.
-    1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input.
+    1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input, equivalent to zero dimensions.
```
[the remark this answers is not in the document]
fdwr: Yeah, this seems like the central point to resolve them so such that the other functions just work.

index.bs
zolkis: Yes, `[=number=]` links to JS `Number`.  Looks like we need to formalize using `{{unsigned long long}}`, eventually with a typedef, and consider including it in normative steps as well, not only asserts (test anchors).

index.bs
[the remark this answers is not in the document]
zolkis: Is the assert enough here? Should we formalize this and describe exactly what should be done (e.g. setting defaults, internal slot, etc).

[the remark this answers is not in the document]
fdwr: > Should we set dimensions to an empty array of unsigned long 

🤔 How should we convey that the original tensor desc's dimensions are not *actually* modified, but rather the dimensions are *effectively* set? Make a copy of the tensor desc in the pseudocode?

> Should we formalize this and describe exactly what should be done

I'm unsure if you're asking me or Ningxin, but "Allow edits by maintainers" is on if there's any particular wording you want to amend. :)

index.bs
fdwr: > Yes, [=number=] links to JS Number

👍

> eventually with a typedef

I'll defer that to someone else's later change, as I want this change to be surgical.

index.bs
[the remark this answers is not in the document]
huningxin: > How should we convey that the original tensor desc's dimensions are not actually modified, but rather the dimensions are effectively set? Make a copy of the tensor desc in the pseudocode?

It should make a copy of the tensor desc. See https://github.com/webmachinelearning/webnn/issues/395

index.bs
```diff
@@ -1610,7 +1610,7 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
     </div>
     1. If |name| is empty, then [=exception/throw=] a {{TypeError}}.
     1. [=Assert=]: the type of |descriptor| is {{MLOperandDescriptor}}.
-    1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input, equivalent to zero dimensions.
+    1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input, and |descriptor|.{{MLOperandDescriptor/dimensions}} is set to an empty array of {{unsigned long long}}.
```
huningxin, line 1613: The dimensions setting operation looks like an algorithm step. I am not sure whether we should put it into an assertion. @inexorabletash, any insights?

inexorabletash: Agreed, this doesn't make sense as an assertion. But it also doesn't make sense as a step, since the following step is "if descriptor.dimensions exists:" which would never be false if this step set it to something. If this is intending to describe the behavior non-normatively, then a NOTE might be appropriate to explain subtle behavior. I'm not sure what's intended, so I'd revisit the whole algorithm to make sure it's coherent.

Also, at this stage in the processing, _dimensions_ would not be an array (a JS concept) or a sequence (an IDL concept) but a [list](https://infra.spec.whatwg.org/#lists) (from Infra).

inexorabletash: Thanks for clarifying! I think this is what you want as text:

> 1. If name is empty, then throw a TypeError.
> 2. If descriptor.dimensions does not exist, then set descriptor.dimensions to an empty list.
> 3. If the check dimensions steps given descriptor.dataType and descriptor.dimensions return false, then throw a "DataError" DOMException.
> 4. If the byte length of descriptor is not supported by the underlying platform, then throw a "DataError" DOMException.
> 5. If any of the following sub-steps fail....

(With all the markdown added back in!)

In case it isn't clear, the _"make an effective copy of the tensor description"_ is implicit with the spec processing model - the JS object passed as the MLOperandDescriptor dictionary is analyzed and all the fields are extracted into some implementation-specific representation (e.g. a C++ object), so the descriptor object is already a copy with modifications unobservable by script.

(Re: sequence/array/list - agreed, separate PR makes sense - cleanup-only PRs are easier to review and understand in the repo history!)

inexorabletash: (Note that in my example text I dropped the assertion about the input type that's currently step 2, tracked in #455 but it's harmless to leave it in for now.)

anssiko:
@fdwr noticed while checking the PR queue this needs a rebase. Good discussion in this PR should be factored in too I believe.

inexorabletash:
Other things that should be rolled into this PR:

- In [constant()](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-constant-value-type) _"Set descriptor.dimensions to undefined."_ - should set it to an empty list instead. And the note _"In the case of a scalar constant, descriptor.dimensions is ignored."_  should either be removed or updated; it's not that dimensions is _ignored_, it's that a rank of 0 _defines_ it as being a scalar.
- [validate MLOperand](https://webmachinelearning.github.io/webnn/#mloperand-validate-mloperand) has _"If desc.dimensions exists"_ - remove that clause?
- [lstm()](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-lstm) accesses dimensions[0] without a size / rank check first (in step 2) - add a check? I couldn't spot any other instances.
- Make the default for [MLOperandDescriptor](https://webmachinelearning.github.io/webnn/#api-mloperanddescriptor)'s dimensions be []

index.bs
fdwr: Thank you for the explicit snippets and links.

index.bs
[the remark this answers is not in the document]
fdwr: > It should make a copy of the tensor desc. See #395

**Ningxin**: Per @inexorabletash's comment below, it sounds like a copy is already implicit `In case it isn't clear, the "make an effective copy of the tensor description" is implicit with the spec processing model`.

index.bs
fdwr: Ningxin: Doh, merge conflict. Working on fixing this line. ⏳

fdwr:
There's a build error (`No 'idl' refs found for 'MLCommandEncoder'`) presumably from https://github.com/webmachinelearning/webnn/pull/546/files. I'll leave the status text there but just unbracket it to hopefully appease bikeshed.

index.bs
```diff
@@ -23,7 +23,7 @@ Logo: https://webmachinelearning.github.io/webmachinelearning-logo.png
 Deadline: 2023-10-01
 Status Text: <p>
   Further implementation experience and user feedback is being gathered for the
-  {{MLCommandEncoder}} interface that proposes to enable more efficient WebGPU
+  MLCommandEncoder interface that proposes to enable more efficient WebGPU
```
zolkis, line 26: If it were introduced in #322, we could leave it here, but it's not mentioned there, and MLCommandEncoder has been removed by another PR, so the name hangs in the void now. We could replace it with "an" (interface that...) or say "the/an experimental MLCommandEncoder interface"?

fdwr: @a-sully What do you say to me just deleting this paragraph for now?

a-sully: This paragraph is soon to be updated in #532, so IMHO any discussions about how exactly to describe our WebGPU interop efforts post-MLCommandEncoder should be had on that PR. For the sake of this PR... could we just this as-is? :)

inexorabletash: Leaving as-is and dropping the `{{}}` to make Bikeshed happy SGTM.

index.bs
```diff
@@ -1815,10 +1815,7 @@ Data truncation will occur when the specified value exceeds the range of the spe
     </div>
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
         1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |type|.
-        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to `undefined`.
-            <div class="note">
-                In the case of a scalar constant, |descriptor|.{{MLOperandDescriptor/dimensions}} is ignored.
-            </div>
+        1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the empty [=/list=] « ».
```
inexorabletash, line 1818: Nit: "an" instead of "the" and no need for « », so just:

`Set |descriptor|.{{MLOperandDescriptor/dimensions}} to an empty [=/list=].`

(Precedent: examples in Infra. Lists can mutate, unlike strings.)

index.bs
```diff
@@ -3997,6 +3994,7 @@ partial interface MLGraphBuilder {
         <div class="issue">
             The shape of |input|, |weight| or |recurrentWeight| could be also checked here.
         </div>
+    1. Require the shape of each tensor to match its expected rank, or else [=exception/throw=] a "{{DataError}}" {{DOMException}}: the rank of |input|, |weight|, |recurrentWeight|, |options|.{{MLLstmOptions/initialHiddenState}}, and |options|.{{MLLstmOptions/initialCellState}} is 3; and the rank of |options|.{{MLLstmOptions/bias}}, |options|.{{MLLstmOptions/recurrentBias}}, and |options|.{{MLLstmOptions/peepholeWeight}} is 2.
```
inexorabletash, line 3997: Some problems with this:
- These checks aren't guarded by "exists" check for the options members.
- The options members are checked below, which would make parts of the overall algorithm redundant.
- Prefer imperative style, i.e. "if... then..."

So I'd suggest copying what [gru()](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-gru)  has:

```
1. If the [=rank=] of |input| or |weight| or |recurrentWeight| is not 3, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
```

index.bs
```diff
@@ -3997,6 +3994,7 @@ partial interface MLGraphBuilder {
         <div class="issue">
             The shape of |input|, |weight| or |recurrentWeight| could be also checked here.
```
inexorabletash, line 3995: Remove this issue, as it's being fixed by the added check!

index.bs
```diff
@@ -5119,12 +5117,10 @@ partial interface MLGraphBuilder {
   </summary>
   <div class=algorithm-steps>
     1. Let |outputShape| be an empty array of {{unsigned long}}.
-    1. If |newShape| is a scalar [=number=], set |outputShape| to the [=/list=] « 1 ».
-    1. Otherwise, if |newShape| is an array of {{unsigned long}}:
-        1. If the [=list/size=] of |newShape| is 0, set |outputShape| to the [=/list=] « 1 » (reshaping to scalar).
-        1. If any value in |newShape| is 0, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-        1. Let |inputElementCount| be the product of all elements in |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
-        1. If product of all values in |newShape| is not equal to |inputElementCount|, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. If the [=list/size=] of |newShape| is 0, set |outputShape| to the empty [=/list=] « » (reshaping to scalar).
```
inexorabletash, line 5120: Per above, prefer "an empty list" and drop the « ».

Also the "(reshaping to scalar)" seems like a non-normative comment. Maybe make this an actual Note? (I don't feel strongly about this.)

index.bs
```diff
@@ -5119,12 +5117,10 @@ partial interface MLGraphBuilder {
...
+    1. If any value in |newShape| is 0, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |inputElementCount| be the product of all elements in |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
```
inexorabletash, line 5122: The product for scalar seems undefined, in this step and the next one.

inexorabletash (changes requested):

index.bs
zolkis: WFM

index.bs
```diff
@@ -6058,7 +6054,7 @@ dictionary MLOperandDescriptor {
   required MLOperandDataType dataType;
 
   // The dimensions field is only required for tensor operands.
-  sequence<unsigned long> dimensions;
+  sequence<unsigned long> dimensions = [];
```
zolkis, line 6057: We need to follow up its checks against empty array vs against "exists"?

inexorabletash: FYI: My audit turned up https://github.com/webmachinelearning/webnn/pull/476#issuecomment-1934885013 which have all been tackled. But please look for more!

index.bs
fdwr: I'll update it to:

`1. Let |inputElementCount| be the product of all elements in |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}. An |inputElementCount| of 0 yields a product of 1.`

In pseudo code, it would be:

```
product = 1
for each dimension in dimensions
    product *= dimension
endfor
```

index.bs
fdwr: Fine with dropping the chevrons. Was following the precedent I saw elsewhere (e.g. `set it to the [=/list=] « 0, 0, 0, 0 »`).

Deparenthesizing to `set |outputShape| to an empty [=/list=] for a scalar`.

index.bs
fdwr: The GRU wording is indeed sufficient an consistent.

index.bs
fdwr: Thank you for the audit Josh. I want to push this through since some of Josh's other changes interact with it, but I'll need to sweep the entire document anyway for coherency, as I've noticed a few changes needed since the transformers change (e.g. the fill sequence `constant` overload is absent from the IDL https://www.w3.org/TR/webnn/#idl-index).

inexorabletash (approved):
LGTM!

zolkis (approved):
Thanks, looks good!

index.bs
fdwr: @huningxin Ningxin, if you're out for Lunar New Year (🏮🎇) until the 24th, I'm considering completing this since I believe all feedback is addressed now, and I can create another CR for any smaller new feedback.

PR #491, opened 2023-12-14
Fix a handful of markdown glitches, probably c/o a find/replace.

inexorabletash:
* Missing spaces in "itthrows"
* "lensizeth" ("size" appearing in the middle of "length")

That said, definition of "bias" is a little sketchy: 

> An MLOperand. Specifies the 1-D tensor of the bias values whose size is equal to the size of the feature dimension of the input. For example, for an input tensor with nchw layout, the size is the value of input.[[descriptor]].dimensions[1].

"size is the value..." reads like it's trying to define size, so a link to Infra's list/size doesn't make sense. I think this is trying to say that the size of the passed bias (a 1-D tensor) must match the dimension of the input, but without phrasing it as normative behavior. 

Maybe this should just be:

> An MLOperand. Specifies the 1-D tensor of the bias values whose size is equal to input.[[descriptor]].dimensions[1].

(This still reads as something informative rather than normative, and I don't see where the behavior would be enforced, but that's probably out of scope for a glitch fix.)

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/491.html" title="Last updated on Dec 15, 2023, 5:25 PM UTC (50ea24b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/491/6ce91ac...inexorabletash:50ea24b.html" title="Last updated on Dec 15, 2023, 5:25 PM UTC (50ea24b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1106,7 +1106,7 @@ When the {{[[contextType]]}} is set to [=default-context|default=] with the {{ML
 </details>
 
 ### Synchronous Execution ### {#api-mlcontext-sync-execution}
-Synchronously carries out the computational workload of a compiled graph {{MLGraph}} on the calling thread, which must be a worker thread, to produce results as defined by the operations in the graph. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it[=exception/throws=] an "{{OperationError}}" {{DOMException}}.
+Synchronously carries out the computational workload of a compiled graph {{MLGraph}} on the calling thread, which must be a worker thread, to produce results as defined by the operations in the graph. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it [=exception/throws=] an "{{OperationError}}" {{DOMException}}.
 
 <script type=idl>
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM, thanks for the fix!

huningxin:
> An MLOperand. Specifies the 1-D tensor of the bias values whose size is equal to input.[[descriptor]].dimensions[1].

"size is equal to" sounds good. But its size doesn't always match `input.[[descriptor]].dimensions[1]`. It depends on which dimension is the feature/channel dimension. For "nchw" input layout, it is `input.[[descriptor]].dimensions[1]`, while for "nhwc" input layout, it is `input.[[descriptor]].dimensions[3]`.

inexorabletash:
PR updated with a change from "size is the value of" to "size is equal to" in two places.

@huningxin can you take another look? Thanks for your patience!

huningxin (approved):
LGTM, thanks!

Issue #497, opened 2024-01-05
Define power preference and device type in terms of WebIDL enums


inexorabletash:
Right now there are definitions for MLContext's [device type](https://www.w3.org/TR/webnn/#device-type) and [power preference](https://www.w3.org/TR/webnn/#power-preference) which are used internally only, but also the WebIDL-defined [MLDeviceType](https://www.w3.org/TR/webnn/#enumdef-mldevicetype) and [MLPowerPreference](https://www.w3.org/TR/webnn/#enumdef-mlpowerpreference). These enumerated types have equivalent values, and the [create a context](https://www.w3.org/TR/webnn/#create-a-context) algorithm relies on these being identical (e.g. in _set context.[[powerPreference]] to options["powerPreference"]_ where the source is a IDL enum value and the target is a internal value)

The spec could be simplified a bit by using the enum values internally as well, and the internal enum value definitions recycled as documentation for the WebIDL enum values.

(Note that this doesn't apply to the [context type](https://www.w3.org/TR/webnn/#context-type) internal enum, as that is never script-exposed)

If there's interest (or if it'd make the spec clearer) I can put up a PR for this.

inexorabletash:
I should note that I ran across this while exploring #498 - the simplifications suggested there leave the internal enum definitions unreferenced.

PR #495, opened 2023-12-19
Adopt Bikeshed best practices and Infra spec conventions

inexorabletash:
A bunch of relatively stand-alone commits addressing issues noted in https://github.com/webmachinelearning/webnn/issues/450

It'll be easiest to review each individual commit as they're each a separate "topic". Reviewers should feel free to cherry-pick changes if anything is deemed controversial.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Dec 19, 2023, 6:48 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fwebmachinelearning%2Fwebnn%2Fd5bee458649f3d6f8acf84fa64596d12a66a57e2%2Findex.bs&force=1&md-warning=not%20ready)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20webmachinelearning/webnn%23495.)._
</details>

PR #499, opened 2024-01-06
Stylistic change: Replace some explicit use of link target naming with scopes

inexorabletash:
Simplify explicit link targets (e.g. power-preference-default) and link naming (e.g. default) with making the values scoped to a definition. Let Bikeshed take care of the linking, which also helps catch errors.

This PR layers on top of PR #495 since they touch similar lines, although they're conceptually separate. So only review the last commit here. Sorry about that!

This PR would also be partially obsoleted if we agree on #497 but it's harmless to merge as-is.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/499.html" title="Last updated on Jan 18, 2024, 6:51 PM UTC (b154e54)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/499/50a00b9...inexorabletash:b154e54.html" title="Last updated on Jan 18, 2024, 6:51 PM UTC (b154e54)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -793,13 +793,13 @@ Its <a>default allowlist</a> is <code>'self'</code>.
 <div class=algorithm-steps>
     1. Let |context| be a new {{MLContext}} object.
     1. If |options| is a {{GPUDevice}} object,
-        1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
-        1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
-        1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+        1. Set |context|.{{[[contextType]]}} to "[=context type/webgpu=]".
+        1. Set |context|.{{[[deviceType]]}} to "[=device type/gpu=]".
+        1. Set |context|.{{[[powerPreference]]}} to "[=power preference/default=]".
     1. Otherwise,
-        1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
-        1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device-type-cpu|cpu=]".
-        1. If |options|["{{powerPreference}}"] [=map/exists=], then set |context|.{{[[powerPreference]]}} to |options|["{{powerPreference}}"]. Otherwise, set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+        1. Set |context|.{{[[contextType]]}} to "[=context type/default=]".
+        1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device type/cpu=]".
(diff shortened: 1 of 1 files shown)


zolkis (commented):
I could not spot an error after two reads. 
@huningxin PTAL, you might have better glasses. :) 

As a general comment, we need to think if want to capture something more when changing constructs like

> Store a reference to |outputImpl| in |output|.{{MLOperand/[[operand]]}}

to 

> Set |output|.{{MLOperand/[[operand]]}} to |outputImpl|.

huningxin (approved):
LGTM, thanks @inexorabletash !

@anssiko and @wchao1115 , please also take a look.

anssiko (approved):
The last and relevant commit https://github.com/webmachinelearning/webnn/pull/499/commits/fe38e88adec3abf808b0f78cc7c323826b0c7229 LGTM, thanks!

inexorabletash:
Rebased so it's just got the one commit.

wchao1115:
I don't mind this change but I'm curious about how to stay consistent throughout. It looks like this change replaces a few `-` with a whitespace. Is this what we want to go about identifier naming of the links going forward? How do we know if these are the only places in the doc that need this treatment?

inexorabletash:
So this comes down to a bit of a philosophical stance. I find that any time something explicit is used - like link target (lt) attributes, or a named identifier - rather than relying on the implicit name and Bikeshed's smarts, it requires extra typing and extra thought to maintain the spec.

There's nothing inherently wrong with the longer form - and in cases like providing an explicit scope for links I'm definitely a fan, but the cases tackled in this CL don't seem to add value. 

Also... the changes proposed in #497 will actually overwrite several of the changes here with even more concise markdown, if that gets consensus.

inexorabletash:
... but if we want to keep e.g. `context-type` as the internal ID I'm happy do update the PR. This would change `<dfn>context type</dfn> ` to `<dfn data-lt="context-type">context type</dfn> `  and then the scoped links become `[=context-type/default=]` and `[=context-type/webgpu=]`. 

The main intent of the PR was the scoping, so I'm happy either way.

wchao1115:
@inexorabletash I have no problem with that reasoning but my question is more towards how to keep the exact naming convention you're envisioning consistent throughout the history of this spec and going forward given multiple contributors. 

This is not different from defining a naming convention in a code project. Since naming consistency affects readability, it is most common for the project's contributors to agree on it once and for all, then codify it in writing e.g. NamingConvention.md, etc. Without such an agreement, next time someone makes a change, it'll break consistency again.

inexorabletash:
Ah, got it. The easy answer is: you don't invent IDs -- unless absolutely necessary, which is rare and usually means you're doing something wrong. If the spec text is "context type", that's what you link to or use as the scope. If the spec text is "default" ensure that it is defined within a scope and link to the scoped term. 

I think some of the previous confusion came from having multiple "default" terms that weren't scoped so you needed to invent unique IDs for them, like "context-type-default". If the terms are properly scoped then you can just link to "context type/default" without having to define any arbitrary IDs.

wchao1115:
Make sense. Would you take up my request to write a simple BikeShedIDNamingConvention.md (or feel free to pick a different file name) to describe how ID naming should be done in different situations with scoping sample? That way we can just have the existing Contributing.md pointing to it, as a convention for ID naming to avoid repeating the same mistake going forward. What do you think?

inexorabletash:
For completeness: text about scoping and link targets was included in the initial landing of `SpecCodingConventions.md` so this should be good to go.

PR #505, opened 2024-01-12
Bug fix: Remove null value support in reshape()'s newShape parameter

### LINKED ISSUE #388: Remove the support of `null` value in `newShape` parameter of `reshape` build method
WebNN [`reshape`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-reshape) op build method currently supports a special `null` value in `newShape` array that instructs the implementation to compute the size of the dimension with the value `null` so that the total size remains constant. 

In PR review https://github.com/webmachinelearning/webnn/pull/367#issuecomment-1498242058, @fwdr proposed WebNN should only support the concrete values in newShape because this can be easily handled by framework through shape inference. Thanks Dwyane!

> Though also, is this some behavior that could be easily resolved by the frameworks calling into WebNN, since presumably they already know the shapes along the way (shape inference) and could instead pass concrete values to WebNN?

I think this would also help simplify the implementation.

/cc @wacky6

inexorabletash:
As noted in issue #388, the reshape op build method currently supports a special null value in newShape array that instructs the implementation to compute the size of the dimension with the value null so that the total size remains constant. @fwdr proposed WebNN should only support the concrete values in newShape because this can be easily handled by framework through shape inference

It appears this change has already landed in the Chromium implementation and WPTs.

Fixes #388

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/505.html" title="Last updated on Jan 18, 2024, 6:49 PM UTC (ab36f0d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/505/50a00b9...inexorabletash:ab36f0d.html" title="Last updated on Jan 18, 2024, 6:49 PM UTC (ab36f0d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5316,7 +5316,7 @@ partial interface MLGraphBuilder {
 Alter the shape of a tensor to a new shape. Reshape does not copy or change the content of the tensor. It just changes the tensor's logical dimensions for the subsequent operations.
 <script type=idl>
 partial interface MLGraphBuilder {
-  MLOperand reshape(MLOperand input, sequence<unsigned long?> newShape);
+  MLOperand reshape(MLOperand input, sequence<unsigned long> newShape);
 };
 </script>
 <div>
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM, thanks for the fix!

inexorabletash:
Rebased - Preview | Diff now available.

anssiko (approved):

wchao1115 (approved):

PR #511, opened 2024-01-17
Wording change: Define power prefs & device type using WebIDL enums

### LINKED ISSUE #497: Define power preference and device type in terms of WebIDL enums

inexorabletash:
Replace definitions for MLContext's device type and power preference which are used internally only, with the identical WebIDL-defined MLDeviceType and MLPowerPreference enums.

Fixes #497

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Jan 17, 2024, 8:38 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fwebmachinelearning%2Fwebnn%2F69aeffa6d777f6a025f05a1edd028a6d3f9fe351%2Findex.bs&force=1&md-warning=not%20ready)

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20webmachinelearning/webnn%23511.)._
</details>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -795,12 +795,12 @@ Its <a>default allowlist</a> is <code>'self'</code>.
     1. Let |context| be a new {{MLContext}} object.
     1. If |options| is a {{GPUDevice}} object,
         1. Set |context|.{{[[contextType]]}} to "[=webgpu-context|webgpu=]".
-        1. Set |context|.{{[[deviceType]]}} to "[=device-type-gpu|gpu=]".
-        1. Set |context|.{{[[powerPreference]]}} to "[=power-preference-default|default=]".
+        1. Set |context|.{{[[deviceType]]}} to {{MLDeviceType/"gpu"}}.
+        1. Set |context|.{{[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
     1. Otherwise,
         1. Set |context|.{{[[contextType]]}} to "[=default-context|default=]".
-        1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device-type-cpu|cpu=]".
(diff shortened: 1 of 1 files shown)


zolkis:
Looks good, thanks.

inexorabletash:
Apologies, it looks like I'd pushed an incomplete version of this PR. Updated to actually update device type too.

Also, since this will conflict with #499 I'm going to close this for now and will re-open if/when that lands, so make things easier for reviewers/editors.

Issue #450, opened 2023-08-15
Use web spec best practices


zolkis:
_Originally posted by @inexorabletash in https://github.com/webmachinelearning/webnn/pull/446#discussion_r1294860601_
            
> Drive by comment: the first step here (_If a or b is not an instance of MLOperand, then throw a "TypeError"..._)  is unnecessary, and I see this repeated throughout the newly introduced algorithms for methods ("operations" in Web IDL). The Web IDL bindings already take care validating input types and throwing TypeError.

> Further, even in cases where an explicit error must be thrown, the intrinsic type TypeError should be thrown, not a DOMException of type "TypeError" (which is not present in https://webidl.spec.whatwg.org/#idl-DOMException-error-names)

(Issue created to document this change).

inexorabletash:
Another example: 

in `constant()` - _If type is undefined, let type be "float32"._ - this is handled automatically by having the default in the Web IDL.

inexorabletash:
Also (and this is not new), dictionary members are more normally references in prose using the infra ordered map syntax: _options_["`foo`"] 

By the time an algorithm "sees" the dictionary, it's gone through the bindings layer and is an ordered map - https://webidl.spec.whatwg.org/#idl-dictionaries

inexorabletash:
> throw a "TypeError" simple exception 

Just write: throw a TypeError. 

In bikeshed markdown:
```
... [=exception/throw=] a {{TypeError}}.
```

inexorabletash:
And (this new is in PR 446) the entire "check linear options" algorithm is unnecessary; the dictionary bindings for ECMAScript take care of all of that, including default values.

inexorabletash:
(apologies for random feedback; I may have more time later this week to do a more thorough pass)

inexorabletash:
Not new: In "create MLActivation" steps

> If name is undefined or null, throw a "TypeError" and abort these steps.

It looks like all call sites are internal to the spec and so this could just be an assert.

And this is more off topic, but: none of the call sites for "create MLActivation" actually pass _builder_ - that should be fixed at the call sites. I also don't see any of the call sites passing _init-steps_ but maybe I'm missing something.

inexorabletash:
Not new: in input(name, descriptor):

> If name is undefined or an empty string, then throw a TypeError and stop.

`name` is defined as DOMString; passing literal `undefined` from script will yield the string `"undefined"`. So this can be reduced to just a test for an empty string.

[...]

PR #512, opened 2024-01-17
Wording change: New up a specific type, rather than just object

inexorabletash:
In algorithms that mint MLActivations and MLOperands, construct the actual type not a generic "object". Also, use Infra wording around optional paramters to those algorithms.

For #450


huningxin (approved):
LGTM, thanks!

anssiko (approved):
Thanks for this strict improvement.

inexorabletash:
Rebased - Preview | Diff now available.

wchao1115 (approved):

PR #519, opened 2024-01-18
Wording change: Streamline context creation and validation algorithms

### LINKED ISSUE #498: Streamline context creation and validation algorithm definitions
A handful of observations:

* The [validate MLContext](https://www.w3.org/TR/webnn/#validate-mlcontext) algorithm has 3 steps that confirm that the internal slots are valid. This could only fail if the spec was internally inconsistent or an implementation was non-compliant, so these don't add value to the spec. See also #455 So there's really only a single useful sentence in the algorithm: _If the user agent cannot support context.[[contextType]], context.[[deviceType]] and context.[[powerPreference]], return false._ 
* The [validate MLContext](https://www.w3.org/TR/webnn/#validate-mlcontext) algorithm is invoked in the [MLGraphBuilder constructor](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-mlgraphbuilder) where an MLContext is passed. This should not be necessary as the MLContext is already created, so there's no way for a non-valid context to be passed in.
* All other invocations of [validate MLContext](https://www.w3.org/TR/webnn/#validate-mlcontext) immediately follow an invocation of [create a context](https://www.w3.org/TR/webnn/#create-a-context). 

Synthesizing the above, we could eliminate [validate MLContext](https://www.w3.org/TR/webnn/#validate-mlcontext) and instead have [create a context](https://www.w3.org/TR/webnn/#create-a-context) include:

> If the user agent cannot support context.[[contextType]], context.[[deviceType]] and context.[[powerPreference]], return failure.

And then the four call sites could simplify to:

> Let context be the result of creating a context given options. If that returns failure, then reject promise with a "NotSupportedError" DOMException.

(All four are slightly different, given sync/async failure and options vs. gpuDevice arguments, but would simplify similarly.)

I'm happy to put up a PR for this. On the other hand, if I'm making a bad assumption somewhere above (e.g. construction of a non-valid context is possible), then we should add some explanatory notes into the spec!

inexorabletash:
Bundle validation right into context creation, rather than making it a separate step.

Fixes #498

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/519.html" title="Last updated on Jan 25, 2024, 5:53 PM UTC (99ca098)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/519/7773794...inexorabletash:99ca098.html" title="Last updated on Jan 25, 2024, 5:53 PM UTC (99ca098)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -800,6 +800,7 @@ Its <a>default allowlist</a> is <code>'self'</code>.
         1. Set |context|.{{[[contextType]]}} to "[=context type/default=]".
         1. If |options|["{{deviceType}}"] [=map/exists=], then set |context|.{{[[deviceType]]}} to |options|["{{deviceType}}"]. Otherwise, set |context|.{{[[deviceType]]}} to "[=device type/cpu=]".
         1. If |options|["{{powerPreference}}"] [=map/exists=], then set |context|.{{[[powerPreference]]}} to |options|["{{powerPreference}}"]. Otherwise, set |context|.{{[[powerPreference]]}} to "[=power preference/default=]".
+    1. If the user agent cannot support |context|.{{[[contextType]]}}, |context|.{{[[deviceType]]}} and |context|.{{[[powerPreference]]}}, return failure.
     1. Return |context|.
 </div>
 </details>
@@ -812,8 +813,7 @@ Its <a>default allowlist</a> is <code>'self'</code>.
     1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" {{DOMException}}.
     1. Let |promise| be [=a new promise=].
     1. Return |promise| and run the following steps [=in parallel=].
-    1. Let |context| be the result of [=creating a context=] given |options|.
-    1. If [=validating MLContext=] given |context| returns false, [=reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}}.
+    1. Let |context| be the result of [=creating a context=] given |options|. If that returns failure, then [=reject=] |promise| with a "{{NotSupportedError}}" {{DOMException}}.
     1. [=Resolve=] |promise| with |context|.
 </div>
 </details>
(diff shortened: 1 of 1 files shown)


inexorabletash:
Note that there's still discussion in #498 (e.g. about the concept of context validity) so this is probably not a slam dunk to merge, but I wanted the artifact up to spur the discussion.

zolkis (approved):

index.bs
```diff
@@ -1066,9 +1063,9 @@ The <dfn>power preference</dfn> indicates preference as related to power consump
 <dl>
 <dt>"<code><dfn data-lt="power-preference-default">default</dfn></code>"</dt>
 <dd>Let the user agent select the most suitable behavior.</dd>
-<dt>"<code><dfn data-lt="power-preference-high-performance">high-performance</dfn></code>"</dt>
+<dt>"<code><dfn ignore data-lt="power-preference-high-performance">high-performance</dfn></code>"</dt>
```
huningxin, line 1066: What's the intention of adding "ignore"? Its usage is missed in the commit log?

inexorabletash: That's to silence a warning... because with the "validate MLContext" algorithm is removed, the enum member is never actually referenced in the spec! That's because there's an implicit conversion of the enum value to the corresponding `MLPowerPreference` enum value. The proposed follow-on work discussed in #497 would dedupe the enums, and turn this definition into a definition for the `MLPowerPreference` enum, and "ignore" can go away.

I'll update the commit log to capture this.

huningxin: Thanks for the explanation. The updated commit log looks good.

huningxin (approved):
LGTM

inexorabletash:
Rebased, should be ready to go now.

anssiko (approved):

wchao1115:
@inexorabletash Just checking if you can merge the change yourself, or whether you need me to click the button for you.

inexorabletash:
> @inexorabletash Just checking if you can merge the change yourself, or whether you need me to click the button for you.

Nope, I can't merge - someone with ["Write"](https://docs.github.com/en/get-started/learning-about-github/access-permissions-on-github) permissions needs to do it. I don't even have "Triage" !

inexorabletash:
To be explicit, @wchao1115 - a click on this and any other "approved" PRs with no open discussion would be very appreciated.

Issue #470, opened 2023-10-18
1-D input tensors for `matmul` op are not widely supported by native ML APIs


huningxin:
(raised by @wacky6 Jiewei in Chromium CL review https://chromium-review.googlesource.com/c/chromium/src/+/4940628/comments/88fd209e_269f2c93)

[WebNN's matmul](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-matmul) supports 1-D input tensors, in particular  for
```
c = builder.matmul(a, b)
```

> If a is 1-dimensional, it is converted to a 2-dimensional tensor by prepending a 1 to its dimensions.
> If b is 1-dimensional, it is converted to a 2-dimensional tensor by by appending a 1 to its dimensions.
> If both a and b are 1-dimensional, the operation is a vector dot-product, which produces a scalar output.

This feature aligns with Pytorch ([`torch.matmul`](https://pytorch.org/docs/stable/generated/torch.matmul.html)) and ONNX ([`MatMul`](https://onnx.ai/onnx/operators/onnx__MatMul.html)).

However, 1-D input tensors are not widely supported by native ML APIs:
* DirectML's [`DML_GEMM_OPERATOR_DESC`](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_gemm_operator_desc) supports input tensors rank from 2 to 4.
* BNNS's [`BroadcastMatMul`](https://developer.apple.com/documentation/accelerate/3546628-bnnsfiltercreatelayerbroadcastma) requires input tensors rank >= 2.
* TensorFlow's [`BatchMatMulV2`](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/batch-mat-mul-v2) which TensorFlow-Lite and NNAPI follows, requires input tensors x and y are 2-D or higher.

The open is whether WebNN matmul should drop the support of 1-D input tensors. This would help simplify the implementation. Frameworks can still support 1-D input tensor by reshaping those tensors to 2-D with prepending or appending 1 to its dimensions, as ONNXRuntime DirectML EP [`MatMulShapeMapping()`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/dml/OperatorAuthorHelper/OperatorHelper.cpp#L555) does.

/cc @wchao1115 @fdwr

fdwr:
> Frameworks can still support 1-D input tensor by reshaping those tensors to 2-D

[...]

PR #523, opened 2024-01-20
Wording change: Tighten up output shape calculation algorithms

inexorabletash:
For gemm() - make lists that are mutated be explicit clones using Infra terminology, and simplify the wording reversing the lists.

For matmul() - make lists that are mutated be explicit clones using Infra terminology, use append/prepend definitions from Infra, convert a variable change from "let" to "set" and drop use of "array".

For #395

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/523.html" title="Last updated on Feb 14, 2024, 6:39 AM UTC (9deaeb5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/523/8503e24...inexorabletash:9deaeb5.html" title="Last updated on Feb 14, 2024, 6:39 AM UTC (9deaeb5)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4354,11 +4354,9 @@ partial interface MLGraphBuilder {
     1. Let |colsB| be |shapeB|[|sizeB| - 1].
     1. Let |rowsB| be |shapeB|[|sizeB| - 2].
     1. If |colsA| is not equal to |rowsB|, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-    1. Let |slicedA| be a [=list/clone=] of |shapeA| with the last 2 items [=list/removed=].
-    1. Let |slicedB| be a [=list/clone=] of |shapeB| with the last 2 items [=list/removed=].
-    1. If both |sizeA| and |sizeB| are greater than 2, then let |outputShape| be the result of [=bidirectionally broadcasting the shapes=] |slicedA| and |slicedB|. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-    1. Otherwise, if both |sizeA| and |sizeB| are equal to 2, then let |outputShape| be an empty [=/list=].
-    1. Otherwise, let |outputShape| be the [=list/size|longer=] of |slicedA| or |slicedB|.
+    1. Let |batchShapeA| be a [=list/clone=] of |shapeA| with the spatial dimensions (last 2 items) [=list/removed=].
+    1. Let |batchShapeB| be a [=list/clone=] of |shapeB| with the spatial dimensions (last 2 items) [=list/removed=].
+    1. Let |outputShape| be the result of [=bidirectionally broadcasting the shapes=] |batchShapeA| and |batchShapeB|. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. [=list/Append=] « |rowsA|, |colsB| » to |outputShape|.
     1. Return |outputShape|.
   </div>
--- a/index.bs
+++ b/index.bs
@@ -4326,8 +4326,8 @@ partial interface MLGraphBuilder {
 
 <div>
     **Arguments:**
-        - *a*: an {{MLOperand}}. The first N-dimensional input tensor.
-        - *b*: an {{MLOperand}}. The second N-dimensional input tensor.
+        - *a*: an {{MLOperand}}. The first input tensor which is at least 2-D.
+        - *b*: an {{MLOperand}}. The second input tensor which is at least 2-D.
 
     **Returns:** an {{MLOperand}}. The output tensor that contains the matrix
     product of two input tensors.
@@ -4337,9 +4337,6 @@ partial interface MLGraphBuilder {
         - If both *a* and *b* are 2-dimensional, they are multiplied like conventional
             matrices and produce a 2-dimensional tensor as the output.
         - If either *a* or *b* is `N`-dimensional where `N > 2`, it is treated as a stack of matrices with dimensions corresponding to the last two indices. The matrix multiplication will be broadcasted accordingly by following the [[!numpy-broadcasting-rule]]. The output is a `N`-dimensional tensor whose rank is the maximum [=rank=] of the input tensors. For each dimension, except the last two, of the output tensor, its size is the maximum size along that dimension of the input tensors.
-        - If *a* is 1-dimensional, it is converted to a 2-dimensional tensor by prepending a 1 to its dimensions.
-        - If *b* is 1-dimensional, it is converted to a 2-dimensional tensor by by appending a 1 to its dimensions.
-        - If both *a* and *b* are 1-dimensional, the operation is a vector dot-product, which produces a scalar output.
 </div>
 
 <details open algorithm>
@@ -4351,16 +4348,19 @@ partial interface MLGraphBuilder {
     1. Let |sizeA| be the [=list/size=] of |shapeA|.
     1. Let |shapeB| be a [=list/clone=] of |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
     1. Let |sizeB| be the [=list/size=] of |shapeB|.
-    1. If |sizeA| and |sizeB| is 1, return «  1  ».
-    1. If |sizeA| is 1, then [=list/prepend=] 1 to |shapeA| and set |sizeA| to 2.
-    1. If |shapeA|[0] is not equal to |shapeB|[|sizeB| - 2], then [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
-    1. If |sizeB| is 1, then [=list/append=] 1 to |shapeB| and set |sizeB| to 2.
-    1. If |shapeA|[|sizeA| - 1] is not equal to |shapeB|[0], then [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
-    1. Let |size| be the maximum of |sizeA| and |sizeB|.
-    1. Let |shape| be a new [=/list=].
-    1. [=list/For each=] |index| in [=the range=] 0 to |size|, exclusive:
-        1. [=list/Append=] the maximum of |shapeA|[|index|] and |shapeB|[|index|] to |shape|.
-    1. Return |shape|.
+    1. If either |sizeA| or |sizeB| is less than 2, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |colsA| be |shapeA|[|sizeA| - 1].
+    1. Let |rowsA| be |shapeA|[|sizeA| - 2].
+    1. Let |colsB| be |shapeB|[|sizeB| - 1].
+    1. Let |rowsB| be |shapeB|[|sizeB| - 2].
+    1. If |colsA| is not equal to |rowsB|, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |slicedA| be a [=list/clone=] of |shapeA| with the last 2 items [=list/removed=].
+    1. Let |slicedB| be a [=list/clone=] of |shapeB| with the last 2 items [=list/removed=].
+    1. If both |sizeA| and |sizeB| are greater than 2, then let |outputShape| be the result of [=bidirectionally broadcasting the shapes=] |slicedA| and |slicedB|. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Otherwise, if both |sizeA| and |sizeB| are equal to 2, then let |outputShape| be an empty [=/list=].
+    1. Otherwise, let |outputShape| be the [=list/size|longer=] of |slicedA| or |slicedB|.
+    1. [=list/Append=] « |rowsA|, |colsB| » to |outputShape|.
+    1. Return |outputShape|.
   </div>
 </details>
 
--- a/index.bs
+++ b/index.bs
@@ -2892,11 +2892,13 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>gemm(|a|, |b|, |options|)</dfn> method steps are:
   </summary>
   <div class=algorithm-steps>
-    1. Let |shapeA| be |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |sizeA| the [=list/size=] of |shapeA|.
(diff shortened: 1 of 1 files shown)


inexorabletash:
For matmul(), the current text is:

> If shapeA[0] is not equal to shapeB[sizeB - 2], then throw an "OperationError" DOMException.
> If shapeA[sizeA - 1] is not equal to shapeB[0], then throw an "OperationError" DOMException.

Should the former be `- 1` as well? If not, what happens if sizeB is < 2 ?

zolkis (commented):
Looks good. Thanks.

index.bs
```diff
@@ -3084,11 +3084,13 @@ partial interface MLGraphBuilder {
...
-    1. Let |shapeA| be |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |sizeA| the [=list/size=] of |shapeA|.
-    1. Let |shapeB| be |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |sizeB| the [=list/size=] of |shapeB|.
+    1. Let |shapeA| be a [=list/clone=] of |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+    1. Let |sizeA| be the [=list/size=] of |shapeA|.
+    1. Let |shapeB| be a [=list/clone=] of |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+    1. Let |sizeB| be the [=list/size=] of |shapeB|.
     1. If |sizeA| is not 2 or |sizeB| is not 2, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-    1. If |options|.{{MLGemmOptions/aTranspose}} is true, then let |shapeA| be the reverse array of |shapeA|.
-    1. If |options|.{{MLGemmOptions/bTranspose}} is true, then let |shapeB| be the reverse array of |shapeB|.
+    1. If |options|.{{MLGemmOptions/aTranspose}} is true, then reverse the order of the items in |shapeA|.
```
huningxin, line 3092: Should Infra define "reverse the order of the items" algorithm?

index.bs
```diff
@@ -4571,16 +4573,19 @@ partial interface MLGraphBuilder {
...
     1. If |shapeA|[0] is not equal to |shapeB|[|sizeB| - 2], then [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
-    1. If |sizeB| is 1 and |sizeA| is not, then append 1 to |shapeB| to become [ |shapeB| | 1 ] and let |sizeB| be 2.
+    1. If |sizeB| is 1, then [=list/append=] 1 to |shapeB| and set |sizeB| to 2.
     1. If |shapeA|[|sizeA| - 1] is not equal to |shapeB|[0], then [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
-    1. Let |shape| be an array whose size |size| is the maximum of |sizeA| and |sizeB|.
+    1. Let |size| is the maximum of |sizeA| and |sizeB|.
+    1. Let |shape| be a new [=/list=].
     1. [=map/For each=] |index| in [=the range=] 0 to |size|, exclusive:
-        1. Set |shape|[|index|] to the maximum of |shapeA|[|index|] and |shapeB|[|index|].
+        1. [=list/Append=] the maximum of |shapeA|[|index|] and |shapeB|[|index|] to |shape|.
```
huningxin, line 4588: Nit: [=list/append=] for consistency?

huningxin: Would `index` be out-of-bounds if the size of a particular shape is less than `size`? 

And I think the previous algorithm itself has issue. If the size of a shape is less than maximum size, this shape should be prepended 1 until its size is equal to maximum size before entering this loop.

index.bs
```diff
@@ -4571,16 +4573,19 @@ partial interface MLGraphBuilder {
...
+    1. If |sizeA| and |sizeB| is 1, return «  1  ».
+    1. If |sizeA| is 1, then [=list/prepend=] 1 to |shapeA| and set |sizeA| to 2.
     1. If |shapeA|[0] is not equal to |shapeB|[|sizeB| - 2], then [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
-    1. If |sizeB| is 1 and |sizeA| is not, then append 1 to |shapeB| to become [ |shapeB| | 1 ] and let |sizeB| be 2.
+    1. If |sizeB| is 1, then [=list/append=] 1 to |shapeB| and set |sizeB| to 2.
     1. If |shapeA|[|sizeA| - 1] is not equal to |shapeB|[0], then [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
-    1. Let |shape| be an array whose size |size| is the maximum of |sizeA| and |sizeB|.
+    1. Let |size| is the maximum of |sizeA| and |sizeB|.
+    1. Let |shape| be a new [=/list=].
     1. [=map/For each=] |index| in [=the range=] 0 to |size|, exclusive:
```
huningxin, line 4587: Should it use [=list/iterate=]?

index.bs
inexorabletash: Possibly; I'm not sure how often it occurs across specs. Usually for Infra the policy is to incorporate algorithms like that once they're used in multiple specs.

I replaced the text here because "the reverse array" sounds like a term that should be introduced, whereas "reverse the order of the items" seems like a fairly straightforward explanation.

index.bs
inexorabletash: Oh, good catch. There are several places where map/For each is used on lists. I'll spin up a new PR for those.

inexorabletash: Hmmm, although they're not lists either, just a basic "for" loop. I don't think there's a specific term for those, so I'll unlink (in a new PR)

inexorabletash: Tackled in #527

index.bs
huningxin: Thanks for the explanation. "reverse the order of the items" sounds good to me.

index.bs
inexorabletash: Re: previous algorithm - agreed, this looks "sus".

With the caveat that I'm still new to tensors, my intuition at least according to the prose ("If either a or b is N-dimensional where N > 2, it is treated as a stack of matrices with dimensions corresponding to the last two indices...") is that we need the end result to be:

* sizeA == sizeB (achieved by prepending - how does appending come in???)
* sizeA > 1 (via appending/prepending; alternately we reject per #470)
* sizeB > 1 (via appending/prepending; alternately we reject per #470)
* shapeA[sizeA - 2] = shapeB[sizeB - 1] (i.e. columns in first matrix = rows in second matrix)

... and the algorithm as written doesn't seem to either try to achieve this or enforce it. 

Maybe someone that knows could weigh in with a more precise description or pseudocode? I can translate to spec language!

inexorabletash: Chromium implementation does this:

* if sizeA < 2 || sizeB < 2, throw -- per #470
* colsA = shapeA[sizeA - 1]
* rowsA = shapeA[sizeA - 2]
* colsB = shapeB[sizeB - 1]
* rowsB = shapeB[sizeB - 2]
* if colsA != rowsB, throw -- the 2D matrices must match
* slicedA = shapeA[0 : sizeA - 2]
* slicedB = shapeB[0 : sizeB - 2]
* if sizeA > 2 && sizeB > 2
    * b = broadcastShapes(slicedA, slicedB, bidirectional = true) or throw
    * outputShape  = b with [rowsA, colsB] appended
* else if size A == 2 && size B == 2
    * outputShape = [rowsA, colsB]
* else (sizeA == 2 or sizeB == 2)
    * outputShape = (sizeA > sizeB) ? slicedA : slicedB) with [rowsA, colsB] appended

... where broadcastShapes() is a whole 'nother can of worms.

I notice https://webmachinelearning.github.io/webnn/#mlgraphbuilder-broadcast-shapes says it is implementation-defined - is that true!?!?!?

inexorabletash: I put #534 up to define broadcasting shapes, so for this PR we could wait for that to land, then rebase this on top of it and I can translate the above into spec text. I'm feeling close to actually understanding what's going on!

huningxin: > I notice https://webmachinelearning.github.io/webnn/#mlgraphbuilder-broadcast-shapes says it is implementation-defined - is that true!?!?!?

broadcast shouldn't be implementation-defined, thanks for putting a PR to fix it.

inexorabletash: Latest update includes a commit that rewrites the matmul output shape calc to match Chromium's impl. PTAL?

inexorabletash:
Rebased and force-pushed - sorry about that. There are two commits - the first is unchanged  from before, the second is a rewrite of the matmul shape calculations to match the Chromium impl. 

Please take a look?

index.bs
```diff
@@ -2892,11 +2892,13 @@ partial interface MLGraphBuilder {
...
-    1. Let |shapeA| be |a|'s [=MLOperand/shape=] and |sizeA| be |a|'s [=MLOperand/rank=].
-    1. Let |shapeB| be |b|'s [=MLOperand/shape=] and |sizeB| be |b|'s [=MLOperand/rank=].
+    1. Let |shapeA| be a [=list/clone=] of |a|'s [=MLOperand/shape=].
+    1. Let |sizeA| be the [=list/size=] of |shapeA|.
+    1. Let |shapeB| be a [=list/clone=] of |b|'s [=MLOperand/shape=].
+    1. Let |sizeB| be the [=list/size=] of |shapeB|.
     1. If |sizeA| is not 2 or |sizeB| is not 2, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-    1. If |options|.{{MLGemmOptions/aTranspose}} is true, then let |shapeA| be the reverse array of |shapeA|.
-    1. If |options|.{{MLGemmOptions/bTranspose}} is true, then let |shapeB| be the reverse array of |shapeB|.
+    1. If |options|.{{MLGemmOptions/aTranspose}} is true, then reverse the order of the items in |shapeA|.
```
fdwr, line 2900: Yes, I prefer this active voice that you changed it to (just "reverse the thing") rather than passive ("allow/permit/let the thing to be the reverse of the thing").

index.bs
```diff
@@ -4324,8 +4326,8 @@ partial interface MLGraphBuilder {
 
 <div>
     **Arguments:**
-        - *a*: an {{MLOperand}}. The first N-dimensional input tensor.
-        - *b*: an {{MLOperand}}. The second N-dimensional input tensor.
+        - *a*: an {{MLOperand}}. The first input tensor which is at least 2-D.
+        - *b*: an {{MLOperand}}. The second input tensor which is at least 2-D.
```
fdwr, line 4331: ~~Should we also mention that the two input tensors must have the same rank~~ Oh never mind, we simplified GEMM *some*, but we are still supporting broadcasting for it (part of me wanted to even have this higher-level broadcasting policy broadcasting be resolved by the caller, such that the lower level WebNN API only accepts consistent rank tensors here which is more easily testable/spec'able, but that probably borders on the side of caller inconvenience).

index.bs
```diff
@@ -4335,27 +4337,30 @@ partial interface MLGraphBuilder {
...
+    1. Let |rowsA| be |shapeA|[|sizeA| - 2].
+    1. Let |colsB| be |shapeB|[|sizeB| - 1].
+    1. Let |rowsB| be |shapeB|[|sizeB| - 2].
+    1. If |colsA| is not equal to |rowsB|, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |slicedA| be a [=list/clone=] of |shapeA| with the last 2 items [=list/removed=].
+    1. Let |slicedB| be a [=list/clone=] of |shapeB| with the last 2 items [=list/removed=].
+    1. If both |sizeA| and |sizeB| are greater than 2, then let |outputShape| be the result of [=bidirectionally broadcasting the shapes=] |slicedA| and |slicedB|. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Otherwise, if both |sizeA| and |sizeB| are equal to 2, then let |outputShape| be an empty [=/list=].
+    1. Otherwise, let |outputShape| be the [=list/size|longer=] of |slicedA| or |slicedB|.
+    1. [=list/Append=] « |rowsA|, |colsB| » to |outputShape|.
```
fdwr, line 4362: 🤔 I don't think the explicit `if`/`else if`/`else` are needed here because broadcasting already works correctly either way, even with scalars and empty dimensions (it must). So if you strip off the trailing spatial dimensions and take just the leading batch dimensions of `A` and `B` for bidirectional broadcasting, then if `batchShapeA` and `batchShapeB` are empty, you will correctly get an empty `[]` back, and for any non-empty cases, you'll get the correct result too. So I think it would just be this:

```suggestion
    1. Let |batchShapeA| be a [=list/clone=] of |shapeA| with the spatial dimensions (last 2 items) [=list/removed=].
    1. Let |batchShapeB| be a [=list/clone=] of |shapeB| with the spatial dimensions (last 2 items) [=list/removed=].
    1. Let |outputShape| be the result of [=bidirectionally broadcasting the shapes=] |batchShapeA| and |batchShapeB|. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
    1. [=list/Append=] « |rowsA|, |colsB| » to |outputShape|.
```

Right?

fdwr (approved):
We can reduce some redundant GEMM steps, but otherwise LGTM sir.

fdwr (approved):
👍

index.bs
a-sully: [XLA has a stance on this](https://openxla.org/xla/broadcasting#principles) which I find very appealing:

> The XLA language is as strict and explicit as possible, avoiding implicit "magical" features. Such features might make some computations slightly easier to define, but at the cost of more assumptions baked into user code that will be difficult to change in the long term. If necessary, implicit magical features can be added in client-level wrappers.

If there's a tradeoff to be made between:
- caller inconvenience, and
- spec maintainability/testability, while removing ambiguity

... I think we should strive for the latter in WebNN :)

fdwr: Yeah, generally we've been moving more that direction. For example, `reshape` no longer takes a special null value as a wildcard to compute based on the tensor element count and the remaining elements. `matmul` (as you see here) lost its additional if/else if/else ~~1-D~~ 1D logic, with some cost to the caller (reshape calls). As someone who's implemented a number of specs (Unicode bidi, Unicode line breaking...) I definitely value spec clarity and clear conformance, and I appreciate every operator definition having clear pseudocode or a decomposition of equivalent operators. There *may* be some implementation-defined aspects for certain data types near the edges (depending on device capability), like saturation vs infinity, rounding of subnormals, and NaN representation, but generally yeah, we all want to lessen that.

PR #530, opened 2024-01-26
Meta: Add a gitignore

a-sully:
`make` generates an `index.html` file which should not be uploaded to the repo

In another spec I worked on, this [resulted in the `index.html` file accidentally being uploaded](https://github.com/WICG/file-system-access/pull/398) (and no one noticed for a while, whoops). It would be nice to avoid that outcome here :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/index.html


anssiko:
Yes, this is a good guardrail. Thanks!

Issue #503, opened 2024-01-10
concat() assertions around input shape/datatypes seem invalid


inexorabletash:
The MLGraphBuilder  concat(sequence&lt;MLOperand&gt; inputs, unsigned long axis) method has:

> Assert: the shape, i.e. dimensions of each operand in inputs is the same, except on the dimension given by axis on which they are concatenated.
> Assert: the dataType of each operand in inputs is the same.

Since the inputs are provided by script, these don't seem to be invariants. Are these intended to be validation steps instead?

Issue #502, opened 2024-01-10
Convert "If descriptor.dimensions does not exist, then descriptor defines a scalar input." from assertion to note


inexorabletash:
In MLGraphBuilder's input() method steps:

> Assert: If descriptor.dimensions does not exist, then descriptor defines a scalar input.

This doesn't appear to be an assertion, rather it seems to be describing (non-normatively) how to interpret the descriptor.

Convert this to a note?

Issue #500, opened 2024-01-10
Define "calculating output dimensions"


inexorabletash:
Various graph builder methods (7 of them?) contain this or similar phrasing:

> Let outputShape be the result of invoking the underlying implementation for calculating output dimensions, given options.

Is this truly expected to be implementation dependent, or can it be rigorously specified on a per-op basis?

huningxin:
I think the output shape inference should be implementation independent and be specified per-op. The spec mentions the output shape calculation for some ops. For example, 

for [Element-wise unary operations](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-unary)

> The shape of the output tensor is the same as the shape of input tensor.

for [conv2d](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-conv2d), 

> More specifically, the spatial dimensions or the sizes of the last two dimensions of the output tensor for the nchw input layout can be calculated as follow:

> outputSize = 1 + (inputSize - (filterSize - 1) * dilation - 1 + beginningPadding + endingPadding) / stride

They should be reflected in algorithm steps.

Issue #494, opened 2023-12-19
The input of `triangular` should be at least 2D tensor


BruceDai:
This is raised by @fdwr on reviewing implementation for `triangular` op on https://github.com/webmachinelearning/webnn-baseline/pull/64. (Thanks Dwayne!)

> it should say the input is at least 2D. You can see other ML libraries just treat each 2D subsection as a separate batch:
> 
> https://numpy.org/doc/stable/reference/generated/numpy.triu.html
> https://pytorch.org/docs/stable/generated/torch.tril.html
> https://onnx.ai/onnx/operators/onnx__Trilu.html
> 
> So shape [4,3,5,7] would apply the triangular masking to the last two dimensions.

fdwr:
```diff
...

  Arguments:
- input: an [MLOperand](https://www.w3.org/TR/webnn/#mloperand). The input 2-D tensor.
+ input: an [MLOperand](https://www.w3.org/TR/webnn/#mloperand). The input tensor which is at least 2D.

   options: an optional [MLTriangularOptions](https://www.w3.org/TR/webnn/#dictdef-mltriangularoptions). The optional parameters of the operation.

- Returns: an [MLOperand](https://www.w3.org/TR/webnn/#mloperand). The output 2-D tensor representing a triangular matrix.
+ Returns: an [MLOperand](https://www.w3.org/TR/webnn/#mloperand). The output tensor representing a masked triangular matrix which is the same shape as the input.

...

+ // lower triangular matrix with two batches:
+ //   [[[7, 0, 0],
+ //     [9, 4, 0],
+ //     [2, 6, 3]],
+ //    [[1, 0, 0],
+ //     [4, 5, 0],
+ //     [7, 8, 9]]]
+ const lower = builder.triangular(input, { upper: false });
```

BruceDai:
Thanks @fdwr !
I think you can create a PR for this!

PR #533, opened 2024-01-26
Process: Add documentation for labels, current and proposed

inexorabletash:
For review - I left lots of notes which should get resolved before this ever merges. Editors - please make edits directly if you can!

Before merging we should also:
- [x] Add links from CONTRIBUTING.md etc
- [x] Consider creating a `docs/` directory to capture these files
- [x] Do a pass over existing issues to adjust labels if they don't match, e.g. a few "question" issues should maybe be "feature request" ?

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/533.html" title="Last updated on Jan 30, 2024, 7:48 PM UTC (7caa3b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/533/ce7deb3...inexorabletash:7caa3b1.html" title="Last updated on Jan 30, 2024, 7:48 PM UTC (7caa3b1)">Diff</a>


IssueTriage.md
```diff
@@ -0,0 +1,104 @@
+# Triage Guidance for WebNN Repo
+
+TODO: Consider removing these, or fit these into the documentation below.
+
+- **dependencies** - unused?
```
anssiko, line 5: I deleted this currently unused label.

The fewer the labels the more likely the group will use them (and use them correctly).

IssueTriage.md
```diff
@@ -0,0 +1,104 @@
...
+- **tag** - redundant with **tag-needs-resolution** and **tag-tracker** ?
```
anssiko, line 6: Ditto. This is now gone.

IssueTriage.md
```diff
@@ -0,0 +1,104 @@
...
+
+The working group chairs and spec editors should regularly review bugs and ensure that labels are accurate, and ensure that issues are getting appropriate attention; for example, scheduling discussion of new feature requests, discussion to resolve outstanding questions, and drawing attention to issues that are ready for a contributor to author a PR.
+
+
+## Types
+
+Every issue should have one of these issue types, and only rarely more than one.
+
+- **bug** - a flaw in the specification that will require a normative fix; for example, an algorithm computes an output incorrectly
+- **conventions-integration** - where the spec does not conform to specification best practices from Web IDL, Bikeshed, Infra, etc.
```
anssiko, line 27:
```suggestion
- **conventions** - where the spec does not conform to specification best practices from Web IDL, Bikeshed, Infra, etc.
```
Renamed this. Short and sweet.

IssueTriage.md
```diff
@@ -0,0 +1,104 @@
...
+## Spec Impact
+
+These broad categories describe the projected impact on the specification and implementation of an issue.
+
+- **Editorial** - spec text could be improved, but does not imply normative behavior changes to the proposed API
+- **Normative** - _PROPOSED_ - resolving the issue would require behavior changes, likely affecting implementations and tests
+
+TODO: I'm not convinced of the utility of these; maybe drop them?
+
+TODO: Make these lowercase for consistency?
```
anssiko, line 45:
```suggestion
- **editorial** - spec text could be improved, but does not imply normative behavior changes to the proposed API
- **normative** - _PROPOSED_ - resolving the issue would require behavior changes, likely affecting implementations and tests

TODO: I'm not convinced of the utility of these; maybe drop them?
```
Made editorial lowercase.

IssueTriage.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Issue Scope
+
+Every issue should have at least one of these issue types, but occasionally multiple.
+
+- **operation set** - discussions about the overall operator coverage of WebNN; examples include alignment with other published operator sets, use cases that require multiple new operators, compatibility with implementations, etc.
+- **operator specific** - _PROPOSED_ - issues regarding the specification of a single operator or small number of operators
```
anssiko, line 52:
```suggestion
- **operation set** - discussions about the overall operation coverage of WebNN; examples include alignment with other published operation sets (op set), use cases that require multiple new operations, compatibility with implementations, etc.
- **operation specific** - _PROPOSED_ - issues regarding the specification of a single operation or small number of operations
```

Are we making someone unhappy if we standardize on _operation_? :)

I note the spec needs some regexp action for this because we use operation and operator interchangeably.

anssiko (changes requested):
Thanks @inexorabletash!

I dropped some unused labels, did some renames and proposed some changes to this PR accordingly. This doc will be very helpful for the group.

IssueTriage.md
inexorabletash: It looks like most frameworks (TF, ONNX, PyTorch) use "operator" more frequently in this context, e.g. "operator set", "operators", "operator compatibility", "operator coverage".  (thanks @a-sully) 

_Operators_ perform _operations_, so the phrase _"a platform operator for the operation"_ correctly uses both. 

For the labels, I think "operator" would align with more common usage. But I'm still new to all this!

anssiko: Let’s go with the most popular name for the labels. We can adjust these later as needed.

anssiko (approved):

inexorabletash:
Given that it already warrants special consideration, we should decide and document how [new operator proposals](https://github.com/webmachinelearning/webnn/blob/main/CONTRIBUTING.md#proposing-and-adding-a-new-operation) should be labeled. 

Ignoring operation/operator spelling for the moment, is that **operation set** (because it's claiming the current op set is incomplete, and an addition to the op set), **operator specific** (because it's about specific new ops), or a new label e.g. "new operator" ?

If we expect this to be common, we could introduce an issue template that automatically applies labels, which would ease the triage burden.

anssiko:
How about if we settle on short and sweet "opset" and "op" to sidestep the naming discussion, while allow to filter out global from specific?

inexorabletash:
I worry that "op" without a qualifier like "specific" could end up being applied to everything relating to ops. I guess that's why we have docs.

I'll defer to others who know the space better than I do re: "opset" vs. "op set" vs. "operator set" etc.

IssueTriage.md
```diff
@@ -0,0 +1,103 @@
...
+## Types
+
+Every issue should have one of these issue types, and only rarely more than one.
+
+- **bug** - a flaw in the specification that will require a normative fix; for example, an algorithm computes an output incorrectly
+- **conventions** - where the spec does not conform to specification best practices from Web IDL, Bikeshed, Infra, etc.
+- **use case** - a new use case for the API that should be documented or considered; may spawn other issues
+- **process** - a meta issue about how the specification is evolved; for example, a discussion of issue labels
+- **testing** - discussion of test coverage
+- **feature request** - _PROPOSED_ - suggestion for an addition to the proposed API
```
anssiko, line 29: Added https://github.com/webmachinelearning/webnn/labels/feature%20request to test drive this.

IssueTriage.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+## Spec Impact
+
+These broad categories describe the projected impact on the specification and implementation of an issue.
+
+- **editorial** - spec text could be improved, but does not imply normative behavior changes to the proposed API
+- **normative** - _PROPOSED_ - resolving the issue would require behavior changes, likely affecting implementations and tests
+
+TODO: I'm not convinced of the utility of these; maybe drop them?
```
anssiko, line 41: One use case for "editorial" is to filter out issues that are https://www.w3.org/2023/Process-20231103/#editorial-change to assess whether the spec satisfies Rec Track transition requirements at various maturity stages.

I think we don't need normative.

IssueTriage.md
```diff
@@ -0,0 +1,103 @@
+# Triage Guidance for WebNN Repo
+
+TODO: Consider removing these, or fit these into the documentation below.
+
+- **enhancement** - unclear if this is about the spec text, the proposed API, etc. Prefer **feature request** etc.
```
anssiko, line 5: I did some further triage, and the issues with https://github.com/webmachinelearning/webnn/labels/enhancement are only a few. We should probably drop this label entirely.

IssueTriage.md
inexorabletash: I did a pass and labeled issues that looked like requests for pure functionality additions. How does it look to others?

inexorabletash:
A good process question: do we actually want to tag "the normative spec text has a flaw" with [bug](https://github.com/webmachinelearning/webnn/labels/bug), or is that considered the default?

IssueTriage.md
fdwr: > I'll defer to others who know the space better than I do re: "opset" vs. "op set" vs. "operator set" etc.

I've always just said `opset` as a shorthand for `operator set`. 🤷‍♂️

huningxin: `opset` for label sgtm.

huningxin:
Tagging "the normative spec text has a flaw" with bug sounds good to me. I think "lack of normative spec text" is also a bug, for example, https://github.com/webmachinelearning/webnn/issues/283

anssiko:
There was a request for a new https://github.com/webmachinelearning/webnn/labels/webgpu%20interop label so I created one, @inexorabletash to be reflected in this PR.

huningxin:
> There was a request for a new [ webgpu interop](https://github.com/webmachinelearning/webnn/labels/webgpu%20interop) label so I created one, @inexorabletash to be reflected in this PR.

MLBuffer is also useful for chained inference use case. Should we also create a tag for it?

inexorabletash:
> There was a request for a new [ webgpu interop](https://github.com/webmachinelearning/webnn/labels/webgpu%20interop) label so I created one, @inexorabletash to be reflected in this PR.

Added! (I don't get the nice formatting of labels in comments, it seems.) And renamed the section to "Workstreams" but ideas welcome.

Also removed the **normative** suggestion and added a link to the Process doc. 

New suggestion: drop **cr** and **v2** and use Milestones instead, which have nice progress indicators.

anssiko:
>MLBuffer is also useful for chained inference use case. Should we also create a tag for it?

Thanks, this would be a good feature-specific label for consideration.

@inexorabletash would it makes sense to use a combo of type: https://github.com/webmachinelearning/webnn/labels/feature%20request + feature: "name" label for significant new features? This suggest we'd have a few feature-specific label for bigger features. Are there good experiences from the spec-land for this?

>New suggestion: drop cr and v2

No open issue associated with cr label (we've addressed them!) so I dropped https://github.com/webmachinelearning/webnn/labels/cr 👋 

https://github.com/webmachinelearning/webnn/labels/v2 label is quite popular. This is "work deferred to the next version" which means this collection becomes a mixed bag. It is something we should look from time to time and especially after each major spec milestone to plan for the next.

> and use Milestones instead, which have nice progress indicators.

What is the best example of the GH Milestones feature used right in a spec project so we could learn from it and evaluate if this feature would help the broader group? I've used this feature for some spec projects long time ago so asking for more recent view on how things have improved.

inexorabletash:
> Do we want to label proposed feature removals, the reverse of "feature request"? On the impact scale they're similar.

My concern here is that a feature request is (in most cases) purely additive. It can be deferred to "v.next", allowed to linger indefinitely, wait for use cases, etc. A removal, on the other hand, needs to consider compatibility, probably needs to happen within a version, acknowledge that use cases have been de-prioritized, etc. A feature addition that doesn't happen is neutral for the long term maintenance of implementations. But a feature removal that doesn't happen implies long term work for implementations, tests, interop, etc. So I agree that in terms of impact scale to the spec itself they're similar, but the attention needed and urgency seem quite different. I'd really hate for a suggested removal to not get the attention it deserves.

inexorabletash:
> Are there good experiences from the spec-land for this?

For both feature-specific labels and milestones - I'm asking around. I have anecdotes but not data. :)

IssueTriage.md
inexorabletash: Updated proposal w/ `opset`

inexorabletash:
From asking around, more anecdotes:

- an issue can only be associated with 1 milestone, you can't use them for any overlapping deadlines. So if you have clear unambiguous milestones those can work, but pick one scale for your milestones, and use labels for any other scale of grouping.
- CSS consistently uses "topic" labels for separate specs (multiple in one repo, a very different pattern than here); use of feature-level labels is infrequent but does happen.

In personal projects I like the use of Milestones to indicate when an issue is "accepted" as a requirement for the current release or explicitly "deferred" to a future release, and issues with no Milestone are undecided (so: action is required). Using labels is fine but you don't get the nice bar graphs. The current situation is that we have a handful of [v2](https://github.com/webmachinelearning/webnn/labels/v2) but (1) IMHO many of those should be considered blockers for TR and (2) anything not marked v2 is ambiguous.  Ultimately it all comes down to: I want a clear view on what work remains and what next action is needed, so I'll be happy with anything that helps with that!

Labels for _"significant new features"_ SGTM, on a case-by-case basis, like has happened for [webgpu interop](https://github.com/webmachinelearning/webnn/labels/webgpu%20interop).

IssueTriage.md
anssiko: Renamed to https://github.com/webmachinelearning/webnn/labels/opset

inexorabletash:
For the proposed "operator specific" label, this is the list of open issues I'd apply it to:

- #503
- #502
- #500
- #494
- #493
- #492
- #490
- #487
- #486
- #484
- #481
- #475
- #474
- #472
- #471
- #470
- #467
- #466
- #465
- #452
- #449
- #444
- #396
- #392
- #383
- #377
- #374
- #368
- #360
- #358
- #353
- #326
- #324
- #307
- #278
- #270
- #84

inexorabletash:
I dropped the proposals for "samples" and "interop" as I couldn't identify enough proposed issues in each category for it to make sense.

anssiko:
Created https://github.com/webmachinelearning/webnn/labels/operator%20specific and populated per triage https://github.com/webmachinelearning/webnn/pull/533#issuecomment-1930736774

anssiko:
I've updated https://github.com/webmachinelearning/webnn/labels to match. A few notes:
- did not yet add PROPOSED "needs PR", "has PR", "needs WPT", "has WPT" labels
- kept TODO https://github.com/webmachinelearning/webnn/labels/enhancement until we remap those issues, removed TODO "help wanted" that was empty

@inexorabletash, do you also feel like we could merge this soon? Please check things are cohesive e.g. remove "help wanted".

anssiko:
Let's call #448 a https://github.com/webmachinelearning/webnn/labels/bug for the time being and remove the TODO from here and merge this PR.

Triage is always an imperfect practice and the best way to improve is to try in production and adjust. Thanks for iterating on this PR!

inexorabletash:
TODOs removed - merge at your leisure.

Issue #462, opened 2023-09-08
Improve "underlying platform" references in spec


inexorabletash:
The spec currently uses algorithms of the form (just examples copy/pasted from different places

1. Make a request to the underlying platform to:
    1. Let opImpl be an [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined) platform operator for the slice operation, given starts and sizes.
    2. ...
    3. Create an implementation-defined platform operand outputImpl to represent the output, given output and opImpl.
    4. ...
2. Connect input.[[operand]] as input to opImpl.
3. Connect output.[[operand]] as output to opImpl.

* "implementation-defined" isn't really helpful here, since it never explains what this does.
* "platform operator" is never defined
* "the slice operation" isn't defined
* "connect" isn't defined

Ideally this would all be defined. Here's one way to approach it, imagining a much simpler interface:

First we define the abstract notion of an op:

An **_op_** represents a mathematical operation as part of a graph. An **op** has zero or more **_inputs_** (**ops**) and one **_output_** (an **op**), and an **_algorithm_** which, when **_invoked_**, takes a corresponding number of **_input values_** and produces an **_output value_**.

In a dedicated section of the spec, we'd rigorously define each op's algorithm:

An **_addition op_** is an **op** with two **inputs** (both numbers) and the following **algorithm**:
1. Let _a_ be the first *input value*.
2. Let _b_ be the second *input value*.
3. Return _a_ + _b_.

An **_average op_** is an **op** with one **input** (a list) and the following **algorithm**:
1. Let _a_ be the input.
2. Let _sum_ be 0.
3. For each _value_ of _a_:
    1. Let _sum_ be _sum_ + _value_.
4. Return _sum_.

Obviously these are overly simple, but in the spec right now the more complex ops are currently only defined by links. I'd tackle this incrementally, but it should be in the spec itself! See Web Audio for examples e.g. https://webaudio.github.io/web-audio-api/#fft-windowing-and-smoothing-over-time. These could be inline with the builder methods or as a separate section of the spec.

And then in a builder method for `add()` it'd be something like:

1. Let _output_ be the result of creating an MLOperand given this and desc.
2. Let _opImpl_ be a new **addition op**.
3. Set _opImpl_'s **inputs** to _inputs_.
4. Set _opImpl_'s **outputs** to _output_.[[operand]].
5. Set _output_.[[operator]] to _opImpl_.

That's probably not quite correct but hopefully the intention is clear.

More notes:
[...]

Issue #324, opened 2023-01-09
Simplify the operand layout support of conv2d and pooling 2d operations


huningxin:
In the existing WebNN spec, [conv2d](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-conv2d) supports two input operand layouts defined by [MLInputOperandLayout](https://www.w3.org/TR/webnn/#enumdef-mlinputoperandlayout) and four filter operand layouts defined by [MLConv2dFilterOperandLayout](https://www.w3.org/TR/webnn/#enumdef-mlconv2dfilteroperandlayout). 

```js
enum MLInputOperandLayout {
  "nchw",
  "nhwc"
};

enum MLConv2dFilterOperandLayout {
  "oihw",
  "hwio",
  "ohwi",
  "ihwo"
};
```

This may make the implementation more complicated especially if a native ML framework or OS API doesn't support some of these layouts. If one layout is unsupported, the implementation may need to insert the `transpose` operations into the graph around the `conv2d` operation that transposes the unsupported layout to supported one. This would easily lead to an inefficient graph representation that may have redundant `transpose` operations. Or the implementation may need to optimize the graph by techniques such as "transpose sink" which may require a more complex implementation. This issue was raised in Chromium [CL](https://chromium-review.googlesource.com/c/chromium/src/+/4031876/14..16/third_party/blink/renderer/modules/ml/webnn/ml_graph_xnnpack.cc#b460) review.

To simplify the implementation, the proposal is to reduce the supported operand layouts, for example, just keep the default one. Because WebNN supports `transpose` operation, the layout adaption and graph level optimization can be handled by ML frameworks that usually already support such functionalities.

Thanks @wacky6 for this idea.

fdwr:
@huningxin Which one is being proposed as preferred - NCHW or NHWC? Do we have a holistic list of the preferred layouts used by popular ML libraries to favor one over the other? I know some libraries support both, some default to NHWC but actually perform better with NCHW, some only support one (e.g. ONNX == NCHW).

[...]

PR #534, opened 2024-01-26
New content: Add definition for shape broadcasting

inexorabletash:
This change introduces a new section for Algorithms, following APIs, to collect algorithms referenced throughout the specification.

A section for Broadcasting is introduced, which defines broadcasting shapes and gives an explicit algorithm matching WebNN implementations of NumPy's General Broadcasting Rules. Definitions for "broadcastable" and "unidirectionally broadcastable" are introduced. The previous definition of "broadcast-shapes" is removed in favor of these new algorithms.

For #324, #462, and potentially #523.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/534.html" title="Last updated on Feb 13, 2024, 4:40 AM UTC (145bff1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/534/55183f7...inexorabletash:145bff1.html" title="Last updated on Feb 13, 2024, 4:40 AM UTC (145bff1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4718,7 +4718,7 @@ partial interface MLGraphBuilder {
   <div class=algorithm-steps>
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}}.
-    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=unidirectionally broadcasting the shapes=] |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |slope|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=unidirectionally broadcasting the shapes=] |slope|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
         1. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. If any of the following sub-steps fail, [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
         1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |descriptor|.
--- a/index.bs
+++ b/index.bs
@@ -2240,8 +2240,8 @@ partial interface MLGraphBuilder {
     1. If |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}} is not equal to |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}}, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}}.
-    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of running the [=MLGraphBuilder/broadcast-shapes=] steps given |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
+    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=bidirectionally broadcasting the shapes=] |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+        1. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. If any of the following sub-steps fail, [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
         1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |descriptor|.
         1. Make a request to the underlying platform to:
@@ -2252,21 +2252,6 @@ partial interface MLGraphBuilder {
         1. Connect |a|.{{MLOperand/[[operand]]}} and |b|.{{MLOperand/[[operand]]}} as inputs to |opImpl|.
         1. Connect |output|.{{MLOperand/[[operand]]}} as output to |opImpl|.
     1. Return |output|.
-  </div>
-</details>
-
-<details open algorithm>
-  <summary>
-    To <dfn for="MLGraphBuilder">broadcast-shapes</dfn> given [=/list=] |shape1| and [=/list=] |shape2|, run the following steps:
-  </summary>
-  <div class=algorithm-steps>
-    1. [=Assert=]: The type of |shape1| and |shape2| is `sequence of unsigned long`.
-    1. Let |output| be the result of invoking the [=implementation-defined=] shape broadcast on |shape1| and |shape2|.
-            1. If that fails, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-    1. Return |output|.
-        <div class = "note">
-        The most common implementation is that two shapes are compatible, when each of their corresponding dimensions are equal, or one of them is 1. The output shape consists of the maximum of the corresponding dimensions.
-        </div>
   </div>
 </details>
 
@@ -2376,8 +2361,8 @@ Although operations *greaterOrEqual* and *lesserOrEqual* can each be implemented
         1. If |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}} is not equal to |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}}, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to {{MLOperandDataType/"uint8"}}.
-    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of running the [=MLGraphBuilder/broadcast-shapes=] steps given |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
+    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=bidirectionally broadcasting the shapes=] |a|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |b|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+        1. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. If any of the following sub-steps fail, [=exception/throw=] an "{{OperationError}}" {{DOMException}}.
         1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |descriptor|.
         1. Make a request to the underlying platform to:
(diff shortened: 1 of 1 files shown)


inexorabletash:
I'm still pretty hazy on definitions, so please look at this very skeptically. 

Re: introducing a new "Algorithms" section - this is a pattern I've adopted from other specs. I thought I'd give a whirl here and see what everyone thinks. It's easy enough to drop and put these definitions anywhere else to keep the overall spec structure unchanged.

inexorabletash:
Hold off on reviews, I was confusing unidirectional (mentioned in several places in the spec) and bidirectional (mentioned in the Chromium implementation). Fridays, am I right?

inexorabletash:
Okay, this is updated to match the Chromium impl, I think. But... that may not be what's desired @a-sully mentions that unidirectional broadcasting seems specific to ONNX and other frameworks do things to work around this during export.

a-sully:
AFAICT ONNX is an outlier in making a distinction between "unidirectional" and "multidirectional" broadcasting. This distinction is not made by:
- Numpy: https://numpy.org/doc/stable/user/basics.broadcasting.html
- PyTorch: https://pytorch.org/docs/stable/notes/broadcasting.html (which references Numpy)
- Tensorflow/XLA/StableHLO: https://www.tensorflow.org/xla/broadcasting (which also references Numpy, though it has some other restrictions)

The ["unidirectional broadcastable"](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md#unidirectional-broadcasting) constraint of some ONNX ops (e.g. [`prelu()`](https://github.com/onnx/onnx/blob/main/docs/Operators.md#PRelu)) requires workarounds when exporting from other formats to ONNX - like in this example of using TVM to export PyTorch to ONNX: https://github.com/pytorch/pytorch/issues/70570#issuecomment-1034379620. The folks on that issue discovered that the "unidirectional broadcastable" constraint for `prelu()` was not documented (which has since been fixed) and that there were inconsistencies between ONNX implementations (which... hopefully has also been fixed ¯\\_(ツ)_/¯ )

Numpy's broadcasting rules are the clear standard across the industry and IMHO is the standard that we should expose to the web. It then becomes up to the user agent to ensure that the constraints of the underlying framework (e.g. unidirectional broadcasting for ONNX, [lack of inferred broadcasting specifications for XLA](https://www.tensorflow.org/xla/broadcasting#principles)) are satisfied.

wacky6:
We follow Numpy broadcast rules for ops (in the chromium implementation), I think it's the golden "standard". +1 to adopt this throughout the spec (I didn't realize the spec matmul referred to an implementation defined broadcast 😱).


> @a-sully mentions that unidirectional broadcasting seems specific to ONNX and other frameworks do things to work around this during export

Is ONNX unidirectional broadcast issue easy to workaround? The linked prelu() patch seems "easy enough". 

Should we (from the spec's perspective) expect implementations (e.g. chromium DML backend)to do asid easy workarounds?

index.bs
```diff
@@ -2879,6 +2864,7 @@ partial interface MLGraphBuilder {
         1. If the sequence length of |newShape| is not equal to the [=rank=] of |inputDesc|, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
         1. Let |outputDesc| be a copy of |inputDesc|.
         1. [=list/For each=] |index| in [=the range=] 0 to the [=rank=] of |input|, exclusive:
+            Issue: Can this be replaced with [=unidirectionally broadcasting the shapes=] |inputDesc| and |newShape|.
```
huningxin, line 2867: Because it checks size of new shape is equal to input rank, I suppose it can unidirectionally broadcast input shape to new shape.

huningxin (approved):
LGTM with a comment, thanks much!

huningxin:
@fdwr and @wchao1115 , please also take a look. Thanks!

index.bs
inexorabletash: I pushed a second commit that makes this update. Please take a look?

index.bs
```diff
@@ -6004,9 +5985,9 @@ partial interface MLGraphBuilder {
...
-    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of running the [=MLGraphBuilder/broadcast-shapes=] steps given |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |other|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
-    1. If |condition| is not unidirectionally broadcastable to |descriptor|.{{MLOperandDescriptor/dimensions}} according to the [[!numpy-broadcasting-rule]], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=bidirectionally broadcasting the shapes=] |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |other|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+        1. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. If |condition| is not [=unidirectionally broadcastable=] to |descriptor|.{{MLOperandDescriptor/dimensions}}, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
```
fdwr, line 5990:
```suggestion
    1. If |condition| is not [=bidirectionally broadcastable=] to |descriptor|.{{MLOperandDescriptor/dimensions}}, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
```

Oops, existing spec typo, as the 3 input tensors (`condition`, `trueValues`, `falseValues`) are actually *multidirectionally* broadcastable with each other, which is functionally equivalent to `bidirectionalBroadcast(a, bidirectionalBroadcast(b, c))`. That's what the reference and implementation do:

https://github.com/webmachinelearning/webnn-baseline/pull/56/files#diff-1ab0e4e354fcd892d24b937bb8aad28636b554a25f248d3433573b57603ea411

```
  const valueShape = getBroadcastShape(trueValuesReshape.shape, falseValuesReshape.shape);
  const outputShape = getBroadcastShape(condition.shape, valueShape);
```

https://chromium-review.googlesource.com/c/chromium/src/+/5039655

```
  const auto value_shape =
      BroadcastShapes(true_value.dimensions, false_value.dimensions, true);
  if (!value_shape) {
    return base::unexpected(
        "The shapes of true_value and false_value are not broadcastable.");
  }

  const auto output_shape =
      BroadcastShapes(condition.dimensions, value_shape.value(), true);
  if (!output_shape) {
    return base::unexpected(
        "The condition shape is not broadcastable to the shape broadcasted "
        "from true_value and false_value.");
  }
```

https://chromium-review.googlesource.com/c/chromium/src/+/5069382

```
// * The shapes of all input operands must be bidirectionally broadcastable. (with each other)
```

index.bs
```diff
@@ -2854,16 +2839,12 @@ partial interface MLGraphBuilder {
     <div class="note">
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
-    1. If any of the following steps fail, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-        1. Let |inputDesc| be |input|.{{MLOperand/[[descriptor]]}}.
-        1. If the sequence length of |newShape| is not equal to the [=rank=] of |inputDesc|, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
```
fdwr, line 2859: Yeah, it appears the rule about sequence length matching input rank is not necessary, nor is it so in either Chromium implementation or the reference baseline.

index.bs
```diff
@@ -6219,6 +6200,48 @@ dictionary MLOperandDescriptor {
...
+## Broadcasting ## {#algorithms-broadcasting}
+
+Broadcasting refers to how operations treat tensors with different shapes, and follow the precedent set by [[!numpy-broadcasting-rule]].
+
+<div algorithm>
+To <dfn data-lt="unidirectionally broadcasting the shapes">unidirectionally broadcast the shapes</dfn> |A| and |B|, perform the following steps. |A| and |B| are [=/lists=] of positive integers, representing the dimensions of tensors, and the steps return a new [=/list=] of positive integers, or failure.
+
+1. Let |sizeA| be the [=list/size=] of |A|.
+1. Let |sizeB| be the [=list/size=] of |B|.
+1. Let |output| be a new [=/list=].
```
fdwr, line 6215:
```suggestion
1. Let |outputShape| be a new [=/list=].
```
Just for added clarity, since `output` often is used for the *tensor*, and consistency with `outputShape` elsewhere in spec.

index.bs
```diff
@@ -6219,6 +6200,48 @@ dictionary MLOperandDescriptor {
...
+
+<div algorithm>
+To <dfn data-lt="bidirectionally broadcasting the shapes">bidirectionally broadcast the shapes</dfn> |A| and |B|, perform the following steps. |A| and |B| are [=/lists=] of positive integers, representing the dimensions of tensors, and the steps return a new [=/list=] of positive integers, or failure.
+
+1. Let |sizeA| be the [=list/size=] of |A|.
+1. Let |sizeB| be the [=list/size=] of |B|.
+1. Let |outputSize| be the maximum of |sizeA| and |sizeB|.
+1. Let |output| be a new [=/list=].
+1. [=list/For each=] |index| in [=the range=] 0 to |outputSize|, exclusive:
+    1. Let |dimA| be |A|[|sizeA| - |index| - 1] if |index| < |sizeA|, or 1 otherwise.
```
fdwr, line 6237: Another option that might be clearer and functionally equivalent, because then the loop becomes a more straight-forward foward pass using Append and without the negation, is to just pad A and B upfront (we do this often inside our internal code to avoid going crazy). e.g.

```
1. Let |outputSize| be the maximum of |sizeA| and |sizeB|.
1. Let |output| be a new [=/list=].
1. Let |paddedA| be a copy of |A| with enough leading 1's to pad up to length |outputSize|.
1. Let |paddedB| be a copy of |B| with enough leading 1's to pad up to length |outputSize|.
1. [=list/For each=] |index| in [=the range=] 0 to |outputSize|, exclusive:
    1. Let |dimA| be |paddedA|[|index|] if |index| < |sizeA|, or 1 otherwise.
    1. Let |dimB| be |paddedB|[|index|] if |index| < |sizeB|, or 1 otherwise.
    1. If |dimA| is not equal to |dimB|, and |dimA| is not equal to 1, and |dimB| is not equal to 1, then return failure.
    1. [=list/Append=] the maximum of |dimA| and |dimB| to |output|.
1. Return |output|.
```
🤷‍♂️ (no strong feeling either way)

index.bs
```diff
@@ -6219,6 +6200,48 @@ dictionary MLOperandDescriptor {
...
+
+## Broadcasting ## {#algorithms-broadcasting}
+
+Broadcasting refers to how operations treat tensors with different shapes, and follow the precedent set by [[!numpy-broadcasting-rule]].
+
+<div algorithm>
+To <dfn data-lt="unidirectionally broadcasting the shapes">unidirectionally broadcast the shapes</dfn> |A| and |B|, perform the following steps. |A| and |B| are [=/lists=] of positive integers, representing the dimensions of tensors, and the steps return a new [=/list=] of positive integers, or failure.
+
+1. Let |sizeA| be the [=list/size=] of |A|.
+1. Let |sizeB| be the [=list/size=] of |B|.
```
fdwr, line 6214: It's an error for tensor B.rank > A.rank using undirectional broadcasting (B.rank <= A.rank is fine). Propose:
```suggestion
1. Let |sizeB| be the [=list/size=] of |B|.
1. If |sizeB| > |sizeA|, then return failure.
```

fdwr (commented):
Thank you Josh.

fdwr:
> The unidirectional broadcastable constraint of some ONNX ops (e.g. prelu() requires workarounds when exporting from other formats ...

Huh, I would have presumed `prelu` was bidirectionally broadcastable too. TIL... For operator behavior changes, I'd like it to be separate from this CR that's just adding definitions.

> Is ONNX unidirectional broadcast issue easy to workaround? The linked prelu() patch seems "easy enough".

Yep, easy with a little reshape of leading 1's.

> Should we (from the spec's perspective) expect implementations (e.g. chromium DML backend)to do asid easy workarounds?

Whichever broadcasting behavior we have for prelu, unidirectional or bidirectional, it's trivial either way for backends to insert leading 1's as needed (which the Chromium DML backend already does for DML, since it expects elementwise tensor descriptions of each input to have been massaged to consistent rank first). In the worst case, backends can utilize whatever logic they have for `expand` to pre-expand tensors before forwarding them along.

index.bs
inexorabletash: Yeah I get too clever sometimes. Suggestion SGTM

inexorabletash:
History was getting messy - I force-pushed a new squashed commit. It rolls in #564 to get the build working, but I'll rebase once that merges.

fdwr:
>  I force-pushed a new squashed commit. 

**Josh**: 🤔 Shoot, I'm unsure if I merged your layernorm CR via squash commit like I was supposed. Except larger branch integrations where we want to retain history, that's the general policy for cleaner history. Then the CR author doesn't need to ever explicitly squash their changes, because force pushes in GitHub yield this unfortunate result for reviewers to minimize the review delta since last time...

image

fdwr (approved):
LGTM. 👞✨

inexorabletash:
No worries... I tend to squash locally, force-push, and always review the full diff anyway to see what the actual change is going to be. I'll try to refrain from squashing/force-pushing here though, going forward!

index.bs
fdwr: Appears to me this comment is resolved, and Ningxin already signed off for the rest. Given he is IIRC out \~Feb 10th to 24th for the lunar new year, I propose completing this, and having Ningxin ask for any smaller changes in a later delta.

index.bs
```diff
@@ -4737,8 +4718,8 @@ partial interface MLGraphBuilder {
   <div class=algorithm-steps>
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dataType}}.
-    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of running the [=MLGraphBuilder/broadcast-shapes=] steps given |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |slope|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
+    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=unidirectionally broadcasting the shapes=] |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |slope|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
```
fdwr, line 4721: If A is the first parameter and B is the second parameter (where A is broadcasted to B), then based on the definition below, these two should be swapped so `slope` is broadcasted to `input`, consistent with `expand`:

```suggestion
    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the result of [=unidirectionally broadcasting the shapes=] |slope|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} and |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
```

fdwr (approved):
One more tiny fix after re-review, which I'll just (per your earlier permission) submit and complete this. Thanks Josh.

Issue #211, opened 2021-09-09
Define algorithms for dictionaries with lists as default values


anssiko:
Via @domenic https://github.com/webmachinelearning/webnn/pull/204/files#r703625252

> You need to have a normative algorithm which checks if input["scales"] exists, and if not, use the defaults. I.e. something like
> 
> Let scales be the [list](https://infra.spec.whatwg.org/#list) « 1.0, 1.0, 1.0, 1.0 ».
> If input["scales"] exists, then set scales to input["scales"].
> 
> (and maybe also validate that input["scales"] is of length 4, or whatever you need?)
> 
> This seems difficult since right now your spec has no algorithms (i.e. no [method steps](https://heycam.github.io/webidl/#method-steps)) for its operations, just a non-normative description of the inputs and return value.

We should make this a global update to all affected dictionaries. We probably want to define a reusable algorithm we reference from all these places.

Given this issue is specific to dictionaries with list defaults, for "method steps" convention in general, I opened https://github.com/webmachinelearning/webnn/issues/210

inexorabletash:
This is mostly taken care of by 0fe9e5ca0891bc79a882a153c682be94ae63b580 and prior changes, albeit with slightly different wording than suggested by @domenic

Suggested:
> Let scales be the list « 1.0, 1.0, 1.0, 1.0 ».
> If input["scales"] exists, then set scales to input["scales"].

Current:
> If options.padding does not exist, set it to « 0, 0, 0, 0 ».

Differences:
* Literal list syntax « 0, 0, 0, 0 » is used without "the list" prefix (linked to the Infra definition) - I missed this!
* Dictionary member syntax - see https://github.com/webmachinelearning/webnn/issues/450#issuecomment-1681830276 for some discussion
* Unpacking to a local variable vs. mutating the dictionary. Either is fine; while the latter looks like it would modify the passed JS object but the WebIDL processing model means that doesn't happen. In most cases in the spec an  _options_ dict is passed verbatim to _"... an implementation-defined platform operator for the ..., given options."_  (see https://github.com/webmachinelearning/webnn/issues/462 for some thoughts)

I'll put up a PR for the "the list" wording change, after that perhaps this can be closed?

zolkis:
> I'll put up a PR for the "the list" wording change, after that perhaps this can be closed? 

I like the most compact formulation, which simplifies reading of a long algorithm: 

> If options.padding does not exist, set it to the list « 0, 0, 0, 0 ».

[...]

PR #538, opened 2024-01-29
Bug fix: Provide default for MLLayerNormalizationOptions axes in steps

inexorabletash:
This adds an explicit algorithm step in layerNormalization() to provide a default described only in prose for the axes option.

For #211

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/538.html" title="Last updated on Feb 10, 2024, 2:11 AM UTC (b6c8489)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/538/d1a02f2...inexorabletash:b6c8489.html" title="Last updated on Feb 10, 2024, 2:11 AM UTC (b6c8489)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3820,6 +3820,7 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>layerNormalization(|input|, |options|)</dfn> method steps are:
   </summary>
   <div class=algorithm-steps>
+    1. If |options|.{{MLLayerNormalizationOptions/axes}} does not [=map/exist=], set |options|.{{MLLayerNormalizationOptions/axes}} to the [=/list=] « 1, 2, 3 ».
     1. If the [=rank=] of |options|.{{MLLayerNormalizationOptions/scale}} is not equal to the [=list/size=] of |options|.{{MLLayerNormalizationOptions/axes}}, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. If the [=rank=] of |options|.{{MLLayerNormalizationOptions/bias}} is not equal to the [=list/size=] of |options|.{{MLLayerNormalizationOptions/axes}}, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. [=list/For each=] |index| in [=the range=] 0 to the [=list/size=] of |options|.{{MLLayerNormalizationOptions/axes}}, exclusive:
--- a/index.bs
+++ b/index.bs
@@ -3800,7 +3800,7 @@ partial interface MLGraphBuilder {
 
     : <dfn>axes</dfn>
     ::  
-        The indices to the input dimensions to reduce. When this member is not present, it is assumed to be [1,2,3] that is, the reduction for the mean and variance values are calculated across all the input features for each individual sample in the batch.
+        The indices to the input dimensions to reduce. When this member is not present, it is treated as if all dimensions except the first were given (e.g. for a 4-D input tensor, axes = [1,2,3]). That is, the reduction for the mean and variance values are calculated across all the input features for each independent batch.
 
     : <dfn>epsilon</dfn>
     ::
--- a/index.bs
+++ b/index.bs
@@ -3820,7 +3820,7 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>layerNormalization(|input|, |options|)</dfn> method steps are:
   </summary>
   <div class=algorithm-steps>
-    1. If |options|.{{MLLayerNormalizationOptions/axes}} does not [=map/exist=], set |options|.{{MLLayerNormalizationOptions/axes}} to the [=/list=] « 1, 2, 3 ».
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM, thanks for the fix!

inexorabletash:
@fdwr can you review (and merge if it looks good?) - thanks!

index.bs
```diff
@@ -3820,6 +3820,7 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>layerNormalization(|input|, |options|)</dfn> method steps are:
   </summary>
   <div class=algorithm-steps>
+    1. If |options|.{{MLLayerNormalizationOptions/axes}} does not [=map/exist=], set |options|.{{MLLayerNormalizationOptions/axes}} to the [=/list=] « 1, 2, 3 ».
```
fdwr, line 3823: Hmm, the spec wording doesn't quite match the intention nor the Chromium implementation, as that example of [1,2,3] is only accurate for *4D* input tensors. The intention is that if not specified, axes defaults to [1,2,...N-1], and so for 3D, effective axes = [1,2].

Chromium: https://chromium-review.googlesource.com/c/chromium/src/+/5068275
```
    if (input_rank > 1) {
      // Default to `input_dimensions[1, ... , N-1]` where N is the input rank.
      reduction_dimensions.assign(input_dimensions.begin() + 1,
                                  input_dimensions.end());
```

WPT: https://github.com/webmachinelearning/webnn-baseline/pull/60/files#diff-0f9a981097ee5210d3ed917824f3a9fe72917311e38bc99d7972ea9009634f9dR32
```
  if (axes === undefined) {
    axes = Array.from({length: input.rank - 1}, (_, index) => index + 1);
  }
```

inexorabletash: The Infra [exclusive range](https://infra.spec.whatwg.org/#the-exclusive-range) for that. So naively it'd be:

```
1. If |options|.{{MLLayerNormalizationOptions/axes}} does not [=map/exist=], then set |options|.{{MLLayerNormalizationOptions/axes}} to a new [=/list=] equal to [=the range=] 1 to |input|'s [=rank-=], exclusive. 
```

However... what if _input_'s rank is 0 (scalar) or 1 ? Empty list in both cases?

Meta comment: although we're slowly evolving "standardese" into a more formal language, it's a long long road. The important thing is that it is as unambiguous as we can make it, and doesn't assume domain knowledge... because engineers are creative! (Also that it has tests.) So don't feel beholden to specific spec syntax, although if it exists we should use it!

fdwr: > what if input's rank is 0 (scalar) or 1 ? Empty list in both cases?

Correct.

inexorabletash: If empty list then I'd probably write:
```
1. If |options|.{{MLLayerNormalizationOptions/axes}} does not [=map/exist=], then set |options|.{{MLLayerNormalizationOptions/axes}} to a new [=/list=] equal to [=the range=] 1 to |input|'s [=rank-=], exclusive, if |input|'s [=rank=] is greater than 1, and an empty [=/list=] otherwise.
```
... or break it down into substeps, but I think that's overkill.

fdwr: Substeps seems overkill. How about adding "from" and either "rank - 1" or "end exclusive". e.g.

`...equal to [=a range=] from 1 to |input|'s [=rank-=] - 1, if |input|'s [=rank=] is greater than 1, ..."

inexorabletash: Editor's call. :) Infra suggests "The range n to m, exclusive" https://infra.spec.whatwg.org/#the-exclusive-range or "The range n to m, inclusive" https://infra.spec.whatwg.org/#the-exclusive-range - without "from" and the inclusive/exclusive applies to the end implicitly, but whatever makes it clearer for readers.

(FYI: Infra says a range is an "ordered set", but that's still just a list.)

inexorabletash: BTW, with Write permissions should be able to make edits to the PR directly on GitHub. Click on the **Files changed** tab, click on the "···" menu for the `index.bs` file, click **Edit**, make a change, click **Commit changes...**. Please do so if it will save some round-trips!

fdwr: ⏳ Reading the links you supplied (thanks).

> Infra suggests ... without "from"
> set options.axes (to) a new list equal (to) the range 1 (to) input's rank

Yeah, normally I'd be fine without the "from", but the series of 3 `to`'s in a row just made it harder mentally parse 😅.

> Edit, make a change, click Commit changes

Alrighty, if you're comfortable with that (I usually don't stomp over other's CR's). There's also the "```suggestion" boxes which are really convenient for both parties and allow some negotiation rather than a direct edit. I'll add one now...

fdwr, line 3823:
```suggestion
    1. If |options|.{{MLLayerNormalizationOptions/axes}} does not [=map/exist=], then set |options|.{{MLLayerNormalizationOptions/axes}} to a new [=/list=], either equal to [=the range=] from 1 to |input|'s [=rank-=], exclusive, if |input|'s [=rank=] is greater than 1, or an empty [=/list=] otherwise.
```

fdwr: Does read clearly enough @huningxin and @inexorabletash?

inexorabletash (commented):
LGTM!

fdwr:
Oh, we still ought to update the *prose* too, as it still says: "When this member is not present, it is assumed to be [1,2,3] that is..."

I'll push another iteration...

inexorabletash (commented):
LGTM

inexorabletash:
(The build error is due to the lingering {{MLCommamdEncoder}} reference)

PR #539, opened 2024-01-29
Bug fix: Optional sequences shouldn't be defaulted to null

inexorabletash:
MLArgMinMaxOptions and MLReduceOptions have an 'axes' member that the IDL declares as defaulting to null. They have type sequence<unsigned long> which is not nullable. The intent is that these are optional members with default behavior if not present, which is handled in the algorithm already.

If the desire is to be able to explicitly pass null, the IDL fix would be to make the type nullable: `sequence<unsigned long>? axes = null;` but that's inconsistent with the rest of the spec and not the convention for optional dictionary members.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/539.html" title="Last updated on Feb 7, 2024, 5:43 PM UTC (44d5e0b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/539/d1a02f2...inexorabletash:44d5e0b.html" title="Last updated on Feb 7, 2024, 5:43 PM UTC (44d5e0b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1477,7 +1477,7 @@ Return the index location of the minimum or maxmium values of all the input valu
 
 <script type=idl>
 dictionary MLArgMinMaxOptions {
-  sequence<unsigned long> axes = null;
+  sequence<unsigned long> axes;
   boolean keepDimensions = false;
   boolean selectLastIndex = false;
 };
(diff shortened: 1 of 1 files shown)


fdwr:
Looks right to me. Like you said, the intention is that you may omit optional `axes` completely in the `MLReduceOptions` to get a form of "global reduce", which is treated equivalent to having passed *all* axes (e.g. for 5D, `{axes: [0,1,2,3,4]}`). Note that complete omission of `axes` is also distinct from passing an *explicitly* empty sequence (e.g `{axes: [], keepDimensions: true}`) which is effectively treated as a nop.

huningxin (approved):
LGTM

null value doesn't have any semantics for axes, we should remove it.

fdwr (approved):

fdwr:
Merging this. Chai is on the reviewers list for this CR, but we don't need to wait for his review (after speaking with him, there's a select set of reviews and topics he's interested in, but otherwise he won't be looking at every CR).

PR #555, opened 2024-02-05
Content change: Add definitions for MLOperand's shape and dataType

inexorabletash:
Simplify text that digs the dimensions and dataType out of an MLOperand's [[descriptor]] by introducing definitions, and use them consistently.

Also simplify the definition for MLOperand's rank, and use it in more places.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/555.html" title="Last updated on Feb 13, 2024, 7:07 PM UTC (ef7355e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/555/d48e394...inexorabletash:ef7355e.html" title="Last updated on Feb 13, 2024, 7:07 PM UTC (ef7355e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1611,16 +1611,16 @@ partial interface MLGraphBuilder {
   </summary>
   <div class=algorithm-steps>
     1. If |options|.axis is not in [=the range=] 0 to the [=rank=] of |input|, exclusive, then [=exception/throw=] a {{TypeError}}.
-    1. If the [=list/size=] of |mean|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} is not 1, then [=exception/throw=] a {{TypeError}}.
-    1. If |mean|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[0] is not equal to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
-    1. If the [=list/size=] of |variance|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} is not 1, then [=exception/throw=] a {{TypeError}}.
-    1. If |variance|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[0] is not equal to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
+    1. If |mean|'s [=MLOperand/rank=] is not 1, then [=exception/throw=] a {{TypeError}}.
+    1. If |mean|'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
+    1. If |variance|'s [=MLOperand/rank=] is not 1, then [=exception/throw=] a {{TypeError}}.
+    1. If |variance|'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLBatchNormalizationOptions/scale}} [=map/exists=]:
         1. If its [=list/size=] is not 1, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLBatchNormalizationOptions/scale}}.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[0] is not equal to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
+        1. If |options|.{{MLBatchNormalizationOptions/scale}}'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLBatchNormalizationOptions/bias}} [=map/exists=]:
         1. If its [=list/size=] is not 1, then [=exception/throw=] a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


zolkis:
Looks good.

index.bs
```diff
@@ -6077,12 +6077,11 @@ interface MLOperand {};
...
 
-<div algorithm>
-To get the <dfn for="MLOperand">rank</dfn> of an {{MLOperand}} |operand|, run the following steps:
-<div class=algorithm-steps>
-    1. Return the [=list/size=] of |operand|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
-</div>
-</div>
+An {{MLOperand}}'s <dfn for=MLOperand>shape</dfn> is its {{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}.
+
+An {{MLOperand}}'s <dfn for=MLOperand>rank</dfn> is its [=MLOperand/shape=]'s [=list/size=].
```
huningxin, line 6082: {{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} might not present, if MLOperand is a scalar. A scalar's rank is 0.

inexorabletash: Good point. Is it fair to say that the shape of a scalar is [ ] ? Or do we want to introduce an assertion, e.g. make getting the shape a sequence of steps that assert the shape exists and the size is non-zero? (and make rank smarter)

From reviewing the spec it looks like there are these cases:

* operand's descriptor's `dimensions` are used without checking -- mostly replaced here by using **shape**, so a neutral change
* operand's descriptor's `dimensions` are checked with "exists" -- that's fine

Also, MLOperand's [shape()](https://webmachinelearning.github.io/webnn/#dom-mloperand-shape) io defined as `sequence<unsigned long>` but the steps are "Return this.[[descriptor]].dimensions." - what's the intended behavior for a scalar? I'll file a bug.

inexorabletash: This should wait for #476 to land first, which better defines scalars.

fdwr: > Is it fair to say that the shape of a scalar is [ ] 

Correct.

> Or do we want to introduce an assertion, e.g. make getting the shape a sequence of steps that assert the shape exists and the size is non-zero?

A shape's rank can definitely be zero.

> what's the intended behavior for a scalar?

Intended behavior is a sequence of 0 elements.

index.bs
```diff
@@ -1611,16 +1611,16 @@ partial interface MLGraphBuilder {
   </summary>
   <div class=algorithm-steps>
     1. If |options|.axis is not in [=the range=] 0 to the [=rank=] of |input|, exclusive, then [=exception/throw=] a {{TypeError}}.
-    1. If the [=list/size=] of |mean|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} is not 1, then [=exception/throw=] a {{TypeError}}.
-    1. If |mean|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[0] is not equal to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
-    1. If the [=list/size=] of |variance|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}} is not 1, then [=exception/throw=] a {{TypeError}}.
-    1. If |variance|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[0] is not equal to |input|.{{MLOperand/[[descriptor]]}}.{{MLOperandDescriptor/dimensions}}[|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
+    1. If |mean|'s [=MLOperand/rank=] is not 1, then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 1614: 👍 I like using the existing conceptual definition of `rank`, rather than repeating `operand.list size of [[descriptor]].dimensions` everywhere - more concise.

index.bs
```diff
@@ -1611,16 +1611,16 @@ partial interface MLGraphBuilder {
...
+    1. If |mean|'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 1615: Does the `shape` function need `()` after it, `mean.shape()[0]`? Otherwise JS would give `undefined` to you if you tried `mean.shape[0]` directly, no? Or does bikeshed automagically add them? Or am I misinterpreting MLOperand/shape here as a bikeshed-specific *helper* function rather than the IDL `shape` function.

index.bs
```diff
@@ -1611,16 +1611,16 @@ partial interface MLGraphBuilder {
...
+    1. If |variance|'s [=MLOperand/rank=] is not 1, then [=exception/throw=] a {{TypeError}}.
+    1. If |variance|'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 1617: I don't mind using `'s` form instead (tis shorter), but I see many places that say "property of thing" (e.g. rank of input, rank of desc, shape of filter tensor, size of options.dilations), and so if we're changing that broadly, then consistency is nice (updating "rank of input" --> "input's rank"), unless you actually got them all here and now. Either way, incrementally is okay too, as it could make this CR large.

index.bs
fdwr: With 476's update, the dimensions should always be present and initialized to at least `[]` by default.

fdwr (commented):
"is this an improvement?" - I think so. Thanks.

index.bs
inexorabletash: Bikeshed helper - really, just spec convenience - not IDL. Invoking JS from within a spec algorithm is generally a no-no, due to side effects.

index.bs
inexorabletash: Yeah, I was thinking a follow-on. Modern specs use the shorter form even when a bit weird, like "this's"

fdwr (approved):

fdwr:
@inexorabletash : Please promote from draft when ready. 0D is merged.

huningxin (approved):
LGTM, thanks!

inexorabletash:
Okay, merged the changes from main. I didn't force-push so the history has a merge commit. Hopefully I did it right!

Should be ready for review.

fdwr (approved):
SLGTM.

Issue #456, opened 2023-08-24
Define the maximum number of operand dimensions (maximum rank)


huningxin:
Regarding to the current definition of [`MLOperandDescriptor`](https://www.w3.org/TR/webnn/#api-mloperanddescriptor)

```js
dictionary MLOperandDescriptor {
  // The operand type.
  required MLOperandType type;

  // The dimensions field is only required for tensor operands.
  sequence<unsigned long> dimensions;
};
```

there is no definition of the maximum number of the `dimensions` sequence in spec.

However, for implementation, the native ML APIs usually have the maximum supported size. For example:
- [XNNPACK](https://github.com/google/XNNPACK/blob/master/include/xnnpack.h#L28C1-L28C30): #define XNN_MAX_TENSOR_DIMS 6
- [DirectML](https://learn.microsoft.com/en-us/windows/ai/directml/directml-constants): 
<html>
<body>

Constant | Value | Description
-- | -- | --
DML_TENSOR_DIMENSION_COUNT_MAX | 5 | DirectML tensors support a maximum of 5 dimensions for DML_TARGET_VERSION < DML_FEATURE_LEVEL_3_0.
DML_TENSOR_DIMENSION_COUNT_MAX1 | 8 | DirectML tensors support a maximum of 8 dimensions for DML_TARGET_VERSION >= DML_FEATURE_LEVEL_3_0.

</body>
</html>

And there may be per operator definitions, such as for [convolution operator](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_convolution_operator_desc), the maximum dimensions count is 5, for [element-wise add operator](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_element_wise_add_operator_desc), the maximum dimensions count is 8.  Thanks @fdwr [sharing](https://chromium-review.googlesource.com/c/chromium/src/+/4728963/comment/0568a1f0_6af3da02/) this information!

In a Chromium [CL review](https://chromium-review.googlesource.com/c/chromium/src/+/4778299/comment/4d270070_c95ee9c4/), @RafaelCintron (Thanks!) mentions
> we should have a better solution for web developers to know ahead of time which operator parameters are expected to fail and which will not so they can know which models to use. Requiring JS error string parsing for each browser vendor is not a great solution.

Rafael also shared that WebGPU solved a similar problem with "limits" https://gpuweb.github.io/gpuweb/#limits.

fdwr:
[...]

PR #575, opened 2024-02-16
Bug fix: Fix MLGraphBuilder.input()'s handling of scalars. Fixes #502

### LINKED ISSUE #502: Convert "If descriptor.dimensions does not exist, then descriptor defines a scalar input." from assertion to note

inexorabletash:
MLOperandDescriptor was updated in c320472d to always have dimensions, defaulted to an empty list for scalars. That makes the current prose for input() incorrect. Issue #502 already tracked correcting it, so let's simplify - just change the logic for "is a scalar?" and drop the bogus assert.

Note that I didn't leave a note about empty dimensions being a scalar, but can add that if desired.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3318,8 +3318,7 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
     1. If |name| is empty, then [=exception/throw=] a {{TypeError}}.
-    1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input.
-    1. If |descriptor|.{{MLOperandDescriptor/dimensions}} [=map/exists=]:
+    1. If |descriptor|.{{MLOperandDescriptor/dimensions}}'s [=list/size=] is not 0:
         1. If [=checking dimensions=] given |descriptor|.{{MLOperandDescriptor/dataType}} and |descriptor|.{{MLOperandDescriptor/dimensions}} returns false, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
         1. If |descriptor|'s [=byte length=] is not supported by the underlying platform, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. If any of the following sub-steps fail, [=exception/throw=] an "{{OperationError}}" {{DOMException}}.


index.bs
```diff
@@ -3318,8 +3318,7 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
     1. If |name| is empty, then [=exception/throw=] a {{TypeError}}.
-    1. [=Assert=]: If |descriptor|.{{MLOperandDescriptor/dimensions}} does not [=map/exist=], then |descriptor| defines a scalar input.
-    1. If |descriptor|.{{MLOperandDescriptor/dimensions}} [=map/exists=]:
+    1. If |descriptor|.{{MLOperandDescriptor/dimensions}}'s [=list/size=] is not 0:
```
fdwr, line 3321: Can we just delete this extra conditional line? The 0-size case shouldn't need special handling here because "checking dimensions" should handle it correctly, but I should have deleted this line too in it:

`If dimensions’s [size](https://infra.spec.whatwg.org/#list-size) is 0, return false.`

The "checking dimensions" step is apparently only used by `input` and `constant(descriptor, bufferView)`.

(https://www.w3.org/TR/webnn/#ref-for-check-dimensions)

Then we could unindent line 3322 and just always process it.

index.bs
```diff
@@ -3318,8 +3318,7 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
...
         1. If [=checking dimensions=] given |descriptor|.{{MLOperandDescriptor/dataType}} and |descriptor|.{{MLOperandDescriptor/dimensions}} returns false, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
         1. If |descriptor|'s [=byte length=] is not supported by the underlying platform, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
```
fdwr, line 3323: > If |descriptor|'s [=byte length=] is not supported by the underlying platform, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.

I wonder if this line makes more sense to move into "check dimensions" to apply generically, as it already has this closely related bit:

`If ... it is too large to be supported by the implementation given type, return false.`

index.bs
inexorabletash: done, pushed a new commit

index.bs
inexorabletash: Yeah, and it's there in constant() too...

inexorabletash: Okay, this was a bit more intrusive so I did it as a separate commit, but I liked where it ended up. Take a look?

inexorabletash: (uh, last two commits together)

index.bs
```diff
@@ -5783,10 +5779,12 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
 
 <details open algorithm>
   <summary>
-    To <dfn>check dimensions</dfn> given [=/list=] |dimensions| and {{MLOperandDataType}} |type|, run the following steps:
+    To <dfn>check dimensions</dfn> given {{MLOperandDescriptor}} |descriptor|, run the following steps:
```
fdwr, line 5782: I appreciate bundling of both params into a single descriptor, but I wonder if it was that way before for cases like with https://www.w3.org/TR/webnn/#api-mlgraphbuilder-expand which has a separate `newShape` parameter (but not a full descriptor since the type remains the same as the input). Granted, `expand` doesn't currently call "check dimensions" anyway, instead relying on the bidirectional broadcasting, but then I'm not seeing the broadcasting function call "check dimensions" either. So somewhere along the chain `expand` should call check dimensions. Though, that's not part of this CR, and we can always split it back if needed later. So signing off.

fdwr (approved):
👍 Ningxin said he'd be back next week if you can wait a little longer (I don't want to complete *too* many CR's without my editorial partner 😅), but LGTM.

index.bs
inexorabletash: Yeah, the reason for combining is that only a descriptor has a "byte length" so we'd need to rework that definition too if we kept it split.

index.bs
```diff
@@ -5783,10 +5779,12 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
...
   </summary>
   <div class=algorithm-steps>
-    1. If |dimensions|'s [=list/size=] is 0, return false.
+    1. Let |dimensions| be |descriptor|.{{MLOperandDescriptor/dimensions}}.
+    1. Let |type| be |descriptor|.{{MLOperandDescriptor/dataType}}.
+    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
     1. If |dimensions|'s [=list/size=] is too large to be supported by the implementation, return false.
     1. If any element of |dimensions| is not a positive number, or it is too large to be supported by the implementation given |type|, return false.
```
huningxin, line 5789: Given the dimension data type is [defined](https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/575.html#dom-mloperanddescriptor-dimensions) as unsigned long, would this check “If any element of |dimensions| is not a positive number” be unnecessary? 

And I suppose the remaining check "or it is too large to be supported by the implementation given |type|" should be covered by above descriptor's byte length check. Should we remove it? 

I understand this is an old issue. I am fine that we want to fix it in a separate PR.

index.bs
```diff
@@ -5783,10 +5779,12 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
 
 <details open algorithm>
   <summary>
-    To <dfn>check dimensions</dfn> given [=/list=] |dimensions| and {{MLOperandDataType}} |type|, run the following steps:
+    To <dfn>check dimensions</dfn> given {{MLOperandDescriptor}} |descriptor|, run the following steps:
   </summary>
   <div class=algorithm-steps>
-    1. If |dimensions|'s [=list/size=] is 0, return false.
+    1. Let |dimensions| be |descriptor|.{{MLOperandDescriptor/dimensions}}.
+    1. Let |type| be |descriptor|.{{MLOperandDescriptor/dataType}}.
```
huningxin, line 5786: Nit: would |dataType| be more explicit?

huningxin (approved):
The fix of MLGraphBuilder.input() LGTM, thanks!

Just left a comment for check dimensions algorithm, I am fine we fix it in a separate PR.

index.bs
inexorabletash: Yeah I'd prefer dataType too, was just using type as it was already there for one of the lines. I'll update it.

index.bs
inexorabletash: Yes but just to confirm the progression here...

```
    1. Let |dimensions| be |descriptor|.{{MLOperandDescriptor/dimensions}}.
    1. Let |dataType| be |descriptor|.{{MLOperandDescriptor/dataType}}.
    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
    1. If |dimensions|'s [=list/size=] is too large to be supported by the implementation, return false.
    1. If any element of |dimensions| is not a positive number, or it is too large to be supported by the implementation given |dataType|, return false.
    1. Return true.
```
simplifies to:
```
    1. Let |dimensions| be |descriptor|.{{MLOperandDescriptor/dimensions}}.
    1. Let |dataType| be |descriptor|.{{MLOperandDescriptor/dataType}}.
    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
    1. If |dimensions|'s [=list/size=] is too large to be supported by the implementation, return false.
    1. If |dimensions| is too large to be supported by the implementation given |dataType|, return false.
    1. Return true.
```
simplifies to:
```
    1. Let |dimensions| be |descriptor|.{{MLOperandDescriptor/dimensions}}.
    1. Let |dataType| be |descriptor|.{{MLOperandDescriptor/dataType}}.
    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
    1. If |dimensions|'s [=list/size=] is too large to be supported by the implementation, return false.
    1. Return true.
```
simplifies to:
```
    1. Let |dimensions| be |descriptor|.{{MLOperandDescriptor/dimensions}}.
    1. Let |dataType| be |descriptor|.{{MLOperandDescriptor/dataType}}.
    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
    1. Return true.
```
simplifies to:
```
    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
    1. Return true.
```
This works for me, but I want to make sure I didn't miss a nuance here.

inexorabletash: I guess `If |dimensions|'s [=list/size=] is too large to be supported by the implementation, return false.` is distinct and could be kept, per #456

So maybe the minimal version is:

```
    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
    1. If |descriptor|.{{MLOperandDescriptor/dimensions}}'s [=list/size=] is too large to be supported by the implementation, return false.
    1. Return true.
```

huningxin (approved):
LGTM, thanks much!

index.bs
fdwr: I like that, because the two limits are technically distinct, and we encounter them with real products:
- Dimensions can be limited (e.g. only supporting lengths per dimensions up to uint32 max), even if it can create a backing resource with more bytes than uint32 max.
- Byte size can be limited (not creating a resource of more than 16GBs), even if all the dimensions are in range.

index.bs
```diff
@@ -5887,6 +5873,23 @@ dictionary MLOperandDescriptor {
...
+  </summary>
+  <div class=algorithm-steps>
+    1. If |descriptor|'s [=byte length=] is not supported by the implementation, then return false.
+    1. If |descriptor|.{{MLOperandDescriptor/dimensions}}'s [=list/size=] is too large to be supported by the implementation, return false.
+
+        Issue(456): The maximum number of operand dimensions is not defined, but native ML APIs usually have a maximum supported size.
+
+    1. Return true.
+  </div>
+
```
fdwr, line 5888: extra spacing intended?

fdwr (approved):

index.bs
inexorabletash: Yeah, otherwise the Issue is not recognized and formatted correctly.

index.bs
fdwr: Oh yes, that should be re-added. I misread that last night "If |dimensions|'s [=list/size=] is too large" as meaning "if any of the dimensions were too large", but you meant the total rank there (e.g. \<=8D). I'd just remove this `given |dataType|`, because the dataType of the tensor data doesn't make any difference for the limit of the extents. TY.

```
1. If any element of |dimensions| is too large to be supported by the implementation, return false.
```

fdwr: I just restored that line and move the byte check last (typically the rank and dimension are checked first before even trying to compute the total number of bytes of the tensor). PTAL.

inexorabletash: Failure due to:
```
      LINE ~5881: The var 'dimensions' (in algorithm 'check dimensions' for MLOperandDescriptor) is only used once.
      If this is not a typo, please add an ignore='' attribute to the <var>.
```

...since I deleted the `Let |dimensions| be ...`  local var declaration. Either restore that (and use it in the other line), or use the `|descriptor|.{{MLOperandDescriptor/dimensions}}` longhand.

inexorabletash:
LGTM thanks @fdwr !

huningxin (commented):
The latest change LGTM, thanks @fdwr !

PR #577, opened 2024-02-21
Bug fix: Make concat() validation and initialization more precise.

### LINKED ISSUE #503: concat() assertions around input shape/datatypes seem invalid

inexorabletash:
- Validate that inputs is non-empty, present in the Chromium prototype implementation.
- Add local "first" variable, to simplify steps.
- Drop assertion about dataTypes being the same, already covered by the algorithm.
- Add step validating that the ranks are the same, present in the Chromium prototype implementation, and drop the assertion.
- Make output descriptor creation more explicit.
- Drop "If any of the following steps fail" since the steps explicitly threw already.
- Init desc.dimensions[axis] from "first" and start loop at 1.

Fixes #503

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/577.html" title="Last updated on Feb 21, 2024, 6:15 PM UTC (4b3f05e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/577/4ad9a63...inexorabletash:4b3f05e.html" title="Last updated on Feb 21, 2024, 6:15 PM UTC (4b3f05e)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1554,22 +1554,24 @@ partial interface MLGraphBuilder {
     <div class="note">
         The permissions and context validity have been checked by [[#api-mlgraphbuilder-constructor]] steps.
     </div>
-    1. [=Assert=]: the shape, i.e. {{MLOperandDescriptor/dimensions}} of each operand in |inputs| is the same, except on the dimension given by |axis| on which they are concatenated.
-    1. [=Assert=]: the {{MLOperandDescriptor/dataType}} of each operand in |inputs| is the same.
-    1. If any of the following steps fail, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-        1. Let |desc| be |inputs|[0].{{MLOperand/[[descriptor]]}}.
-        1. If |axis| is greater than or equal to |desc|'s [=rank=], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-        1. Set |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] to 0.
-        1. [=list/For each=] |index| in [=the range=] 0 to |inputs|'s [=rank=], exclusive:
-            1. Let |input| be |inputs|[|index|].
-            1. If [=validating MLOperand=] given |input| and [=this=] returns false, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-            1. If |input|'s [=MLOperand/dataType=] is not equal to |inputs|[0]'s [=MLOperand/dataType=], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-            1. [=list/For each=] |dim| in [=the range=] 0 to |input|'s [=rank=], exclusive:
-                <div class="note">
-                    If the shape of each corresponding dimension and type of the operands, except for those of the dimension given by |axis|, is not the same, fail.
-                </div>
-                1. If |dim| is not equal to |axis| and if |input|'s [=MLOperand/shape=][|dim|] is not equal to |inputs|[0]'s [=MLOperand/shape=][|dim|], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1554,22 +1554,24 @@ partial interface MLGraphBuilder {
...
-                </div>
-                1. If |dim| is not equal to |axis| and if |input|'s [=MLOperand/shape=][|dim|] is not equal to |inputs|[0]'s [=MLOperand/shape=][|dim|], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-                1. If |dim| is equal to |axis|, add to |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] the value of |input|'s [=MLOperand/shape=][|dim|].
+    1. If |inputs| [=list/is empty=], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |first| be |inputs|[0].
+    1. If |axis| is greater than or equal to |first|'s [=rank=], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |desc| be a new {{MLOperandDescriptor}}.
+    1. Set |desc|.{{MLOperandDescriptor/dataType}} to |first|'s [=MLOperand/dataType=].
+    1. Set |desc|.{{MLOperandDescriptor/dimensions}} to a [=list/clone=] of |first|'s [=MLOperand/shape=].
+    1. Set |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] to 0.
```
huningxin, line 1563: Does it make sense by setting |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] to |first|'s [=MLOperand/shape=][|axis|]? And the following loop can start from index 1.

huningxin (approved):
LGTM with a comment, thanks so much!

index.bs
inexorabletash: Done!

huningxin (approved):
LGTM!

huningxin:
@fdwr , would you also like to take a look? Thanks!

index.bs
```diff
@@ -1554,22 +1554,24 @@ partial interface MLGraphBuilder {
...
-            1. [=list/For each=] |dim| in [=the range=] 0 to |input|'s [=rank=], exclusive:
-                <div class="note">
-                    If the shape of each corresponding dimension and type of the operands, except for those of the dimension given by |axis|, is not the same, fail.
-                </div>
-                1. If |dim| is not equal to |axis| and if |input|'s [=MLOperand/shape=][|dim|] is not equal to |inputs|[0]'s [=MLOperand/shape=][|dim|], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-                1. If |dim| is equal to |axis|, add to |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] the value of |input|'s [=MLOperand/shape=][|dim|].
+    1. If |inputs| [=list/is empty=], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |first| be |inputs|[0].
+    1. If |axis| is greater than or equal to |first|'s [=rank=], then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |desc| be a new {{MLOperandDescriptor}}.
```
fdwr, line 1560: I see we call it slightly different in a few places:

`Let desc be a new MLOperandDescriptor.`
`Let descriptor be a new MLOperandDescriptor.`

A name like `outputDesc` would make it clear up front that this isn't just *any* temporary tensor description, but that it will specifically become the primary output description (as opposed to one invented for intermediate processing). but that would be too big a cascading change for here. So resolving my own comment.

fdwr (approved):
👍

Issue #572, opened 2024-02-15
Synchronously validate input operands/activations


inexorabletash:
When calling a builder method on [MLGraphBuilder](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder) that takes an [MLOperand](https://webmachinelearning.github.io/webnn/#api-mloperand) or [MLActivation](https://webmachinelearning.github.io/webnn/#api-mlactivation) as an input (i.e. anything but [input()](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-input) and [constant()](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-constant)), there is no check that the input is from the same builder. 

The only place in the spec that does this is [validate MLOperand](https://webmachinelearning.github.io/webnn/#mloperand-validate-mloperand) steps that take an operand and builder.

This algorithm is only called in:
* [concat()](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-concat) with _"If validating MLOperand given input and this returns false, then throw a "DataError" DOMException."_ which is bogus - it should pass input and this.[[builder]].
* [build()](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-build) asynchronously validates that the output operands are from the same builder, with _"If validating MLOperand given operand and this returns false, then reject promise with a TypeError, and abort these steps."_

@a-sully raised this in #552 and it was previously mentioned (but without specifics) in #234

Should failure here be synchronous? This is implied by the behavior of concat() and the open ended _"If any of the following sub-steps fail, ..."_ text in the builder methods. I don't see a reason for async failure (except in build itself, since you can't synchronously fail a method that returns a promise) so I'd strongly prefer synchronous failure. (If async, the build() should be augmented to traverse the whole graph.)

Assuming we want synchronous failure we could spec this in a few ways:

* Add explicit steps like concat() to all builder methods, e.g.: _If validating MLOperand given input and this.[[builder]] returns false, then throw a "DataError" DOMException._ - repeated for every input (required and optional if given). 
* Define steps for [connect](https://github.com/webmachinelearning/webnn/issues/549#issuecomment-1921901735) which validate and return failure; currently the builder method steps all already include _If any of the following sub-steps fail, throw an "OperationError" DOMException._

Either way, concat() needs to be aligned with the other methods (fix to pass this.[[builder]] or remove redundant steps)

This also raises the question of what exception type it should be:
* "DataError" DOMException, like concat(), or
[...]

Issue #326, opened 2023-01-10
Define the algorithm of calculating the effective padding for "same-upper" and "same-lower" option


huningxin:
WebNN [conv2d](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-conv2d) operation allows to set [`MLConv2dOptions.autoPad`](https://www.w3.org/TR/webnn/#dom-mlconv2doptions-autopad) option to "same-upper" or "same-lower" of [`MLAutoPad`](https://www.w3.org/TR/webnn/#enumdef-mlautopad) enum.

The spec mentions:

> With the "same-upper" option, the padding values are automatically computed such that the additional ending padding of the spatial input dimensions would allow all of the input values in the corresponding dimension to be filtered. The "same-lower" option is similar but padding is applied to the beginning padding of the spatial input dimensions instead of the ending one.

The spec should define the algorithm of how the padding values are automatically computed.

This issue was raised in Chromium [CL](https://chromium-review.googlesource.com/c/chromium/src/+/4031876/comment/2b0411bb_72492bfe/) review. Thanks @wacky6 for the feedback.

fdwr:
I agree with this issue's intent to better document padding behavior.

Though, I also wonder whether more of this higher-level policy can be resolved by the caller? I just finished looking through Bin's [WebNN: Define convTranspose2d operator in mojo](https://chromium-review.googlesource.com/c/chromium/src/+/4952434), and it's a lot of complexity to validate and compute the output shapes. Like how we want to [simplify `matmul`'s complicated 1D reshaping higher layer policy](https://github.com/webmachinelearning/webnn/issues/470), is there anything here too we can simplify to reduce the implementation, testing burden, and spec complexity for `conv`? The calling frameworks (e.g. TF/ORT) already have this logic anyway for their own needs and could resolve some of those policies like massaging automagic padding into explicit padding (or even to an explicit output shape) before reaching WebNN, and so it *shouldn't* be an undue burden on them, possibly even just a 1 or 2-line function call for them to resolve explicit padding and output.

huningxin:
> is there anything here too we can simplify to reduce the implementation, testing burden, and spec complexity for conv?

Do you propose we should drop the support of [`MLAutoPad`](https://www.w3.org/TR/webnn/#enumdef-mlautopad) and only support explicit padding? If the frameworks can handle that, that would definitely reduce the implementation complexity.

Honry:
[...]

PR #582, opened 2024-02-24
Content: Specify output size calculations

### LINKED ISSUE #500: Define "calculating output dimensions"

inexorabletash:
This covers:
* Reduction ops (including argMin/argMax)
* conv2d
* convTranspose2d
* Pooling ops (which rely on conv2d)

Fixes #500

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/582.html" title="Last updated on Feb 27, 2024, 9:51 PM UTC (2f9868c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/582/f269174...inexorabletash:2f9868c.html" title="Last updated on Feb 27, 2024, 9:51 PM UTC (2f9868c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1216,7 +1216,8 @@ partial interface MLGraphBuilder {
   </summary>
   <div class=algorithm-steps>
     1. [=Assert=]: |op| is one of "argMin", "argMax".
-    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating reduction output sizes=] given |input|'s [=MLOperand/shape=], |options|.{{MLArgMinMaxOptions/axes}} (if it [=map/exists=]), and |options|.{{MLArgMinMaxOptions/keepDimensions}}. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. If |options|.{{MLArgMinMaxOptions/axes}} is given, then if any its [=list/items=] are not in [=the range=] 0 to |input|'s [=MLOperand/rank=], exclusive, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating reduction output sizes=] given |input|'s [=MLOperand/shape=], |options|.{{MLArgMinMaxOptions/axes}} (if it [=map/exists=]), and |options|.{{MLArgMinMaxOptions/keepDimensions}}.
     1. Let |desc| be a new {{MLOperandDescriptor}}.
     1. Set |desc|.{{MLOperandDescriptor/dataType}} to {{MLOperandDataType/"int64"}}.
     1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
@@ -4848,11 +4849,10 @@ partial interface MLGraphBuilder {
 
 <details open algorithm>
   <summary>
-    To <dfn for="MLGraphBuilder">calculate reduction output sizes</dfn>, given a [=/list=] of unsigned integers |inputShape|, a optional [=/list=] of unsigned integers |axes|, and [=/boolean=] |keepDimensions|, perform the following steps. They return a new [=/list=] of unsigned integers, or failure.
+    To <dfn for="MLGraphBuilder">calculate reduction output sizes</dfn>, given a [=/list=] of unsigned integers |inputShape|, a optional [=/list=] of unsigned integers |axes|, and [=/boolean=] |keepDimensions|, perform the following steps. They return a new [=/list=] of unsigned integers.
   </summary>
   <div class=algorithm-steps>
     1. Let |inputSize| be |inputShape|'s [=list/size=].
-    1. If |axes| is given, then if any its [=list/items=] are not in [=the range=] 0 to |inputSize|, exclusive, return failure.
     1. If |axes| is not given, let |axes| be [=the range=] 0 to |inputSize|, exclusive.
     1. If |keepDimensions| is true, then:
         1. Let |outputShape| be a [=list/clone=] of |inputShape|.
@@ -4872,7 +4872,8 @@ partial interface MLGraphBuilder {
   </summary>
   <div class=algorithm-steps>
     1. [=Assert=]: |op| is one of "reduceL1", "reduceL2", "reduceLogSum", "reduceLogSumExp", "reduceMax", "reduceMean", "reduceMin", "reduceProduct", "reduceSum", "reduceSumSquare".
-    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating reduction output sizes=] given |input|'s [=MLOperand/shape=], |options|.{{MLReduceOptions/axes}} (if it [=map/exists=]), and |options|.{{MLReduceOptions/keepDimensions}}. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. If |options|.{{MLReduceOptions/axes}} is given, then if any its [=list/items=] are not in [=the range=] 0 to |input|'s [=MLOperand/rank=], exclusive, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating reduction output sizes=] given |input|'s [=MLOperand/shape=], |options|.{{MLReduceOptions/axes}} (if it [=map/exists=]), and |options|.{{MLReduceOptions/keepDimensions}}.
     1. Let |desc| be a new {{MLOperandDescriptor}}.
     1. Set |desc|.{{MLOperandDescriptor/dataType}} to |input|'s [=MLOperand/dataType=].
     1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
--- a/index.bs
+++ b/index.bs
@@ -1216,8 +1216,7 @@ partial interface MLGraphBuilder {
   </summary>
   <div class=algorithm-steps>
     1. [=Assert=]: |op| is one of "argMin", "argMax".
-    1. If |options|.{{MLArgMinMaxOptions/axes}} [=map/exists=], if any of its elements is not in [=the range=] 0 to |input|'s [=MLOperand/rank=], exclusive, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
-    1. Let |outputShape| be the result of invoking the underlying implementation for calculating reduction output dimensions, given |options|.
+    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating reduction output sizes=] given |input|'s [=MLOperand/shape=], |options|.{{MLArgMinMaxOptions/axes}} (if it [=map/exists=]), and |options|.{{MLArgMinMaxOptions/keepDimensions}}. If that returns failure, then [=exception/throw=] a "{{DataError}}" {{DOMException}}.
     1. Let |desc| be a new {{MLOperandDescriptor}}.
     1. Set |desc|.{{MLOperandDescriptor/dataType}} to {{MLOperandDataType/"int64"}}.
     1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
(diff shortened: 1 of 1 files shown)


zolkis:
Wow, this is big! Thanks!

inexorabletash:
> Wow, this is big! Thanks!

Yeah... I didn't know what I was getting myself into.

index.bs
```diff
@@ -1810,7 +1809,71 @@ partial interface MLGraphBuilder {
 </div>
 
 <details open algorithm>
+  <summary>
+    To <dfn for=MLGraphBuilder>calculate conv2d padding</dfn> given unsigned integers |inputSize|, |filterSize|, |stride| and |dilation| and {{MLAutoPad}} |autoPad|, perform these steps. They return a [=/list=] of 2 numbers.
```
huningxin, line 1813: @inexorabletash 

There is a proposal to drop the support of [`MLAutoPad`](https://www.w3.org/TR/webnn/#enumdef-mlautopad) and only support explicit padding. Should we coordinate with that proposal? For example, I can send a PR to remove `MLAutoPad`. What do you think?

@fdwr

index.bs
```diff
@@ -1834,7 +1897,47 @@ partial interface MLGraphBuilder {
...
+                    1. Let |filterHeight| be |filterShape|[2].
+                    1. Let |filterWidth| be |filterShape|[3].
+            </dl>
+        1. Let |outputSizes| be the result of [=MLGraphBuilder/calculating conv2d output sizes=] given |inputHeight|, |inputWidth|, |filterHeight|, |filterWidth|, |options|.{{MLConv2dOptions/padding}}, |options|.{{MLConv2dOptions/strides}}, |options|.{{MLConv2dOptions/dilations}} and |options|.{{MLConv2dOptions/autoPad}}.
+        1. Switch on |options|.{{MLConv2dOptions/inputLayout}}:
+            <dl class=switch>
+                : {{MLInputOperandLayout/"nchw"}}
+                :: Let |outputShape| be « |batches|, |channels|, floor( |outputSizes|[0] ), floor( |outputSizes|[1] ) ».
+                : {{MLInputOperandLayout/"nhwc"}}
+                :: Let |outputShape| be « |batches|, floor( |outputSizes|[0] ), floor( |outputSizes|[1], |channels| ) ».
```
huningxin, line 1939: Typo?
```suggestion
                :: Let |outputShape| be « |batches|, floor( |outputSizes|[0] ), floor( |outputSizes|[1] ), |channels|  ».
```

index.bs
```diff
@@ -1810,7 +1809,71 @@ partial interface MLGraphBuilder {
...
+</details>
+
+
+<details open algorithm>
+  <summary>
+    To <dfn for=MLGraphBuilder>calculate conv2d output size</dfn> given unsigned integers |inputSize|, |filterSize|, |beginningPadding|, |endingPadding|, |stride| and |dilation|, perform these steps. They return a number.
+  </summary>
+  <div class=algorithm-steps>
+    1. Let |effectiveFilterSize| be ( |filterSize| - 1 ) * |dilation| + 1.
+    1. Let |outputSize| be ( |inputSize| - |effectiveFilterSize| + |beginningPadding| + |endingPadding| ) / |stride| + 1.
```
huningxin, line 1843: Should we handle any errors? Like |outputSize| is overflow or underflow?

index.bs
```diff
@@ -1999,7 +2169,46 @@ partial interface MLGraphBuilder {
...
+                    1. Let |filterHeight| be |filterShape|[1].
+                    1. Let |filterWidth| be |filterShape|[2].
+            </dl>
+        1. Let |outputSizes| be the result of [=MLGraphBuilder/calculating convtranspose2d output sizes=] given |inputHeight|, |inputWidth|, |filterHeight|, |filterWidth|, |options|.{{MLConvTranspose2dOptions/padding}}, |options|.{{MLConvTranspose2dOptions/strides}}, |options|.{{MLConvTranspose2dOptions/dilations}}, |options|.{{MLConvTranspose2dOptions/outputPadding}}, and |options|.{{MLConvTranspose2dOptions/autoPad}}.
+        1. Switch on |options|.{{MLConvTranspose2dOptions/inputLayout}}:
+            <dl class=switch>
+                : {{MLInputOperandLayout/"nchw"}}
+                :: Let |outputShape| be « |batches|, |channels|, floor( |outputSizes|[0] ), floor( |outputSizes|[1] ) ».
+                : {{MLInputOperandLayout/"nhwc"}}
+                :: Let |outputShape| be « |batches|, floor( |outputSizes|[0] ), floor( |outputSizes|[1], |channels| ) ».
```
huningxin, line 2210: Same here, a typo?
```suggestion
                :: Let |outputShape| be « |batches|, floor( |outputSizes|[0] ), floor( |outputSizes|[1] ), |channels| ».
```

index.bs
```diff
@@ -4446,6 +4655,54 @@ partial interface MLGraphBuilder {
     </pre>
 </div>
 
+<details open algorithm>
+  <summary>
+    To <dfn for=MLGraphBuilder>calculate pool output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, {{MLRoundingType}} |roundingType|, [=/list=] of 2 unsigned integers |windowDimensions|, [=/list=] of 4 unsigned integers |padding|, [=/list=] of 2 unsigned integers |strides|, [=/list=] of 2 unsigned integers |dilations|, optional [=/list=] of 2 unsigned integers |outputSizes|, and {{MLAutoPad}} |autoPad|, perform these steps. They return a [=/list=] of 4 unsigned integers.
```
huningxin, line 4660: Maybe be more explicit about pool2d? "calculate pool2d output sizes"

index.bs
```diff
@@ -4629,14 +4886,33 @@ partial interface MLGraphBuilder {
         - *SumSquare*: Compute the sum of the square of all the input values along the axes.
 </div>
 
+<details open algorithm>
+  <summary>
+    To <dfn for="MLGraphBuilder">calculate reduction output sizes</dfn>, given a [=/list=] of unsigned integers |inputShape|, a optional [=/list=] of unsigned integers |axes|, and [=/boolean=] |keepDimensions|, perform the following steps. They return a new [=/list=] of unsigned integers, or failure.
+  </summary>
+  <div class=algorithm-steps>
+    1. Let |inputSize| be |inputShape|'s [=list/size=].
```
huningxin, line 4894: Would |inputRank| be more consistent within the spec?

index.bs
```diff
@@ -4629,14 +4886,33 @@ partial interface MLGraphBuilder {
...
+    1. If |axes| is given, then if any its [=list/items=] are not in [=the range=] 0 to |inputSize|, exclusive, return failure.
```
huningxin, line 4895: According to this [proposal](https://github.com/webmachinelearning/webnn/issues/572#issuecomment-1962198054), should the |axes| be validated in "Validate arguments:" section? Then, this method could assume the axes is present and valid.

index.bs
inexorabletash: SGTM - it would definitely simplify this a bunch!

index.bs
inexorabletash: If I search the spec I see "size be" turns up 11 times, "rank be" 0 times. 

(Rank might be better generally, but seems like that's a separate issue?)

index.bs
inexorabletash: done!

index.bs
inexorabletash: Ah, good point - I pulled it in to dedupe, but you're right - we need to be consistent. I pulled the failure mode back out.

index.bs
inexorabletash: That's a good question... the specs I've been involved in usually operate in an abstract space with infinite precision, unlimited range, etc., with very limited checking. 

I noticed the Chromium prototype impl is using checked math (i.e. testing underflow/overflow), and given the sizes that ML deals with it, it seems like this is going to be a practical concern. 

We can spec this any way we want... ideas welcome. We should look at other specs for inspiration, too.

index.bs
fdwr: Yes please.

index.bs
fdwr: Hmm, I've always found it cleaner when implementing specs (like bidi, line breaking, vertical layout...) that I was grateful they left out that degree of fine validation and instead focused on the algorithm itself. Any validation specific to the nature of the operators (e.g. maybe it only supports even sizes) belongs in there, but too much otherwise muddies an already complex document. Maybe we can even call that out somewhere in a common section, that the implementation should handle overflow/underflow/safe limit checks, but not as extra prose for each operator.

index.bs
fdwr: Good point. `inputSize` is quite ambiguous alone because many ML libraries have a `Size()` operator which (for better or worse, and probably should have been called `ElementCount()` instead) returns the total *number of elements*, not the count of dimensions. So, `inputRank` or `inputShapeSize` or `inputDimensionCount` would all clearer names (with `inputRank` being the most concise), but deferring that to a separate CR that replaces the other 11 occurrences would be fine too.

index.bs
```diff
@@ -4629,14 +4887,33 @@ partial interface MLGraphBuilder {
...
+    1. Let |inputSize| be |inputShape|'s [=list/size=].
+    1. If |axes| is not given, let |axes| be [=the range=] 0 to |inputSize|, exclusive.
+    1. If |keepDimensions| is true, then:
+        1. Let |outputShape| be a [=list/clone=] of |inputShape|.
+        1. [=list/For each=] |axis| of |axes|:
+            1. Set |outputShape|[|axis|] to 1.
+    1. Otherwise:
+        1. Let |outputShape| be an empty [=/list=].
+        1. [=list/For each=] |index| in [=the range=] 0 to |inputSize|, exclusive:
+            1. If |axes| does not [=list/contain=] |index|, then [=list/append=] |inputShape|[|index|].
```
fdwr, line 4904: Yep, this looks right, handling the empty `axes` too (nop case), the all axes case (scalar result case), and degenerate (but not illegal) cases like axes in arbitrary order [4,2,3] and repeats [2,3,2].

index.bs
```diff
@@ -4446,6 +4656,54 @@ partial interface MLGraphBuilder {
...
+                1. Let |inputHeight| be |inputShape|[2].
+                1. Let |inputWidth| be |inputShape|[3].
+            : {{MLInputOperandLayout/"nhwc"}}
+            ::
+                1. Let |batches| be |inputShape|[0].
+                1. Let |inputHeight| be |inputShape|[1].
+                1. Let |inputWidth| be |inputShape|[2].
+                1. Let |channels| be |inputShape|[3].
+        </dl>
+    1. If |outputSizes| is not given, then:
```
fdwr, line 4679:
```suggestion
    1. If |outputSizes| is given, then:
```

fdwr (approved):
😮 This must have taken a while. It's certainly an improvement over "Let outputShape be the result of invoking the underlying implementation for calculating output dimensions, given options". Thanks J.

index.bs
fdwr: Here it was: https://github.com/webmachinelearning/webnn/issues/326

index.bs
anssiko: @fdwr a W3C parlance nit: TL;DR: s/CR/PR

Confusingly CR has a [domain-specific meaning](https://www.w3.org/2023/Process-20231103/#RecsCR) in the W3C context. It does not refer to the usual generic term Change Request that is familiar from change management process. A W3C spec junkie might interpret CR in this context to mean Candidate Recommendation. To avoid this name collision, we've used PR to refer to a Pull Request.

To not make things too simple, each successful merge at this stage of the spec evolution will in fact [produce](https://www.w3.org/standards/history/webnn/) a new [Candidate Recommendation Draft](https://www.w3.org/2023/Process-20231103/#publishing-crud) of the spec so basically what you're saying above is accurate.

(I don't want to mention also [PR is overloaded](https://www.w3.org/2023/Process-20231103/#RecsPR), but let's just pretend PR means Pull Request and feel good :-))

index.bs
huningxin: https://github.com/webmachinelearning/webnn/pull/587

inexorabletash:
Marking as draft. Once #587 I'll push a rebased version. I have it locally and it's much smaller; most of the lines now handle the nchw/nhwc layout unpacking/packing.

index.bs
fdwr: 🧐 Then under webnn/*, I'll just always spell it out (CR -> code review / change request, PR ->  pull request).

inexorabletash:
Okay, rebased - sorry about the forced push, but it should be ready for another review now.

index.bs
inexorabletash: Captured as #588 so resolving this conversation.

huningxin (approved):
LGTM, thanks much!

PR #587, opened 2024-02-27
Remove MLAutoPad

### LINKED ISSUE #326: Define the algorithm of calculating the effective padding for "same-upper" and "same-lower" option

huningxin:
Fix #326

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/587.html" title="Last updated on Feb 27, 2024, 1:22 PM UTC (cb26ac4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/587/a904425...huningxin:cb26ac4.html" title="Last updated on Feb 27, 2024, 1:22 PM UTC (cb26ac4)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1709,17 +1709,10 @@ enum MLConv2dFilterOperandLayout {
   "ihwo"
 };
 
-enum MLAutoPad {
-  "explicit",
-  "same-upper",
-  "same-lower"
-};
-
 dictionary MLConv2dOptions {
   sequence<unsigned long> padding;
   sequence<unsigned long> strides;
   sequence<unsigned long> dilations;
-  MLAutoPad autoPad = "explicit";
   unsigned long groups = 1;
   MLInputOperandLayout inputLayout = "nchw";
   MLConv2dFilterOperandLayout filterLayout = "oihw";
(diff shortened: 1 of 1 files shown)


fdwr (approved):
Thanks Ningxin.

Issue #574, opened 2024-02-16
Consider alternate styling/linking for method argument definitions


inexorabletash:
Method arguments are currently detailed with this markdown (using compute() as an example):

```
<div>
    **Arguments:**
      - *graph*: an {{MLGraph}}. The compiled graph to be executed.
      - *inputs*: an {{MLNamedArrayBufferViews}}. The resources of inputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.
      - *outputs*: an {{MLNamedArrayBufferViews}}. The pre-allocated resources of required outputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.
</div>
```
which renders as:
<kbd><img src="https://github.com/webmachinelearning/webnn/assets/771547/aeb6face-1976-46c3-87ff-41eb2d805c40"></kbd>

There are a few other forms used in other specs worth considering to improve maintainability.

## Alternative 1: Make args into definitions

A minimal change makes the argument names become definitions which are get automatically linked, to help catch errors.
```
<div dfn-for="MLContext/compute(graph, inputs, outputs)" dfn-type="argument">
    **Arguments:**
      - <dfn>graph</dfn>: an {{MLGraph}}. The compiled graph to be executed.
      - <dfn>inputs</dfn>: an {{MLNamedArrayBufferViews}}. The resources of inputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.
      - <dfn>outputs</dfn>: an {{MLNamedArrayBufferViews}}. The pre-allocated resources of required outputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.
</div>
```
which renders as:
<kbd><img src="https://github.com/webmachinelearning/webnn/assets/771547/425ece34-f7ff-4248-a4d6-aa7ada0b7468"></kbd>

## Alternative 2: Style as definition list

Extending the above using a definition list, similar to what is currently done for dictionaries:
```
**Arguments:**
<dl dfn-for="MLContext/compute(graph, inputs, outputs)" dfn-type="argument">

: <dfn>graph</dfn>
:: an {{MLGraph}}. The compiled graph to be executed.

: <dfn>inputs</dfn>
:: an {{MLNamedArrayBufferViews}}. The resources of inputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.

: <dfn>outputs</dfn>
[...]

PR #592, opened 2024-02-28
Apply [EnforceRange] to all integer types

### LINKED ISSUE #562: Apply [EnforceRange] to integer types?
The prototype Chromium implementation applies [`[EnforceRange]`](https://webidl.spec.whatwg.org/#EnforceRange) to all places where integer (specifically all the `unsigned long`) values are used as inputs.

The spec does not currently use `[EnforceRange]` anywhere.

The behavior difference is that with `[EnforceRange]` a value outside the range results in an exception, so a value less than 0 or greater than 4294967295. Note that values are rounded so this doesn't result in an exception if e.g. `2.2` is passed, but it catches `-2`. I'd strongly encourage throwing (i.e. including `[EnforceRange]`) since conversion is otherwise non-intuitive.

Note that Chromium doesn't yet implement `MLTriangularOptions` which has a `long` (not `unsigned`); for consistency we'd probably want to annotate that as well.

a-sully:
Fixes #562

See the linked issue for details

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/592.html" title="Last updated on Feb 28, 2024, 10:32 PM UTC (6023741)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/592/0de084d...a-sully:6023741.html" title="Last updated on Feb 28, 2024, 10:32 PM UTC (6023741)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1086,7 +1086,7 @@ dictionary MLOperandDescriptor {
 
   // The dimensions field is empty for scalar operands,
   // and non-empty for tensor operands.
-  sequence<unsigned long> dimensions = [];
+  sequence<[EnforceRange] unsigned long> dimensions = [];
 };
 </script>
 
@@ -1531,7 +1531,7 @@ Return the index location of the minimum or maxmium values of all the input valu
 
 <script type=idl>
 dictionary MLArgMinMaxOptions {
-  sequence<unsigned long> axes;
+  sequence<[EnforceRange] unsigned long> axes;
   boolean keepDimensions = false;
   boolean selectLastIndex = false;
 };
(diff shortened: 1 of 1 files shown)


inexorabletash:
Spot check LGTM. 

* The one occurrence of `unsigned long` without `[EnforceRange]` is the return value of a method where it would be meaningless. 
* The auto-generated dictionary definitions **do** get `[EnforceRange]` added ([example](https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/592.html#dom-mlargminmaxoptions-axes)) but the manually written method argument definitions are **not** updated ([example](https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/592.html#api-mlgraphbuilder-concat)). I'm personally fine with that as IMHO the annotation is noise and I don't think of these definitions as normative. But editors should weigh in! (#574 is maybe relevant here)

fdwr (approved):
👍 Arigatou Austin.

huningxin (approved):
LGTM, thanks much @a-sully !

PR #593, opened 2024-03-04
Define task source, construct with realms, improve parallel usage

inexorabletash:
- Use a more common phrasing for "in parallel" steps.
- Make "in parallel" operations interact with script via queued tasks.
- Define an "ml task source" used for queued tasks.
- Ensure the correct Realm is used for asynch construction.

Relevant requirements/examples:
- https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel
- https://webidl.spec.whatwg.org/#js-promise-examples

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/593.html" title="Last updated on Mar 5, 2024, 8:24 PM UTC (8510ffc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/593/ca0b903...inexorabletash:8510ffc.html" title="Last updated on Mar 5, 2024, 8:24 PM UTC (8510ffc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1006,13 +1006,11 @@ In accordance with the [=ArrayBufferView/write|Web IDL warning=], to prevent the
         1. If [=validating graph resources=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=queue an ml task=] with |global| to [=reject=] |promise| with a {{TypeError}}.
         1. Let |transferredInputs| be the result of [=MLNamedArrayBufferViews/transfer|transferring=] {{MLNamedArrayBufferViews}} |inputs| with |realm|.
         1. Let |transferredOutputs| be the result of [=MLNamedArrayBufferViews/transfer|transferring=] {{MLNamedArrayBufferViews}} |outputs| with |realm|.
-        1. Invoke [=execute graph=] given |graph|, |transferredInputs| and |transferredOutputs|.
-        1. If that [=exception/throws=] an error, [=queue an ml task=] with |global| to [=reject=] |promise| with the error.
-        1. Otherwise, when [=execute graph=] has completed:
-            1. Let |result| be a new {{MLComputeResult}} with |realm|.
-            1. Set |result|.{{MLComputeResult/inputs}} to |transferredInputs|.
-            1. Set |result|.{{MLComputeResult/outputs}} to |transferredOutputs|.
-            1. [=Queue an ml task=] with |global| to [=resolve=] |promise| with |result|.
+        1. Invoke [=execute graph=] given |graph|, |transferredInputs| and |transferredOutputs|. If that [=exception/throws=] an error, [=queue an ml task=] with |global| to [=reject=] |promise| with the error and abort these steps.
+        1. Let |result| be a new {{MLComputeResult}} with |realm|.
+        1. Set |result|.{{MLComputeResult/inputs}} to |transferredInputs|.
+        1. Set |result|.{{MLComputeResult/outputs}} to |transferredOutputs|.
+        1. [=Queue an ml task=] with |global| to [=resolve=] |promise| with |result|.
     1. Return |promise|.
   </div>
 </details>
--- a/index.bs
+++ b/index.bs
@@ -996,16 +996,16 @@ In accordance with the [=ArrayBufferView/write|Web IDL warning=], to prevent the
     The <dfn method for=MLContext>compute(|graph|, |inputs|, |outputs|)</dfn> method steps are:
   </summary>
   <div class=algorithm-steps>
-    1. Let |promise| be [=a new promise=].
     1. Let |global| be [=this=]'s [=relevant global object=].
     1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. If |graph|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[contextType]]}} is not "[=context type/default=]", then return [=a new promise=] [=rejected=] with an "{{OperationError}}" {{DOMException}}.
+    1. If [=validating graph resources=] given |inputs| and |graph|.{{MLGraph/[[inputDescriptors]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If [=validating graph resources=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. Let |transferredInputs| be the result of [=MLNamedArrayBufferViews/transfer|transferring=] {{MLNamedArrayBufferViews}} |inputs| with |realm|.
+    1. Let |transferredOutputs| be the result of [=MLNamedArrayBufferViews/transfer|transferring=] {{MLNamedArrayBufferViews}} |outputs| with |realm|.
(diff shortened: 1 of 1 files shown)


inexorabletash:
@reillyeon @a-sully - can you take an initial look?

One thing I didn't touch that should probably be improved: "execute graph" still throws (DOMException, TypeError) which shouldn't happen in parallel, even though compute() takes that "throw" and turns it into a reject.

a-sully:
> One thing I didn't touch that should probably be improved: "execute graph" still throws (DOMException, TypeError) which shouldn't happen in parallel, even though compute() takes that "throw" and turns it into a reject.

FWIW the [File System spec](https://fs.spec.whatwg.org/) solves this by [making the respective algorithm optionally return a `DOMException` error name](https://fs.spec.whatwg.org/#file-system-access-result-error-name) and [expecting the caller to queue a task to reject, as appropriate](https://fs.spec.whatwg.org/#:~:text=Dependent%20specifications%20may,the%20empty%20string.)

index.bs
```diff
@@ -835,11 +846,14 @@ The <dfn>power preference</dfn> indicates preference as related to power consump
     The <dfn method for=ML>createContext(|options|)</dfn> steps are:
 </summary>
 <div class=algorithm-steps>
-    1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" {{DOMException}}.
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. If |global|'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, return [=a new promise=] [=rejected=] with a "{{SecurityError}}" {{DOMException}}.
+    1. Let |realm| be [=this=]'s [=relevant realm=].
     1. Let |promise| be [=a new promise=].
```
a-sully, line 852: optional cleanup here and below: declare `|promise|` before checking the webnn-feature, as to not repeat creating `[=a new promise=]` (could also defer this to a follow-up, if you'd like)

a-sully (approved):
LGTM! I'm happy with deferring the "execute graph" problem to a follow-up

reillyeon (approved):
I have only one question which is whether it should be "ml task source" or "ML task source".

index.bs
```diff
@@ -736,6 +738,15 @@ In a situation when a GPU context executes a graph with a constant or an input i
 
 When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's [=power preference=] and [=device type=] specified in the {{MLPowerPreference}} and {{MLDeviceType}} options.
 
+## Task Source ## {#programming-model-task-source}
+
+The <dfn>ml task source</dfn> is a [=task source=] to be used for all [=tasks=] related to asynchronous compilation and execution of {{MLGraph}}s.
```
huningxin, line 743: "ml task source" seems also to be used for context creation, should it also be mentioned here?

huningxin (approved):
LGTM with a question, thanks much!

inexorabletash:
> I have only one question which is whether it should be "ml task source" or "ML task source".

Capitalized "ml task" to "ML task" everywhere.

index.bs
inexorabletash: Added!

index.bs
inexorabletash: I'll defer, and FWIW I prefer the current structure where `|promise|` is only used for the "in parallel" steps, and the almost-synchronous failures are handled in a single line of the spec.

It also doesn't save text:

* If ... then return a new promise rejected with a ....
* If ... then reject promise with a ... and return promise.

a-sully: SGTM 👍

inexorabletash:
> > One thing I didn't touch that should probably be improved: "execute graph" still throws (DOMException, TypeError) which shouldn't happen in parallel, even though compute() takes that "throw" and turns it into a reject.
> 
> FWIW the [File System spec](https://fs.spec.whatwg.org/) solves this by [making the respective algorithm optionally return a `DOMException` error name](https://fs.spec.whatwg.org/#file-system-access-result-error-name) and [expecting the caller to queue a task to reject, as appropriate](https://fs.spec.whatwg.org/#:~:text=Dependent%20specifications%20may,the%20empty%20string.)

I decided to tackle this and made a slight variation - the steps now return an error, and if an error is returned the invoker rejects with an equivalent error in the right realm. I think that's probably precise enough without being too wordy. I didn't just use an error name because TypeError and "OperationError" DOMException are both used.

a-sully:
> > > One thing I didn't touch that should probably be improved: "execute graph" still throws (DOMException, TypeError) which shouldn't happen in parallel, even though compute() takes that "throw" and turns it into a reject.
> > 
> > 
> > FWIW the [File System spec](https://fs.spec.whatwg.org/) solves this by [making the respective algorithm optionally return a `DOMException` error name](https://fs.spec.whatwg.org/#file-system-access-result-error-name) and [expecting the caller to queue a task to reject, as appropriate](https://fs.spec.whatwg.org/#:~:text=Dependent%20specifications%20may,the%20empty%20string.)
> 
> I decided to tackle this and made a slight variation - the steps now return an error, and if an error is returned the invoker rejects with an equivalent error in the right realm. I think that's probably precise enough without being too wordy. I didn't just use an error name because TypeError and "OperationError" DOMException are both used.

Seems reasonable to me 👍

index.bs
```diff
@@ -980,20 +997,21 @@ In accordance with the [=ArrayBufferView/write|Web IDL warning=], to prevent the
...
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. Run the following steps [=in parallel=]:
+        1. If |graph|.{{MLGraph/[[context]]}} is not [=this=], then [=queue an ML task=] with |global| to [=reject=] |promise| with a {{TypeError}} and abort these steps.
+        1. If |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[contextType]]}} is not "[=context type/default=]", then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{OperationError}}" {{DOMException}} and abort these steps.
+        1. If [=validating graph resources=] given |inputs| and |graph|.{{MLGraph/[[inputDescriptors]]}} returns false, then [=queue an ML task=] with |global| to [=reject=] |promise| with a {{TypeError}} and abort these steps.
+        1. If [=validating graph resources=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=queue an ML task=] with |global| to [=reject=] |promise| with a {{TypeError}} and abort these steps.
+        1. Let |transferredInputs| be the result of [=MLNamedArrayBufferViews/transfer|transferring=] {{MLNamedArrayBufferViews}} |inputs| with |realm|.
```
reillyeon, line 1007: As a follow-up, I don't think you can transfer a buffer during "in parallel" because that would change the state of the buffer in parallel with script execution.

Most of these steps in fact should happen before going "in parallel".

reillyeon (approved):

index.bs
inexorabletash: Doing it here would simplify the steps, so... at the risk of feature creep...

inexorabletash: Updated to move the steps out of the "in parallel"

reillyeon (approved):

huningxin (approved):
LGTM!

fdwr (approved):

Issue #595, opened 2024-03-07
Update anchor link for compute method


ibelem:
The link for `compute() ` method is `https://www.w3.org/TR/webnn/#api-mlcontext-async-execution`.

**Expected Result:**
`https://www.w3.org/TR/webnn/#api-mlcontext-compute` , like other link `https://www.w3.org/TR/webnn/#api-mlgraphbuilder-build` for `build()` method.

Issue #531, opened 2024-01-26
Drop the support of synchronous execution


huningxin:
The current WebNN spec supports both [asynchronous](https://www.w3.org/TR/webnn/#api-mlcontext-async-execution) and [synchronous](https://www.w3.org/TR/webnn/#api-mlcontext-sync-execution) execution modes. In particular, the synchronous execution mode, including [`MLContext.computeSync()`](https://www.w3.org/TR/webnn/#dom-mlcontext-computesync), [`ML.createContextSync()`](https://www.w3.org/TR/webnn/#dom-ml-createcontextsync) and [`MLGraphBuilder.buildSync()`](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-buildsync) methods, were introduced (only available in dedicated worker) for easy integration with ML frameworks written in C++ and compiled to Wasm, for example ONNXRuntime WebNN EP (Execution Provider) used the sync execution before. 

Chromium [WebNN prototype](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/ml_context.idl) supports both execution modes for implementation feedback. Chrome team encouraged WG to check [whether sync APIs are really necessary before its launch](https://bugs.chromium.org/p/chromium/issues/detail?id=1488162).

Recently, ONNXRuntim WebNN EP experimented ([onnxruntime#19145](https://github.com/microsoft/onnxruntime/pull/19145)) the async execution mode and compared the performance with sync. For sync execution, ONNXRuntime runs WebNN EP in a dedicated worker and calls WebNN `computeSync()` method there, the JavaScript user code in main thread communicate (via `postMessage`) with WebNN EP in worker thread through ONNXRuntime [Wasm proxy](https://onnxruntime.ai/docs/api/js/interfaces/Env.WebAssemblyFlags.html#proxy). For async execution, ONNXRuntime runs WebNN EP in main thread and calls WebNN async `compute()` method through [asyncify](https://emscripten.org/docs/porting/asyncify.html#usage-with-embind). 

[...]

PR #596, opened 2024-03-07
Update anchor link for compute() after dropping synchronous execution support

ibelem:
Fix https://github.com/webmachinelearning/webnn/issues/595 .

After dropping the support of synchronous execution in https://github.com/webmachinelearning/webnn/issues/531 , we'd better to update the anchor link of `compute()` from `https://www.w3.org/TR/webnn/#api-mlcontext-async-execution` to `https://www.w3.org/TR/webnn/#api-mlcontext-compute` .

***
<a href="https://pr-preview.s3.amazonaws.com/ibelem/webnn/pull/596.html" title="Last updated on Mar 7, 2024, 4:07 AM UTC (c01e4c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/596/480b094...ibelem:c01e4c1.html" title="Last updated on Mar 7, 2024, 4:07 AM UTC (c01e4c1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -975,7 +975,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
   </div>
 </details>
 
-### {{MLContext/compute()}}  ### {#api-mlcontext-async-execution}
+### {{MLContext/compute()}}  ### {#api-mlcontext-compute}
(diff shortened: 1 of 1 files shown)


anssiko (approved):
@ibelem thanks!

We usually don't tweak anchor names without a good reason, because [cool URIs don't change](https://www.w3.org/Provider/Style/URI), but in this case this change is justified. 

This is a good example to highlight why selecting future-proof anchor names is important. Other spec might link here, resources such as MDN etc. So naming is not just bikeshedding here :-)

huningxin (approved):
LGTM, thanks for the fix, @ibelem !

PR #598, opened 2024-03-08
Content: Introduce emulation section, move reshape examples there

### LINKED ISSUE #551: Deprecated squeeze with emulation path is missing its own section
Since the [previous CR](https://www.w3.org/TR/2023/CR-webnn-20230330/) we removed `squeeze` (in PR #532) that can be generically emulated and expressed in terms of other low-level operations. Now the informative `squeeze` emulation note is plugged into the `lstm` section which is wrong. We should keep this note in its own `squeeze` section.

Considering deprecations may be expected in the future too, we should agree on how to best structure the spec for this:

- keep deprecated ops in the top-level API section: remove all the normative content and leave only the informative emulation note in place, maybe annotate the title with "deprecated"
- add a new top-level section for deprecated ops next to the API top-level

Thoughts? Is the first option better?

inexorabletash:
Per discussion in #551, move the squeeze, unsqueeze, and flatten examples from the reshape definition into a dedicated section.

Fixes #551

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5308,42 +5308,6 @@ partial interface MLGraphBuilder {
   </div>
 </details>
 
-<div class="note">
-  <details open>
-    <summary>
-    Many shape-related operations such as [squeeze](https://pytorch.org/docs/stable/generated/torch.squeeze.html), [unsqueeze](https://pytorch.org/docs/stable/generated/torch.unsqueeze.html), and [flatten](https://pytorch.org/docs/stable/generated/torch.flatten.html) can be generically implemented using the {{MLGraphBuilder/reshape()}} operation as follows:
-    </summary>
-    <pre highlight="js">
-    // Returns a tensor with all specified dimensions of input of size 1 removed.
-    function squeeze(input, axes) {
-      if (!axes) axes = [];
-      if (!axes.length)
-        input.shape().forEach((item, i) => { axes.push(i); });
-      shape = Array.from(input.shape());
-      for (let axis in axes.sort().reverse())
-        if (axis < shape.length && shape[axis] == 1)
-          shape.splice(axis, 1);
-      return builder.reshape(input, shape);
-    }
-
-    // Returns a new tensor with a dimension of size one inserted at the specified position.
-    function unsqueeze(input, axes) {
-      shape = Array.from(input.shape());
-      for(let axis in axes.sort())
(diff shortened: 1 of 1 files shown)


inexorabletash:
Notes: 

* The **Appendices - MLOperandDataType and ArrayBufferView compatibility** section is normative but now follows two non-normative sections. Can we rename and move it up?
* This PR only tackles part of the discussion in #551 namely moving squeeze (etc) out of reshape. It doesn't move any example decompositions out. If we want to do that, here's an idea
   * Give the new "**Operator Emulation**" section two subsections: "**Emulation**" (what's there now) and "**Decomposition**"
   * Move the 23 "generically emulated" blocks into "decomposition". Reword the preambles to be _"The behavior of the hardSwish operation..."_
   * Leave non-normative links in a standard format, e.g. "The behavior of this operation can be generically emulated." (which links to the emulation)
   * I am worried about bloating the Table of Contents with the decompositions, though, so I'd propose we leave those out. 

I don't feel strongly about any of this, it just seemed an easy issue to resolve...

zolkis:
Makes sense to me.

inexorabletash:
@huningxin - can you take a look? Maybe we should merge as-is or with minor tweaks, rather than trying to "boil the ocean" with other spec structure changes? I'm happy to incorporate any suggested changes, though.

huningxin (approved):
LGTM!

huningxin:
> @huningxin - can you take a look? Maybe we should merge as-is or with minor tweaks, rather than trying to "boil the ocean" with other spec structure changes?

+1 to merge as-is.

anssiko (approved):
Happy to see this editorial detail get addressed. Now seeing this in the spec preview helps a bunch and I agree this is a good compromise i.e. keep decompositions close to ops and for those ops that are not (anymore) part of the WebNN op set put them into this Operator Emulation section.

index.bs
```diff
@@ -6063,6 +6022,77 @@ Compile the graph up to the output operand.
...
 
+Operator Emulation {#emulation}
+===============================
+
+*This section is non-normative.*
+
+Operations present in other neural network inference APIs can often be emulated using operations present in WebNN.
+
+
+## squeeze ## {#emulation-squeeze}
```
fdwr, line 6033: I like the added clearer/more searchable headings.

fdwr (approved):
👍

PR #605, opened 2024-03-16
Synchronously validate input operands/validations

inexorabletash:
Previously the spec had a "validate `MLOperand`" helper that (1) ensured the operand was from the passed `MLGraphBuilder` and (2) that the operand was internally consistent, and this was called during (3) `build()` and (4) only `concat()` among the vending methods.

- (1) is needed but can be done when the `MLOperand` is created, giving better feedback, so (3) isn't needed.

- (2) is not needed - `MLOperands` are immutable so they can't be created in a bad state.

- (4) should be expanded to all `MLOperand` creations that take input `MLOperand`s.

This renames the helper, ensures it is called by every `MLOperand` vending method that takes `MLOperand` inputs, and drops it from `build()` (although that will probably collide with PR #603 which should land first). Similar validation is added for `MLActivation`s.

For #572

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1116,15 +1116,9 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
     1. Return |result|.
 </details>
 
-<details open algorithm>
-  <summary>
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM!

fdwr (approved):
👏

PR #591, opened 2024-02-28
Content: Define operand concept, simplify graph connection steps

inexorabletash:
As discussed in #572:

- Define an "operator" concept, with inputs, outputs, activations.
- Defer "platform operator" and "platform operand" to build.
- Standardize the graph connection steps across builder methods.
- Simplify activation, input and constant steps.

Not covered in this change:

- Erroring if input's [[builder]] doesn't match `this`
- Build algorithm - covered by #448 and #457
- gru() is missing steps to populate output. Added "Issue" note.
- Introducing "Validate arguments" section for each method.
- Introducing "Calculate output shape" section for each method.

For #549 and #572.

PR #606, opened 2024-03-18
Drop definitions for "platform operator" and "platform operand"

inexorabletash:
This fell out of making build() more specific - these terms are no longer needed, and the makeup of the MLGraph can be described more abstractly.

Discussed here: https://github.com/webmachinelearning/webnn/pull/603#pullrequestreview-1939733168

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/606.html" title="Last updated on Mar 18, 2024, 4:18 PM UTC (cfb8733)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/606/1062297...inexorabletash:cfb8733.html" title="Last updated on Mar 18, 2024, 4:18 PM UTC (cfb8733)">Diff</a>

PR #619, opened 2024-03-22
Bugfix: Drop "re-throw" in MLActivation creation steps

inexorabletash:
The "create an MLActivation" steps don't throw, so there's no need to re-throw.

Also, nothing specifies _init-steps_ so drop those, and simplify how the steps handle _options_, to align with how things are done for `MLOperands`. 

Discussed in https://github.com/webmachinelearning/webnn/pull/591#issuecomment-1970013201

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/619.html" title="Last updated on Mar 23, 2024, 6:37 PM UTC (c6ee63a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/619/91cb734...inexorabletash:c6ee63a.html" title="Last updated on Mar 23, 2024, 6:37 PM UTC (c6ee63a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1721,7 +1721,6 @@ partial interface MLGraphBuilder {
   </summary>
     1. If [=checking clamp options=] given |options| returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |op| be the result of [=creating an MLActivation=] given [=this=], "clamp" and |options|.
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
     1. Return |op|.
 </details>
 
@@ -3373,7 +3372,6 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>hardSigmoid(|options|)</dfn> method steps are:
   </summary>
     1. Let |op| be the result of [=creating an MLActivation=] given [=this=], "hardSigmoid" and |options|.
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
     1. Return |op|.
 </details>
 
@@ -3445,7 +3443,6 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder id=hardswish-noargs>hardSwish()</dfn> method steps are:
   </summary>
     1. Let |op| be the result of [=creating an MLActivation=] given [=this=] and "hardSwish".
-        1. If that [=exception/throws=] an error, re-[=exception/throw=] the error.
     1. Return |op|.
 </details>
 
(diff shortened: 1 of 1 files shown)


fdwr (approved):

huningxin (approved):
LGTM with two questions:

https://www.w3.org/TR/webnn/#create-an-mlactivation has a step to run init-steps if it presents

> If init-steps are given, run init-steps with options.

I am not sure whether it would throw. However, it looks like there are no any call sites that supply the init-steps, should we remove this step accordingly?

The following step:

> Otherwise, initialize activation.[[[operator]]](https://www.w3.org/TR/webnn/#dom-mlactivation-operator-slot) given options in an [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined) way for the given name operation.

I am not sure whether this step would fail. However, as https://github.com/webmachinelearning/webnn/pull/606 drops the platform operand and operator, should we drop this step as well?

inexorabletash:
I don't know the context around init-steps. Are there plans to add them? Or was it an idea that turned out not to be needed?

Either way - can either merge this and follow up, or I can revise - either way, this is not blocking anything else.

huningxin:
@inexorabletash 

> Are there plans to add them? Or was it an idea that turned out not to be needed?

I suppose the latter. I dig the commits, and found it was introduced by @zolkis in https://github.com/webmachinelearning/webnn/commit/30ba12ece0937263df81c1dd90f06214bdfbf802. I try to recall and it seems to be intended to run the activation initialization steps supplied by different activation creation caller. But it turns out there are no callers using it. @zolkis , please correct me if I am wrong.

> Either way - can either merge this and follow up, or I can revise - either way, this is not blocking anything else.

This PR proposes to remove the "re-throw" because "create an MLActivation" steps don't throw. I think we may need to ensure all "create an MLActivation" steps won't throw. So, if "init-steps" is useful and may throw, the "re-throw" may need tol be kept. Otherwise, the "init-steps" should be removed and that makes "create an MLActivation" steps don't throw at all, then the "re-throw" could be removed. WDYT?

inexorabletash:
> This PR proposes to remove the "re-throw" because "create an MLActivation" steps don't throw. I think we may need to ensure all "create an MLActivation" steps won't throw. So, if "init-steps" is useful and may throw, the "re-throw" may need tol be kept. Otherwise, the "init-steps" should be removed and that makes "create an MLActivation" steps don't throw at all, then the "re-throw" could be removed. WDYT?

Makes sense. I removed the steps in c6ee63af45c248942129279ed97de1d0a4878f6d, but if we decide to keep them or re-add them then we can abandon this PR and do one instead that ensures all invocations re-throw, and make it explicit that init-steps themselves may throw and the MLActivation creation steps should re-throw.

zolkis:
Back then we wanted to keep optional init steps around until we figured how we end up specifying other builder algorithms.
Since now the usage has became more clear, it's indeed time to simplify and LGTM.

huningxin (approved):
LGTM!

Issue #396, opened 2023-06-09
Clarify the restriction for `minValue` and `maxValue` of `MLClampOptions`


huningxin:
WebNN [`clamp`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-clamp) limits the input tensor element-wise within a range specified by the minimum and maximum values. The minimum and maximum values could be specified by `MLClampOptions`.

This open item was raised by @quidity (Thanks Alex!) in Chromium [CL review](https://chromium-review.googlesource.com/c/chromium/src/+/4597587/comment/d1368beb_9079fbb8/) which is

> can min & max be == floats and is that ok or not?

The current [Chromium implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/webnn/ml_graph_builder.cc;l=70)  restricts the min value should be less than or equal to the max value. TensorFlow.js also implements [this restriction](https://github.com/tensorflow/tfjs/blob/tfjs-v4.7.0/tfjs-core/src/ops/clip_by_value.ts#L47). However, some backends implement more strict restriction that the min value should be less than max value, like [XNNPACK](https://github.com/google/XNNPACK/blob/master/src/operators/unary-elementwise-nc.c#L328).

WebNN spec should clarify the restriction.

/cc @fdwr, for inputs on DirectML .

fdwr:
> WebNN spec should clarify the restriction.

We have a few options, but I think we should:

- a. Not exclude infinities. Positive infinity should be a perfectly legal `maximum`, and negative infinity a legal `minimum`.
- b. Consider rather than rejecting inverted ranges at the WebNN API layer, set min/max explicitly to whatever behavior we want before passing the values to the lower layer backends. Though, if we're not able to decide, then rejecting at the WebNN API layer and forcing the caller to massage min/max values before calling WebNN is reasonable. Since `clamp` only accepts scalar min/max values, it's a much easier situation to handle than if clip accepted entire *tensors* for min/max.

# Framework comparisons

Given `[0,5,10,15]` and the inverted range `min=10` and `max=5`:

- TensorFlow: 10, 10, 10, 10
- PyTorch: 5, 5, 5, 5
- C++ `std::clamp`: 10, 10, 5, 5
- DirectML: 10, 10, 10, 10

As you can see, there's far from universal consensus on how to treat inverted ranges 😅🙃.

# TensorFlow

https://www.tensorflow.org/api_docs/python/tf/clip_by_value

```
[...]

Issue #566, opened 2024-02-15
Transferring constant ArrayBuffers


reillyeon:
Right now the input and output `ArrayBuffer`s  passed to `MLContext`'s `compute()` method are transferred so that their contents cannot be modified by script during the computation. It seems appropriate that an `ArrayBuffer` passed to `MLGraphBuilder`'s `constant()` method should be treated the same way, though it would be better if the transfer were delayed until `build()` is called so that multiple `ArrayBufferView`s using the same backing `ArrayBuffer` can be passed as constants. `build()` would construct a list of unique backing `ArrayBuffer`s to transfer rather than the naive approach of simply iterating over the `ArrayBufferView`s themselves.

I'll note that transferring `ArrayBuffer` parameters makes an API harder to call from WebAssembly, but I think that any argument for why compute inputs and outputs are transferred should also apply to constants.

huningxin:
@reillyeon 

> but I think that any argument for why compute inputs and outputs are transferred should also apply to constants.

Agreed.

> I'll note that transferring ArrayBuffer parameters makes an API harder to call from WebAssembly

This is true. 

Should we consider copying the content of input (and constant) `ArrayBuffer` instead? Notice that the Chromium implementation actually copies the input and constant `ArrayBuffer`s for [XNNPACK backend](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/webnn/ml_graph_xnnpack.cc;l=534?q=ml_graph_xnnpack&ss=chromium%2Fchromium%2Fsrc) and [DirectML (mojo) backend](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/webnn/ml_graph_mojo.cc;l=192). It means for each input ArrayBuffer, WebAssembly path may need two copies, one is done by JS that copies from Wasm heap to an ArrayBuffer that can be transferred, another one is done by Chromium implementation that copies from ArrayBuffer to internal storage.

reillyeon:
Given the constraints on implementations based on the underlying platform APIs I think it's reasonable to define constant and input handling such that `MLGraphBuilder` and `compute()` copy any provided `ArrayBuffer`s. If developers want to avoid a copy they can use `MLBuffer`s instead.

In theory Chromium's XNNPACK backend could avoid copies but I'm also planning to delete it in favor of a more flexible TFLite backend. The zero-copy option only works for CPU backends anyways (without `MLBuffer`).

[...]

PR #621, opened 2024-03-26
Conventions: Ensure all dict members have definitions

### LINKED ISSUE #483: Consistency for dictionary member definitions
Most dictionary members are given explicit definitions, e.g. for
```webidl
dictionary MLBufferResourceView {
  required GPUBuffer resource;
  unsigned long long offset = 0;
  unsigned long long size;
};
```
... there's a normative prose block:

> MLBufferResourceView has the following members:
> **resource**, of type GPUBuffer
> A GPUBuffer object. Specifies the GPU buffer source.
> **offset**, of type unsigned long long, defaulting to 0
> Specifies an unsigned long long offset in the buffer source.
> **size**, of type unsigned long long
> Specifies the unsigned long long size of the buffer view.

This is not necessary and IMHO is basically a redundant definition that will make spec maintenance more difficult and risks introducing inconsistencies. The types are captured in the dictionary definition, and the behavior needs to be captured in the algorithms where the members are used. But... since these definitions are provided mostly consistently in the spec, it's worth ensuring _all_ such dictionaries get definitions.

This can be audited in the developer console with:
```js
[...document.querySelectorAll('.idl dfn[data-dfn-type=dict-member]')].map(d => d.innerText).sort()
```
...which in the current ED lists 10 examples, but after PR #478 (which should merge soon) lists only 8. That shorter list is:

- "alpha"
- "dataType"
- "deviceType"
- "dimensions"
- "inputs"
- "maxValue"
- "minValue"
- "outputs"
- "powerPreference"

inexorabletash:
- Documents the convention.
- Dedupes `MLEluOptions`'s `alpha` definitions
- Dedupes `MLClampOptions`'s `minValue` and `maxValue`, and references #396
- Moves `MLOperandDescriptor` member documentation out of IDL comments
- Introduces simple definitions for `MLComputeResult`'s `inputs` and `outputs`
- Converts "device type" and "power preference" into definitions for `MLContextOptions`'s `deviceType` and `powerPreference`.

Also includes these adjacent changes to improve the document flow:

- Moves the "context type" definition into the `MLContext` section.
- Moves the Permission Policy Integration section from API into Programming Model which seems like a slightly better home for it.

Fixes #483

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/621.html" title="Last updated on Mar 27, 2024, 6:42 PM UTC (1f1b3c1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/621/edc9e52...inexorabletash:1f1b3c1.html" title="Last updated on Mar 27, 2024, 6:42 PM UTC (1f1b3c1)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -616,9 +616,9 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring; 
 
 The WebGPU API identifies <a href="https://gpuweb.github.io/gpuweb/#privacy-machine-artifacts">machine-specific artifacts</a> as a privacy consideration. Similarly, the WebNN API's compute unit scheduling may under certain circumstances introduce a fingerprint. However, similarly to WebGPU, such fingerprints are identical across most or all of the devices of each vendor, mitigating the concern. Furthermore, software implementations can be used to further eliminate such artifacts.
 
-The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Device type=] normatively indicates the kind of device and is either {{MLDeviceType/"cpu"}} or {{MLDeviceType/"gpu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
+The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. An {{MLDeviceType}} normatively indicates the kind of device and is either {{MLDeviceType/"cpu"}} or {{MLDeviceType/"gpu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. An {{MLPowerPreference}} indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
 
-If a future version of this specification introduces support for new a  [=device type=] that can only support a subset of {{MLOperandDataType}}s, that may introduce a new fingerprint.
+If a future version of this specification introduces support for a new {{MLDeviceType}} that can only support a subset of {{MLOperandDataType}}s, that may introduce a new fingerprint.
 
 In general, implementers of this API are expected to apply <a href="https://gpuweb.github.io/gpuweb/#privacy-considerations">WebGPU Privacy Considerations</a> to their implementations where applicable.
 
@@ -668,7 +668,7 @@ An {{MLContext}} interface represents a global state of neural network execution
 
 In a situation when a GPU context executes a graph with a constant or an input in the system memory as an {{ArrayBufferView}}, the input content is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of the graph execution. This data upload and download cycles will only occur whenever the execution device requires the data to be copied out of and back into the system memory, such as in the case of the GPU. It doesn't occur when the device is a CPU device. Additionally, the result of the graph execution is in a known layout format. While the execution may be optimized for a native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert the content back to a known layout format at the end of the graph in order to maintain the expected behavior from the caller's perspective.
 
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's [=power preference=] and [=device type=] specified in the {{MLPowerPreference}} and {{MLDeviceType}} options.
+When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's {{MLPowerPreference}} and {{MLDeviceType}} options.
 
 ## Task Source ## {#programming-model-task-source}
 
@@ -677,6 +677,12 @@ The <dfn>ML task source</dfn> is a [=task source=] to be used for all [=tasks=] 
 <div algorithm>
 <p>To <dfn>queue an ML task</dfn> given a [=global object=] |global| and a series of steps |steps|, [=queue a global task=] on the [=ML task source=] with |global| and |steps|.
 </div>
+
+## Permissions Policy Integration ## {#permissions-policy-integration}
+
+This specification defines a [=policy-controlled feature=] identified by the
+string "<code><dfn data-lt="webnn-feature">webnn</dfn></code>".
+Its [=policy-controlled feature/default allowlist=] is <code>'self'</code>.
 
 
 API {#api}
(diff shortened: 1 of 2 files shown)


inexorabletash:
@zolkis can you take an initial look at this?

index.bs
```diff
@@ -720,31 +726,15 @@ interface ML {
...
+The <dfn dfn-for=MLContextOptions dfn-type=dict-member>deviceType</dfn> option is an <dfn dfn-type=enum>MLDeviceType</dfn> and indicates the application's preference for the kind of device used for the context. It is one of the following:
 <dl dfn-for="MLDeviceType">
 <dt>"<dfn enum-value>cpu</dfn>"</dt>
 <dd>Provides the broadest compatibility and usability across all client devices with varying degrees of performance.</dd>
 <dt>"<dfn enum-value>gpu</dfn>"</dt>
 <dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations.</dd>
 </dl>
 
-The <dfn>power preference</dfn> indicates preference as related to power consumption. It is one of the following:
+The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> option is an <dfn dfn-type=enum>MLPowerPreference</dfn> and indicates the application's preference as related to power consumption. It is one of the following:
```
zolkis, line 737: I checked this section with the preview, and I wonder if would it make sense to group these definitions under an `MLContextOptions` sub-section, which is either under 7.2 The ML interface (here), or eventually/preferably under `createContext()` above the algorithm?

index.bs
```diff
@@ -754,6 +744,8 @@ The <dfn>power preference</dfn> indicates preference as related to power consump
 <dd>Prioritizes power consumption over other considerations such as execution speed.</dd>
 </dl>
 
+### {{ML/createContext()}} ### {#api-ml-createcontext}
+
```
zolkis, line 748: (here could come an `MLContextOptions` sub-section?) -- see the comment above

zolkis (approved):
This is nice, thanks!

I left a suggestion, but it's up to your judgement.

index.bs
inexorabletash: I added a subsection in ed0efec0edb614592712df79a70faed07f786d76

I initially tried putting it under createContext() and above the algorithm, but as a subsection it was weird because then the document structure is:

7.2 ML interface
(the IDL)
7.2.1 createContext()
7.2.1.1 MLContextOptions &larr; **new!**
(the dictionary/enum descriptions)
(the "create a context" algorithm)
(the "createContext()" algorithms)

...which makes the new subsection seem to contain the algorithms. That implies something like:

7.2 ML interface
(the IDL)
7.2.1 createContext()
7.2.1.1 MLContextOptions &larr; **new!**
(the dictionary/enum descriptions)
7.2.1.2 Creating a context &larr; **new!**
(the "create a context" algorithm)
(the "createContext()" algorithms)

... which is confusing so it should really be:

7.2 ML interface
(the IDL)
7.2.1 Creating a context &larr; **new!**
7.2.1.1 MLContextOptions &larr; **new!**
(the dictionary/enum descriptions)
7.2.1.2 createContext() &larr; **moved!**
(the "create a context" algorithm)
(the "createContext()" algorithms)

... which seems like overkill. In the end, this seemed the least bad:

7.2 ML interface
(the IDL)
7.2.1 MLContextOptions &larr; **new!**
(the dictionary/enum descriptions)
7.2.2 createContext()
(the "create a context" algorithm)
(the "createContext()" algorithms)

But maybe I'm misunderstanding your suggestion?

index.bs
```diff
@@ -1015,15 +1025,19 @@ enum MLOperandDataType {
...
-
-  // The dimensions field is empty for scalar operands,
-  // and non-empty for tensor operands.
   sequence<[EnforceRange] unsigned long> dimensions = [];
 };
 </script>
 
+<dl dfn-type=dict-member dfn-for=MLOperandDescriptor>
+    : <dfn>dataType</dfn>
+    :: The operand type.
```
huningxin, line 1035: Nit:
```suggestion
    :: The operand data type.
```

index.bs
```diff
@@ -820,22 +814,38 @@ interface MLContext {
...
+<dd>Context created from WebGPU device.</dd>
+</dl>
+
 <div class="note">
 When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with the {{MLContextOptions}}.{{MLContextOptions/deviceType}} set to {{MLDeviceType/"gpu"}}, the user agent is responsible for creating an internal GPU device that operates within the context and is capable of ML workload submission on behalf of the calling application. In this setting however, only {{ArrayBufferView}} inputs and outputs are allowed in and out of the graph execution since the application has no way to know what type of internal GPU device is being created on their behalf. In this case, the user agent is responsible for automatic uploads and downloads of the inputs and outputs to and from the GPU memory using this said internal device.
 </div>
 
+<dl dfn-type=dict-member dfn-for=MLComputeResult>
+    : <dfn>inputs</dfn>
+    :: An object where the keys are the graph input names, and the values are the bound {{ArrayBufferView}}s for the supplied input tensor values.
```
huningxin, line 843: Should we say "transferred {{ArrayBufferView}}s" instead? The `ArrayBufferView`s are the result of ["transfer an MLNamedArrayBufferViews" step](https://www.w3.org/TR/webnn/#mlnamedarraybufferviews-transfer) rather than the original user-bound ones.

index.bs
```diff
@@ -820,22 +814,38 @@ interface MLContext {
...
+
+    : <dfn>outputs</dfn>
+    :: An object where the keys are the graph output names, and the values are the bound {{ArrayBufferView}}s for the computed output tensor values.
```
huningxin, line 846: ditto.

huningxin (approved):
LGTM with nits, thanks!

index.bs
zolkis: Right, for now let's leave it like after  ed0efec.

index.bs
inexorabletash: Sure, although #566 might change things.

index.bs
inexorabletash: Done in 1f1b3c13762ec4366ebb9e233fec09f6ae432ddf

index.bs
inexorabletash: Done in 1f1b3c13762ec4366ebb9e233fec09f6ae432ddf

index.bs
inexorabletash: Done in 1f1b3c13762ec4366ebb9e233fec09f6ae432ddf

inexorabletash:
@fdwr - final review and merge if it LG2U? Thanks!

huningxin (approved):
LGTM!

index.bs
```diff
@@ -1625,6 +1639,14 @@ partial interface MLGraphBuilder {
 };
 </script>
 
+<dl dfn-type=dict-member dfn-for=MLClampOptions>
+    : <dfn>minValue</dfn>
+    :: The minimum value of the range. When it is not specified, the clamping is not performed on the lower limit of the range.
+
+    : <dfn>maxValue</dfn>
+    :: The maximum value of the range. When it is not specified, the clamping is not performed on the upper limit of the range.
```
fdwr, line 1647: I had to stop and think if "the clamping is not performed" was really equivalent to what happens in actual practice that "the unspecified limit is treated as infinity for the data type" (or rather negative infinity for the minimum), and I realize it is, because in both cases if the input is less than infinity, then it's equivalent, but also if NaN, then the output will be NaN, and if the input is infinity, the output will be infinity. So, it's equivalent.

index.bs
```diff
@@ -1658,6 +1680,9 @@ partial interface MLGraphBuilder {
     To <dfn>check clamp options</dfn> given {{MLClampOptions}} |options|, run the following steps:
   </summary>
     1. If |options|.{{MLClampOptions/minValue}} is greater than |options|.{{MLClampOptions/maxValue}}, then return false.
+
+        Issue(396): Not all implementations support {{MLClampOptions/minValue}} equal to {{MLClampOptions/maxValue}}.
```
fdwr, line 1684: Hmm, is TF.js the only remaining outlier now? Every other ML library I know (include TF-main and XNNPack) support `min` and `max` fine. Well, there's always the decomposition `clamp(x) = min(max(x, maxValue), minValue)` too, but I'll open an issue at TF.js for better consistency across libraries (https://github.com/tensorflow/tfjs/blob/tfjs-v4.7.0/tfjs-core/src/ops/clip_by_value.ts#L47).

fdwr (approved):
👍

PR #548, opened 2024-02-01
Drop the support of synchronous execution

huningxin:
Remove the definition and algorithm steps for
- ML.createContextSync()
- MLGraphBuilder.buildSync()
- MLContext.computeSync()

This PR also removes the synchronous execution description from Programming Model section.

The previous algorithm steps of `MLGraphBuilder.build()` depends on `MLGraphBuilder.buildSync()`. The algorithm steps of `MLGraphBuilder.buildSync()` have issue tracked by #457. That is left to a separate PR. This PR just merges the algorithm steps of `MLGraphBuilder.buildSync()` into `MLGraphBuilder.build()` with some modifications:
- Remove "The permissions and context validity have been checked..." note regarding to #501.
- Change throwing exception to rejecting the promise with exception.

Fix #531

@wchao1115 @anssiko @inexorabletash , PTAL. Thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/548.html" title="Last updated on Feb 15, 2024, 8:52 AM UTC (7fb3187)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/548/eb06ccf...huningxin:7fb3187.html" title="Last updated on Feb 15, 2024, 8:52 AM UTC (7fb3187)">Diff</a>

PR #622, opened 2024-03-26
Revise graph resource validation

### LINKED ISSUE #602: Is "validate graph resources" backwards?
Currently, [compute()](https://webmachinelearning.github.io/webnn/#dom-mlcontext-compute) calls [validate graph resources](https://webmachinelearning.github.io/webnn/#validate-graph-resources) with the passed input (or output) maps and the graph's input (or output) descriptors.

I believe the intent is to ensure that all of a graph's input (or output) descriptors are matched to a (name, buffer) pair in the passed input (or output) resources.

The steps are:

> To validate graph resources, given MLNamedArrayBufferViews resources and ordered map descriptors, run the following steps:
> 1. For each name → resource of resources:
>     1. If descriptors[name] does not exist, return false.
>     1. If validating buffer with descriptor given resource and descriptors[name] returns false, then return false.
> 1. Return true.

This instead seems to be ensuring that each (name, buffer) pair in the passed resources has a corresponding input (or output) descriptor - which is similar but backwards.

i.e.

```js
const builder = new MLGraphBuilder(context);
const input1 = builder.input('input1', desc1);
const input2 = builder.input('input2', desc2);
const output1 = builder.mul(input1, input2);
const graph = await builder.build({'output1': output1});
// graph has two inputs (input1, input2) and one output (output1)

// Expected usage
await context.compute(graph, {input1: inbuf1, input2: inbuf2}, {output1: outbuf1});

// Should fail, because input2 is missing - but it *doesn't*.
await context.compute(graph, {input1: inbuf1}, {output1: outbuf1});

// Maybe fail, because output2 is passed but unused. Currently it *does* fail. 
await context.compute(graph, {input1: inbuf1, input2: inbuf2}, {output1: outbuf1, output2: outbuf2});
```

It seems like we've got two options:
* Invert the test - iterate over the descriptors, and verify that there are matching resources
* Make the test reflexive - the set of expected and passed resources must exactly match

inexorabletash:
`compute()` calls validate graph resources with the passed input/output maps and the graph's input/output descriptors. The intent is to ensure that all of a graph's descriptors are matched to a passed resource. However, the steps were only validating that a passed resource had a corresponding descriptor - it wouldn't flag if a resource was missing!

The Chromium prototype implementation made the test reflexive - all resources much have matching descriptors and vice versa. After deliberation we settled on a more helpful behavior: all inputs must be provided, but extra inputs are ignored. A requested output must be present, but dropping outputs is allowed. 

Fixes #602

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/622.html" title="Last updated on Apr 16, 2024, 7:52 AM UTC (2384d12)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/622/954ed19...inexorabletash:2384d12.html" title="Last updated on Apr 16, 2024, 7:52 AM UTC (2384d12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -848,17 +848,6 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 <details open algorithm>
   <summary>
-    To <dfn>validate graph resources</dfn>, given {{MLNamedArrayBufferViews}} |resources| and [=ordered map=] |descriptors|, run the following steps:
-  </summary>
-    1. If |resources|'s [=map/size=] does not equal |descriptors|'s [=map/size=], then return false.
-    1. [=map/For each=] |name| → |resource| of |resources|:
-        1. If |descriptors|[|name|] does not [=map/exist=], return false.
-        1. If [=validating buffer with descriptor=] given |resource| and |descriptors|[|name|] returns false, then return false.
-    1. Return true.
-</details>
-
-<details open algorithm>
-  <summary>
     To <dfn>validate buffer with descriptor</dfn> given {{ArrayBufferView}} |bufferView| and {{MLOperandDescriptor}} |descriptor|, run the following steps:
   </summary>
     1. If |bufferView|'s [=element type=] does not match to |descriptor|.{{MLOperandDescriptor/dataType}}  according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility), return false.
@@ -920,6 +909,8 @@ In accordance with the [=ArrayBufferView/write|Web IDL warning=], to prevent the
     **Returns:** Promise<{{MLComputeResult}}>.
 </div>
 
+Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext/compute(graph, inputs, outputs)/graph}}'s inputs are not provided as {{MLContext/compute(graph, inputs, outputs)/inputs}}, or if any requested {{MLContext/compute(graph, inputs, outputs)/outputs}} do not match the {{MLContext/compute(graph, inputs, outputs)/graph}}'s outputs.
+
 <details open algorithm>
   <summary>
     The <dfn method for=MLContext>compute(|graph|, |inputs|, |outputs|)</dfn> method steps are:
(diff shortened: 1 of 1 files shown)


inexorabletash:
@huningxin and @fdwr - can you please review? Thanks!

index.bs
```diff
@@ -840,6 +840,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
   <summary>
     To <dfn>validate graph resources</dfn>, given {{MLNamedArrayBufferViews}} |resources| and [=ordered map=] |descriptors|, run the following steps:
   </summary>
+    1. If |resources|'s [=map/size=] does not equal |descriptors|'s [=map/size=], then return false.
```
zolkis, line 843: Isn't this implied by the loop below? (but a quick test is fine)

inexorabletash: If _resources_'s keys are [ "a", "b" ] but _descriptors_'s keys are [ "a", "b", "c" ] the loop below won't fail:

> 1. For each name → resource of resources:

So this would loop over [ "a", "b" ]

>> 1. If descriptors[name] does not exist, return false.

_descriptors_["a"] and _descriptors_["b"] exist, so this does not fail

>> 2. If validating buffer with descriptor given resource and descriptors[name] returns false, then return false.

Presumably this also doesn't fail

> 2. Return true.

So we return true, even though "c" is not provided.

fdwr:
Before reviewing, I'm just thinking aloud the various cases, given a graph that expects 2 inputs and generates 2 outputs...

- `await context.compute(graph, {input1: inbuf1, input2: inbuf2}, {output1: outbuf1, output2: outbuf2});`
  ✅ Exactly matching. All required inputs satisfied. All generated outputs bound.
- `await context.compute(graph, {input1: inbuf1}, {output1: outbuf1, output2: outbuf2});`
  ❌ Missing an input.
- `await context.compute(graph, {input1: inbuf1, input2: inbuf2, input3: inbuf3}, {output1: outbuf1, output2: outbuf2});`
  🤷‍♂️✔ Extra input specified. I foresee useful cases to pass potentially ignorable input so an app can more generically provide inputs regardless of the model flavor. Say one diffusion model doesn't use a mask input, whereas another inpainting model does use the mask, and just having the non-inpainting ignore that parameter is cleaner than branching between the cases, especially when you have many different permutations of potential inputs.
- `await context.compute(graph, {input1: inbuf1, input2: inbuf2}, {output1: outbuf1});`
  🤷‍♂️✔ An output was generated by the graph was ignored and discarded. This seems fine to me too. The caller doesn't always care about all outputs, and needing to create a dummy variable just to discard that dummy seems pointless.
- `await context.compute(graph, {input1: inbuf1, input2: inbuf2}, {output1: outbuf1, output2: outbuf2, output3: outbuf3});`
  🤷‍♂️❌ If you require an output, and the graph doesn't provide it, that's an error. Though, by my same logic above for optionally unused *inputs*, there can be a set of closely related model flavors where some provide an output and others do not, and the output in that case would be missing from the return `results.outputs`. So maybe this is 🤷‍♂️✔ too, where the caller has to check the dictionary. I suppose there isn't an IDL dictionary syntax to express {requiredOutput1: outbuf1, *optionalOutput2*: outbuf2}. 🤔...

Anyway, we definitely both agree (per 602) on the case where it "Should fail, because input2 is missing - but it *doesn't*".

huningxin:
> * `await context.compute(graph, {input1: inbuf1, input2: inbuf2}, {output1: outbuf1});`
> 🤷‍♂️✔ An output was generated by the graph was ignored and discarded. 

Computing a subset of outputs is supported. There was an example using sync compute API https://www.w3.org/TR/2024/CRD-webnn-20240214/#example-4a21156c. Unfortunately, my PR https://github.com/webmachinelearning/webnn/pull/548 that removes the sync APIs deleted this example together. We may want to bring it back by adapting to async compute API.

inexorabletash:
> there isn't an IDL dictionary syntax to express {requiredOutput1: outbuf1, _optionalOutput2_: outbuf2}. 

Correct.

Based on above ✔🤷‍♂️❌ it sounds like we might want:

* Inputs: 
    * every input descriptor must have matching input resource: _can't run inference with missing inputs!_
    * an input resource without a matching input descriptor is okay: _maybe an some graphs don't use it_
* Outputs:
    * every output resource must have a matching output descriptor: _asking for an output that's not provided by the inference is a bug!_
    * an output descriptor without a matching output resource is okay: _if it's not needed, avoid extra work_

Given this we'd want either one algorithm that takes a flag (is input output?) or two algorithms. The latter is probably easier. And since they're only used in one place this could be in-line in "execute graph", something like this:

> 5. For each name → descriptor of graph.[[inputDescriptors]]:
>> 1. If inputs[name] does not exist, then return a new promise rejected with a TypeError.
>> 2. If validating buffer with descriptor given inputs[name] and descriptor returns false, then return a new promise rejected with a TypeError.
> 6. For each name → resource of outputs:
>> 1. If graph.[[outputDescriptors]][name] does not exist, then return a new promise rejected with a TypeError.
>> 2. If validating buffer with descriptor given resource and graph.[[outputDescriptors]][name] returns false, then return a new promise rejected with a TypeError.

... and then a non-normative note saying something like:

> NOTE: Invocations of compute() will fail if any of the graph's inputs are not provided as input resources, or if any requested output resources do not match the graph's outputs.

huningxin:
> The latter is probably easier. And since they're only used in one place this could be in-line in "execute graph", something like this:

SGTM!

inexorabletash:
1e7e2b988cc8141c347510012d97ccd7b7c9139b captures the algorithm discussed above. We should probably wait for the WG telecon to decide if this is what we want.

huningxin (approved):
LGTM!

inexorabletash:
I updated this PR's name/description to match the changes. I think we're probably good to merge now?

huningxin:

fdwr:
Minor merge conflict from the transferred inputs/outputs CR. So just resolved.

PR #632, opened 2024-04-02
add a note for empty input

### LINKED ISSUE #615: Graph with no input
Hi, it looks like currently we don't prohibit graph without any input.  I was working on Chromium CoreML backend and realized that CoreML doesn't support model without inputs. 

It doesn't seem to be useful to have a model that takes no input, as that will just generate the same result every time. So I'd propose to let `builder.build` error out for no-input graph.  What do you think?

philloooo:
per discussion in  #615 , describe that WebNN allows graph without inputs, and suggest implementations how to support that.

fix #615

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/webnn/pull/632.html" title="Last updated on Apr 2, 2024, 7:34 PM UTC (3630cae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/632/60b8e0a...philloooo:3630cae.html" title="Last updated on Apr 2, 2024, 7:34 PM UTC (3630cae)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1284,6 +1284,10 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
     1. Return |operand|.
 </details>
 
+<div class="note">
+  The {{MLGraphBuilder}} API allows creating an {{MLGraph}} without input operands. If the underlying platform doesn't support that, implementations can work around this limitation by adding a stub input or passing constants as inputs to the graph.
+</div>
+
 ### constant operands ### {#api-mlgraphbuilder-constant}
 Create a constant {{MLOperand}} that can be used in {{MLGraphBuilder}} methods.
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1284,6 +1284,10 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
     1. Return |operand|.
 </details>
 
+<div class="note">
+  The {{MLGraphBuilder}} API allows creating an {{MLGraph}} without input operands. If the underlying platform doesn't support that, implementations can work around this limitation by adding a stub input or passing constants as inputs to the graph.
```
zolkis, line 1288: or, in a more spec'ese formulation, 

> implementations may add a stub input or pass constants as inputs to the graph

huningxin (approved):
LGTM, thanks!

inexorabletash:
@fdwr - can you review and merge if it looks good? Thanks!

fdwr (approved):
Thanks - yes, 0 inputs is a valid case imo for generators like random/fillSequence... (now, 0 *outputs* would be more questionable 😉).

PR #638, opened 2024-04-05
Editorial: Make "generically emulated" text a macro, update wording

### LINKED ISSUE #187: BatchNormalization should be an optional operation
BatchNormalization usually are optimized away for inference (fused into conv2d op filter and bias), for example TFLite does not have this op, for compatibility purpose we shall move this op to optional section of the spec.

inexorabletash:
Many complex operations in the spec are given decompositions, showing how the operation could be replaced through the use of more primitive operations. This text is repeated in many places. Improve this in two ways:

* Use Bikeshed's Text Macro[1] to reduce the repetition.

* Streamline the text, and make it explicit that if the underlying platform doesn't support an operation, WebNN API implementations can use the decomposition as a guide to emulate it.

The macro [EMULATED] is used at the end of an intro sentence since there are variations - some of the decompositions are grouped, and some make assumptions (e.g. activation functions, layouts, etc).

If we assume that implementations must implement all operations in the spec, either via the underlying platform or emulation, this fixes #187

[1] https://speced.github.io/bikeshed/#metadata-text-macro

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/638.html" title="Last updated on Apr 5, 2024, 9:13 PM UTC (61355d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/638/1f60107...inexorabletash:61355d9.html" title="Last updated on Apr 5, 2024, 9:13 PM UTC (61355d9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -36,7 +36,7 @@ Status Text: <p>
   <li>it has at least two independent, interoperable implementations of every feature defined in the specification, where interoperability can be verified by passing open test suites, and two or more implementations interoperating with each other;</li>
   <li>it has an open test suite of every feature defined in the specification.</li>
   </ul>
-
+Text Macro: EMULATED generically emulated from the usage of other operations as follows, although user agents typically have a more efficient implementation. In cases where the underlying platform does not directly support an operation, this decomposition can be used as a template to guide the implementation.
 </pre>
 <pre class="anchors">
 urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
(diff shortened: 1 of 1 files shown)


reillyeon (approved):

anssiko (approved):
I note we have one outlier case in https://www.w3.org/TR/webnn/#api-mlgraphbuilder-reshape-method where a similar (but _not_ the same) note is added for squeeze, unsqueeze, and flatten ops that are _not_ part of the API:
>Many shape-related operations such as [squeeze](https://pytorch.org/docs/stable/generated/torch.squeeze.html), [unsqueeze](https://pytorch.org/docs/stable/generated/torch.unsqueeze.html), and [flatten](https://pytorch.org/docs/stable/generated/torch.flatten.html) can be generically implemented using the [reshape()](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-reshape) operation as follows:

Quickly glimpsing over the spec a reader might think those three ops are also part of the API. There's a subtle difference in the wording: "generically implemented" vs. "generically emulated". This is probably intentional but could mislead readers until they click the links and land on PyTorch API docs.

Noting this here because it caught my eye while reviewing. The PR is good as is. My comment can be addressed here or separately as deemed appropriate.

inexorabletash:
> I note we have one outlier case in https://www.w3.org/TR/webnn/#api-mlgraphbuilder-reshape-method where a similar (but _not_ the same) note is added for squeeze, unsqueeze, and flatten ops that are _not_ part of the API:

https://github.com/webmachinelearning/webnn/pull/598 should make that clearer; wording can also be tweaked.

huningxin (approved):
LGTM!

huningxin:

fdwr (approved):

Issue #636, opened 2024-04-05
Behavior for numeric overflows/underflows/etc in algorithms


inexorabletash:
Some spec algorithms steps perform math that could overflow/underflow in implementations that do the arithmetic using particular types, such as `uint32_t`. This has been raised in https://github.com/webmachinelearning/webnn/issues/610#issue-2196705182 and https://github.com/webmachinelearning/webnn/pull/582#discussion_r1503641101 and probably elsewhere.

There are a handful of nuances here:
* This issue _isn't_ talking about places where there are input parameter constraints that can't be satisfied by Web IDL types, e.g. a floating point parameter must be greater than zero, or where one parameter must be greater than another, etc. Those need explicit steps.
* This issue _isn't_ talking about implementation-specific limits on operator behavior, e.g. maximum number of tensor dimensions, etc.
* This issue _is_ talking about places where intermediate value could overflow, just because the implementation is doing the math in a specific space e.g. using 32- or 64-bit intermediate values, because 128-bit values aren't supported or have poor performance.
* This issue is _maybe_ talking about places where the final value of a calculation has limitations imposed by constraints in other parts of the platform, e.g. running into 2^32, 2^53, 2^64 or other specified limits for objects (Arrays, ArrayBuffers, etc). Some of those limitations have relaxed over time (e.g. maximum ArrayBuffer size).

https://source.chromium.org/chromium/chromium/src/+/main:components/ml/webnn/graph_validation_utils.cc;l=36 is an implementation example where a filter size is constrained to be a uint32 even though larger values could be supported (I think?)

I checked with some other spec experts. Approaches really vary depending on the spec's domain.

Most specs implicitly follow the guidance from [Infra](https://infra.spec.whatwg.org/) which says this on [algorithm limits](https://infra.spec.whatwg.org/#algorithm-limits):
> A document using the Infra Standard generally should not enforce specific limits on algorithm inputs with regards to their size, resource usage, or equivalent. This allows for competition among user agents and avoids constraining the potential computing needs of the future.

[...]

Issue #391, opened 2023-05-23
Allow 0 size dimensions


huningxin:
In Chromium CL [review](https://chromium-review.googlesource.com/c/chromium/src/+/4528431/4/components/ml/webnn/graph_validation_utils.cc#21), @fdwr mentioned (Thanks Dwayne!)

> 0 size dimensions really should just be treated as nops. e.g. memcpy(dest, src, 0) is valid and does nothing, and adding two empty tensors just returns an empty tensor. There are legitimate cases within a graph where a tensor may be temporarily sliced down to emptiness and then reconcatenated later with other data (I've come across at least one ONNX model that does this).

From native ML API perspective, Dwayne also mentioned

> Now, there are certain backends that are not prepared for empty dimensions (e.g. DML doesn't accept them and rejects the operator creation call), and those operators will need to be skipped inside the backend, but the front-end model builder imo should treat them validly and return 0 elements.

We may want to investigate more native ML APIs to understand what's the status of the support.

I am opening this issue to start tracking, e.g. adding TODO in the implementation. @fdwr, feel free to share more details. Thanks!

fdwr:
# References

All the major Python ML APIs support it:

```python
################################################################################
import numpy

x = numpy.ones(shape=(2,0,2), dtype=numpy.float32)
y = numpy.add(x, x)
print("NumPy:")
print("value:", y)
print("shape:", y.shape)

# Prints:
# value: []
# shape: (2, 0, 2)

################################################################################
import tensorflow as tf

x = tf.ones(shape=(2,0,2), dtype=tf.float32)
y = tf.add(x, x);
print("TensorFlow:")
print("value:", y)
print("shape:", y.shape)

# Prints:
# value: tf.Tensor([], shape=(2, 0, 2), dtype=float32)
# shape: ()

################################################################################
import torch

x = torch.ones(size=(2,0,2), dtype=torch.float)
y = torch.add(x, x)
print("PyTorch:")
print("value:", y)
print("shape:", y.shape)

# Prints:
# value: tensor([], size=(2, 0, 2))
# shape: torch.Size([2, 0, 2])

################################################################################
import onnx

# Scalar via [].
x = onnx.helper.make_tensor(
    name="value", data_type=onnx.TensorProto.FLOAT, dims=[2,0,2], vals=[]
)
print(x)

# Prints:
# dims: 2
# dims: 0
# dims: 2
# data_type: 1
# name: "value"
```

# Other Implementation Experience

[...]

PR #641, opened 2024-04-08
Introduce "valid dimension", used as needed when calculating operand shapes

### LINKED ISSUE #610: Need clarify scale factor for resample2d
Link to **resample2d** definitions: https://webmachinelearning.github.io/webnn/#mlgraphbuilder-calculate-resample-output-sizes.

1) The **output sizes** by calculating with **scales** are not correct, and the `desc's shape` is not accurate since the **shape** is defined as "An MLOperand's shape is its [[descriptor]].dimensions.". - I submitted a pr #611 to fix it, please take a review, thanks.

2) And it needs clarify the limitations for the product of scale factor and spatial dimension's size, I write such tests in my [CL-5382118](https://chromium-review.googlesource.com/c/chromium/src/+/5382118/1/third_party/blink/web_tests/external/wpt/webnn/validation_tests/resample2d.https.any.js), please take a look, thanks.

inexorabletash:
Certain operands like reshape2d() calculate dimensions in their output shape with expressions like shape[i] * scales[i] which could be larger than can be represented as a dimension in MLOperandDescriptor.

Introduce a "valid dimension" term, use it when calculating output dimensions in the following ops, and throw if needed.

- constant() (sequential fill)
- concat()
- conv2d()
- convTranspose2d()
- pad()
- pooling operators
- resample2d()
- reshape()

Fixes #610

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/641.html" title="Last updated on Apr 16, 2024, 3:44 AM UTC (67027f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/641/1f60107...inexorabletash:67027f2.html" title="Last updated on Apr 16, 2024, 3:44 AM UTC (67027f2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1370,6 +1370,7 @@ Data truncation will occur when the values in the range exceed the range of the 
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
         1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |type|.
         1. Let |size| be max(0, ceil((|end| - |start|)/|step|)).
+        1. If |size| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
         1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to the [=/list=] « |size| ».
     1. *Make graph connections:*
         1. Let |operand| be the result of [=creating an MLOperand=] given [=this=] and |descriptor|.
@@ -1777,7 +1778,10 @@ partial interface MLGraphBuilder {
                 If the shape of each corresponding dimension and type of the operands, except for those of the dimension given by |axis|, is not the same, fail.
             </div>
             1. If |dim| is not equal to |axis| and if |input|'s [=MLOperand/shape=][|dim|] is not equal to |first|'s [=MLOperand/shape=][|dim|], then [=exception/throw=] a {{TypeError}}.
-            1. If |dim| is equal to |axis|, add to |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] the value of |input|'s [=MLOperand/shape=][|dim|].
+            1. If |dim| is equal to |axis|:
+                1. Let |size| be the sum of |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] and |input|'s [=MLOperand/shape=][|dim|].
+                1. If |size| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
+                1. Set |desc|.{{MLOperandDescriptor/dimensions}}[|axis|] to |size|.
     1. *Make graph connections:*
         1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |desc|.
         1. Let |operator| be an [=operator=] for the concat operation, given |inputs| and |axis|.
@@ -1976,6 +1980,7 @@ partial interface MLGraphBuilder {
                 : {{MLInputOperandLayout/"nhwc"}}
                 :: Let |outputShape| be « |batches|, floor( |outputSizes|[0] ), floor( |outputSizes|[1] ), |outputChannels|  ».
             </dl>
+        1. If any [=list/item=] in |outputShape| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
         1. Let |desc| be a new {{MLOperandDescriptor}}.
         1. Set |desc|.{{MLOperandDescriptor/dataType}} to |input|'s [=MLOperand/dataType=].
         1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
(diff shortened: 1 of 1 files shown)


inexorabletash:
Up for discussion. A few more thoughts here:

* We'll want to roll in any change needed by #391
* It might be cleaner to inline the steps, rather than having "...then return failure." and "If that returns failure..." I kept the change minimal for now. Discuss!
* If we anticipate this occurring elsewhere in the spec then we could introduce a definition like:

> A **valid dimension** is an integer greater than zero in the range of [unsigned long](https://webidl.spec.whatwg.org/#idl-unsigned-long). Implementations may impose a lower upper limit.

(dropping "greater than zero" depending on #391)

... and then change [check dimensions](https://webmachinelearning.github.io/webnn/#mloperanddescriptor-check-dimensions) to say:

> If any element of _descriptor_.**dimensions** is not a valid dimension, then return false.

Thoughts?

inexorabletash:
If we introduce a "valid dimension" definition, we probably want to use it in these other places that calculate a dimension:
* [constant](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-constant-start-end-step-type)
* [conv2d](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-conv2d)
* [convTranspose2d](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-convtranspose2d)
* [pad](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-pad)
* [pooling operations](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-pooling-op)
* [reshape](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-reshape)

... which are the obvious places where a dimension is assigned from a calculation. Probably more places.

inexorabletash:
> * [concat](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-concat)

Ah yes, it's a sum of other dimensions, so could overflow.

> * [gru](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-gru) and [gruCell](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-grucell), their steps compare a dimension with `3 * hiddenSize` that might be a invalid unsigned long.
> * [lstm](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-lstm) and [lstmCell](https://webmachinelearning.github.io/webnn/#dom-mlgraphbuilder-lstmcell), their steps compare a dimension with `4 * hiddenSize` that might be a invalid unsigned long.

I _think_ those are cases where the _intermediate_ value could overflow if the math is done in uint32 space, but the _final_ values used for dimensions are always valid uint32, since they come directly from other MLOperands. So those are cases where #636 applies.

I'm gonna update this PR to introduce the "valid dimension" term. Should I go ahead and expand it beyond resample2d() or do that in another PR?

huningxin (approved):
LGTM!

huningxin:
Thanks @inexorabletash !

> Should I go ahead and expand it beyond resample2d() or do that in another PR?

Given "valid dimension" is defined, having all its usages in this PR sounds good to me. But I am also fine if doing it in another PR.

inexorabletash:
> Given "valid dimension" is defined, having all its usages in this PR sounds good to me. But I am also fine if doing it in another PR.

Done in 4f30583c2b18c6fcebb6f178b14ac4ddcc12b725 - I'll update the PR name / description

index.bs
```diff
@@ -5115,7 +5131,7 @@ partial interface MLGraphBuilder {
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |outputShape| be an empty array of {{unsigned long}}.
     1. If |newShape|'s [=list/size=] is 0, set |outputShape| to an empty [=/list=] for a scalar.
-    1. If any value in |newShape| is 0, then [=exception/throw=] a {{TypeError}}.
+    1. If any [=list/item=] in |newShape| is 0 or is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 5134: Nit: "valid dimension" already ensures the dimension is greater than 0.

```suggestion
    1. If any [=list/item=] in |newShape| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
```

huningxin (approved):
LGTM with a nit

index.bs
inexorabletash: I left this alone because if #391 allows 0 widely we might not want it for reshape... but on further thought that's caught by the later steps (_"If product of all values in newShape is not equal to inputElementCount, then throw a TypeError."_) So applying!

inexorabletash:
@fdwr - can you please review and merge if it looks good to you? Thanks!

fdwr (approved):
Tiny wording thought (do you prefer one over the other? I'll go with whatever you like.), else LGTM. Thanks, as I like centralizing this dimension validity rather than it being scattered around.

inexorabletash:
Suggested wording change was great @fdwr - it felt awkward as I was writing it.

Issue #559, opened 2024-02-06
Control flow operations: if, while


philloooo:
Hi, currently WebNN doesn't support control flow operations like `if` and `while`, so I am creating this issue to open the discussion on whether WebNN should support control flows.

We have a [where](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-where) which could be used similar to `if`, but it requires both branches to be computed, then toss away the branch we don't need, this is much less efficient to use it as `if`.

I don't see anything in current spec that can be used for `while`. 

Use case:
For transformer like models, for the decoder part,  it's essentially a while loop that keep feeding the output token back to the decoder, until it reaches the end token. If WebNN doesn't have the `while` op, this loop will exist on the javascript side, meaning that you will have this round trip and synchronization to client side JS for each iteration, much less efficient then encapsulating the whole logic into a single WebNN computation.

Thoughts?

Honry:
**+1.** Enabling modern LLMs and state-space models (SSMs) on WebNN requires a control-flow primitive—specifically a **loop op**—to handle recurrent sequence layers efficiently.

A growing class of shipping models (such as **[Qwen3.5](https://medium.com/@mlabonne/qwen3-5-nobody-agrees-on-attention-anymore-4709e1bd014b)**, **Mamba**, and gated-DeltaNet architectures) replaces most standard full-attention layers with linear-attention / SSM layers. Each such layer runs a recurrence across the sequence length:

$$S_t = f(S_{t-1}, x_t) \quad \text{for } t = 0 \dots T-1$$
 
Because this state transition happens deep inside the model graph across dozens of layers, **it cannot be hoisted out to JavaScript** (this is distinct from the outer generation/decode loop). The correct execution graph is literally `step → step → … → step`, with each step depending on the previous token's state.

Without a `loop` op, WebNN graphs—which are compiled once with a fixed operator count—cannot express $T$ steps when the sequence length $T$ is dynamic at runtime. The only workaround (unrolling for a fixed $T$) forces developers to build and compile separate graphs for every possible sequence length, which is not viable for real-world web applications.

A standard `loop` op — comprising a condition subgraph, body subgraph, and loop-carried state—maps 1:1 to underlying WebNN backends. It covers both per-layer sequence recurrence ($i < T$) and outer decode loops without synthesis overhead.

PR #642, opened 2024-04-08
gather(): Address indices validation and other algorithm nits

### LINKED ISSUE #484: Should Gather's indices data type be integers and support negative value?
In proposal [Add support for operations needed for well-known transformers](https://github.com/webmachinelearning/webnn/pull/478#discussion_r1407038824), the Gather's indices data type is defined as "uint32":

> indices: an [MLOperand](https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/478.html#mloperand). The indices N-D tensor of the input values to gather. The values must be of type "uint32" in the range [0, N-1] where N is the size of the input dimension indexed by options.axis.

However, as discussed in Chromium [CL-5017758](https://chromium-review.googlesource.com/c/chromium/src/+/5017758/comment/3439ff1d_f6ff665a/) review, frameworks and native ML APIs usually use signed integer and some of them support negative index values:

### Frameworks

- ONNX's [Gather](https://onnx.ai/onnx/operators/onnx__Gather.html): indices data type is int32 or int64, supports negative index values, in range [-N, N-1]
- TensorFlow's [Gather](https://www.tensorflow.org/api_docs/python/tf/gather): indices data type is int32 or int64, only supports non-negative index values, in range [0, N-1]

### Native ML APIs

- DirectML's [Gather](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_gather_operator_desc): indices data type is int64, int32, uint64 and uint32, supports negative index values
- CoreML's [Gather](https://apple.github.io/coremltools/mlmodel/Format/NeuralNetwork.html#gatherlayerparams): indices data type is integers, support negative index values
- StableHLO's [Gather](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#gather): indices data type is integers, doesn't support negative index values

Because the indices is an operand, its values may only available at run-time. If WebNN only supports "uint32" and range [0, N-1], frameworks that support negative indices, like ONNX, should polyfill (resolve the negative values) the indices tensor before passing to WebNN. WebNN implementation should enforce the restriction by polyfill the indices received from framework for all backends no matter whether the backend supports negative indices or not.

The best case would be single polyfill (frameworks supporting [0, N) + any backends), such as TF + DML, TF + StableHLO. The worst case might be double polyfill (frameworks support [-N, N) + any backends), such as ONNX + DML, ONNX + StableHLO.

The proposal is to support signed integers for Gather's indices and allow the negative index values, in range [-N, N-1] where N is the size of input dimensions indexed by options.axis. 

With that, frameworks should just pass the indices tensor to WebNN, no polyfill needed. WebNN implementation would only need to polyfill the indices for the backends that don't support negative indices.

The best case would be non polyfill at all (any frameworks + backends supporting [-N, N)), such as ONNX + DML, TF + CoreML. The worst case might be single polyfill (any frameworks + backends supporting [0, N-1)), such as ONNX + StableHLO, TF + StableHLO.

/cc @wchao1115 @fdwr @wacky6

### LINKED ISSUE #486: Add "implementation consideration" about how out-of-bound indices of Gather should be handled
In proposal [Add support for operations needed for well-known transformers ](https://github.com/webmachinelearning/webnn/pull/478#discussion_r1390547694), the validation steps of `indices` operand are defined as:

> 4. Let axis be options.[axis](https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/478.html#dom-mlgatheroptions-axis).
> 5. Let axisSize be input.[[[descriptor]]](https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/478.html#dom-mloperand-descriptor-slot).[dimensions](https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/478.html#dom-mloperanddescriptor-dimensions)[axis]

> 7. [For each](https://infra.spec.whatwg.org/#map-iterate) index → value of indices:
>   i. If index is greater than or equal to axisSize, then [throw](https://webidl.spec.whatwg.org/#dfn-throw) a "[DataError](https://webidl.spec.whatwg.org/#dataerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException).

The index value of `indices` operand may not be available at the graph build time, e.g., when `indices` is an input operand or an intermediate operand produced by another operator. It should define the run-time behavior if the index value of indices is greater than or equal to axisSize.

In that proposal, @wacky6 also [recommended](https://github.com/webmachinelearning/webnn/pull/478#discussion_r1396672041) to add  a "implementation consideration" (i.e. an informative section) mentioning how out-of-bound indices should be handled by browser vendor and platform implementation with some bullet points:
> * Runtime out-of-bounds indices should be explicitly handled by either the browser implementation or the platform implementation, to avoid OOB memory accesses.
> * If the platform implementation doesn't handle out-of-bounds indices, the browser implementation should take steps to ensure the platform operator doesn't receive out-of-bound indices
> * Mention what caller should expect as a result of list item 2: 0, NaN, first/last indices (if implemented with clamp)

/cc @anssiko @wchao1115 @fdwr

inexorabletash:
* #486 points out that indices can't be validated at build-time, and clamping behavior with an implementation note is given instead.

* Fix a typo in the steps.

* Replace several map-like iterations over lists with list iteration.

* #484 discusses support for negative indices. Added a note that if the underlying platform doesn't support them, the implementation can "polyfill" this with some additional logic.

Fixes #486
Fixes #484

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/642.html" title="Last updated on Apr 16, 2024, 4:04 PM UTC (cc0cf8a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/642/05b15b1...inexorabletash:cc0cf8a.html" title="Last updated on Apr 16, 2024, 4:04 PM UTC (cc0cf8a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2714,40 +2714,41 @@ partial interface MLGraphBuilder {
 <div>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input N-D tensor from which the values are gathered.
-        - *indices*: an {{MLOperand}}. The indices N-D tensor of the input values to gather. The values must be of type {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}} in the range [0, N-1] where N is the size of the input dimension indexed by *options.axis*.
+        - *indices*: an {{MLOperand}}. The indices N-D tensor of the input values to gather. The values must be of type {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}}, and are clamped to the range -N (inclusive) to N (exclusive) where N is the size of the input dimension indexed by *options.axis*, and a negative index means indexing from the end of the dimension.
         - *options*: an optional {{MLGatherOptions}}. The optional parameters of the operation.
 
     **Returns:** an {{MLOperand}}. The output N-D tensor of [=MLOperand/rank=] equal to the [=MLOperand/rank=] of *input* + the [=MLOperand/rank=] of *indices* - 1.
+</div>
+
+<div class="note">
+  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce clamping operands in the compiled graph if the required clamping behavior is not provided by the underlying platform.
 </div>
 
 <details open algorithm>
   <summary>
     The <dfn method for=MLGraphBuilder>gather(|input|, |indices|, |options|)</dfn> method steps are:
   </summary>
-    1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input| abd |indices| returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input| and |indices| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |indices|'s [=MLOperand/dataType=] is neither {{MLOperandDataType/"uint32"}} nor {{MLOperandDataType/"int64"}}, then [=exception/throw=] a {{TypeError}}.
     1. Let |shapeInput| be |input|'s [=MLOperand/shape=] and |rankInput| be |shapeInput|'s [=MLOperand/rank=].
     1. Let |shapeIndices| be |indices|'s [=MLOperand/shape=].
     1. Let |axis| be |options|.{{MLGatherOptions/axis}}.
-    1. Let |axisSize| be |input|'s [=MLOperand/shape=][|axis|]
     1. If |axis| is greater than or equal to |rankInput|, then [=exception/throw=] a {{TypeError}}.
-    1. [=map/For each=] |index| → |value| of |indices|:
-        1. If |index| is greater than or equal to |axisSize|, then [=exception/throw=] a {{TypeError}}.
     1. Let |dimCount| be zero.
     1. Let |rankOutput| be zero.
     1. Let |shapeOutput| be an empty list.
-    1. [=map/For each=] |size| → |value| of |shapeInput|:
+    1. [=list/For each=] |size| of |shapeInput|:
         1. If |dimCount| is equal to |axis| then [=iteration/break=].
         1. Set |shapeOutput|[|dimCount|] to |size|.
         1. Increment |dimCount| by one.
     1. Set |rankOutput| to |dimCount|.
     1. Let |dimCount| be zero.
-    1. [=map/For each=] |size| → |value| of |shapeIndices|:
+    1. [=list/For each=] |size| of |shapeIndices|:
         1. Set |shapeOutput|[|rankOutput| + |dimCount|] to |size|.
         1. Increment |dimCount| by one.
     1. Set |rankOutput| to |rankOutput| + |dimCount|.
     1. Let |dimCount| be zero.
-    1. [=map/For each=] |size| → |value| of |shapeInput|:
+    1. [=list/For each=] |size| of |shapeInput|:
         1. If |dimCount| is less than or equal to |axis| then [=iteration/continue=].
         1. Set |shapeOutput|[|rankOutput| + |dimCount| - |axis| - 1] to |size|.
         1. Increment |dimCount| by one.
--- a/index.bs
+++ b/index.bs
@@ -2721,7 +2721,7 @@ partial interface MLGraphBuilder {
 </div>
 
 <div class="note">
-  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce {{MLGraphBuilder/clamp()}} in the compiled graph if the required clamping behavior is not provided by the underlying platform.
+  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce {{MLGraphBuilder/clamp()}} in the compiled graph if the required clamping behavior is not provided by the underlying platform. Similarly, if the underlying platform does not support negative indices, the implementation can introduce operations in the compiled graph to transform a negative index from the end of the dimension into a positive index.
 </div>
 
 <details open algorithm>
--- a/index.bs
+++ b/index.bs
@@ -2721,7 +2721,7 @@ partial interface MLGraphBuilder {
 </div>
 
 <div class="note">
-  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce clamping operands in the compiled graph if the required clamping behavior is not provided by the underlying platform.
+  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce {{MLGraphBuilder/clamp()}} in the compiled graph if the required clamping behavior is not provided by the underlying platform.
 </div>
 
 <details open algorithm>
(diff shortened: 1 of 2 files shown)


inexorabletash:
I noticed the other glitches in gather(), then went looking for open bugs and tackled #486 at the same time. Wheeee!

inexorabletash:
@a-sully can you take a first look?

a-sully (approved):
LGTM 👍

index.bs
```diff
@@ -2714,40 +2714,41 @@ partial interface MLGraphBuilder {
...
+        - *indices*: an {{MLOperand}}. The indices N-D tensor of the input values to gather. The values must be of type {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}}, and are clamped to the range -N (inclusive) to N (exclusive) where N is the size of the input dimension indexed by *options.axis*, and a negative index means indexing from the end of the dimension.
         - *options*: an optional {{MLGatherOptions}}. The optional parameters of the operation.
 
     **Returns:** an {{MLOperand}}. The output N-D tensor of [=MLOperand/rank=] equal to the [=MLOperand/rank=] of *input* + the [=MLOperand/rank=] of *indices* - 1.
 </div>
 
+<div class="note">
+  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce clamping operands in the compiled graph if the required clamping behavior is not provided by the underlying platform.
```
huningxin, line 2724: Referencing `clamp` operator?

```suggestion
  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce {{MLGraphBuilder/clamp()}} in the compiled graph if the required clamping behavior is not provided by the underlying platform.
```

huningxin (approved):
LGTM with a comment, thanks!

inexorabletash:
Do we want to tweak this at all to also tackle #484 or leave that for another PR?

a-sully:
Ah good question... 

To specify support for negative indices in WebNN, we need to assume it's reasonable that every WebNN backend (that doesn't already support negative indices) will be able to transform these values at runtime. For example, the simplest way for our WebNN implementation to polyfill is by inserting `if` and `add` operations:
```
if index > 0:
  return index
else:
  return index + input.dimensions[axis]
```
This assumes that an `if()` operator is available. Looking at #484, TFLite and StableHLO are called out as the backends which don't support negative indices. Both of those backends appear to have an `if` operator [[1](https://crsrc.org/c/third_party/tflite/src/tensorflow/lite/schema/schema_generated.h;drc=229b345cbe5afd07c55c51a5a5325ce9016200d5;l=1342)][[2](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#if)], though I'm not sure whether e.g. `if` is supported on all TFLite backends we care about

This becomes easier if we say that WebNN must support control flow ops #559

There may be some other more complicated way to transform these values, too...

@huningxin any thoughts?

huningxin:
@a-sully 

> This becomes easier if we say that WebNN must support control flow ops #559
> 
> There may be some other more complicated way to transform these values, too...

It might be emulated by `where`:

```js
add(indices, where(lesser(indices, constant(0)), constant(input.dimensions[axis]), constant(0)))
```

a-sully:
Thanks, @huningxin!

> To specify support for negative indices in WebNN, we need to assume it's reasonable that every WebNN backend (that doesn't already support negative indices) will be able to transform these values at runtime

Seems reasonable. @inexorabletash shall we fold that into this PR (a nonnormative note to implementers, perhaps?) and close #484?

inexorabletash:
> Seems reasonable. @inexorabletash shall we fold that into this PR (a nonnormative note to implementers, perhaps?) and close #484?

I added a sentence to the note in 446bc773590c1e8814e7b593c6913da7fd3f4b09 - I kept it short, and didn't provide the decomposition. WDYT?

a-sully (approved):

huningxin (approved):
LGTM!

inexorabletash:
@fdwr can you do a final review and merge if it looks good to you? Thanks!

index.bs
```diff
@@ -2714,40 +2714,41 @@ partial interface MLGraphBuilder {
...
         - *options*: an optional {{MLGatherOptions}}. The optional parameters of the operation.
 
     **Returns:** an {{MLOperand}}. The output N-D tensor of [=MLOperand/rank=] equal to the [=MLOperand/rank=] of *input* + the [=MLOperand/rank=] of *indices* - 1.
 </div>
 
+<div class="note">
+  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built. Implementations can introduce {{MLGraphBuilder/clamp()}} in the compiled graph if the required clamping behavior is not provided by the underlying platform. Similarly, if the underlying platform does not support negative indices, the implementation can introduce operations in the compiled graph to transform a negative index from the end of the dimension into a positive index.
```
fdwr, line 2724:
```suggestion
  The {{MLGraphBuilder/gather(input, indices, options)/indices}} parameter to {{MLGraphBuilder/gather()}} can not be clamped to the allowed range when the graph is built because the inputs are not known until execution. Implementations can introduce {{MLGraphBuilder/clamp()}} in the compiled graph if the required clamping behavior is not provided by the underlying platform. Similarly, if the underlying platform does not support negative indices, the implementation can introduce operations in the compiled graph to transform a negative index from the end of the dimension into a positive index.
```

index.bs
```diff
@@ -2714,40 +2714,41 @@ partial interface MLGraphBuilder {
 <div>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input N-D tensor from which the values are gathered.
-        - *indices*: an {{MLOperand}}. The indices N-D tensor of the input values to gather. The values must be of type {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}} in the range [0, N-1] where N is the size of the input dimension indexed by *options.axis*.
+        - *indices*: an {{MLOperand}}. The indices N-D tensor of the input values to gather. The values must be of type {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}}, and are clamped to the range -N (inclusive) to N (exclusive) where N is the size of the input dimension indexed by *options.axis*, and a negative index means indexing from the end of the dimension.
```
fdwr, line 2717:
```suggestion
        - *indices*: an {{MLOperand}}. The indices N-D tensor of the input values to gather. The values must be of type {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}}, and must be to the range -N (inclusive) to N (exclusive) where N is the size of the input dimension indexed by *options.axis*, and a negative index means indexing from the end of the dimension.
```

Since the primary motivation of this clause is really about security mitigation for the case of garbage in, rather than prescribing algorithm behavior (I don't want people to rely on this assumption that bad indices will be automagically clamped for them when they should never pass them in the first place), then I'd write this sentence as more a statement that the indices *should* be limited to that range, and the note below can recommend a mitigation. I agree that clamping to the range -N (inclusive) to N (exclusive) makes sense, but it's also one possible prescribed mitigation for degenerate input. For example, DML ensures that the computed input offset read is within the bounds of the input tensor (so at least no off-memory reads), but not that each index is constrained to the dimensions.

fdwr (approved):
Approved with two suggestions. Thanks for adding Josh.

fdwr:
> It might be emulated by where:
> add(indices, where(lesser(indices, constant(0)), constant(input.dimensions[axis]), constant(0)))

Yep, that looks right. An alternative would be...

`where(lesser(indices, constant(0)), indices, add(indices, constant(input.dimensions[axis])))`

...which is slightly shorter (one less `constant(0)`), but it's a tossup which one is more efficient 🤔. So that's fine.

> This assumes that an if operator is available

Btw, the former name of `where` was `elementwiseIf`, which makes it more immediately clear to those less familiar with ML that it's the common `select` or `ternary` operator. Though, the new name is more discoverable to people who already know TF and PT ⚖️.

inexorabletash:
Thanks @fdwr  - I incorporated your change then fixed one grammar glitch.

fdwr (approved):
👍 Arigatou Josh.

PR #643, opened 2024-04-16
Fix/simplify some validation steps

inexorabletash:
- Overzealous conversion of "p of x" to "x's p" mangled some cases of "p of x, y or z"; fix them.
- When validating optional arguments, shorten substeps by using "its" more consistently rather than restating the option name.
- A few places were operating on options unconditionally. Make those conditional.
- When comparing two args (e.g. dataTypes), make the order consistent, i.e. "if filter's dataType is not the same as input's data type".
- Prefer "equal to" over "the same as" in algorithms.
- Fix a list literal to use «»

Also, sneak in another case where `[EMULATED]` text macro can be used.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/643.html" title="Last updated on Apr 17, 2024, 8:25 PM UTC (ef4eebc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/643/d846723...inexorabletash:ef4eebc.html" title="Last updated on Apr 17, 2024, 8:25 PM UTC (ef4eebc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5796,7 +5796,7 @@ partial interface MLGraphBuilder {
   <summary>
     The <dfn method for=MLGraphBuilder>where(|condition|, |input|, |other|)</dfn> method steps are:
   </summary>
-    1. If |condition|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"uint8"}}, then [=exception/throw=] a {{TypeError}}.
+    1. If |condition|'s [=MLOperand/dataType=] is not equal to {{MLOperandDataType/"uint8"}}, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/dataType=] is not equal to |other|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |input|'s [=MLOperand/dataType=].
--- a/index.bs
+++ b/index.bs
@@ -1563,13 +1563,13 @@ partial interface MLGraphBuilder {
     1. If |variance|'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLBatchNormalizationOptions/scale}} [=map/exists=]:
         1. If its [=MLOperand/rank=] is not 1, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLBatchNormalizationOptions/scale}}'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
+        1. If its [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLBatchNormalizationOptions/bias}} [=map/exists=]:
         1. If its [=MLOperand/rank=] is not 1, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLBatchNormalizationOptions/bias}}'s [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
+        1. If its [=MLOperand/shape=][0] is not equal to |input|'s [=MLOperand/shape=][|options|.{{MLBatchNormalizationOptions/axis}}], then [=exception/throw=] a {{TypeError}}.
     1. *Make graph connections:*
         1. Let |operator| be an [=operator=] for the batchNormalization operation, given |input|, |mean|, |variance| and |options|.
-        1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |input|.{{MLOperand/[[descriptor]]}}, that may use the same underlying data as |input|.
+        1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |input|.{{MLOperand/[[descriptor]]}}.
         1. If |options|.{{MLBatchNormalizationOptions/activation}} [=map/exists=], then add it to |operator|'s [=operator/activation functions=].
         1. Set |output|.{{MLOperand/[[operator]]}} to |operator|.
         1. Set |operator|'s [=operator/inputs=] to |input|, |mean|, and |variance|.
@@ -1903,7 +1903,7 @@ partial interface MLGraphBuilder {
     1. If |options|.{{MLConv2dOptions/activation}} [=map/exists=], and [=MLGraphBuilder/validating activation=] with [=this=] and it returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/rank=] is not 4, then [=exception/throw=] a {{TypeError}}.
     1. If |filter|'s [=MLOperand/rank=] is not 4, then [=exception/throw=] a {{TypeError}}.
-    1. If |input|'s [=MLOperand/dataType=] is not the same as |filter|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
+    1. If |filter|'s [=MLOperand/dataType=] is not equal to |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLConv2dOptions/padding}} does not [=map/exist=], set it to the [=/list=] « 0, 0, 0, 0 ».
     1. Else if |options|.{{MLConv2dOptions/padding}}'s [=list/size=] is not 4, then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLConv2dOptions/strides}} does not [=map/exist=], set it to the [=/list=] « 1, 1 ».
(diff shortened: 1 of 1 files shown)


inexorabletash:
This is a preamble to a change that introduces dataType validation to all ops where necessary - this cleanup work was noticed along the way, and keeps that upcoming change nice and clean.

inexorabletash:
Please take a look @huningxin and @fdwr ?

huningxin (approved):
LGTM, thanks for the fix!

index.bs
```diff
@@ -1903,7 +1903,7 @@ partial interface MLGraphBuilder {
     1. If |options|.{{MLConv2dOptions/activation}} [=map/exists=], and [=MLGraphBuilder/validating activation=] with [=this=] and it returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/rank=] is not 4, then [=exception/throw=] a {{TypeError}}.
     1. If |filter|'s [=MLOperand/rank=] is not 4, then [=exception/throw=] a {{TypeError}}.
-    1. If |input|'s [=MLOperand/dataType=] is not the same as |filter|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
+    1. If |filter|'s [=MLOperand/dataType=] is not equal to |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 1906: Yeah, this reordering makes sense.

index.bs
```diff
@@ -2730,7 +2730,7 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>gather(|input|, |indices|, |options|)</dfn> method steps are:
   </summary>
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input| and |indices| returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If |indices|'s [=MLOperand/dataType=] is neither {{MLOperandDataType/"uint32"}} nor {{MLOperandDataType/"int64"}}, then [=exception/throw=] a {{TypeError}}.
+    1. If |indices|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"uint32"}} or {{MLOperandDataType/"int64"}}, then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 2733: Hmm, I found `neither nor` less ambiguous than `not x or y` (is that `(not x) or y` or `not (x or y)`?...). Yes, it's a little longer, but clearer imo. Propose reverting:

```suggestion
    1. If |indices|'s [=MLOperand/dataType=] is neither {{MLOperandDataType/"uint32"}} nor {{MLOperandDataType/"int64"}}, then [=exception/throw=] a {{TypeError}}.
```

index.bs
```diff
@@ -3956,31 +3962,31 @@ partial interface MLGraphBuilder {
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input|, |weight|, |recurrentWeight|, |options|.{{MLLstmOptions/bias}} (if it [=map/exists=]), |options|.{{MLLstmOptions/recurrentBias}} (if it [=map/exists=]), |options|.{{MLLstmOptions/peepholeWeight}} (if it [=map/exists=]), |options|.{{MLLstmOptions/initialHiddenState}} (if it [=map/exists=]), and |options|.{{MLLstmOptions/initialCellState}} (if it [=map/exists=]) returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLLstmOptions/activations}} [=map/exists=], and [=MLGraphBuilder/validating activation=] with [=this=] and any [=list/item=] in it returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |numDirections| be 1 if |options|.{{MLLstmOptions/direction}} is {{MLRecurrentNetworkDirection/"forward"}}, or otherwise let it be 2.
-    1. If |input|'s [=MLOperand/rank=] or |weight| or |recurrentWeight| is not 3, then [=exception/throw=] a {{TypeError}}.
+    1. If the [=MLOperand/rank=] of any of |input|, |weight| or |recurrentWeight| is not 3, then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 3965: 👍 Yes, much clearer now.

index.bs
```diff
@@ -3956,31 +3962,31 @@ partial interface MLGraphBuilder {
...
         1. If its [=MLOperand/rank=] is not 2, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLLstmOptions/bias}}'s [=MLOperand/shape=][0] is not |numDirections|, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLLstmOptions/bias}}'s [=MLOperand/shape=][1] is not 4 * |hiddenSize|, then [=exception/throw=] a {{TypeError}}.
+        1. If its [=MLOperand/shape=][0] is not |numDirections|, then [=exception/throw=] a {{TypeError}}.
+        1. If its [=MLOperand/shape=][1] is not 4 * |hiddenSize|, then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLLstmOptions/recurrentBias}} [=map/exists=]:
         1. If its [=MLOperand/rank=] is not 2, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLLstmOptions/recurrentBias}}'s [=MLOperand/shape=][0] is not |numDirections|, then [=exception/throw=] a {{TypeError}}.
-        1. If |options|.{{MLLstmOptions/recurrentBias}}'s [=MLOperand/shape=][1] is not 4 * |hiddenSize|, then [=exception/throw=] a {{TypeError}}.
+        1. If its [=MLOperand/shape=][0] is not |numDirections|, then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 3974: 👍 Yeah, this is more consistent with the precedent of the line above it for rank.

index.bs
```diff
@@ -5793,7 +5796,7 @@ partial interface MLGraphBuilder {
   <summary>
     The <dfn method for=MLGraphBuilder>where(|condition|, |input|, |other|)</dfn> method steps are:
   </summary>
-    1. If |condition|'s [=MLOperand/dataType=] is not equal to {{MLOperandDataType/"uint8"}}, then [=exception/throw=] a {{TypeError}}.
+    1. If |condition|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"uint8"}}, then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 5799: Hmm, if we're eliding `equal to` here, then do we need it above too?

```diff
- If its [=MLOperand/dataType=] is not equal to |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
+ If its [=MLOperand/dataType=] is not |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
```

Meh, shrug.

fdwr (commented):
Small comments, else LGTM. Thanks Josh.

index.bs
inexorabletash: Good test. Compare:

* _"If Bob's height is not 1.6m, then..."_  - seems fine
* _"If Bob's height is not equal to 1.6m, then..."_ - seems fine, but wordier
* _"If Bob's height is not equal to Alice's height, then..."_ - seems fine
* _"If Bob's height is not the same as Alice's height, then..."_ - also seems fine (but was used less frequently)
* _"If Bob's height is not Alice's height, then..."_ - reads a bit awkward to me, especially if it's not clear that this is a simple scalar value (e.g. is this literally a shared reference?)

Ultimately I'm fine with anything but aiming for consistency.

zolkis: These are fine in English, what is important is which is closer to spec'ese, or otherwise ecmascript'ese.
IIRC "is equal" and "is not equal" are used in most standards (e.g. HTML, INFRA, ...), to be checked.

index.bs
inexorabletash: Here are sample cases from the follow-on change, and why I shied away from neither/nor:

* If input’s dataType is not "float32" or "float16", then ...
* If input’s dataType is not "float32", "float16", "int32", or "int8", then ...
* If input’s dataType is not "uint8", then ...
* If allowedDataTypes does not contain input’s dataType, then ... _&larr; used for pooling/reduction/unary ops_

So these would be rephrased as:

* If input’s dataType is neither "float32" nor "float16", then ... _&larr; works but fancy?_
* If input’s dataType is neither "float32", "float16", "int32", nor "int8", then ... _&larr; awkward?_
* If input’s dataType is neither "float32" nor "float16" nor "int32" nor "int8", then ... _&larr; awkward?_
* If input’s dataType is not "uint8", then ... _&larr; inconsistent?_
* If allowedDataTypes does not contain input’s dataType, then ...

We could also go for:

* If input’s dataType is not in « "float32", "float16" », then ...
* If input’s dataType is not in « "float32", "float16", "int32", "int8" », then ...
* If input’s dataType is not "uint8", then ... _&larr; inconsistent?_
* If input's dataType is not in allowedDataTypes , then ... 

Preference?

I'll also offer a table-based approach, e.g.

* If input's dataTypes is not an allowed data type for "op" and "input", then ...

... where in an appendix there's a table of op / arg / allowed types.

index.bs
inexorabletash: It gets more complicated when (1) the test is inverted and (2) there are multiple options - see other comment. But agreed that explicit "equal" is probably clearer.

index.bs
inexorabletash: Additional alternatives, trying to work "equal" in there:

* If input’s dataType is not equal to either "float32" or "float16", then ...
* If input’s dataType is not equal to any of "float32", "float16", "int32", or "int8", then ...
* If input’s dataType is not equal to "uint8", then ...
* If allowedDataTypes does not contain input’s dataType, then ... _&larr; used for pooling/reduction/unary ops_

index.bs
inexorabletash: reverted in ef4eebc618228c4c281952a72aa15a2639f049bf

index.bs
fdwr: Aah, the paradox of choice... I'll accept it as-is and merge, as it sounds like it may well change in the follow-on change anyway, but from the above, I rather like that list approach:

`If input’s dataType is not « "float32", "float16", "int32", "int8" », then ...`

fdwr (approved):
Thanks Josh.

PR #644, opened 2024-04-17
Validate the hidden size of GRU and LSTM operators

### LINKED ISSUE #625: Need clarify the maximum limit of the hidden size of the Gru operator
WebNN defines two parameters bias and recurrent bias for Gru operator. Their shapes are [numDirections, 3 * hiddenSize], but in some backends like [DirectML](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_gru_operator_desc) only defines one bias and the shape is [1, 1, numDirections, 6 * hiddenSize] ], 6*hiddenSize needs to be verified at this time.
We are currently validate it in backed side, but in order to avoid surprise build failures for web developers, we hope to validate and throw error in the verification layer. So should this limit be part of the spec and validated for all backends?
Link to chromium [CL discussion](https://chromium-review.googlesource.com/c/chromium/src/+/5339174/comment/d4ee35e1_89a13efe/).

inexorabletash:
For gru()/gruCell() and lstm()/lstmCell(), a hiddenSize parameter is passed, a multiple of which defines a dimension of the output.

- This is sometimes implicitly validated by the presence of an option with the same dimension - but not always.

- Some underlying platforms operate on a single bias tensor, rather than the two bias/recurrentBias options present in the WebNN API. So the combined size needs to also be a valid dimension.

Introduce validation for all cases, validate the combined size, and add an explanation inline since this is subtle.

Fixes #625

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/644.html" title="Last updated on Apr 29, 2024, 4:26 PM UTC (9feeefb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/644/9f0f45a...inexorabletash:9feeefb.html" title="Last updated on Apr 29, 2024, 4:26 PM UTC (9feeefb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3985,15 +3985,12 @@ partial interface MLGraphBuilder {
     1. If |input|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"float32"}} or {{MLOperandDataType/"float16"}}, then [=exception/throw=] a {{TypeError}}.
     1. If the [=MLOperand/rank=] of any of |input|, |weight| or |recurrentWeight| is not 3, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/shape=][0] is not equal to |steps|, then [=exception/throw=] a {{TypeError}}.
-<<<<<<< HEAD
+    1. If the [=MLOperand/dataType=] of either |weight| or |recurrentWeight| is not equal to |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
     1. If |hiddenSize| * 8 is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
         <details class=note>
           <summary>Why |hiddenSize| * 8 ?</summary>
           Some underlying platforms operate on a single bias tensor which is a concatenation of {{MLLstmOptions/bias}} and {{MLLstmOptions/recurrentBias}}. Therefore, 4 * |hiddenSize| + 4 * |hiddenSize| must also be a [=valid dimension=].
         </details>
-=======
-    1. If the [=MLOperand/dataType=] of either |weight| or |recurrentWeight| is not equal to |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
->>>>>>> refs/heads/review
     1. Let |batchSize| be |input|'s [=MLOperand/shape=][1].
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM, thanks!

zolkis:
Special thanks for the clarifying notes.

index.bs
```diff
@@ -4017,11 +4019,15 @@ partial interface MLGraphBuilder {
     1. If |input|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"float32"}} or {{MLOperandDataType/"float16"}}, then [=exception/throw=] a {{TypeError}}.
     1. If the [=MLOperand/rank=] of any of |input|, |weight| or |recurrentWeight| is not 3, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/shape=][0] is not equal to |steps|, then [=exception/throw=] a {{TypeError}}.
+<<<<<<< HEAD
```
huningxin, line 4022: This seems to be a rebase issue.

index.bs
```diff
@@ -4238,11 +4244,15 @@ partial interface MLGraphBuilder {
     1. If |options|.{{MLLstmCellOptions/activations}} [=map/exists=], and [=MLGraphBuilder/validating activation=] with [=this=] and any [=list/item=] in it returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"float32"}} or {{MLOperandDataType/"float16"}}, then [=exception/throw=] a {{TypeError}}.
     1. If the [=MLOperand/rank=] of any of |input|, |weight|, |recurrentWeight|, |hiddenState| or |cellState| is not 2, then [=exception/throw=] a {{TypeError}}.
+<<<<<<< HEAD
```
huningxin, line 4247: Same here.

index.bs
inexorabletash: Thanks, fixed in 98e3892872e85cfec401624bb9ca1d178b3561f0

index.bs
inexorabletash: fixed in 98e3892872e85cfec401624bb9ca1d178b3561f0

huningxin (approved):
LGTM

fdwr (approved):
👍

Issue #492, opened 2023-12-15
Constant sequential filling operation needs output shape parameter


fdwr:
See `constant(start, end, step, type)` at https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-constant-start-end-step-type

# Problem

Computing the output shape indirectly from a floating-point calculation is brittle because the caller, depending on its equation and rounding, can yield a different size from WebNN which will cause graph execution to fail due to mismatched tensors. Such incompatible off-by-one differences can occur via differing calculation precisions (e.g. the caller computes the output shape in float64 whereas Chromium computes it in float32) and via differing handling of the inclusive/exclusive end value (see a non-hypothetical example below). So it's prudent for the library to pass WebNN the shape explicitly, and we've increasingly been moving toward more explicitness anyway (e.g. removing the special null values in Reshape). The output shape was in the [`fillSequence` operator](https://github.com/webmachinelearning/webnn/issues/375#issuecomment-1674224992), but got dropped, and it reoccurred to me while reviewing Bin's Chromium implementation.

# Proposal

```diff
- constant(start, end, step, type) method steps are:
+ constant(start, step, outputShape, type) method steps are:
...
- Let size be max(0, ceil((end - start)/step)).
- Set descriptor.[dimensions](https://www.w3.org/TR/webnn/#dom-mloperanddescriptor-dimensions) to [size].
+ Set descriptor.[dimensions](https://www.w3.org/TR/webnn/#dom-mloperanddescriptor-dimensions) to [outputShape].
+ Let outputElementCount be the product of all elements in descriptor.[dimensions](https://www.w3.org/TR/webnn/#dom-mloperanddescriptor-dimensions).
...
- [For each](https://infra.spec.whatwg.org/#map-iterate) index in [the range](https://infra.spec.whatwg.org/#the-range) 0 to size, exclusive:
+ [For each](https://infra.spec.whatwg.org/#map-iterate) index in [the range](https://infra.spec.whatwg.org/#the-range) 0 to outputElementCount, exclusive:
```

\* Note the `end` is now redundant, since it was only used for size calculation, not needed for the core `start + (index * step)` formula.

# Problematic Library differences

We see differences between TF and PT with the exact same inputs due to how they handle the end result (included vs excluded), which emphasizes why WebNN should be clearly informed by the caller of the expected shape:

```python
import tensorflow
result = tensorflow.range(5, 5, 2)
print("value:", result)

# output = [], shape= (0,)
```

```python
[...]

Issue #463, opened 2023-09-08
Allow checking whether operators/types are supported for a backend before creating a graph


huningxin:
This feedback is from @RafaelCintron (Thanks!) when reviewing [Chromium CL-4828993 review](https://chromium-review.googlesource.com/c/chromium/src/+/4828993/comment/007b377c_ee84c38e/). 

In the current spec and implementation, if any operators/types combination are not supported by a backend, for example [DirectML backend doesn't support the dilations of average pooling operator](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/dml/graph_impl.cc;l=470?q=dml%2Fgraph_impl&ss=chromium%2Fchromium%2Fsrc), the errors will be thrown when user code calls `MLGraphBuilder.build()`. Rafael mentioned this might be too late because user code may already download all the weights.

> We should add new WebNN APIs to help developers determine which of these operators/type combinations will fail BEFORE they attempt to create a graph. This will help them avoid large model downloads with operators/types the backend does not support.

wacky6:
:thinking: Or perhaps WebNN needs a weight loading step?

This is essentially the pattern I see with popular models (e.g. Real-ESRGAN, Stable diffusion). The graph describes the architecture and doesn't include weights.

If we want to fail before building the grpah, I'm afraid the operand methods (e.g. conv2d) needs to be async (to allow querying the backend), or make every feature feature-detectable (think of sending a feature matrix to the renderer).

huningxin:
@wacky6 

> Or perhaps WebNN needs a weight loading step?

This is a great idea. 

According to the current WebNN spec, the WebGPU context has [`MLCommandEncoder.initializeGraph(graph)`](https://www.w3.org/TR/webnn/#dom-mlcommandencoder-initializegraph) method for preparing and optimizing constant input data before subsequent graph executions. 

We probably can reuse this method for default context and extend it for weights uploading. For example, `MLContext.initializeGraph(graph, weights)`, it would allow to upload the weights to constant operands after building the graph. 

If the constants/weights uploading could be decoupled from graph building, developers can build the graph topology/architecture first before downloading the weights. If there are any unsupported error, they may handle the error or stop there. Only when the graph building succeeds, they proceed to download the weights from network and upload to device by `MLContext.initializeGraph()`.

wacky6:
Weight loading would substantially change how `graphBuilder.constant()` works. This will be a non-trivial change to implementation and usage.

[...]

Issue #283, opened 2022-08-26
Specify the operand data type constraints of operation


huningxin:
The current spec doesn't specify the operand type constraints of an operation. However, some operations, e.g., [`softmax`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-softmax) should only support float32 operand type according to the survey of frameworks and native ML APIs in the following table. The lack of the operand type constraints specification would lead to implementation issue, such as [Chromium CL 3856752](https://chromium-review.googlesource.com/c/chromium/src/+/3856752/comments/267c926c_b14af87f).

| Framework / API | supported data types |
|---|---|
| TensorFlow [tf.nn.softmax](https://www.tensorflow.org/api_docs/python/tf/nn/softmax) | half, float32, float64 |
| ONNX [Softmax](https://github.com/onnx/onnx/blob/main/docs/Operators.md#Softmax) | tensor(float16), tensor(float), tensor(double), tensor(bfloat16) |
| NNAPI [ANEURALNETWORKS_SOFTMAX](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0a2bfbb83a537701e2843a3d5004250c2c) | FLOAT32, FLOAT16 |
| DirectML [DML_ACTIVATION_SOFTMAX_OPERATOR_DESC](https://docs.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_activation_softmax_operator_desc) | FLOAT32, FLOAT16 |
| BNNS [ActivationFunction](https://developer.apple.com/documentation/accelerate/bnns/activationlayer/3649741-init) | float |

Thanks @wacky6 for pointing this out.

huningxin:
[...]

PR #647, opened 2024-04-18
Introduce MLNumber for specifying numeric inputs of any type

### LINKED ISSUE #325: Clarify the usage of 32 bit floating point type and consider using double
This issue was raised during Chromium [CL](https://chromium-review.googlesource.com/c/chromium/src/+/4031876/comment/b5589f90_a4e15806/) review. Thanks Jiewei @wacky6 for the valuable feedback!

The WebIDL 32 bit floating point type [`float`](https://webidl.spec.whatwg.org/#idl-float) is widely used by WebNN spec, such as for setting min and max value of [clamp](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-clamp) operator:

```js
dictionary MLClampOptions {
  float minValue;
  float maxValue;
};
```

However, WebIDL spec has a [warning](https://webidl.spec.whatwg.org/#idl-float) of using `float` and the recommendation is to use `double`. 

> Unless there are specific reasons to use a 32 bit floating point type, specifications should use double rather than float, since the set of values that a double can represent more closely matches an ECMAScript Number.

According to that, Jiewei suggested that the spec should mention reason of using `float` (at least an informative note) and propose  using `double` instead. The reasons include (pasted from the CL thread):

* Don't see particular reasons for limiting this to fp32
* What if (in the future) we have backends that can work with fp64? In this case fp32 becomes a limitation.
* Would it be worthwhile specifying float point casting behavior in the spec? Say, if caller passes in a fp64 number (that is too large to represent in fp32/fp16), how should backend interpret this (cast to infinity, error, find closest value non-infinity, something else)?

### LINKED ISSUE #442: Type of some parameters should match the input data type
1. **pad()**, `MLPadOptions::value` is float in current spec, which is used as padded value if the `MLPadOptions::mode` is "constant", its type should exactly match the input data type, otherwise this may cause precision loss.
(ONNX Pad-18 requires the input data and constant value (a single scalar input tensor) to be the same type T.)

2. **clamp()**, `MLClampOptions::minValue` and `MLClampOptions::maxValue`, current spec states them as a float scalar.

And for the v2 ops, we will also have to consider these: (thanks @fdwr for providing the list)
```
struct DML_FILL_VALUE_CONSTANT_OPERATOR_DESC
{
    const DML_TENSOR_DESC* OutputTensor;
    DML_TENSOR_DATA_TYPE ValueDataType;
    DML_SCALAR_UNION Value;
};

struct DML_FILL_VALUE_SEQUENCE_OPERATOR_DESC
{
    const DML_TENSOR_DESC* OutputTensor;
    DML_TENSOR_DATA_TYPE ValueDataType;
    DML_SCALAR_UNION ValueStart;
    DML_SCALAR_UNION ValueDelta;
};

struct DML_DIAGONAL_MATRIX1_OPERATOR_DESC
{
    _Maybenull_ const DML_TENSOR_DESC* InputTensor;
    const DML_TENSOR_DESC* OutputTensor;
    DML_TENSOR_DATA_TYPE ValueDataType;
    DML_SCALAR_UNION Value;
    INT DiagonalFillBegin;
    INT DiagonalFillEnd;
};
```

### LINKED ISSUE #678: Specifies scalar values casted to match input type.
Some of the operations have scalar parameters, e.g. [linear](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-linear), [gemm](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-gemm)'s `alpha` and `beta`.
These are passed as  `float` numbers.  I think we should specify in the processing algorithm that they get casted to match with the input type.  So that if input is float16, they also get downcast to float16.

Background is - in some cases, CoreML requires the parameter types to match. So we either downcast the scalar params to float16, or upcast input operands to float32, the latter is less ideal because on CoreML only float16 get executed on NPU.

Any concerns with adding this step in the algorithm?

inexorabletash:
For some `MLGraphBuilder` methods the type of a numeric input can vary - e.g. for `constant()` an explicit `MLOperandDataType` is provided; for `clamp()` and `pad()` the data type is implied by input operands. In these cases, specifying the numeric value as either a float/double or int64 type runs into accuracy or range issues - you can't accurately represent all int64 values as a double, and you can't represent the full range of floats as int64. (You also can't represent all int64 values as an long long either - over 2^53 things get wierd. But that's a digression.)

- Some methods that took a float or double argument/option but can operate on integer data types now take `MLNumber` - a "variant" (union) of either a JS Number (equivalent to a double a.k.a. float64) or BigInt (for full precision int64/uint64 use)
  - `constant()` (scalar overload) - `MLNumber`
  - `clamp()` (min/max options) - `MLNumber`
  - `pad()` (padding value) - `MLNumber`
  - This fixes #442
- Other methods that took a `float` argument/option now take a `double`. These are all ops that only operate on floating point types, so no need for `MLNumber` though it would be harmless to use it there to allow BigInt inputs. This follows web API best practices, and allows full precision conversion to float16 (since float64 to float32 to float16 can yield different results than float64 directly to float16)
  - `batchNormalization()`, `instanceNormalization()`, `layerNormalization()` - `epsilon` option
  - `elu()`, `gemm()`, `hardSigmoid()`, `leakyRelu()`, `linear()`  - `alpha` and `beta` options
  - This fixes #325
- In all of these cases, the input number is cast to the input/output operand data type when it is used.
Casting algorithms are spelled out, always have "clamp" semantics (i.e. no weird modulus wrapping), and never fail.
  - For `MLOperand`-vending methods, the conversion can be done eagerly.
  - For `MLActivation`s, the conversion is done at "fusion time"; it's notable that the same `MLActivation` instance could be fused with multiple ops, and cast differently for each.
  - This fixes #678
- Using a BigInt/numeric union is novel here, but WebIDL experts agree it's reasonable (whatwg/webidl/issues/1388). Note that all existing WebIDL/JS APIs that are designed to take just a Number or just a BigInt will throw if the other thing is passed - there is intentionally no silent conversion. So we are in novel territory here and should look for implementation and developer feedback.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/647.html" title="Last updated on Jul 3, 2024, 4:18 PM UTC (2e06299)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/647/08719e1...inexorabletash:2e06299.html" title="Last updated on Jul 3, 2024, 4:18 PM UTC (2e06299)">Diff</a>


inexorabletash:
FYI @fdwr @huningxin @zolkis - marked as "draft" but worth looking at.

- We may want to settle #492 before landing this - and "unrestricted double" may not be appropriate (since we want to preclude Infinity for fill ops?)
- Is there anywhere that we want to error on NaNs but allow Infinities? If so we need explicit prose steps. 
- `bigint` is not _just_ int64 - it allows arbitrary sized ints (e.g. `2n**100n`). If we want to error for a non-valid int64, we need explicit prose steps.
- I put the MLNumber definition and prose in the least bad place I could think of; suggestions welcome!

index.bs
```diff
@@ -4389,7 +4407,7 @@ enum MLPaddingMode {
 
 dictionary MLPadOptions {
   MLPaddingMode mode = "constant";
-  float value = 0;
+  MLNumber value = 0;
```
a-sully, line 4410: Since MLNumber includes `unrestricted double`, this value may now be non-finite. Do we need to update the `pad()` function's algorithm to check for this?

index.bs
```diff
@@ -1633,8 +1651,8 @@ partial interface MLGraphBuilder {
 Clamp the input tensor element-wise within a range specified by the minimum and maximum values.
 <script type=idl>
 dictionary MLClampOptions {
-  float minValue;
-  float maxValue;
+  MLNumber minValue;
+  MLNumber maxValue;
```
a-sully, line 1655: FWIW [the Chromium implementation assumes](https://crsrc.org/c/third_party/blink/renderer/modules/ml/webnn/ml_graph_builder.cc;l=530-532;drc=0fb9efef8267777ba45c3ba8e1d934dfe61a0d41) that these values are finite (not unrestricted) floats. In the case of `clamp` it seems reasonable for infinite values to be passed. Barring any pushback, I'd like to propose we change the implementation to allow infinites

That being said, should the `clamp()` function's algorithm throw on `NaN`?

Now that we're allowing unrestricted doubles there's a separate question of whether `minValue` should default to `-Infinity` and `maxValue` should default to `Infinity`... Probably best to leave that to a follow-up :)

index.bs
```diff
@@ -1133,6 +1135,16 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
 To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |builder| and {{MLOperand}} |operand|, return true if |operand|.{{MLOperand/[[builder]]}} is |builder|, and false otherwise.
 </p>
 
+#### {{MLNumber}} #### {#api-mlnumber-typedef}
+
+<dfn typedef>MLNumber</dfn> is used when specifying the type of a numeric option for an {{MLOperand}} which can be of any {{MLOperandDataType}}, including both 64-bit integer types ({{MLOperandDataType/"uint64"}} and {{MLOperandDataType/"int64"}}) and 32-bit floating point ({{MLOperandDataType/"float32"}}). Implementations must process the value according to the corresponding {{MLOperandDataType}}.
```
a-sully, line 1140: > Implementations must process the value according to the corresponding {{MLOperandDataType}}.

What if there is no corresponding `MLOperandDataType`? For example, `clamp()` can be built as an `MLActivation` without needing to specify any particular dtype. What is the dtype or the activation's associated [operator](https://webmachinelearning.github.io/webnn/#dom-mlactivation-operator-slot)?

The concept of an activation's operator is a bit hand-wavy in the spec, but [it's very concrete in the Chromium implementation](https://crsrc.org/c/third_party/blink/renderer/modules/ml/webnn/ml_activation.h;drc=763ac39bd0d3d8ae0562492b4628d5ed98fc2cbf;l=40) and the data type must be known when we pass the activation as an input to some other builder method anyways (we need to check that the types match, assuming we don't want to allow mixed precision https://github.com/webmachinelearning/webnn/issues/283#issuecomment-2046275085)

index.bs
inexorabletash: I think in one of the bugs @fdwr suggested that we use MLNumber where we support non-finite and just `(bigint or double)` where we only support finite values.

Some options:

* `MLNumber` typedef for the common case  and explicit union `(bigint or double)` otherwise 
* No typedefs - explicit unions `(bigint or double)` or `(bigint or unrestricted double)` as needed
* Multiple typedefs - e.g. 
  * `typedef (bigint or unrestricted double) MLNumber`
  * `typedef (bigint or double) MLFiniteNumber`
* Use `MLNumber` everywhere, with further restrictions in prose.

I'm slightly partial to the last option, because I anticipate we'll need prose for NaN everywhere anyway, so also restricting infinities is just a few more words. But expressing restrictions in IDL is also nice, so multiple typedefs would be my second choice.

index.bs
inexorabletash: Presumably the data type of the eventual operator's input? i.e. for the impl we'd need to hold onto the union in the MLActivation until the operator is constructed?

We need to improve the spec text, but I want to understand the implementation first.

index.bs
fdwr: >  I'd like to propose we change the implementation to allow infinites

Yes, very much so.

> That being said, should the clamp() function's algorithm throw on NaN?

Well NaN propagation is well defined, and there are helper functions specifically intended to check for NaN's during computation ([1](https://www.tensorflow.org/api_docs/python/tf/math/is_nan) [2](https://pytorch.org/docs/stable/generated/torch.isnan.html) [3](https://numpy.org/doc/stable/reference/generated/numpy.isnan.html) [4](https://onnx.ai/onnx/operators/onnx__IsNaN.html)...). So NaN isn't really a foreign concept or illegal concept in ML (then again, maybe it's just because I'm so intimately familiar with dozens of floating-point formats :b). I'd also expect clamp to behave like other libraries. e.g.

```python
x = tensorflow.constant([250], dtype=tensorflow.float32)
y = tensorflow.clip_by_value(x, 13, float('nan'))
print("value:", y)
print("shape:", y.shape)

# Prints:
# value: tf.Tensor([nan], shape=(1,), dtype=float32)
# shape: (1,)
```

> Now that we're allowing unrestricted doubles there's a separate question of whether minValue should default to -Infinity and maxValue should default to Infinity... Probably best to leave that to a follow-up :)

Those are sensible defaults. We just have to be sure that the logic inside Chromium for integer data types like int32 respectively maps +Infinity to INT32_MAX and -Infinity to INT32_MIN. Agreed that it's cleaner to be a separate CR.

index.bs
huningxin: > typedef (bigint or double) MLFiniteNumber

This also sounds good.

index.bs
inexorabletash: Added a short note in 67b5a68bf88e8202f0e915d80d8c7c2def6ee504 but we can probably improve it.

a-sully: > > i.e. for the impl we'd need to hold onto the union in the MLActivation until the operator is constructed?
> 
> I think so.

Does this make sense? What happens if an `MLActivation` is passed to multiple operators with different data types?

index.bs
a-sully: Cool, I'll add support for infinites and NaN in `clamp` when I [add support for bigint in the bindings code](https://crrev.com/c/5466806)

>  I'd also expect clamp to behave like other libraries. e.g.

Yup, SGTM. We should add WPT conformance tests asserting that we get the expected behavior for inf/NaN (FYI @BruceDai)

index.bs
inexorabletash: > What happens if an MLActivation is passed to multiple operators with different data types?

My assumption is that the MLActivation will internally need to hold the union for its whole lifetime, and the value is converted (as a copy) to a value of appropriate data type during either:
* operator creation (e.g. when conv2d() is called)
* during build, when the graph is converted to a compiled graph

Ideally the difference between those would not be observable, which might constrain how we do validation - if we throw on values (e.g. a negative number passed to a unsigned type, Infinity passed to an integer type, etc) then it would be observable if we validate during op creation or during build.

a-sully: For the sake of completeness, there's a third place where validation could happen:

- activation creation

So the levels of validation for a `clamp()` activation might look like:

- On activation creation:
  - [check that `minValue <= maxValue`](https://webmachinelearning.github.io/webnn/#check-clamp-options)
- On operator creation (with the respective activation):
  - [check `MLGraphBuilder` is the same](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-validate-activation)
  - As of #646 we should have enough information to validate data types, too. This is related to the discussion of what the behavior should be if the `MLNumber` does not fit into the respective type (e.g. passing a negative number to a "uint32" operator)
- During build:
  - Currently, this includes platform-specific checks such as whether the operator supports the given activation. Moving this validation forward falls under the umbrella of #463

... this seems fine? What I find unintuitive is squaring this statement:

> the MLActivation will internally need to hold the union for its whole lifetime, and the value is converted (as a copy) to a value of appropriate data type

with the idea that an `MLActivation` has one [operator slot](https://webmachinelearning.github.io/webnn/#dom-mlactivation-operator-slot). I would assume a single "operator" would imply one data type, but then again the whole "operator" concept is pretty hand-wavy anyways so maybe this is fine :)

inexorabletash: To be explicit, here's where I'm envisioning the conversion/validation happens:

- for `constant()` - immediate conversion based on explicitly passed `MLOperandDataType`
- for `MLOperand clamp()` - immediate conversion based on _input_'s data type
- for `MLActivation clamp()` - when the activation passed to an operand-vending method - see 
- for `pad()` - immediate conversion based on _input_'s data type

See https://github.com/webmachinelearning/webnn/pull/649#discussion_r1575166753 for a framework where activation validation could be plugged in.

philloooo: question:
Does this mean for example for `gemm`, if the input `a` and `b` are passed as float16, and if you pass 
`{alpha: MLNumber(3.14123452435)}`, it will automatically convert that to float16 as `3.141`?

inexorabletash: I guess under the hood - yes (although it'd be just `{alpha: 3.14123452435}`)

Presumably that's what the spec does today, even if we didn't introduce `MLNumber`?

index.bs
zolkis: > Multiple typedefs - e.g.
typedef (bigint or unrestricted double) MLNumber
typedef (bigint or double) MLFiniteNumber

I find that the most clear.

index.bs
philloooo: Under the hood it preserves the fp32 precision right now (as what you are suggesting). 
But I'd prefer we explicitly set the expectation that it will be casted to the precision to be the same as the input type.  
In the `gemm` example, since it's emulated as `alpha * matmul(a,b) + beta * c` in CoreML, and CoreML requires all these binary ops to have the same type. so if `a`, `b` are fp16, then it's better to cast `alpha` to fp16 to multiply them. Otherwise we need to cast input to fp32 (which doesn't get executed on ANE).

So right now we just don't support fp16 inputs for gemm on CoreML until this is sorted out.

index.bs
inexorabletash: Added MLFiniteNumber in a3ff823ab44ef44b128234c873a475134c0c5bf4

inexorabletash:
FYI, I updated this PR with changes for explicit casting (#678), and changing `float` inputs to `double` (#325)

index.bs
inexorabletash: So for https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/647.html#converttoint it sounds like we'd always want enforceRange = false, clamp = true ? (i.e. N < min &rarr; -&infin;, N > max &rarr; +&infin;)

And for https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/647.html#converttofloat we'd always want restricted = false? (i.e. NaN and +/- &infin; always pass through, N < min &rarr; -&infin;, N > max &rarr; +&infin;)

If that sound right, I can simplify the PR to remove those options.

anssiko:
Cross-linking this async update from today's agenda:

https://github.com/webmachinelearning/meetings/issues/24#issuecomment-2194957526

(Thanks @inexorabletash!)

index.bs
inexorabletash: PR updated in 6f75f2b7c041c09cd67fc416849e3c46cb7abbc2 to always clamp.

inexorabletash:
Okay - I think this is ready for a review. Probably best not to worry about the commit history (and definitely squash on merge, and edit down the comment!) - I can squash/force-push if desired.  

@a-sully has a work-in-progress CL for the Chromium prototype, and we'll want WPTs too.

index.bs
```diff
@@ -6014,6 +6061,85 @@ To <dfn data-lt="bidirectionally broadcasting the shapes">bidirectionally broadc
...
+1. Otherwise:
+    1. Let |lowerBound| be -2<sup>|bitLength| - 1</sup>.
+    1. Let |upperBound| be 2<sup>|bitLength| - 1</sup> - 1.
+1. If |x| is -0, then set |x| to +0.
+1. If |x| is NaN, then return +0.
+1. Set |x| to min(max(|x|, |lowerBound|), |upperBound|).
+1. Round |x| to the nearest integer, choosing the even integer if it lies halfway between two, and choosing +0 rather than -0.
+1. Return |x|.
+
+NOTE: This is based on a definition in [[WEBIDL]] with these differences: 64-bit integers are not treated specially, the input |x| is an abstract number, and clamping is always performed.
```
a-sully, line 6137: I can see clamping resulting in unexpected behavior from the developer's point of view, though the alternative of throwing an exception may not be desirable either ¯\\_(ツ)_/¯

Seems reasonable to start with clamping for now

a-sully (approved):
LGTM 👍

index.bs
```diff
@@ -6014,6 +6061,85 @@ To <dfn data-lt="bidirectionally broadcasting the shapes">bidirectionally broadc
...
+</div>
+
+<div algorithm>
+The <dfn>ConvertToInt</dfn>(|x|, |bitLength|, |signedness|) steps are:
+
+1. If |signedness| is "unsigned", then:
+    1. Let |lowerBound| be 0.
+    1. Let |upperBound| be 2<sup>|bitLength|</sup> - 1.
+1. Otherwise:
+    1. Let |lowerBound| be -2<sup>|bitLength| - 1</sup>.
```
fdwr, line 6129: > -2<sup>|bitLength| - 1</sup>

Huh, so I see some other mathy programs following this surprising behavior (surprising to me anyway), but this just seems so fundamentally wrong that some operator like multiplication or exponentiation should bind more tightly to a number than a number should bind to that number's *own sign*. Like, "-2" is an atomic entity (the negative sign is not some negating operator, nor is it subtraction in PEMDAS, but rather part of the identity of the numeric literal itself), and thus `-2 ^ 16` *should* mean `pow(-2, 16)`, and not `-pow(2, 16)`. Well, I don't have the energy to fight the historical faux pas of potentially multiple math books. So, mind just clarifying the order operations with some parens? :b

```suggestion
    1. Let |lowerBound| be -(2<sup>|bitLength| - 1</sup>).
```

fdwr (approved):
Great - thank you Joshua :). Just one question and one clarity request. (update: oops, found one more)

index.bs
```diff
@@ -4524,7 +4570,7 @@ enum MLPaddingMode {
 
 dictionary MLPadOptions {
   MLPaddingMode mode = "constant";
-  float value = 0;
+  MLFiniteNumber value = 0;
```
fdwr, line 4573: Padding should just permit any constant value through, consistently with `constant`, unless you encountered a good reason. People may want to fill the borders with infinity or even NaN, and once the MLNumber is cast to the tensor data type, the actual padding operation is just a dumb bit fill (no element math).

huningxin: +1 to be consistent with `constant`.

index.bs
```diff
@@ -6014,6 +6061,85 @@ To <dfn data-lt="bidirectionally broadcasting the shapes">bidirectionally broadc
 |shapeA| is <dfn>bidirectionally broadcastable</dfn> to |shapeB| if [=bidirectionally broadcasting the shapes=] |shapeA| and |shapeB| does not result in failure.
 </p>
 
+## Casting ## {#algorithms-casting}
+
+Explicit numeric casting is used in algorithms where parameters passed as {{MLNumber}}, {{MLFiniteNumber}}, or {{double}} need to be converted to match the {{MLOperandDataType}} of input or output {{MLOperand}}s.
```
huningxin, line 6066: I am wondering whether we can leverage or refer to ECMAScript numeric casting algorithms, say
- [`Math.fround()`](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-math.fround)
- [`Math.f16round()`](https://tc39.es/proposal-float16array/#sec-math.f16round)
- [`ToInt8()`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toint8)
- [`ToUint8()`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-touint8)
- [`ToInt32()`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-toint32)
- [`ToUint32()`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-touint32)
- [`ToBigInt64()`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tobigint64)
- [`ToBigUint64()`](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tobiguint64)

index.bs
inexorabletash: Done in a5f299d - lets us remove MLFiniteNumber entirely. \o/

index.bs
```diff
@@ -1246,7 +1262,7 @@ interface MLGraphBuilder {
   // Create an operand for a graph constant.
   MLOperand constant(MLOperandDescriptor descriptor, ArrayBufferView bufferView);
 
-  // Create a scalar operand from the specified number of the specified type.
+  // Create a single-value operand from the specified number of the specified type.
```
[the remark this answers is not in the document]
inexorabletash: Done in 2e06299 - not sure why I changed that.

index.bs
inexorabletash: @a-sully and I looked at this; the problem with those is that the input is a JS Number (i.e. float64) and they throw if anything else is passed (like a BigInt) due to their call to [ToNumber()](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tonumber). Referencing those algorithms is also somewhat wrong given the [spec processing model](https://garykac.github.io/procspec/#processing-model) where within algorithms we're no longer dealing with JS types. 

Ideally we'd work with WebIDL and export the algorithms defined there, but we can do that later and generally we add stuff to WebIDL only when its used in multiple specs.

index.bs
inexorabletash: Yeah, when I was initially crafting this I was thinking we'd want [EnforceRange](https://webidl.spec.whatwg.org/#EnforceRange) behavior - like we do for integer arguments - but that gets real weird real fast since doing the checks sometimes happens at activation fusion time, and the same code and constants might suddenly throw if you switch from using float32 to float16.

The places in the spec that still take integers and use EnforceRange still make sense, since you'd never want clamping (or worse, wrapping) behavior when specifying an axis, dimension, etc.

inexorabletash:
Since it's relevant here and related to switching pad() to match constant(): for the places that still use `double` - operand options for _epsilon_, _alpha_, and _beta_ - does it ever make sense to support Infinity, -Infinity or NaN? (i.e. should we switch to _unrestricted double_?)

I can't think of a case where those are ever useful, even if the behavior is well defined. Maybe leave it as `double` until someone complains?

index.bs
fdwr: > and they throw if anything else is passed (like a BigInt)

😢 I would love to see Javascript work more cleanly with 64-bit integers, as it's rather messy currently (compared to say C++ where casting between 64-bit integers and 64-bit floats works as expected) where casting in JS from 3.14159 to BigInt throws instead of just truncating o_o (meanwhile, the same language is happy to implicitly and silently cast between strings and numbers, and [sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) a list of numbers *actually* sorts your numbers *as if* they were strings 🙃).

fdwr:
> for the places that still use double - operand options for epsilon, alpha, and beta - does it ever make sense to support Infinity, -Infinity or NaN? (i.e. should we switch to unrestricted double?)

tldr: I don't feel strongly either way.

Does it make sense? No (especially not for epsilon). Though, I also *generally* feel we shouldn't unnecessarily restrict the API moreso than whatever the equivalent op decomposition would produce. So, if `linear` can be expressed as...

```
function linear(builder, input, options) {
  return builder.add(
    builder.mul(input, builder.constant(input.dataType(), options.alpha)),
    builder.constant(input.dataType(), options.beta));
}
```

...then it's nice to be consistent with the math it produces, NaN's and infinities alike; but then in practice, I don't see how these values would be useful, and so I'm fine with just `double` for now. 🤷

huningxin (approved):
LGTM!

fdwr (approved):
👍 Updates appreciated. Do you want a re-review from Phillis and Austin first before merging?

inexorabletash:
I think it's good to merge. Just remember to squash and tidy up the commit message!

Issue #645, opened 2024-04-18
Specify the behavior of sofplus when the steepness is 0


shiyi9801:
It's raised by @a-sully in [CL review](https://chromium-review.googlesource.com/c/chromium/src/+/5445810/3/third_party/blink/renderer/modules/ml/webnn/ml_graph_builder_test.cc#b2313), thanks!

Softplus calculates `ln(1 + exp(steepness * x)) / steepness`, when the steepness is 0 it might result in division by zero.

I tried Pytorch `torch.nn.Softplus(beta=0)` and the results are all inf, and [TF](https://www.tensorflow.org/api_docs/python/tf/math/softplus) and [ONNX](https://onnx.ai/onnx/operators/onnx__Softplus.html) don't have this attribute. Also [DirectML](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_activation_softplus_operator_desc) doesn't support `steepness < 1.0`.

huningxin:
Does negative `steepness` value make sense? `softplus` should produce positive results (as a smooth approximation to relu), but negative `steepness` would result in negative results.

fdwr:
The expected results of division by zero for floating point values is:

- 0 / 0 = indeterminate
- positiveValue / 0 = infinity
- negativeValue / 0 = -infinity

(unlike division by zero for integers, there's nothing ambiguous in the IEEE standard for floating point)

> Also [DirectML](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_activation_softplus_operator_desc) doesn't support steepness < 1.0.

It does now 😉. Coincidentally we relaxed DirectML's SOFTPLUS validation in February to permit < 1 (but that version is not out yet, and the docs are still valid for DML 1.13) which includes negative values, and even 0 for the sake of PyTorch and potentially WebNN.

> Does negative steepness value make sense?

🤔 I don't know the use case, but like you say, the graph is smooth, and PyTorch supports it without complaint:

```python
import torch

s = torch.nn.Softplus(beta=-1.0)
x = torch.tensor([0.5930860043, 0.9014285803, -0.6331304312, 0.4639878273], dtype=torch.float32)
y = s(x)

print("value:", y)
print("shape:", y.shape)
print("dtype:", y.dtype)

# value: tensor([-0.4399, -0.3407, -1.0590, -0.4878])
# shape: torch.Size([4])
# dtype: torch.float32
```

[...]

PR #649, opened 2024-04-18
Add axis argument to softmax()

### LINKED ISSUE #466: Softmax axis absent
(raised by @Honry in review https://github.com/microsoft/onnxruntime/pull/17665/files)

TF/PT/ONNX all take an `axis` parameter:

- [TensorFlow `tf.nn.softmax`](https://www.tensorflow.org/api_docs/python/tf/nn/softmax)
- [PyTorch `torch.nn.Softmax`](https://pytorch.org/docs/stable/generated/torch.nn.Softmax.html)
- [ONNX `Softmax-13`](https://onnx.ai/onnx/operators/onnx__Softmax.html)

...but [WebNN's `softmax`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-softmax-method) does not, making it challenging to implement a caller's softmax in terms of the function of the same name in WebNN. It *is* possible ([see here](https://github.com/microsoft/onnxruntime/pull/17714/files)) via bracketing transposes+reshapes around it, but the transpose+reshape contortions are unfortunate, and they could be more efficiently implemented in the backend rather than in each framework.

- ✅ Apple Metal Performance Shaders [softMax](https://developer.apple.com/documentation/metalperformanceshadersgraph/mpsgraph/3580387-softmax/) has an axis.
- ✅ [Apple MIL activation.softmax](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.activation.softmax) supports an axis.
- ✅ DirectML's [DML_ACTIVATION_SOFTMAX1_OPERATOR_DESC](https://learn.microsoft.com/en-us/windows/ai/directml/api/ns-directml-dml_activation_softmax1_operator_desc) supports an arbitrary axis list and dimensions, just like reduce. The older [DML_ACTIVATION_SOFTMAX_OPERATOR_DESC](https://learn.microsoft.com/en-us/windows/ai/directml/api/ns-directml-dml_activation_softmax_operator_desc) can achieve it via reshapes/transpose/strides.
- ☑ XNNPack - limited to 2D input currently. This can be achieved by updating XNNPack to accept an axis or by using the existing XNNPack operator plus a reshape (in the simple case when the axis is the last dimension) or transpose (if the axis comes before the last dimension).

So it's achievable in each backend, even without any changes to the DML/XNNPack API's, but it would move the pain from the caller down to where it can be handled efficiently.

------------------------

https://www.w3.org/TR/webnn/#api-mlgraphbuilder-softmax-method

```diff
partial interface MLGraphBuilder {
-  MLOperand softmax(MLOperand input);
+  MLOperand softmax(MLOperand input, unsigned long axis);
-  MLActivation softmax();
+  MLActivation softmax(unsigned long axis);
};
```

```diff
The behavior of this operation can be generically emulated from the usage of other operations as follow. However, user agents typically have a more efficient implementation for it, therefore its usage is encouraged from the performance standpoint.
// This sample deploys a well-known implementation trick [1] to compute the
// exponentials of the distances to the max value, instead of the exponentials
// of the input values itself, in order to increase the numerical stability of
// the result.
// [1]: https://cs231n.github.io/linear-classify/#softmax
- const maxX = builder.reduceMax(x, { axes: [1], keepDimensions: true });
+ const maxX = builder.reduceMax(x, { axes: [axis], keepDimensions: true });
const expX = builder.exp(builder.sub(x, maxX));
- return builder.div(expX, builder.reduceSum(expX, { axes: [1], keepDimensions: true }));
+ return builder.div(expX, builder.reduceSum(expX, { axes: [axis], keepDimensions: true }));
```

inexorabletash:
Frameworks (TensorFlow, PyTorch, ONNX) all accept an axis parameter.

Most backends also support an axis, or it can be emulated with a reshape. As @fdwr wrote: So it's achievable in each backend... but it would move the pain from the caller down to where it can be handled efficiently.

Fixes #466

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/649.html" title="Last updated on Apr 25, 2024, 12:31 AM UTC (b80b0b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/649/0df0296...inexorabletash:b80b0b8.html" title="Last updated on Apr 25, 2024, 12:31 AM UTC (b80b0b8)">Diff</a>


fdwr:
A tangential high-level consideration (cc @anssiko too) - as we make changes to existing operators, the divergence between the Chromium implementation and the spec will bite people. I wonder how we point people to specific frozen versions of the spec to say Chrome/Edge v3.14159 matches with spec v42? Otherwise they'll try `softmax` from WebNN with an `axis` parameter, and it fails. The spec is still a "living spec" candidate recommendation, not frozen, but it becomes increasingly challenging the older the operator is (the fillSequence operator is less impactful because it didn't exist yet anyway) and so clear messaging helps users.

It *appears* we have versions produced near daily (or as often as changes are made) like here https://www.w3.org/TR/2024/CRD-webnn-20240416/ with an alarmingly red disclaimer that the spec is red. So maybe we can point to those for a mapping.

inexorabletash:
> Thank you for championing my own issue in :).

Just trying to ~watch the world burn~ burn the issue list down. :wink: 

> ... the divergence between the Chromium implementation and the spec will bite people.

See #650 for a more severe case - it's going to be painful. That's one of the reasons I'm trying to pick off lots of issues - get the pain behind us, instead of in front of us. Tracking Canary vs. spec dates is probably too much work, but we could try and match the big releases (e.g. Chrome/Edge 124 mostly matches spec v20240101) ? I don't know if we can do that historically, but maybe that's work we take on as part of upcoming browser branch dates?

index.bs
```diff
@@ -5274,8 +5274,8 @@ Compute the [softmax](https://en.wikipedia.org/wiki/Softmax_function) values of
 the 2-D input tensor along axis 1.
 <script type=idl>
 partial interface MLGraphBuilder {
-  MLOperand softmax(MLOperand input);
-  MLActivation softmax();
+  MLOperand softmax(MLOperand input, unsigned long axis);
+  MLActivation softmax(unsigned long axis);
```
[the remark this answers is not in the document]
inexorabletash: I was wondering that too. It looks like frameworks either don't have a default, or default to -1 (i.e. last dimension). Would a default of 1 be surprising?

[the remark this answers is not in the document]
fdwr: Fair. Not having a default is safer to avoid transferrable knowledge surprises (e.g. TF defaults to -1), and we could always relax it later, but once added, it's hard to unrelax. I actually *prefer* being explicit here anyway. I was more thinking of transient compat.

huningxin:
> this impacts your [softmax implementation](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/webnn/builders/impl/softmax_op_builder.cc).

The ONNXRuntime WebNN EP softmax implementation also does the reshape N-D input to 2-D and reshape the 2-D output to N-D. Should WebNN softmax supports N-D input? It looks like [DirectML](https://learn.microsoft.com/en-us/windows/ai/directml/api/ns-directml-dml_activation_softmax1_operator_desc) and [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.activation.softmax) support N-D input.

index.bs
```diff
@@ -5301,27 +5301,28 @@ partial interface MLGraphBuilder {
...
         - an {{MLOperand}}. The output 2-D tensor that contains the softmax results, of the same shape as *input*.
 </div>
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLGraphBuilder>softmax(|input|)</dfn> method steps are:
+    The <dfn method for=MLGraphBuilder>softmax(|input|, |axis|)</dfn> method steps are:
   </summary>
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/rank=] is not 2, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 5315: The `axis` needs to be validated according to input's rank.

index.bs
```diff
@@ -5332,9 +5333,9 @@ partial interface MLGraphBuilder {
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLGraphBuilder id=softmax-noargs>softmax()</dfn> method steps are:
+    The <dfn method for=MLGraphBuilder>softmax(|axis|)</dfn> method steps are:
   </summary>
-    1. Let |op| be the result of [=creating an MLActivation=] given [=this=] and  "softmax".
+    1. Let |op| be the result of [=creating an MLActivation=] given [=this=], "softmax", and «[ "axis" → |axis| ]».
```
huningxin, line 5338: > «[ "axis" → |axis| ]»

Should we add this [ordered map](https://infra.spec.whatwg.org/#ordered-map) literal into Coding Conventions?

index.bs
```diff
@@ -640,7 +640,7 @@ The {{MLGraphBuilder}} interface serves as a builder (factory) to construct a [=
...
 
-A key part of the {{MLGraphBuilder}} interface are methods such as {{MLGraphBuilder/gemm()}} and {{MLGraphBuilder/softmax()}} which create an [=operator=] which represents the actual operation to perform on the input data when the computation is run, and return a new {{MLOperand}} or {{MLActivation}} holding the operator. Methods that create an {{MLOperand}} connect any [=operator/inputs=] and [=operator/activations=] to the operator. Each method invocation returns a distinct new value, without changing the value of any other {{MLOperand}}.
+A key part of the {{MLGraphBuilder}} interface are methods such as {{MLGraphBuilder/gemm()}} and {{MLGraphBuilder/hardSwish()}} which create an [=operator=] which represents the actual operation to perform on the input data when the computation is run, and return a new {{MLOperand}} or {{MLActivation}} holding the operator. Methods that create an {{MLOperand}} connect any [=operator/inputs=] and [=operator/activations=] to the operator. Each method invocation returns a distinct new value, without changing the value of any other {{MLOperand}}.
```
huningxin, line 643: Are there any specific reasons to change `softmax` to `hardSwish`?

huningxin (commented):
I cannot add inline comment, but "#### {{MLGraphBuilder/softmax(input)}} #### {#api-mlgraphbuilder-softmax-input}" at line 5300 needs to be updated with `axis` parameter.

index.bs
inexorabletash: Just `softmax()` no longer works as a link... but I'll make the change more minimal and just use explicit linking syntax. Reverted in 83349191d4adda232c13dc568b58493b1e96ceb5

index.bs
inexorabletash: Added for the operator version in 99e87734fa8daa7f69c95d18c8c075fc6ed0b592

For the activation version - would that be done synchronously when the activation is provided in a graph builder call, or at build time? Do we have other cases of activation validation?

inexorabletash:
> I cannot add inline comment, but "#### {{MLGraphBuilder/softmax(input)}} #### {#api-mlgraphbuilder-softmax-input}" at line 5300 needs to be updated with `axis` parameter.

Done in 66bf64a4ce6305cffb2920be626fdd8d4f22b273

index.bs
fdwr: All the information should be available at construction time (pre-`build`), since WebNN requires explicit shapes during construction.

> Do we have other cases of activation validation?

None come to mind, except possibly [softplus steepness](https://github.com/webmachinelearning/webnn/issues/645), but that's not impacted by the tensor shape constraints.

fdwr:
> The ONNXRuntime WebNN EP softmax implementation also does the reshape N-D input to 2-D and reshape the 2-D output to N-D. Should WebNN softmax supports N-D input? It looks like [DirectML](https://learn.microsoft.com/en-us/windows/ai/directml/api/ns-directml-dml_activation_softmax1_operator_desc) and [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.activation.softmax) support N-D input.

@huningxin: Does CoreML? I'm only seeing a `classcoremltools.converters.mil.mil.ops.defs.iOS15.activation.softmax` with an int32 axis. DirectML does, as axes are a more general superset of the single axis case, and since softmax is a `div` and `reduceSumExp` making it consistent with other reduction operators. However, while I like the elegance of axes, no other front-end or back-end I know of does so. So that's why I proposed a single axis.

index.bs
huningxin: > > For the activation version - would that be done synchronously when the activation is provided in a graph builder call,
> 
> All the information should be available at construction time (pre-`build`), since WebNN requires explicit shapes during construction.

Agreed. In particular, I suppose the validation should be done at the build method of operators that accept a fused activation function, like conv2d.

huningxin:
> Does CoreML? I'm only seeing a `classcoremltools.converters.mil.mil.ops.defs.iOS15.activation.softmax` with an int32 axis.

@fdwr , I meant whether the input should be N-D. I am fine with the single axis proposal.

fdwr:
@huningxin : I misread - yes, the input should be ND - I'll add a comment to that line.

index.bs
```diff
@@ -5294,38 +5294,40 @@ partial interface MLGraphBuilder {
...
         - an {{MLOperand}}. The output 2-D tensor that contains the softmax results, of the same shape as *input*.
 </div>
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLGraphBuilder>softmax(|input|)</dfn> method steps are:
+    The <dfn method for=MLGraphBuilder>softmax(|input|, |axis|)</dfn> method steps are:
   </summary>
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/rank=] is not 2, then [=exception/throw=] a {{TypeError}}.
```
fdwr, line 5319:
```suggestion
    1. If |input|'s [=MLOperand/rank=] is less than 1, then [=exception/throw=] a {{TypeError}}.
```

index.bs
```diff
@@ -5294,38 +5294,40 @@ partial interface MLGraphBuilder {
...
   </pre>
 </details>
 </div>
 
-#### {{MLGraphBuilder/softmax(input)}} #### {#api-mlgraphbuilder-softmax-input}
+#### {{MLGraphBuilder/softmax(input, axis)}} #### {#api-mlgraphbuilder-softmax-input-axis}
 <div>
     **Arguments:**
         - *input*: an {{MLOperand}}. The input 2-D tensor.
+        - *axis*: an {{unsigned long}} scalar. The dimension the softmax will be performed on.
```
fdwr, line 5308: 🤔 What do you think? The softmax operation as a whole is actually applied to all dimensions, not just one, but the reduction *aspect* of softmax is applied to only one. Should we say:

```suggestion
        - *axis*: an {{unsigned long}} scalar. The dimension the reduction will be performed on.
```

Or:
```suggestion
        - *axis*: an {{unsigned long}} scalar. The dimension the softmax reduction will be performed on.
```

No change, because I'm being pedantic:
```suggestion
        - *axis*: an {{unsigned long}} scalar. The dimension the softmax will be performed on.
```

index.bs
```diff
@@ -5294,38 +5294,40 @@ partial interface MLGraphBuilder {
     // of the input values itself, in order to increase the numerical stability of
     // the result.
     // [1]: https://cs231n.github.io/linear-classify/#softmax
-    const max_x = builder.reduceMax(x, { axes: [1], keepDimensions: true });
+    const max_x = builder.reduceMax(x, { axes: [axis], keepDimensions: true });
     const exp_x = builder.exp(builder.sub(x, max_x));
-    return builder.div(exp_x, builder.reduceSum(exp_x, { axes: [1], keepDimensions: true }));
+    return builder.div(exp_x, builder.reduceSum(exp_x, { axes: [axis], keepDimensions: true }));
```
fdwr, line 5299:
```suggestion
    const maxX = builder.reduceMax(x, { axes: [axis], keepDimensions: true });
    const expX = builder.exp(builder.sub(x, maxX));
    return builder.div(expX, builder.reduceSum(expX, { axes: [axis], keepDimensions: true }));
```
I don't know why just this place uses `snake_case` whereas the others use camelCase.

fdwr (commented):
3 suggestions, and rest looks good to me. Thanks Josh.

index.bs
inexorabletash: Done in 7f33accb64e89ba4c9ce41fce14ee0b7b120304a

index.bs
inexorabletash: Don't we need `> 0` in there, to preclude scalars?

index.bs
inexorabletash: Here's a sketch of what this could look like:

* In [create an `MLActivation`](https://webmachinelearning.github.io/webnn/#create-an-mlactivation), add optional _validation steps_ parameter, store the validation steps in an internal slot. The default validation steps are to return true.
* When creating softmax activation, pass these steps as the _validation steps_:
  1. If _axis_ is greater than or equal to _input_’s rank, then return false. 
  1. Otherwise, return true.
  * Note that if this were C++, this would be a lambda function, capturing _axis_. 
* In all 7 methods that take activation functions (singular or multiple), add steps like this:
  1. If _options_.activation exists, and running its _validation steps_ with _input_ returns false, then throw a `TypeError`.

 Thoughts:
* This could also be bundled into the [validate activation](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-validate-activation) steps - make those take an `MLOperand` in addition to or instead of `MLGraphBuilder`.
* Is passing a single `MLOperand` (i.e. _input_) sufficient for validation?

WDYT?

index.bs
fdwr: > Don't we need `> 0` in there, to preclude scalars?

The line below it says:

`1. If |axis| is greater than or equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.`

Scalars have a rank == 0. So passing axis == 0 will cause a `TypeError` since 0 >= 0.
We can add > 0 for redundancy 🤷.

inexorabletash: Oh, duh, thanks for reading the condition for me. Unnecessary line removed in 1877f2eb13c35a375c7ca86b112dc27ace7d9a4e

index.bs
huningxin: > * Is passing a single `MLOperand` (i.e. _input_) sufficient for validation?

I suppose it should pass _output_ MLOperand since the activation follows the operator that fuses it.

index.bs
```diff
@@ -5294,38 +5294,39 @@ partial interface MLGraphBuilder {
...
         - an {{MLOperand}}. The output 2-D tensor that contains the softmax results, of the same shape as *input*.
 </div>
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLGraphBuilder>softmax(|input|)</dfn> method steps are:
+    The <dfn method for=MLGraphBuilder>softmax(|input|, |axis|)</dfn> method steps are:
   </summary>
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If |input|'s [=MLOperand/rank=] is not 2, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 5318: Thanks for remove the 2-D restriction in the steps. I think the other "2-D tensor" need to be replaced by "N-D tensor" for the software text, including

> Compute the softmax values of the 2-D input tensor along axis 1.

The axis "1" also needs to be removed.

> input: an MLOperand. The input 2-D tensor.

> Returns: an MLOperand. The output 2-D tensor that contains the softmax results, of the same shape as input.

index.bs
inexorabletash: I guess we'll need to pass the output descriptor.

fdwr: > In particular, I suppose the validation should be done at the build method of operators that accept a fused activation function, like conv2d.

Yep, since stand-alone activations adopt whatever the size is of the operator they are joined to or used inside. So full validation must be deferred for these.

```
MLConv2dOptions::activation
MLConvTranspose2dOptions::activation
MLBatchNormalizationOptions::activation
MLGruOptions::activations
MLGruCellOptions::activations
MLLstmOptions::activations
MLLstmCellOptions::activations
```

> Is passing a single MLOperand (i.e. input) sufficient for validation?

Sounds fine, the MLActivation and at least either the MLOperand.shape or the MLOperand itself for more generality.

inexorabletash: Okay - 7981fa52a0cdceba56c5026ab2e1021451d35306 sketches out the validation. Notes:

* I went with passing an `MLOperandDescriptor`, because earlier we decided that we'd do all validation _before_ creating the operator and output `MLOperand` to pass. 
* The phrasing around creating/passing a lambda is based on [Fetch](https://fetch.spec.whatwg.org/#body-fully-read)
* The GRU/LSTM outputs are complicated. I'm not sure I'm passing the right thing, so please please double check.

fdwr (approved):
LGTM. TYJB. 👌

index.bs
```diff
@@ -3089,6 +3095,7 @@ partial interface MLGraphBuilder {
             1. Let |desc1| be a new {{MLOperandDescriptor}}.
             1. Set |desc1|.{{MLOperandDescriptor/dataType}} to |input|'s [=MLOperand/dataType=].
             1. Set |desc1|.{{MLOperandDescriptor/dimensions}} to the [=/list=] « |steps|, |numDirections|, |batchSize|, |hiddenSize| ».
+    1. If running the [=MLActivation/validation steps=] of any [=list/item=] in |options|.{{MLGruOptions/activations}} with |desc0| returns false, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 3098: I suppose this step should pass a descriptor with dimensions `[batchSize, hiddenSize]`, because this is the shape of the gates (update, new, reset) where the activations would be applied when running the gruCell().

index.bs
```diff
@@ -3997,6 +4005,7 @@ partial interface MLGraphBuilder {
             1. Let |desc2| be a new {{MLOperandDescriptor}}.
             1. Set |desc2|.{{MLOperandDescriptor/dataType}} to |input|'s [=MLOperand/dataType=].
             1. Set |desc2|.{{MLOperandDescriptor/dimensions}} to the [=/list=] « |steps|, |numDirections|, |batchSize|, |hiddenSize| ».
+        1. If running the [=MLActivation/validation steps=] of any [=list/item=] in |options|.{{MLLstmOptions/activations}} with |desc| returns false, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 4008: Same here, I suppose this step should pass a descriptor with dimensions `[batchSize, hiddenSize]`.

index.bs
inexorabletash: Done in 622ce7535c9a4ded50526a12e905b5bf75b36ecc

index.bs
inexorabletash: Done in 622ce7535c9a4ded50526a12e905b5bf75b36ecc

inexorabletash:
I noticed some related glitches when putting 622ce7535c9a4ded50526a12e905b5bf75b36ecc together - I'll factor those out into a separate PR

inexorabletash:
- Converting this to a draft since #655 should land first.

fdwr (approved):
👍

inexorabletash:
Okay, back to a real PR. Final review @huningxin ? Do we want to discuss this more in the WG before merging?

huningxin (approved):
LGTM

huningxin:
@inexorabletash 

> Okay, back to a real PR. Final review @huningxin ?

Thanks! Done.

> Do we want to discuss this more in the WG before merging?

I think we can merge it. Any thoughts @fdwr ?

fdwr:
> I think we can merge it. Any thoughts @fdwr ?

What are we discussing? It already looked good to me. So I'll merge it now (and then sleep -_-).

PR #656, opened 2024-04-25
Remove sequential filling overload of constant()

a-sully:
Discussion on #492 (specifically, this comment https://github.com/webmachinelearning/webnn/issues/492#issuecomment-2076061596) concluded that this overload was not needed. Its behavior can be emulated by filling a buffer in JavaScript to be passed in with the buffer-based `constant()` overload

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/656.html" title="Last updated on Apr 25, 2024, 4:33 PM UTC (36e73bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/656/d57e4ef...a-sully:36e73bc.html" title="Last updated on Apr 25, 2024, 4:33 PM UTC (36e73bc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1343,40 +1343,6 @@ Data truncation will occur when the specified value exceeds the range of the spe
     1. Return |operand|.
 </details>
 
-#### {{MLGraphBuilder/constant(start, end, step, type)}} #### {#api-mlgraphbuilder-constant-range}
-Create a constant {{MLOperand}} of the specified data type and shape that contains the data as specified by the range. 
-
-<div class="note">
-Data truncation will occur when the values in the range exceed the range of the specified output data type e.g. when a float value is assigned to an {{MLOperandDataType/"int8"}} data type, etc.
-</div>
-
-<div>
-    **Arguments:**
-        - *start*: a {{float}} scalar. The starting value of the range.
-        - *end*: a {{float}} scalar. The ending value of the range.
-        - *step*: a {{float}} scalar. The gap value between two data points in the range.
-        - *type*: an optional {{MLOperandDataType}}. If not specified, it is assumed to be {{MLOperandDataType/"float32"}}.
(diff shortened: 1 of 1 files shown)


fdwr:
I originally proposed adding it, but model analysis (~700 models, 19 of which have the op) shows the:
- (a) tiny tensors being generated, not worth the beefy GPU effort.
- (b) increasing sequences are usually used for shape computation which can actually make it worse when the sequential fill is done on the GPU only to be copied back to the CPU for shape computation (meaning it would be better to keep that computation on the CPU to begin with). These cases should be ultimately optimized away by constant folding after shape inference too.

Additionally this sidesteps any of the implementation difference issues we encountered between TF and PyTorch and validation ambiguities about infinities in the data - if the caller wants infinity in the buffer, they upload their buffer with infinity (and WebNN doesn't stop them).

This operator was not implemented yet in Chromium and is not used currently by known callers (e.g. [here](https://github.com/microsoft/onnxruntime/tree/main/onnxruntime/core/providers/webnn/builders/impl)).

fdwr (approved):
👍

fdwr:
@inexorabletash I can't find your comment, but removal of this overload entirely make moot its current absence in the IDL definition:

```
interface MLGraphBuilder {
  ...
  // Create an operand for a graph constant.
  MLOperand constant(MLOperandDescriptor descriptor, ArrayBufferView bufferView);

  // Create a single-value operand from the specified number of the specified type.
  MLOperand constant(double value, optional MLOperandDataType type = "float32");
  ...
};
```
https://www.w3.org/TR/webnn/#idl-index

@miaobin 👀

huningxin (approved):
LGTM, thanks!

a-sully:
Thanks for the reviews! Please merge at your convenience :)

Issue #375, opened 2023-04-05
Support for transformers


dontcallmedom:
While our [draft charter](https://www.w3.org/2023/03/proposed-webmachinelearning-charter.html) says that the group:
> priority on building blocks required by well-known model architectures such as recurrent neural network (RNN), long short-term memory (LSTM) and transformers

and while the first two are directly mentioned in WebNN, the latter aren't.

dani-lbnl:
It would be valuable to briefly state 2 main kinds of applications of transformers, namely for predictive and generative AI, and each one if not both will be included in the charter.

anssiko:
Support for transformers was discussed on today's call:
https://www.w3.org/2023/04/27-webmachinelearning-minutes.html#t11

The WG felt positive about the prospects of supporting transformers in WebNN and in accordance to the [contribution guidelines](https://github.com/webmachinelearning/webnn/blob/main/CONTRIBUTING.md#proposing-and-adding-a-new-operation) decided to start explore applicable use cases in this issue first, then moving to investigation of samples models, cross-framework support and cross-platform implementability.

anssiko:
@fdwr has been working on a Chromium WebNN prototype https://github.com/fdwr/chromium-src-webnn-dml/pull/1 to inform what additional operators are needed in WebNN to support a well-known generative AI model, Stable Diffusion. I expect this prototyping effort to help inform this discussion on use cases. I believe this prototype is WIP, so @fdwr feel free to drop a comment here when appropriate to share your findings.

@dani-lbnl do you have specific predictive or generative AI models in mind that are in use in your area of research? We could look into them more closely in a similar fashion.

anssiko:
We've discussed this topic on a few of our bi-weekly calls and so far proposed investigation paths include Stable Diffusion (informed by @fdwr's Chromium experiment), SegmentAnything (thanks @huningxin!), Transformers.js/HuggingFace's transformers. Want to propose something else? Drop a comment here.

We should use this issue to discuss the most promising use cases enabled by transformers that are a good fit to be run in the browser then decomp to see what new ops would be needed in WebNN. Based on this update the use cases in the spec as appropriate.

I'm proposing we try to identify a few key use cases first to keep the spec and implementation close to each other.

I'll try to keep this topic on our bi-weekly agenda so folks can also bring their input on the call.

anssiko:
@xenova may be able to provide insights from Transformers.js :-)

xenova:
[...]

PR #657, opened 2024-04-26
Data type table

inexorabletash:
- Variation on #646 - only batchNormalization is updated.


zolkis:
I'd prefer this style than the one in #646

inexorabletash:
A few questions to consider:

**dataType**
- For most "secondary" operands, the type is constrained to be the same as the primary operand. The relevant steps could be written as either:
  -  _"If mean’s dataType is not one of its allowed data types, then throw a TypeError."_ - more consistent.
  -  _"If mean’s dataType is not the same as input's dataType, then throw a TypeError."_ - clearer.

**rank**
- In most cases there's only one allowed rank. So do we prefer:
  - _"If mean’s rank is not its allowed rank, then throw a TypeError."_ - more consistent.   
  - _"If mean’s rank is not 1, then throw a TypeError."_ - clearer.
- If we prefer the former (reference the table), what about #659 which proposes simplifying some validation e.g. _"If mean’s shape is not equal to « input’s shape[options.axis] »..."_ &mdash; this implicitly validates the rank. If we had that, would we still want a step reading _"If mean’s rank is not ..."_ ?

**General**
- Should we add a _"validate the dataType and rank of an operand"_ helper algorithm? The steps for each operands could then be collapsed to just: _"If validating the dataType and rank of mean returns false, then throw a TypeError."_  - less text, but more is hidden behind the algorithm.
- When there are multiple input operands, do we prefer separate steps or combined steps? At an extreme we could even write:  _"If validating the dataType and rank of any of mean, variance, options.scale (if it exists) or options.bias (if it exists) returns false, then throw a TypeError."_

zolkis:
IMHO the "consistent" option is clear enough. :)

inexorabletash:
Other notes:

- For the op groups (element-wise binary/unary/logical, pooling, reduction) a table is not given and the inline steps are retained. Any better suggestions?
- Do we keep a rank validation step if there's a subsequent shape validation step? i.e. keep both of these lines:
    1. If mean’s rank is not its allowed rank, then throw a TypeError.
    1. If mean’s shape is not equal to « input’s shape[options.axis] », then throw a TypeError.

zolkis:
> For the op groups (element-wise binary/unary/logical, pooling, reduction) a table is not given and the inline steps are retained.

Looking at the preview, I think that makes sense.

> Do we keep a rank validation step if there's a subsequent shape validation step?  

Shape should include rank... I'd leave rank out and let impl optionally handle that as a quick check, if it makes sense somewhere.

fdwr:
>  I'm holding off until we're closer to specifying that, so we can avoid churn.

Understood.

>  I've been keeping this PR up-to-date

It will be beautiful.

inexorabletash:
Marking this as "ready for review" - it looks like it didn't get very stale, and `opSupportLimits()` doesn't seem to require integration as currently written (the actual behavior of that method isn't specified in detail). @fdwr & @huningxin - please take a look?

index.bs
```diff
@@ -5583,6 +6045,25 @@ partial dictionary MLOpSupportLimits {
...
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}</td>
+    <td>[=/any rank|N=]</td>
+  <tr>
+    <td>{{slope}}</td>
+    <td>[=/same as=] {{input}}</td>
+    <td>[=/any rank|N=]</td>
```
huningxin, line 6063: Should slope's rank in the range [0, input rank]? Slope's shape must be bidirectionally broadcastable to the shape of input.

huningxin (commented):
Thanks @inexorabletash !

It seems some operators are not covered, such as argMin/Max. Do you plan to do them in a separate CL?

inexorabletash:
> It seems some operators are not covered, such as argMin/Max. Do you plan to do them in a separate CL?

My approach was to very mechanically migrate explicit rank/shape validation steps from the prose steps that contain the constraints to the table format where the prose steps reference the table. That has these implications:

* If an op didn't have either explicit data type or rank validation steps, the table wasn't added since nothing would reference it. So e.g. argMin/Max didn't get a table.
* If an op's data type and/or rank was provided as parameters to the algorithm steps, the steps were not touched, and the table wasn't added for that op.
* If there wasn't an explicit step validating the data type (e.g. gather), the data type in the table is given as: any data type/any
* If there wasn't an explicit step validating the rank (e.g. gemm), or that validate the rank against something other than an explicit number (e.g. layerNormalization), the rank in the table is given as: any rank/N
    * Note that steps that validate the *shape* (e.g. where it must be broadcastable to something) are considered to not validate the rank.

In other words, only introduce the table where it will be normatively referenced.

Obviously we can change this! 

* We can include the table for all ops, which would have "any" / "N" for the allowed data types/ranks
    * argMin/argMax, cast, clamp, concat, element-wise binary, element-wise logical, expand, pad, reshape, slice, split, transpose
    * 🤔 Do we also add steps to the algorithm that reference the table, even though it's a no-op?
* We can add in tables for the cases where data type/rank are passed in to the algorithm
    * element-wise unary, averagePool2d, l2Pool2d, maxPool2d, reduceL1, reduceL2, reduceLogSum, reduceLogSumExp, reduceMax, reduceMean, reduceMin, reduceProduct, reduceSum, reduceSumSquare
    * 🤔 This implies we'd have a table per op (e.g. one for abs, one for ceil, one for cos, etc), and modify the algorithms to pass/take both data types and ranks. That's... a lot of tables.
* We can expand the definition of "allowed ranks" beyond "any" or an explicit number to include:
    * a reference to other parameters, e.g. "same as input"
    * a range
    * 🤔 Do we introduce the implied duplicate steps? e.g. for layerNormalization, do we have steps that validate rank against the table (i.e. scale is in [0, input rank]) and a specific value (i.e. scale is equal to options.axes's size) ?

> I am not sure whether it should point to the allowed data types table of batchNorm. It currently points to the allowed data types definition.

Yeah, I wasn't sure how much effort to put into explicitly linking things. Currently nothing actually links to the tables; it's implied that if a step mentions "allowed data types" or "allowed ranks" then the reader should find the nearby table and look up the appropriate value for the named input operand. It might be better to link those phrases in steps to the table cells.

huningxin:
@inexorabletash 

> In other words, only introduce the table where it will be normatively referenced.

Makes sense to me! Thanks for the explanation!

> We can include the table for all ops, which would have "any" / "N" for the allowed data types/ranks
> * argMin/argMax, cast, clamp, concat, element-wise binary, element-wise logical, expand, pad, reshape, slice, split, transpose

logicalNot may need steps to validate input data type is "uint8"? (and logicalAnd, logicalOr, logicalXor in @fdwr 's [wave3 proposal](https://github.com/webmachinelearning/webnn/issues/375#issuecomment-2292466613)) 

And wave3 will introduce "int4"/"uint4" data types, I assume "any" data types may not apply for some ops, we may need "anyDataTypesAtLeast8Bits" ([`kAllDataTypesAtLeast8bits`](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/public/cpp/supported_data_types.h;bpv=1;bpt=1;l=52?q=services%2Fwebnn%2Fpublic%2Fcpp%2Fsupported_data_types.h&ss=chromium&gsn=kAllDataTypesAtLeast8bits&gs=KYTHE%3A%2F%2Fkythe%3A%2F%2Fchromium.googlesource.com%2Fcodesearch%2Fchromium%2Fsrc%2F%2Fmain%3Flang%3Dc%252B%252B%3Fpath%3Dservices%2Fwebnn%2Fpublic%2Fcpp%2Fsupported_data_types.h%234sFWh36FjzCKZMdFiC1Lch16NjKcaMyzJjLSbqkUeXs) in Chromium prototype)

Of course we can handle the wave3 ops / data types after its PR landed.

> * 🤔 Do we also add steps to the algorithm that reference the table, even though it's a no-op?

I think it's unnecessary to add no-op steps.

> We can add in tables for the cases where data type/rank are passed in to the algorithm
> * element-wise unary, averagePool2d, l2Pool2d, maxPool2d, reduceL1, reduceL2, reduceLogSum, reduceLogSumExp, reduceMax, reduceMean, reduceMin, reduceProduct, reduceSum, reduceSumSquare
> * 🤔 This implies we'd have a table per op (e.g. one for abs, one for ceil, one for cos, etc), and modify the algorithms to pass/take both data types and ranks. That's... a lot of tables.

Yes, it's a lot. Can we group some tables? Like a table for float-pointing element-wise unary, including ceil, floor, cos, sin, erf, exp, log, reciprocal, tan etc.,. However it may require linking the "allowed data types" to a particular table which is another discussion.

> We can expand the definition of "allowed ranks" beyond "any" or an explicit number to include:
> * a reference to other parameters, e.g. "same as input"
> * a range
> * 🤔 Do we introduce the implied duplicate steps? e.g. for layerNormalization, do we have steps that validate rank against the table (i.e. scale is in [0, input rank]) and a specific value (i.e. scale is equal to options.axes's size) ?

Duplicating steps is not the intention. I just feel user should not expect validation failure if the supplied tensor is allowed according to "allowed ranks". We have "same as input" for "allowed data types". Could we have something like "maximum to input rank" for "allowed ranks"?

index.bs
```diff
@@ -1689,18 +1727,18 @@ partial dictionary MLOpSupportLimits {
   </summary>
     1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input|, |mean|, |variance|, |options|.{{MLBatchNormalizationOptions/scale}} (if it [=map/exists=]), and |options|.{{MLBatchNormalizationOptions/bias}} (if it [=map/exists=]) returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If |input|'s [=MLOperand/dataType=] is not {{MLOperandDataType/"float32"}} or {{MLOperandDataType/"float16"}}, then [=exception/throw=] a {{TypeError}}.
+    1. If |input|'s [=MLOperand/dataType=] is not one of its [=allowed data types=], then [=exception/throw=] a {{TypeError}}.
```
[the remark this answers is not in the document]
inexorabletash: 8d9b3a575422eb16c3d1857242c3fe46544e9a86 introduces links between these phrases and the appropriate table

inexorabletash:
> It seems some operators are not covered, such as argMin/Max. 

f61292d64b86ca4088d9a4e886d634ea65634fce adds tables for all ops (or op categories)

inexorabletash:
> > * 🤔 This implies we'd have a table per op (e.g. one for abs, one for ceil, one for cos, etc), and modify the algorithms to pass/take both data types and ranks. That's... a lot of tables.
> 
> Yes, it's a lot. Can we group some tables? Like a table for float-pointing element-wise unary, including ceil, floor, cos, sin, erf, exp, log, reciprocal, tan etc.,. However it may require linking the "allowed data types" to a particular table which is another discussion.

For now, in f61292d64b86ca4088d9a4e886d634ea65634fce I left it at one table per "op category" and introduced the phrase "specified as part of operation steps" in the table.

index.bs
```diff
@@ -3357,6 +3496,31 @@ partial dictionary MLOpSupportLimits {
...
+  </tr>
+  <tr>
+    <td>{{b}}</td>
+    <td>[=/same as=] {{a}}</td>
+    <td>2</td>
+  </tr>
+  <tr>
+    <td>{{MLGemmOptions/c}}</td>
+    <td>[=/same as=] {{a}}</td>
+    <td>[=/any rank|N=]</td>
```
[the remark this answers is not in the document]
inexorabletash: 4bf14f6504ba0587d33ee4439cc5e05f7f081085 updates allowed ranks to be a range, while requiring that the operation steps validate it explicitly

index.bs
```diff
@@ -4284,6 +4575,31 @@ partial dictionary MLOpSupportLimits {
...
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>[=/any rank|N=]</td>
+  </tr>
+  <tr>
+    <td>{{MLLayerNormalizationOptions/scale}}</td>
+    <td>{{input}}</td>
+    <td>[=/any rank|N=]</td>
```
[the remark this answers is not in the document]
inexorabletash: 4bf14f6504ba0587d33ee4439cc5e05f7f081085 updates allowed ranks to be a range, while requiring that the operation steps validate it explicitly

index.bs
```diff
@@ -4284,6 +4575,31 @@ partial dictionary MLOpSupportLimits {
...
+  </tr>
+  <tr>
+    <td>{{MLLayerNormalizationOptions/bias}}</td>
+    <td>{{input}}</td>
+    <td>[=/any rank|N=]</td>
```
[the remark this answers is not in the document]
inexorabletash: 4bf14f6504ba0587d33ee4439cc5e05f7f081085 updates allowed ranks to be a range, while requiring that the operation steps validate it explicitly

index.bs
inexorabletash: 4bf14f6504ba0587d33ee4439cc5e05f7f081085 updates allowed ranks to be a range... but given that input and slope are **bidirectionally** broadcast, can't _slope_'s rank be larger than _input_'s?

inexorabletash:
> I just feel user should not expect validation failure if the supplied tensor is allowed according to "allowed ranks".

Makes sense. Added in 4bf14f6504ba0587d33ee4439cc5e05f7f081085 for the cases you called out - I didn't audit the ops for more, though. How does that look?

index.bs
huningxin: > but given that input and slope are bidirectionally broadcast, can't slope's rank be larger than input's?

You are correct, slope's rank can be larger than input's. I made a mistake. The original "any rank" is fine. Apologize for that.

index.bs
```diff
@@ -2692,6 +2780,28 @@ partial dictionary MLOpSupportLimits {
     **Returns:** an {{MLOperand}}. The output tensor that contains the result of element-wise comparison of the two input tensors.
 </div>
 
+<table id=constraints-elementwise-logical class='data' link-for="MLGraphBuilder/equal(a, b, options), MLGraphBuilder/greater(a, b, options), MLGraphBuilder/greaterOrEqual(a, b, options), MLGraphBuilder/lesser(a, b, options), MLGraphBuilder/lesserOrEqual(a, b, options), MLGraphBuilder/logicalNot(a, options)">
```
huningxin, line 2783: `logicalNot` only allows "uint8" input data type. Would you like to create a separate table or say "specified as part of operation steps"?

index.bs
```diff
@@ -6124,6 +6314,22 @@ partial dictionary MLOpSupportLimits {
...
+  <thead>
+    <tr>
+      <th>input operand</th>
+      <th>[=/allowed data types=]</th>
+      <th>[=/allowed ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>[=/any data type|any=]</td>
```
huningxin, line 6328: "specified as part of operation steps"? For example `reduceLogSum` only takes floating-point input.

index.bs
inexorabletash: Oops - I meant to make that "specified...". Will fix.

index.bs
inexorabletash: Ditto!

index.bs
inexorabletash: Fixed in db8d1e21ef42e7897fd7143d41dea07194dee96c

index.bs
inexorabletash: Fixed in db8d1e21ef42e7897fd7143d41dea07194dee96c

inexorabletash:
As always, having the table contain "specified as part of operation steps" is an intentionally minimal change. There are a few more things we could do:

* Have separate tables; e.g. for element-wise logical we could have a generic table with "any" and a table specific to logicalNot; element-wise unary would need 3 (signed types, float types, and any for identity); pooling would need 2; reduction ops would need 3
* Keep a single table, but expand the text with casual developer-focused commentary, e.g. for element-wise logical it could read: _"specified as part of operation steps; most element-wise logical ops support any data type, but logicalNot() only supports uint8"_

huningxin (approved):
LGTM!

index.bs
```diff
@@ -5531,6 +6182,26 @@ partial dictionary MLOpSupportLimits {
...
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}</td>
+    <td>[=/any rank|N=]</td>
+  <tr>
+    <td>{{slope}}</td>
+    <td>[=/same as=] {{input}}</td>
+    <td>0 to {{input}}'s [=MLOperand/rank=]</td>
```
huningxin, line 6201: Sorry, could you please restore the change to "any rank"? As we discussed in https://github.com/webmachinelearning/webnn/pull/657#discussion_r1837478497

inexorabletash: done in 8e540f3b047ec16f49eee284c4686d59bd917925 !

huningxin (approved):
LGTM!

fdwr:
(should finish reviewing Thursday ⏳)

index.bs
```diff
@@ -2110,6 +2213,32 @@ partial dictionary MLOpSupportLimits {
...
+    <td>4</td>
+  </tr>
+  <tr>
+    <td>{{filter}}</td>
+    <td>[=/same as=] {{input}}</td>
+    <td>4</td>
+  </tr>
+  <tr>
+    <td>{{MLConv2dOptions/bias}}</td>
+    <td>{{input}}</td>
```
fdwr, line 2237:
```suggestion
    <td>[=/same as=] {{input}}</td>
```

index.bs
```diff
@@ -2322,6 +2451,32 @@ partial dictionary MLOpSupportLimits {
...
```
fdwr, line 2475:
```suggestion
    <td>[=/same as=] {{input}}</td>
```

fdwr (commented):
Thanks Josh! Minor comments, else LGTM.

index.bs
```diff
@@ -1761,6 +1816,22 @@ partial dictionary MLOpSupportLimits {
...
+      <th>input operand</th>
+      <th>[=/allowed data types=]</th>
+      <th>[=/allowed ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>[=/any data type|any=]</td>
+    <td>[=/any rank|N=]</td>
+  </tr>
```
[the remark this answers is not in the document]
inexorabletash: This makes sense to me, but can we do it in a follow-up PR? I'm expecting we might iterate on it a bit e.g. I think we should include it for every op, etc.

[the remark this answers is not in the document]
fdwr: Of course!

fdwr (approved):
👍

PR #663, opened 2024-04-29
Fix unidirectionally broadcast the shapes steps

### LINKED ISSUE #662: "Unidirectionally broadcast the shapes" steps have issues
The current [unidirectionally broadcasting the shapes](https://www.w3.org/TR/webnn/#unidirectionally-broadcasting-the-shapes) steps don't allow `sizeA` < `sizeB` according to the following step

> If sizeB > sizeA, then return failure.

And the steps should pad `shapeA`, the current steps pad `shapeB` instead

> Let paddedB be a [clone](https://infra.spec.whatwg.org/#list-clone) of shapeB.
> While paddedB’s [size](https://infra.spec.whatwg.org/#list-size) is less than sizeA, [prepend](https://infra.spec.whatwg.org/#list-prepend) 1 to paddedB.

This would cause the following examples to fail
1. ShapeA `[]`, ShapeB `[3, 4, 5]`
2. shapeA `[5]`, shapeB `[3, 4, 5]`

When creating the output shape, the steps should use `dimB`. The current step uses `dimA`:

> [Append](https://infra.spec.whatwg.org/#list-append) dimA to outputShape. 

This would produce incorrect output shape. For example, assume ShapeA `[3, 1, 5]`, ShapeB `[3, 4, 5]`, the output shape should be `[3, 4, 5]`, the current steps produce `[3, 1, 5]`.

huningxin:
Fix #662

***
<a href="https://pr-preview.s3.amazonaws.com/huningxin/webnn/pull/663.html" title="Last updated on Jun 5, 2024, 2:24 PM UTC (a3976ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/663/932b603...huningxin:a3976ae.html" title="Last updated on Jun 5, 2024, 2:24 PM UTC (a3976ae)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5882,26 +5882,26 @@ Broadcasting refers to how operations treat tensors with different shapes, and f
 
 <details open algorithm>
 <summary>
-To <dfn data-lt="unidirectionally broadcasting the shapes">unidirectionally broadcast the shapes</dfn> |shapeA| and |shapeB|, perform the following steps. |shapeA| and |shapeB| are [=/lists=] of positive integers, representing the dimensions of tensors, and the steps return a new [=/list=] of positive integers, or failure.
+To <dfn data-lt="unidirectionally broadcasting the shapes">unidirectionally broadcast the shapes</dfn> |shapeTo| and |shapeFrom|, perform the following steps. |shapeTo| and |shapeFrom| are [=/lists=] of positive integers, representing the dimensions of tensors, and the steps return a new [=/list=] of positive integers, or failure.
 </summary>
 
-1. Let |sizeA| be |shapeA|'s [=list/size=].
-1. Let |sizeB| be |shapeB|'s [=list/size=].
-1. If |sizeB| > |sizeA|, then return failure.
-1. Let |paddedB| be a [=list/clone=] of |shapeB|.
-1. While |paddedB|'s [=list/size=] is less than |sizeA|, [=list/prepend=] 1 to |paddedB|.
+1. Let |sizeTo| be |shapeTo|'s [=list/size=].
+1. Let |sizeFrom| be |shapeFrom|'s [=list/size=].
+1. If |sizeFrom| > |sizeTo|, then return failure.
+1. Let |paddedShapeFrom| be a [=list/clone=] of |shapeFrom|.
+1. While |paddedShapeFrom|'s [=list/size=] is less than |sizeTo|, [=list/prepend=] 1 to |paddedShapeFrom|.
 1. Let |outputShape| be a new [=/list=].
-1. [=list/For each=] |index| in [=the range=] 0 to |sizeA|, exclusive:
-    1. Let |dimA| be |shapeA|[|index|].
-    1. Let |dimB| be |paddedB|[|index|].
-    1. If |dimA| is not equal to |dimB| and |dimA| is not equal to 1, then return failure.
-    1. [=list/Append=] |dimA| to |outputShape|.
+1. [=list/For each=] |index| in [=the range=] 0 to |sizeTo|, exclusive:
+    1. Let |dimTo| be |shapeTo|[|index|].
+    1. Let |dimFrom| be |paddedShapeFrom|[|index|].
+    1. If |dimTo| is not equal to |dimFrom| and |dimFrom| is not equal to 1, then return failure.
+    1. [=list/Append=] |dimTo| to |outputShape|.
 1. Return |outputShape|.
 
 </details>
 
 <p algorithm>
-|shapeA| is <dfn>unidirectionally broadcastable</dfn> to |shapeB| if [=unidirectionally broadcasting the shapes=] |shapeA| and |shapeB| does not result in failure.
+|shapeFrom| is <dfn>unidirectionally broadcastable</dfn> to |shapeTo| if [=unidirectionally broadcasting the shapes=] |shapeTo| and |shapeFrom| does not result in failure.
 </p>
 
 <details open algorithm>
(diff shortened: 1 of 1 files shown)


fdwr:
The current definition is correct *given* the interpretation of the first/input tensor (A) being broadcasted to the second/target tensor (B), and the text elsewhere in the spec matches that interpretation. So for `expand`, the `inputShape` (A) is broadcasted to `newShape` (B), and for `prelu`, the `slope` (A) is broadcasted to the `input` (B) shape. Current text:

```
  MLOperand expand(MLOperand input, sequence<[EnforceRange] unsigned long> newShape);
};
...
Set outputDescriptor.dimensions to the result of unidirectionally broadcasting the shapes input’s shape and newShape.
...
  MLOperand prelu(MLOperand input, MLOperand slope);
};
...
Set descriptor.dimensions to the result of unidirectionally broadcasting the shapes slope’s shape and input’s shape.
```

So if we're going to flip the interpretation, then we'd need to adjust the wording in two places too:

```diff
- Set outputDescriptor.dimensions to the result of unidirectionally broadcasting the shapes input’s shape and newShape.
+ Set outputDescriptor.dimensions to the result of unidirectionally broadcasting the shapes newShape and input’s shape.
...
- Set descriptor.dimensions to the result of unidirectionally broadcasting the shapes slope’s shape and input’s shape.
+ Set descriptor.dimensions to the result of unidirectionally broadcasting the shapes input’s shape and slope’s shape.
```

I [see for ONNX](https://onnx.ai/onnx/repo-docs/Broadcasting.html#unidirectional-broadcasting), rather than the parameter order of (input -> output) or (original shape -> target shape), it lists parameters as (target shape <- original shape), which feels oddly backwards to me, but the bigger point of confusion is that `A` and `B` are simply unclear names, and we could pick some that are clearer. How about `originalShape` and `targetShape` or `oldShape` and `newShape` or another less ambiguous pair?

huningxin:
@fdwr 

> So if we're going to flip the interpretation, then we'd need to adjust the wording in two places too:

We also need to change ["unidirectionally broadcastable"](https://www.w3.org/TR/webnn/#unidirectionally-broadcastable):

```diff
-shapeA is unidirectionally broadcastable to shapeB if [unidirectionally broadcasting the shapes](https://www.w3.org/TR/webnn/#unidirectionally-broadcasting-the-shapes) shapeA and shapeB does not result in failure.
+shapeA is unidirectionally broadcastable to shapeB if [unidirectionally broadcasting the shapes](https://www.w3.org/TR/webnn/#unidirectionally-broadcasting-the-shapes) shapeB and shapeA does not result in failure.
```

And we need to change the step of ["unidirectionally broadcast the shapes"](https://www.w3.org/TR/webnn/#unidirectionally-broadcasting-the-shapes):

```diff
-If dimA is not equal to dimB and dimA is not equal to 1, then return failure.
+If dimB is not equal to dimA and dimB is not equal to 1, then return failure.
```

huningxin:
@fdwr 

> How about `originalShape` and `targetShape` or `oldShape` and `newShape` or another less ambiguous pair?

+1. I suppose this ambiguity comes from the readers cannot easily tell the broadcasting direction. How about using `fromShape` and `toShape`?

huningxin:
> The bigger point of confusion is that A and B are simply unclear names, and we could pick some that are clearer.

https://github.com/webmachinelearning/webnn/pull/663/commits/7f2c783c1225a2ed2632becc96479780b3585175 tries to use `shapeFrom` and `shapeTo`. Hope they are clearer. Please take another look. Thanks!

index.bs
```diff
@@ -5881,15 +5881,15 @@ To <dfn data-lt="unidirectionally broadcasting the shapes">unidirectionally broa
...
+1. Let |paddedA| be a [=list/clone=] of |shapeA|.
+1. While |paddedA|'s [=list/size=] is less than |sizeB|, [=list/prepend=] 1 to |paddedA|.
 1. Let |outputShape| be a new [=/list=].
-1. [=list/For each=] |index| in [=the range=] 0 to |sizeA|, exclusive:
-    1. Let |dimA| be |shapeA|[|index|].
-    1. Let |dimB| be |paddedB|[|index|].
+1. [=list/For each=] |index| in [=the range=] 0 to |sizeB|, exclusive:
+    1. Let |dimA| be |paddedA|[|index|].
+    1. Let |dimB| be |shapeB|[|index|].
     1. If |dimA| is not equal to |dimB| and |dimA| is not equal to 1, then return failure.
```
inexorabletash: I want to defer to @fdwr but call out:

If the claim _"[The current definition is correct given the interpretation of ...](https://github.com/webmachinelearning/webnn/pull/663#issuecomment-2081749493)"_ is correct, and this change is turning into a strict (1) reorder and (2) rename of the params, then it seems like this line needs to be altered to maintain the current behavior. 

Also, @fdwr's note about updating the "call sites" needs to happen as well?

inexorabletash:
Can we move forward here? I'd recommend starting from a fresh change (force-push) that just renames the parameters from `A` and `B` (or is that `B` and `A`?) to `shapeFrom` and `shapeTo` and then we can decide if we want to swap the order or make other changes.

fdwr (approved):
👍 Thanks Ningxin - it reads much clearer with `from` and `to` than `a` and `b` (ordering still feels backwards putting a `to` parameter before the `from`, but shrug). Merge at your convenience.

huningxin:
@fdwr 

> ordering still feels backwards putting a `to` parameter before the `from`

I also feel backwards. https://github.com/webmachinelearning/webnn/pull/663/commits/a3976aeb8eda7434d9b29decc78afda54797ca71 reverses the order of the two parameters. That actually is the same as my previous proposal: https://github.com/webmachinelearning/webnn/pull/663#issuecomment-2082885894.

inexorabletash (approved):
Thanks for walking through this patiently. LGTM!

fdwr (approved):
👍 Thanks N.

PR #665, opened 2024-04-29
Add note about no-op graphs

### LINKED ISSUE #614: Allow no-op graphs?
In #603 a step was added [build()](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-build) to match the Chromium implementation, which errors out if an input operand or a constant operands is specified as an output.

@fdwr [writes](https://github.com/webmachinelearning/webnn/pull/603#discussion_r1527099873) that this

> ...would nullify the possibility of a nop graph (say one with just a constant and output), which conceptually seems valid to me, as one scenario is where you have a chained sequence of models with one step that can be an action or a nop, and it may be easier to express by always passing a valid (but empty) graph to that component which executes them all.

But also :

> I'm not too worried about it either, since a caller can always insert a dummy `identity` node to satisfy this constraint.

Raising this [per WG telecon discussion](https://www.w3.org/2024/03/21-webmachinelearning-minutes.html)

inexorabletash:
Resolves #614

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/665.html" title="Last updated on Apr 29, 2024, 9:44 PM UTC (e2abe1c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/665/9f0f45a...inexorabletash:e2abe1c.html" title="Last updated on Apr 29, 2024, 9:44 PM UTC (e2abe1c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1390,6 +1390,8 @@ Build a composed graph up to a given output operand into a computational graph a
     1. Return |promise|.
 </details>
 
+NOTE: Specifying an [=computational graph/input=] operand or [=computational graph/constant=] operand as a graph {{MLGraphBuilder/build(outputs)/outputs|output}} results in an error, as this is usually an incorrect usage of the API. Callers can work around this by introducing an {{MLGraphBuilder/identity()}} operator.
+
 ### argMin/argMax operations ### {#api-mlgraphbuilder-argminmax}
 Return the index location of the minimum or maxmium values of all the input values along the axes.
 


inexorabletash:
Linking/wording changes welcome.

huningxin (approved):
LGTM

fdwr (approved):
👍

PR #672, opened 2024-05-02
Handful of algorithm and convention fixes

inexorabletash:
- Bug fix: Shapes should not be broadcast for logical "not" op.
- Bug fix: Validate transpose()'s permutation as a list, not operand.
- Conventions: Use "be" and "let" rather than "denote" and "store".
- Conventions: Use "otherwise" instead of "else" in steps.
- Conventions: Use "for each" not just "for" when iterating.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/672.html" title="Last updated on May 2, 2024, 8:10 PM UTC (8132deb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/672/aa5fac9...inexorabletash:8132deb.html" title="Last updated on May 2, 2024, 8:10 PM UTC (8132deb)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -855,7 +855,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
   <summary>
     To <dfn>execute graph</dfn>, given {{MLGraph}} |graph|, {{MLNamedArrayBufferViews}} |inputs| and {{MLNamedArrayBufferViews}} |outputs|, run the following steps. They return {{undefined}}, or an error.
   </summary>
-    1. Let |inputResources| denote the input resources of |graph|.{{MLGraph/[[implementation]]}}.
+    1. Let |inputResources| be the input resources of |graph|.{{MLGraph/[[implementation]]}}.
     1. [=map/For each=] |name| → |inputValue| of |inputs|:
         1. Let |inputDescriptor| be |graph|.{{MLGraph/[[inputDescriptors]]}}[|name|].
         1. Let |inputTensor| be a new tensor for |graph|.{{MLGraph/[[implementation]]}} as follows:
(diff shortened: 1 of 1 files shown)


inexorabletash:
Some minor fixes noted along the way. All editorial - either purely wording changes, or where the algorithm was clearly wrong but the intent was obvious. Please take a look @huningxin and @fdwr - thanks!

index.bs
```diff
@@ -866,7 +866,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
     1. [=map/For each=] |name| → |outputValue| of |outputs|:
         1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |name| and |inputResources| and wait for completion.
             1. If that returns an error, then return an "{{OperationError}}" {{DOMException}}.
-            1. Otherwise, store the result in |outputTensor|.
+            1. Otherwise, let |outputTensor| be the result.
```
fdwr, line 869: 🤔 We have a mix of various ways of *setting* variables, from an explicit "set" and "store" to passive voice "let".

- Set |descriptor|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
- Let |desc| be a new {{MLOperandDescriptor}} initialized to |input|.{{MLOperand/[[descriptor]]}}.
- Otherwise, store the result in |outputTensor|.
- Otherwise, let |outputTensor| be the result.

I suppose a "set" is used when something already exists, and a "let" is used when something new is being *declared* *and* set?

fdwr (approved):
- Bug fix: Shapes should not be broadcast for logical "not" op.
    👍 Indeed, `bitwiseNot` and `logicalNot` do *not* broadcast.
- Bug fix: Validate transpose()'s permutation as a list, not operand.
    👍 Good blunder fix.
- Conventions: Use "be" and "let" rather than "denote" and "store".
    👍 Sounds more natural.
- Conventions: Use "otherwise" instead of "else" in steps.
    🫤 Though I find "else" more natural/conventional in coding steps.
- Conventions: Use "for each" not just "for" when iterating.
    👍

index.bs
inexorabletash: Correct. Per [Infra](https://infra.spec.whatwg.org/#variables)

> A variable is declared with "let" and changed with "set".

I have a lint tool able to scrub for variables, and it ensures we're using either "let" or "for each" at least once per variable in each algorithm, to avoid "uninitialized" usage. With these changes, the spec is clean. I'll put up a "tools" PR soon.

inexorabletash:
And re: "else" vs. "otherwise" - [Infra talks about this](https://infra.spec.whatwg.org/#algorithm-conditional-statements), but we also used "otherwise" at least 2x as frequently as "else".

fdwr:
> And re: "else" vs. "otherwise" - [Infra talks about this](https://infra.spec.whatwg.org/#algorithm-conditional-statements), but we also used "otherwise" at least 2x as frequently as "else".

Ah yes, I remember those BASIC keywords well, `IF`, `THEN`, and `OTHERWISE` 😅... Alrighty, if our spec already uses `otherwise` 2x more than `else`, then consistency argument win.

index.bs
fdwr: ✅

huningxin (approved):
LGTM!

PR #674, opened 2024-05-03
Use consistent phrasing for operator creation

### LINKED ISSUE #572: Synchronously validate input operands/activations

inexorabletash:
Places that create ops use an inconsistent mixed of simple phrases, camelCase, Title Case, ACRONYMS, and "quoted strings". The most common was camelCase, but the wording can be weird, and the bulk-defined binary/unary/logical/pooling/reduction ops and activations use a "quotedCamelCase", so I went with that.

See https://github.com/webmachinelearning/webnn/pull/591#issuecomment-1970083381 for the most commentary.

Resolves #572

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/674.html" title="Last updated on May 7, 2024, 4:57 PM UTC (ca5cf52)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/674/b369a88...inexorabletash:ca5cf52.html" title="Last updated on May 7, 2024, 4:57 PM UTC (ca5cf52)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1377,7 +1377,7 @@ Build a composed graph up to a given output operand into a computational graph a
     1. [=set/For each=] |operand| in |inputs|:
         1. Set |graph|.{{MLGraph/[[inputDescriptors]]}}[|operand|.{{MLOperand/[[name]]}}] to |operand|.{{MLOperand/[[descriptor]]}}.
 
-            Issue(566): If {{MLGraphBuilder/constant(descriptor, bufferView)|constants'}} {{ArrayBuffer}}s are not [=transferred=], make copies for [=MLGraphBuilder/graph=]'s [=computational graph/constants=] here.
+            Issue(566): If {{MLGraphBuilder/constant(descriptor, bufferView)|constants'}} {{ArrayBuffer}}s are not [=ArrayBuffer/transferred=], make copies for [=MLGraphBuilder/graph=]'s [=computational graph/constants=] here.
 
     1. [=map/For each=] |name| → |operand| of |outputs|:
         1. Set |graph|.{{MLGraph/[[outputDescriptors]]}}[|name|] to |operand|.{{MLOperand/[[descriptor]]}}.


inexorabletash:
@zolkis can you take a look? Low priority.

zolkis (approved):
Looks good, thanks!

inexorabletash:
Thanks @zolkis !

@huningxin & @fdwr - could you please review and merge if it looks good? Thanks!

index.bs
```diff
@@ -4693,7 +4693,7 @@ partial interface MLGraphBuilder {
     1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
     1. *Make graph connections:*
         1. Let |output| be the result of [=creating an MLOperand=] given [=this=] and |desc|.
-        1. Let |operator| be an [=operator=] for the |op| pooling operation, given |options|.
+        1. Let |operator| be an [=operator=] for the |op| operation, given |options|.
```
fdwr, line 4696: Intentional? "the op operation" sounds funny (kinda like saying PIN number or ATM machine).

fdwr (approved):

index.bs
inexorabletash: Intentional, for consistency - everything ends up being `for the "string" operation` once variables are expanded. There are worse examples e.g. "realm _realm_". I had to make my lint tool's check for duplicate words look at the source not final text.

huningxin (approved):
LGTM, thanks!

Issue #666, opened 2024-04-30
Reconsider `MLOperand` methods


a-sully:
The [`MLOperand`](https://webmachinelearning.github.io/webnn/#mloperand) interface looks like:

```diff
interface MLOperand {
  MLOperandDataType dataType();
  sequence<unsigned long> shape();
};
```

Each of these methods meets the criteria for being an attribute (https://w3ctag.github.io/design-principles/#attributes-vs-methods), and are [const](https://crsrc.org/c/third_party/blink/renderer/modules/ml/webnn/ml_operand.h;l=114-116;drc=53c1d06fee5a3e98e7f265419f6f84df2c2c2d62) once the `MLOperand` is constructed. Is we are to have an accessor on `MLOperand` for each key in the `MLOperandDescriptor` which created it, then these should be `readonly attribute`s.

That being said, once we open this can of worms, there are several options we could choose from:

### 1. Getter which returns an `MLOperandDescriptor`

```diff
interface MLOperand {
-  MLOperandDataType dataType();
-  sequence<unsigned long> shape();
+  MLOperandDescriptor descriptor();
};
```

This is my preferred option. Unfortunately, JavaScript dictionaries are not read-only so (unless we want to push to introduce `FrozenDictionary`) we should use a getter.

### 2. Naive conversion to `readonly attribute`

```diff
interface MLOperand {
-  MLOperandDataType dataType();
+  readonly attribute MLOperandDataType dataType;
-  sequence<unsigned long> shape();
+  readonly attribute sequence<unsigned long> shape;
};
```

Seems reasonable. But why have an attribute called `shape` which refers to the same thing as `MLOperandDescriptor.dimensions`?

### 3. Rename `shape` to `dimensions`, or vice versa

```diff
dictionary MLOperandDescriptor {
  required MLOperandDataType dataType;
-  sequence<[EnforceRange] unsigned long> dimensions = [];
+  sequence<[EnforceRange] unsigned long> shape = [];
};

interface MLOperand {
-  MLOperandDataType dataType();
+  readonly attribute MLOperandDataType dataType;
-  sequence<unsigned long> shape();
+  readonly attribute sequence<unsigned long> shape;
};
```

`MLOperandDescriptor` takes a `dimensions` field, while querying this field from an `MLOperand` is done with the `shape()` method. As per https://w3ctag.github.io/design-principles/#attribute-reuse, we should re-use terms rather than unnecessarily introducing new vocabulary.

[...]

PR #676, opened 2024-05-03
Rename MLOperandDescriptor's "dimensions" key to "shape"

### LINKED ISSUE #669: Rename `MLOperandDescriptor`'s `dimensions` to `shape`
As discussed in #666: 

> `MLOperandDescriptor` takes a `dimensions` field, while querying this field from an `MLOperand` is done with the `shape()` method. As per https://w3ctag.github.io/design-principles/#attribute-reuse, we should re-use terms rather than unnecessarily introducing new vocabulary.
>
> A quick CTRL-F of the spec text shows:
> 
> * 139 instances of "dimensions", mostly due to `MLOperandDescriptor.dimensions`
> * 483 instances of "shape", which seems to be used more commonly in prose (e.g. "the shape of the input tensor")
> 
> We should decide to use one or the other (and possibly continue this discussion on another issue regardless)

"shape" is the more widely used term, so I propose we align on that. WDYT?

a-sully:
Fixes #669

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/676.html" title="Last updated on Sep 18, 2024, 4:29 AM UTC (870885a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/676/d7d0363...a-sully:870885a.html" title="Last updated on Sep 18, 2024, 4:29 AM UTC (870885a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -925,7 +925,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
         1. Let |inputDescriptor| be |graph|.{{MLGraph/[[inputDescriptors]]}}[|name|].
         1. Let |inputTensor| be a new tensor for |graph|.{{MLGraph/[[implementation]]}} as follows:
             1. Set the data type of |inputTensor| to the one that matches |inputValue|'s [=element type=].
-            1. Set the dimensions of |inputTensor| to |inputDescriptor|.{{MLOperandDescriptor/dimensions}}.
+            1. Set the shape of |inputTensor| to |inputDescriptor|.{{MLOperandDescriptor/shape}}.
             1. Set the values of elements in |inputTensor| to the values of elements in |inputValue|.
         1. Request the underlying implementation of |graph| to bind |inputResources|[|name|] to |inputTensor|.
     1. [=map/For each=] |name| → |outputValue| of |outputs|:
@@ -1011,7 +1011,7 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
   <pre highlight="js">
     const operandType = {
       dataType: 'float32',
-      dimensions: [2, 2]
+      shape: [2, 2]
     };
     const context = await navigator.ml.createContext();
     const builder = new MLGraphBuilder(context);
@@ -1095,7 +1095,7 @@ enum MLOperandDataType {
 
 dictionary MLOperandDescriptor {
   required MLOperandDataType dataType;
-  sequence<[EnforceRange] unsigned long> dimensions = [];
+  sequence<[EnforceRange] unsigned long> shape = [];
 };
 </script>
 
@@ -1103,17 +1103,17 @@ dictionary MLOperandDescriptor {
     : <dfn>dataType</dfn>
     :: The operand data type.
 
-    : <dfn>dimensions</dfn>
+    : <dfn>shape</dfn>
     :: The shape of the operand. It is empty for scalar operands, and non-empty for tensor operands.
 </dl>
 
 <details open algorithm>
   <summary>
-  To <dfn>create an {{MLOperandDescriptor}}</dfn> given {{MLOperandDataType}} |dataType| and [=/list=] |dimensions|, run the following steps:
+  To <dfn>create an {{MLOperandDescriptor}}</dfn> given {{MLOperandDataType}} |dataType| and [=/list=] |shape|, run the following steps:
   </summary>
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |dataType|.
-    1. Set |descriptor|.{{MLOperandDescriptor/dimensions}} to a [=list/clone=] of |dimensions|.
+    1. Set |descriptor|.{{MLOperandDescriptor/shape}} to a [=list/clone=] of |shape|.
     1. Return |descriptor|.
 </details>
 
@@ -1122,7 +1122,7 @@ dictionary MLOperandDescriptor {
     The <dfn for="MLOperandDescriptor">byte length</dfn> of an {{MLOperandDescriptor}} |desc| is the value returned by the following steps:
   </summary>
     1. Let |elementLength| be 1.
-    1. [=list/For each=] |dimension| of |desc|.{{MLOperandDescriptor/dimensions}}:
+    1. [=list/For each=] |dimension| of |desc|.{{MLOperandDescriptor/shape}}:
         1. Set |elementLength| to |elementLength| * |dimension|.
     1. Let |elementSize| be the [=element size=] of one of the {{ArrayBufferView}} types that matches |desc|.{{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility).
     1. Return |elementLength| * |elementSize|.
@@ -1138,8 +1138,8 @@ Issue(391): Should 0-size dimensions be supported?
   <summary>
     To <dfn for="MLOperandDescriptor">check dimensions</dfn> given {{MLOperandDescriptor}} |descriptor|, run the following steps:
   </summary>
-    1. If any element of |descriptor|.{{MLOperandDescriptor/dimensions}} is not a [=valid dimension=], return false.
-    1. If |descriptor|.{{MLOperandDescriptor/dimensions}}'s [=list/size=] is too large to be supported by the implementation, return false.
+    1. If any element of |descriptor|.{{MLOperandDescriptor/shape}} is not a [=valid dimension=], return false.
+    1. If |descriptor|.{{MLOperandDescriptor/shape}}'s [=list/size=] is too large to be supported by the implementation, return false.
 
         Issue(456): The maximum number of operand dimensions is not defined, but native ML APIs usually have a maximum supported size.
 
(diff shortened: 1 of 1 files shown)


a-sully:
> P.S. I'd like to follow up this change with https://github.com/webmachinelearning/webnn/issues/758 - I'll put up a PR shortly after this merges

P.P.S. I'd like to follow up that change with the changes proposed here https://github.com/webmachinelearning/webnn/issues/666#issuecomment-2342019337

fdwr:
Sorry for the delay, and I realize this is tedious work. @huningxin and I talked about this some last week, and most importantly, we want to be sure it's done in a way that smoothly transitions existing samples, ORT Web, and people using the former to work on drivers, in a way that doesn't cause hiccups by allowing a grace period with the old name.

My personal preference aside (because to me dimensions remains the grammatically clearer name, with any rogue places currently misusing dimensions as rank to be fixed with *rank*, and *shape* being the more ambiguous term - triangle? circle?), we all agree with *intra*-API consistency is the most important, and so `shape()` and `dimensions()` should be resolved. Intra-API consistency is also more important than consistency with other API's. Though, if we look around and *consider* those other API's, we see that *higher-level* Pythonish libraries (NumPy, TF, PT, MLX) use `shape`, whereas lower-level API's (XNNPACK dims, ANN dimensions, DML dimensionCount and sizes, MPS MPSNDArraySizes, CudNN CUDNN_ATTR_TENSOR_DIMENSIONS, OneDNN dnnl_dims_t) use dimensions/sizes. So, is WebNN a higher-level Python library, or a lower-level backend implementation?

reillyeon:
I don't want to spend too much time bikeshedding on this but my reasons for preferring "shape" are:

1. "shape" is shorter and easier to spell.
2. "dimensions" could mean either "shape" or "rank" depending on whether it's interpreted as a vector or a scalar.

zolkis:
(Thinking aloud). The way tensor definitions use these terms (for instance [here](https://dmol.pub/math/tensors-and-shapes.html)) is that `shape` combines the use of `rank` (the number of indices required to get individual elements of a tensor), `axis` (the composing indices in the rank) and `dimension`(the number or elements on an axis).

Scalar: rank 0, no axes, no dimension, shape: ().
Euclidean vector (x, y, z): rank 1, one axis with 3 dimensions (on the 0th axis). Shape: (3).
2D matrix of 5 rows and 4 columns: rank 2, i.e. 2 axes (0 and 1), with given dimensions on rows (axis 0) and columns (axis 1). Shape: (5, 4).

In this sense, `dimensions()` would be ambiguous and `shape()` would be more exact (contrary to intuition in English).

index.bs
```diff
@@ -1095,25 +1095,25 @@ enum MLOperandDataType {
...
 };
 </script>
 
 <dl dfn-type=dict-member dfn-for=MLOperandDescriptor>
     : <dfn>dataType</dfn>
     :: The operand data type.
 
-    : <dfn>dimensions</dfn>
+    : <dfn>shape</dfn>
     :: The shape of the operand. It is empty for scalar operands, and non-empty for tensor operands.
```
fdwr, line 1107: Can we define "shape" more clearly rather than a self-referential definition (e.g. "a cat is a cat")? Maybe:

```
    : <dfn>shape</dfn>
    :: The list of dimensions for the operand. It is empty for scalar operands, and non-empty for tensor operands.
```
or
```
    : <dfn>shape</dfn>
    :: The list of dimension sizes for the operand. It is empty for scalar operands, and non-empty for tensor operands.
```

Also existing, but it's funky that we say "non-empty for tensor operands" even though scalars *are* tensor operands, just zero order operands. Maybe just "It is empty for scalar tensors.".

index.bs
```diff
@@ -6128,7 +6128,7 @@ Given the following build graph:
     // Create MLOperandDescriptor object.
     const desc = {
       dataType: 'float32',
-      dimensions: TENSOR_DIMS
+      shape: TENSOR_DIMS
```
fdwr, line 6131: `TENSOR_DIMS` -> `TENSOR_SHAPE`

(or maybe better, `TENSOR_DIMS` -> `TENSOR_SIZES`, which has the nice benefit that `TENSOR_SIZES` applies to the multidimensional case, and the existing `TENSOR_SIZE` definition below it applies to the 1D case of linear elements, both of which equal the same product)

fdwr:
🤔 Have you considered `sizes`? It avoids the concern with `dimensions` (list of dimensions vs dimension count), it's equally short as `shape` and easy to spell, and it avoids the grammatical issues with `shape` ("... shape excludes information about the object's location, scale, orientation, and reflection... [unlike a] figure [which] is a representation including both shape and size..." [\[1\]](https://en.wikipedia.org/wiki/Shape)).

After collecting the pro's/con's above and elsewhere, and updating [my comment here](https://github.com/webmachinelearning/webnn/issues/666#issuecomment-2314544937), I see why this was difficult, because there was no clear winner in this toss-up (if there was, there wouldn't be any discussion ⚖️).

> Having a grace period where either dimensions or shape are accepted seems reasonable

Cool, that's really the most important aspect of this. I'm going to approve this because you're clearly intent your direction to avoid `dimensions`, and you're willing to do the breakage fixup work, but do consider `sizes`.

fdwr (approved):
👀

index.bs
a-sully: Updated to use `TENSOR_SHAPE` 👍 

> 🤔 Have you considered `sizes`?

Hmm I don't like `sizes` for the same reason I don't like `dimensions`:

> since "dimensions" - while referring to _one set of dimensions_ (i.e. one shape) - is plural, referring to a collection of dimensions is awkward and results in use of variable names like `dimensionsArray`, rather than just `shapes`

And since the "size" of an object generally refers to its _total size_, `TENSOR_SIZE` would become ambiguous. Is it referring to a single dimension (the size of a size), or the entire tensor (the size of the sizes)? 🙃

index.bs
a-sully: Ack. Updated to:
```
:: The list of dimensions of the operand. It is empty for scalar operands.
```

a-sully:
Thank you for the review! Please merge this at your convenience (the button is not available to me)

fdwr:
> Please merge this at your convenience

For stylistic and minor wording choices, I'm content to merge changes, but for bug fixes and content changes, I need my co-editor Ningxin's approval too.

index.bs
fdwr: > the size of the sizes?

ML libraries differ on many aspects, but I've never seen them diverge or be ambiguous on this. A tensor's size always means the total number of elements (just like you're used to with `std::vector`), never the extent of a single dimension. A tensor's *sizes* (in those libraries that use sizes) refers to the *list* of extents, with the total size being the product of sizes.

fdwr: \<ignorable-p-s\>Apparently the word `figure` means something's shape *and* size, but `tensor.figure` sounds kinda weird, and another word I just recalled (that we actually use inside DML a few places, but forgot) is the `extent`, "the degree to which something has spread; the size or scale of something". The `tensor.extent` doesn't suffer your plurality concern, and unlike shape, it actually means the sizes too, not just the ratio of the sizes. Maybe I'll use extent for my own ML toy library 😉. (I am reminded of The Giver's quote "Precision of language").\<ignorable-p-s\>

index.bs
```diff
@@ -2736,7 +2736,7 @@ partial interface MLGraphBuilder {
 </div>
 
 ### expand ### {#api-mlgraphbuilder-expand}
-Expand any dimension of size 1 of the input tensor to a larger size according to the new shape. The expansion is consistent with [[!numpy-broadcasting-rule]]. The input tensor must be [=unidirectionally broadcastable=] to the new shape; its dimensions must have the size of 1 or match the sizes of the corresponding output dimensions according to the new shape.
+Expand any dimension of size 1 of the input tensor to a larger size according to the new shape. The expansion is consistent with [[!numpy-broadcasting-rule]]. The input tensor must be [=unidirectionally broadcastable=] to the new shape; its shape must be of size 1 or match the sizes of the corresponding output dimensions according to the new shape.
```
huningxin, line 2739: Because shape is defined as the list of dimensions, the "shape must be of size 1" seems confusing, would it be read as `shape.length == 1`?

index.bs
```diff
@@ -3833,7 +3833,7 @@ partial interface MLGraphBuilder {
           builder.sub(input, mean), builder.constant(input.dataType(), 2)),
         reduceOptions);
 
-      // The scale and bias tensors are of the shape of the input dimensions
+      // The scale and bias tensors are of the shape of the input shape
```
huningxin, line 3836: Is the last "shape" redundant? 
```suggestion
      // The scale and bias tensors are of the shape of the input
```

huningxin (commented):
Thanks for putting this PR together. Just some minor clarification questions.

And could you also replace `dimensions` with `shape` in WebNN explainer: https://github.com/webmachinelearning/webnn/blob/main/explainer.md

index.bs
fdwr: ``diff
...its shape must be of size 1 or match...
...each dimension must be of size 1 or match...
```

index.bs
a-sully: Fixed 👍

a-sully:
> And could you also replace `dimensions` with `shape` in WebNN explainer: https://github.com/webmachinelearning/webnn/blob/main/explainer.md

Sure, I'll put up a follow-up PR shortly 👍

huningxin (approved):
LGTM!

PR #690, opened 2024-05-20
Add missing validation for pad(), slice(), and split()

inexorabletash:
Noticed during a review of the Chromium prototype. These are all pretty obvious except for slice() where there is subtlety for 0-size dimensions. I added an issue linking to #391 since the steps will need to be revised depending on how that issue is resolved.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/690.html" title="Last updated on May 21, 2024, 4:09 PM UTC (962ca97)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/690/5e7d8cd...inexorabletash:962ca97.html" title="Last updated on May 21, 2024, 4:09 PM UTC (962ca97)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4462,7 +4462,8 @@ partial interface MLGraphBuilder {
     The <dfn method for=MLGraphBuilder>pad(|input|, |beginningPadding|, |endingPadding|, |options|)</dfn> method steps are:
   </summary>
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If |beginningPadding|'s [=list/size=] and |endingPadding|'s [=list/size=] are not both equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a "{{TypeError}}".
+    1. If |input|'s [=MLOperand/rank=] is 0, then [=exception/throw=] a {{TypeError}}
+    1. If |beginningPadding|'s [=list/size=] and |endingPadding|'s [=list/size=] are not both equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be a copy of |input|.{{MLOperand/[[descriptor]]}}.
     1. Let |outputShape| be the result of [=MLGraphBuilder/calculating padding output sizes=] given |input|, |beginningPadding| and |endingPadding|.
     1. If any [=list/item=] in |outputShape| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


inexorabletash:
@huningxin - can you take a first look? Thanks!

index.bs
```diff
@@ -5539,10 +5547,12 @@ partial interface MLGraphBuilder {
   </summary>
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |axis| be |options|.{{MLSplitOptions/axis}}.
+    1. If |axis| is greater than or equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
     1. If |splits| is an {{unsigned long}}:
         1. If |input|'s [=MLOperand/shape=][|axis|] % |splits| is not 0, then [=exception/throw=] a {{TypeError}}.
         1. Otherwise, let |splitCount| be |splits|.
     1. If |splits| is a sequence of {{unsigned long}}:
+        1. If any of its elements is equal to 0, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 5555: Should we also mention #391 here?

index.bs
```diff
@@ -5294,6 +5295,13 @@ partial interface MLGraphBuilder {
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If any of |sizes|'s [=list/items=] are 0, then [=exception/throw=] a {{TypeError}}.
     1. If |starts|'s [=list/size=] and |sizes|'s [=list/size=] are not both equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
+    1. [=list/For each=] |index| in [=the range=] 0 to |input|'s [=MLOperand/rank=], exclusive:
+        1. If |sizes|[|index|] is 0, then [=exception/throw=] a {{TypeError}}.
+
+            Issue(391): If 0-size dimensions are allowed, revise these steps.
+
+        1. If |starts|[|index|] is greater than or equal to |input|'s [=MLOperand/shape=][|index|], then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 5303: If the 0-size dimensions are allowed, would it be allowed that the |starts|[|index|] is equal to |input|'s [=MLOperand/shape=][|index|]? Or is the next validation step good enough?

inexorabletash: I _think_ we'd want to allow `starts[index]` equal to `input's shape[index]` so the resulting slice is 0, so yeah this step would need to be revised too. The note is written to be pretty generic and mentioned "these steps" - were you wanting something more here?

index.bs
inexorabletash: Makes sense - I'll add a note here too.

inexorabletash: Added in 962ca97be0adc271367cce16c2ff8867edad9c9b

huningxin (approved):
LGTM, thanks!

index.bs
huningxin: > The note is written to be pretty generic and mentioned "these steps"

Makes sense, it is good for now.

huningxin:
@fdwr is OOO this week and might not have bandwidth to review this PR. 

I think it's straightforward and could be merged.

@inexorabletash , WDYT?

index.bs
fdwr, line 5303: Is this one technically redundant because the line below will validate it?
Otherwise if we keep this line, then the same issue 391 applies to it too, as it starts[index] == shape[index] is okay given a 0 size.

fdwr (approved):
👍 Thanks as always Joshua.

index.bs
fdwr: Shrug, I wager when we update line 5299, we'll notice this line too. So, let's just merge it now.

inexorabletash: Agreed that it's technically redundant, but I thought we'd revise everything after 391 and in that case we would want to consider the validation carefully.  Thanks for the merge!

Issue #623, opened 2024-03-27
WebNN should support NPU and QDQ operations


wchao1115:
Related to issue #128 and #302, we've been talking about supporting the NPU for the last few years. Now that more commercial NPU platforms become available (e.g. with the more recent arrival of Intel Core Ultra NPU), it is time to formally define NPU support in the WebNN spec. There are two key elements of this specification:
1. *An ability to specify a device type for the NPU*. Unlike more general-purpose devices such as the GPU and CPU, an NPU supports a limited finite set of operations without programmability support. To an extent needed to keep model execution stable and more predictable, the notion of a *fallback device* is needed to support NPU acceleration during model inference. 
2. *A minimum set of operators required to support quantized models*. Because most NPU utilizes a much simpler and less power-hungry low-bit integer arithmetic units, models targeting the NPU almost always need to be quantized first. The bare minimal support here in terms of operators are just two -- the `quantizeLinear` and `dequantizeLinear` operators. These two will be enough to handle quantized models by pairing them up at the right places in the model graph, the so-called tensor-oriented [*QDQ*](https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html#onnx-quantization-representation-format) format used in ONNX. Additionally, two more prominent quantized operators, one for convolution, and another for matmul will allow more quantized models not already expressed in the QDQ format to function i.e. `conv2dInt` and `matmulInt`.

anssiko:
@wchao1115, thanks for this proposal that outlines key elements for NPU support. I'll schedule this important topic for discussion at our upcoming meeting.

As you noted, these topics (NPU device type, support for quantized models) have been explored in the group prior and have been awaiting implementation experience. The timing is now appropriate for the group to reinvigorate this topic with NPU platforms more widely available and in the hands of consumers. Most importantly, the group can now validate proposed spec designs with implementation experience per our established work mode.

I'm looking forward to this discussion. Meanwhile, questions and comments are welcome in this issue from everyone.

philloooo:
Hi thanks for bringing this up!  I'd like to highlight a couple things based on my current implementation experience:
1. I don't think we can assume CPU and GPU device type always support everything, and only NPU is the outlier. For chromium implementations, tflite has a [subset](https://www.tensorflow.org/lite/performance/gpu#supported_ops) of ops supported for GPU backend. On CoreML, also opset coverage is like CPU > GPU > ANE. So if we want to provide fallback, we will need that for both GPU and NPU.
[...]

PR #696, opened 2024-05-30
Add MLDeviceType npu

fdwr:
This change only adds the `MLDeviceType` *enum* (currently [implemented in Chromium-based browsers](https://chromium-review.googlesource.com/c/chromium/src/+/5330647) for early testing) and not yet quantize/dequantize operators, as those are valuable but technically orthogonal given they are *also* useful for GPU and given the enum alone is still useful for non-quantized models using float16 weights.

Of the 4 proposals from https://github.com/webmachinelearning/webnn/issues/623, we're starting with the simplest \#‌1 (simple enum with system-decided fallback), and if additional experience warrants more complexity, we'll re-evaluate the other options later.

Wording recommendations for device fallback are welcome. Neural processing units are novel compared with more generic compute devices like CPU's/GPU's, given their limited operator support and inability to execute the *entire* graph at once.

***
<a href="https://pr-preview.s3.amazonaws.com/fdwr/webnn/pull/696.html" title="Last updated on Jun 17, 2024, 9:09 PM UTC (3b83466)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/696/c00560f...fdwr:3b83466.html" title="Last updated on Jun 17, 2024, 9:09 PM UTC (3b83466)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -960,7 +960,7 @@ The <dfn>device type</dfn> indicates the kind of device used for the context. It
 <dt>"<dfn enum-value>gpu</dfn>"</dt>
 <dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations.</dd>
 <dt>"<dfn enum-value>npu</dfn>"</dt>
-<dd>Provides narrower functionality on dedicated machine learning devices which can perform between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results. The backend implementation may fall back to other devices for certain operators and parts of the graph.</dd>
+<dd>Provides narrower functionality on dedicated machine learning devices which perform between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results. The backend implementation may fall back to other devices for certain operators and parts of the graph.</dd>
 </dl>
 
 The <dfn>power preference</dfn> indicates preference as related to power consumption. It is one of the following:
--- a/index.bs
+++ b/index.bs
@@ -898,7 +898,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 </details>
 
 ### {{MLContext/compute()}}  ### {#api-mlcontext-compute}
-Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it [=exception/throws=] an "{{OperationError}}" {{DOMException}}.
+Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU/NPU timeline for submitting a workload onto the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it [=exception/throws=] an "{{OperationError}}" {{DOMException}}.
 
 <div class="note">
 In accordance with the [=ArrayBufferView/write|Web IDL warning=], to prevent the calling thread from modifying the input and output resources while the computation is ongoing, this method [=MLNamedArrayBufferViews/transfer|transfers=] the input and output {{MLNamedArrayBufferViews}} to new views that share the same backing memory allocations. The transferred views are returned to the caller via the promise fulfillment with the computation result written into the backing memory of the output views.
--- a/index.bs
+++ b/index.bs
@@ -960,7 +960,7 @@ The <dfn>device type</dfn> indicates the kind of device used for the context. It
 <dt>"<dfn enum-value>gpu</dfn>"</dt>
 <dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations.</dd>
 <dt>"<dfn enum-value>npu</dfn>"</dt>
-<dd>Provides narrower functionality on dedicated machine learning devices which can have performance in between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results.</dd>
+<dd>Provides narrower functionality on dedicated machine learning devices which can perform between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results.</dd>
 </dl>
 
 The <dfn>power preference</dfn> indicates preference as related to power consumption. It is one of the following:
--- a/index.bs
+++ b/index.bs
@@ -960,7 +960,7 @@ The <dfn>device type</dfn> indicates the kind of device used for the context. It
 <dt>"<dfn enum-value>gpu</dfn>"</dt>
 <dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations.</dd>
 <dt>"<dfn enum-value>npu</dfn>"</dt>
-<dd>Provides narrower functionality on dedicated machine learning devices which can perform between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results.</dd>
+<dd>Provides narrower functionality on dedicated machine learning devices which can perform between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results. The backend implementation may fall back to other devices for certain operators and parts of the graph.</dd>
 </dl>
 
 The <dfn>power preference</dfn> indicates preference as related to power consumption. It is one of the following:
--- a/index.bs
+++ b/index.bs
@@ -672,7 +672,7 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring; 
 
 The WebGPU API identifies <a href="https://gpuweb.github.io/gpuweb/#privacy-machine-artifacts">machine-specific artifacts</a> as a privacy consideration. Similarly, the WebNN API's compute unit scheduling may under certain circumstances introduce a fingerprint. However, similarly to WebGPU, such fingerprints are identical across most or all of the devices of each vendor, mitigating the concern. Furthermore, software implementations can be used to further eliminate such artifacts.
 
-The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Device type=] normatively indicates the kind of device and is either {{MLDeviceType/"cpu"}} or {{MLDeviceType/"gpu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
+The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Device type=] normatively indicates the kind of device is one of: {{MLDeviceType/"cpu"}}, {{MLDeviceType/"gpu"}}, {{MLDeviceType/"npu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, and thus this type can in some cases add two bits of entropy to the fingerprint. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
 
 If a future version of this specification introduces support for new a  [=device type=] that can only support a subset of {{MLOperandDataType}}s, that may introduce a new fingerprint.
 
@@ -782,7 +782,8 @@ WorkerNavigator includes NavigatorML;
 <script type=idl>
 enum MLDeviceType {
   "cpu",
-  "gpu"
+  "gpu",
+  "npu"
 };
 
 enum MLPowerPreference {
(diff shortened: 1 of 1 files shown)


fdwr:
FYI @mingmingtasd.

index.bs
```diff
@@ -958,6 +959,8 @@ The <dfn>device type</dfn> indicates the kind of device used for the context. It
 <dd>Provides the broadest compatibility and usability across all client devices with varying degrees of performance.</dd>
 <dt>"<dfn enum-value>gpu</dfn>"</dt>
 <dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations.</dd>
+<dt>"<dfn enum-value>npu</dfn>"</dt>
+<dd>Provides narrower functionality on dedicated machine learning devices which perform between CPU and GPU, favoring other factors over raw performance like lower power usage for longer battery life or lower latency for results. The backend implementation may fall back to other devices for certain operators and parts of the graph.</dd>
```
a-sully, line 963: > The backend implementation may fall back to other devices for certain operators and parts of the graph.

That this may be true of the "gpu" device type, too, as @philloooo pointed out in https://github.com/webmachinelearning/webnn/issues/623#issuecomment-2025767956

index.bs
```diff
@@ -672,7 +672,7 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring;
...
+The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. [=Device type=] normatively indicates the kind of device is one of: {{MLDeviceType/"cpu"}}, {{MLDeviceType/"gpu"}}, {{MLDeviceType/"npu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, and thus this type can in some cases add two bits of entropy to the fingerprint. [=Power preference=] indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
```
inexorabletash, line 675: This is based on a quite old (Feb 8) commit, and there's a merge conflict affecting the _"The WebNN API..."_ paragraph. With that fixed locally, my lint tool doesn't detect any errors. 🥳

inexorabletash (changes requested):
The definition of [compute()](https://webmachinelearning.github.io/webnn/#api-mlcontext-compute) in the tip-of-tree version of the spec says:

> ... either on a worker thread for the CPU execution, or on a GPU timeline for the submission of GPU workload on the command queue.

That sentence implies either CPU or GPU, and should be rewritten to accommodate NPU support.

index.bs
huningxin, line 963: I don't quite get the meaning of "perform between CPU and GPU". Does it refer to its power and performance characteristic?

anssiko: Here's a proposal that avoids mentioning CPU and GPU. Feel free to use it, or parts of it, if this sounds reasonable:

"Provides power efficiency for sustained workloads across hardware platforms with purpose-built accelerators. The backend implementation may fall back to other devices for certain operators and parts of the graph."

My expectation is that this descriptive text will receive updates in the future when more products of this emerging category become available. That is to say, I'd expect us to put in here text that reflects the group's current best understanding.

index.bs
anssiko: @fdwr can you push an update to resolve this conflict?

And thanks @inexorabletash for checking this with your lint tool. I'd like to test it when you're ready to share it with the world :-)

index.bs
zolkis: > may fall back to other devices

I think the problem was not the mentioning the CPU and GPU, but the expression of what happens ("perfom between").
The `other devices` might be misleading, should we rather use the same words (in your suggested text, "other accelerators"), eventually "other accelerators, e.g. CPU or GPU"?

index.bs
```diff
@@ -782,7 +782,8 @@ WorkerNavigator includes NavigatorML;
 <script type=idl>
 enum MLDeviceType {
   "cpu",
-  "gpu"
+  "gpu",
+  "npu"
 };
```
zolkis, line 787: I wonder if it makes sense to swap the order, also depicting a possible default fallback path according to [comment](https://github.com/webmachinelearning/webnn/issues/623#issuecomment-2165827072) (eventually).

index.bs
zolkis: Would it make sense to swap "gpu" and "npu", and allow backwards fallback (from GPU to NPU or CPU, and from NPU to CPU). Then, "gpu" would mean that the implementation can choose the device and the fallback order. AFAICT, this supports the current use cases. 
If we want an error if the given device cannot be used, then we might need to add manual fallback options later.

fdwr, line 675: @anssiko also proposed adding:

""ISSUE: MLContextOptions is under active development and the design is expected to change informed by further implementation experience and new use cases from the wider web community. The Working Group is considering additional API surface to allow definition of a fallback device, multiple devices in a preferred order, or an exclusion of a specific device. Other considerations under discussion include error handling, ultimate
fallback and quantized operators. Feedback is welcome on any of these design considerations from web developers, library authors, OS and hardware vendors, and other stakeholders via GitHub issue [#623](https://github.com/webmachinelearning/webnn/issues/623)."

anssiko: Here's the Bikeshed syntax to create such an in-line issue block:
```
Issue(623): MLContextOptions is under active development and the design is expected to change informed by further implementation experience and new use cases from the wider web community. The Working Group is considering additional API surface to allow definition of a fallback device, multiple devices in a preferred order, or an exclusion of a specific device. Other considerations under discussion include error handling, ultimate
fallback and quantized operators. Feedback is welcome on any of these design considerations from web developers, library authors, OS and hardware vendors, and other stakeholders via GitHub:
```
Bikeshed magic appends the following link at the end of the block (thus trailing ":"):
[[Issue #623]](https://github.com/webmachinelearning/webnn/issues/623)

If we want more control over formatting or where the link is placed, we can use a plain old `<a>`.

I'd propose to place this block as the first thing in this section: https://www.w3.org/TR/webnn/#api-mlcontextoptions

anssiko (approved):
Per WG discussion https://www.w3.org/2024/06/13-webmachinelearning-minutes.html#t06 we agreed to merge this PR #696 after adding the note to MLContextOptions section. Wording in that note can be adjusted to ensure all perspectives discussed are considered.

index.bs
fdwr: If you mean ...

```
enum MLDeviceType {
  "cpu",
  "npu",
  "gpu"
};
```

...I'm *okay* with reordering them, but this enum just specifies the primary preferred device with no connection to fallback. If we want to add more control over fallback later, per one of the other options [listed here](https://github.com/webmachinelearning/webnn/issues/623#issuecomment-2063954107), that would be a separate attribute.

These devices have differing levels of generality:

- CPU - most general (can handle anything you throw it at, albeit not as quickly as others)
- GPU - less general (cannot handle branchy operators well or operators that dynamically allocate output memory, but tis very good with trivially parallelizable data)
- NPU - least general (more fixed function pipelines, less memory than either GPU or CPU...)

And different performance characteristics:

- CPU - slowest throughput
- NPU - faster throughput than CPU for parallelizable operators (especially convolution/GEMM) and less battery usage than the equivalent CPU or GPU execution
- GPU - fastest throughput but higher power usage, and usage can interfere with other GPU tasks like rendering

And different fallback characteristics (where fallback means you cannot execute all tasks on the given preferred device because of limitations like operator support or data type support):

- CPU - since the CPU is the most generic and can handle everything thrown at it, there is no need to "fallback" down to another device.
- GPU - unsupported operators or data types may fall back to CPU. "falling back" from GPU to NPU doesn't generally make sense because a GPU is a more generic superset to an NPU anyway, and if you've already ramped up the GPU for execution, then you might as well keep execution on it and avoid memory copies between devices.
- NPU - fallback to either GPU or CPU can make sense, especially if the NPU is on the same die and shares memory with either of those.

So if you want the enum order to correspond to fallback (from most generic to least generic), then the current order of {cpu, gpu, npu} makes sense.

index.bs
fdwr: > I don't quite get the meaning of "perform between CPU and GPU". Does it refer to its power and performance characteristic?

**Ningxin**: Yep, I meant the execution performs faster than CPU and slower than GPU (but moot point now since I used Anssi's suggested text).

> Here's a proposal that avoids mentioning CPU and GPU. Feel free to use it, or parts of it, if this sounds reasonable

**Anssi**:👍 Used as-is.

> should we rather use the same words ...  "other accelerators, e.g. CPU or GPU"?

**Zoltan**: The CPU isn't thought of as an "accelerator", and using the word "device" is consistent with the enum name `MLDeviceType`. So I'd keep it as device.

fdwr: True - added same comment to "gpu".

index.bs
fdwr: Curious, I don't see the clickable pipeline result at the bottom of these code reviews anymore. Now I have to:
- click "Show all checks"
- click "Details"
- click "Summary"
- download `spec-prod-result.zip` from the artifacts
- extract `webnn.w3c/index.html`
- double click the .html file

So from 1-click step to 6 steps. Anyway, the issue text looks right.

anssiko:
@fdwr thank you for updating the PR. It now reflects the WG's latest discussion.

@huningxin @inexorabletash can you review the latest.

The Fallback topic warrants it own (non-blocking) issue. I'd propose to use @fdwr's overview https://github.com/webmachinelearning/webnn/pull/696#discussion_r1642201778 as a starting point for that discussion. @fdwr please feel free to open a separate issue copying your fallback overview there and cross-link to #623.

index.bs
zolkis: @fdwr thanks, makes sense!

inexorabletash: Minor point but: enum definition order in the WebIDL is not web-observable, so this is an "editorial" concern only, and there's not going to be any web-compat issue with reordering later. 

I would avoid relying on the IDL definition implying anything to either implementers or web developers. So adding prose explanation SGTM if we want to. But also, as noted in the last telecon I'm not sure we'll want to keep this API beyond the prototyping phase, so I also don't think we need to spend much time on it. But also also, for the same reason I think it's fine to add text and not agonize about making it perfect.

index.bs
```diff
@@ -725,12 +726,16 @@ interface ML {
 
 ### {{MLContextOptions}} ### {#api-mlcontextoptions}
 
+Issue(623): MLContextOptions is under active development, and the design is expected to change, informed by further implementation experience and new use cases from the wider web community. The Working Group is considering additional API controls to allow the definition of a fallback device, multiple devices in a preferred order, or an exclusion of a specific device. Other considerations under discussion include error handling, ultimate fallback, and quantized operators. Feedback is welcome on any of these design considerations from web developers, library authors, OS and hardware vendors, and other stakeholders via GitHub:
```
inexorabletash, line 729: Nit: I'd format it as `{{MLContextOptions}}` here.

index.bs
```diff
@@ -725,12 +726,16 @@ interface ML {
...
+
 The <dfn dfn-for=MLContextOptions dfn-type=dict-member>deviceType</dfn> option is an <dfn dfn-type=enum>MLDeviceType</dfn> and indicates the application's preference for the kind of device used for the context. It is one of the following:
 <dl dfn-for="MLDeviceType">
 <dt>"<dfn enum-value>cpu</dfn>"</dt>
 <dd>Provides the broadest compatibility and usability across all client devices with varying degrees of performance.</dd>
 <dt>"<dfn enum-value>gpu</dfn>"</dt>
-<dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations.</dd>
+<dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations. The backend implementation may fall back to other devices for certain operators and parts of the graph.</dd>
+<dt>"<dfn enum-value>npu</dfn>"</dt>
+<dd>Provides power efficiency for sustained workloads across hardware platforms with purpose-built accelerators. The backend implementation may fall back to other devices for certain operators and parts of the graph.</dd>
```
inexorabletash, line 738: Nit: we use "underlying platform" more often than "backend" (40 instances vs. 1 instance). Align wording with the rest of the spec?

inexorabletash (approved):
nits but overall LGTM

index.bs
```diff
@@ -613,7 +613,7 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring;
...
+The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. An {{MLDeviceType}} normatively indicates the kind of device and is one of: {{MLDeviceType/"cpu"}}, {{MLDeviceType/"gpu"}}, {{MLDeviceType/"npu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. An {{MLPowerPreference}} indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
```
mwyrzykowski, line 616: Similar to MLContextOptions being under active development as called out on line 729 below, I think we should refer to https://github.com/webmachinelearning/webnn/issues/623 here as well and state that MLDeviceType is subject to change.

mwyrzykowski (changes requested):
It would be good to make a note that MLDeviceType is subject to change similar to the note on line 729, the line right above the comment https://github.com/webmachinelearning/webnn/pull/696/files#r1643272185 

It is outside the scope of this change, but it would be preferable to remove the `If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown,` wording as I'm not sure how that is compatible with the sentence preceding it `and allow the implementation to better select the most appropriate underlying execution device for the workload.`

index.bs
fdwr: Are you wanting the whole block, or a shorter sentence here? Given this is the privacy section, I don't think it needs *quite* as much verbiage as that block in the relevant `MLContextOptions` section. Maybe just `Issue(623): MLContextOptions is under active development, and the design is expected to change, informed by further implementation experience and new use cases from the wider web community.`?

mwyrzykowski: A shorter sentence is great

fdwr: > It is outside the scope of this change, but it would be preferable to remove the If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown.

Yeah, the exact error policy is certainly a question we should shore up. I'm deferring that as part of 623's "Other considerations under discussion include error handling, ultimate fallback and quantized operators".

huningxin (approved):
LGTM!

index.bs
fdwr: Thanks Mike. I'll go ahead and merge this, as I believe all feedback has been addressed now.

Issue #351, opened 2023-02-20
Need to define error handling of MLNamedArrayBufferViews transfer algorithm


huningxin:
According to the steps of the [MLNamedArrayBufferViews transfer algorithm](https://www.w3.org/TR/webnn/#mlnamedarraybufferviews-transfer) in WebNN spec, for each `ArrayBufferView`, it [transfers the underlying `ArrayBuffer`](https://webidl.spec.whatwg.org/#arraybuffer-transfer) of the view. 

However, the `ArrayBuffer` transferring algorithm performs [`DetachArrayBuffer(arrayBuffer [, key])`](https://tc39.es/ecma262/#sec-detacharraybuffer), defined by ECMAScript spec, that would throw a "TypeError" exception if the `ArrayBuffer.[[ArrayBufferDetachKey]]` is not `key` (`key` is set to `undefined` if not present). Because WebNN doesn't set `key`, it means the `ArrayBuffer` whose [[ArrayBufferDetachKey]] is set to a value other than `undefined` would fail, such as the value of [WebAssembly.Memory](https://webassembly.github.io/spec/js-api/#memory)'s [buffer](https://webassembly.github.io/spec/js-api/#dom-memory-buffer) attribute. For another example, the `ArrayBuffer` returned by WebGPU [`GPUBuffer::getMappedRange()`](https://www.w3.org/TR/webgpu/#dom-gpubuffer-getmappedrange) who sets `[[ArrayBufferDetachKey]]` to "WebGPUBufferMapping".

WebNN spec should define how to handle this exception and what's the impact to the `MLNamedArrayBufferViews`. In particular, if the exception occurs when transferring the view in the middle of the loop, what should the implementation deal with the already transferred views, the failing view and the remaining views in the `MLNamedArrayBufferViews`.

This issue was raised by @wacky6 in WebNN [Chromium CL review](https://chromium-review.googlesource.com/c/chromium/src/+/4237384/comment/bd7b27ce_d03b98a9/). Thanks Jiewei!

huningxin:
In [WebML WG Teleconference – 11 May 2023](https://www.w3.org/2023/05/11-webmachinelearning-minutes.html#t09), group discussed this issue, and @RafaelCintron proposed that

> RafaelCintron: we can loop through everything before detaching? to make sure everything can be detached?
… put a pointer to say "this cannot be detached" and only when we know everything can be detached we detach everything

This ideal sounds great. Thanks Rafael! 

[...]

PR #700, opened 2024-06-03
Copy input ArrayBuffers

### LINKED ISSUE #566: Transferring constant ArrayBuffers

a-sully:
Fixes #566

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/700.html" title="Last updated on Jul 7, 2024, 8:30 PM UTC (e52d259)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/700/3a916f2...a-sully:e52d259.html" title="Last updated on Jul 7, 2024, 8:30 PM UTC (e52d259)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -845,7 +845,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 <dl dfn-type=dict-member dfn-for=MLComputeResult>
     : <dfn>inputs</dfn>
-    :: An object where the keys are the graph input names, and the values are the transferred {{ArrayBufferView}}s for the supplied input tensor values.
+    :: An object where the keys are the graph input names, and the values are the copied {{ArrayBufferView}}s for the supplied input tensor values.
 
     : <dfn>outputs</dfn>
     :: An object where the keys are the graph output names, and the values are the transferred {{ArrayBufferView}}s for the computed output tensor values.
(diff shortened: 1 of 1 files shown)


inexorabletash:
My reading of "[get a copy of the bytes](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy)" is that if the buffer is detached then an empty byte sequence is returned but we don't explicitly throw.

Per #351 (and my #723) we want to error if transferring a detached buffer (which will still be used for output buffers). Should we error for detached input buffers too?

I think that just requires adding `If |view| is [=BufferSource/detached=], then throw a {{TypeError}}.` to the steps.

(I think we'd still end up throwing in most cases since the subsequent size check will fail. Although if allow 0 size dimensions (#391) there would be a weird case where it might slip through that check. Harmless, but weird?)

a-sully:
Abandoning since `compute()` is being deprecated in favor of `dispatch()`. See https://github.com/webmachinelearning/webnn/blob/main/mltensor-explainer.md#compute-vs-dispatch

All the logic regarding buffer transferring can be removed alongside `compute()`, which makes this CL irrelevant

PR #704, opened 2024-06-07
Workflows: use ubuntu-latest to get up-to-date python/Bikeshed

inexorabletash:
Context:
https://github.com/speced/bikeshed/issues/2861#issuecomment-2131233140

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   main:
     name: Build, Validate and Deploy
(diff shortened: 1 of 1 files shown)


inexorabletash:
@anssiko - can you review/merge?

See the context link for more - this causes the workflow to run with an old Python, which in turn causes a very old version of Bikeshed to be used. This has been silently plaguing many specs for a while now, causing mysterious issues for authors.

huningxin (approved):
LGTM!

anssiko (commented):
Belated LGTM and thanks for identifying this issue.

PR #708, opened 2024-06-13
Link fix: Link to "is empty" not "empty" when an adjective is intended

inexorabletash:
The Infra term "list/empty" is a verb (to empty); link to "list/is empty" when that's what is intended.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/708.html" title="Last updated on Jun 14, 2024, 9:31 PM UTC (5afa97c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/708/c00560f...inexorabletash:5afa97c.html" title="Last updated on Jun 14, 2024, 9:31 PM UTC (5afa97c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1367,7 +1367,7 @@ Build a composed graph up to a given output operand into a computational graph a
     1. Let |operators| be a new empty [=/set=].
     1. Let |inputs| be a new empty [=/set=].
     1. Let |queue| be a new [=queue=] containing |outputs|'s [=map/values=].
-    1. While |queue| [=queue/is empty|is not empty=]:
+    1. While |queue| [=queue/is not empty=]:
         1. [=queue/Dequeue=] |operand| from |queue|.
         1. [=set/Append=] |operand| to |operands|.
         1. [=set/Append=] |operand|.{{MLOperand/[[operator]]}} to |operators|.
--- a/index.bs
+++ b/index.bs
@@ -1367,7 +1367,7 @@ Build a composed graph up to a given output operand into a computational graph a
     1. Let |operators| be a new empty [=/set=].
     1. Let |inputs| be a new empty [=/set=].
     1. Let |queue| be a new [=queue=] containing |outputs|'s [=map/values=].
-    1. While |queue| is not [=queue/empty=]:
+    1. While |queue| [=queue/is empty|is not empty=]:
         1. [=queue/Dequeue=] |operand| from |queue|.
         1. [=set/Append=] |operand| to |operands|.
         1. [=set/Append=] |operand|.{{MLOperand/[[operator]]}} to |operators|.
--- a/tools/lint.mjs
+++ b/tools/lint.mjs
@@ -273,4 +273,9 @@ for (const match of source.matchAll(/^ *\d+\. .*$/mg)) {
   }
 }
 
+// Avoid incorrect links to list/empty.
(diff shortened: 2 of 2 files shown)


inexorabletash:
@fdwr and @huningxin - please take a look?

tools/lint.mjs
```diff
@@ -273,4 +273,9 @@ for (const match of source.matchAll(/^ *\d+\. .*$/mg)) {
   }
 }
 
+// Avoid incorrect links to list/empty.
+for (const match of source.matchAll(/is( not)? \[=(list\/|stack\/|queue\/|)empty=\]/g)) {
+  error(`Link to 'is empty' (noun) not 'empty' (verb): ${format(match)}`);
```
huningxin, line 278: Should 'is empty' be an (adjective)? Apologizes if I misunderstood as I am not a native speaker.

huningxin (approved):
LGTM with a question, thanks!

index.bs
```diff
@@ -1367,7 +1367,7 @@ Build a composed graph up to a given output operand into a computational graph a
     1. Let |operators| be a new empty [=/set=].
     1. Let |inputs| be a new empty [=/set=].
     1. Let |queue| be a new [=queue=] containing |outputs|'s [=map/values=].
-    1. While |queue| is not [=queue/empty=]:
+    1. While |queue| [=queue/is empty|is not empty=]:
```
fdwr, line 1370: I'm perplexed by the new wording. Is this saying when the queue is either empty _or_ not empty?

tools/lint.mjs
fdwr: Yes, it should be an adjective here.

(supposedly "empty" *can* be used as a noun, calling a "bottle" an "empty", but I've never heard it used this way in my life)

inexorabletash: Uh, yeah...  I'll fix the PR next week unless someone beats me to it.

index.bs
inexorabletash: So Bikeshed's autolink syntax is `term|text` so you can override the term with custom text. The "queue/is empty" part is the (scoped) link to the term, the "is not empty" part after the pipe is what will show in the spec.

It's definitely awkward... and the good news is that Infra **_does_** define "is not empty" as an alias for "is empty" so this can be simplified to just `[=queue/is not empty=]`.

I'll update the PR next week unless someone beats me to it.

fdwr: > so you can override the term with custom text.

👍. So then what's still bending my brain 🙃 is the disagreement between the two sides of the `|`, one linking to "is empty" but the other stating the *opposite* "is not empty" (kinda like saying `[=queue/is red|is blue=]`), whereas if it had been `While [queue] not [=queue/is empty=]`, it would agree. I'll update it.

inexorabletash: LGTM!

tools/lint.mjs
inexorabletash: Total aside: it's common in the restaurant/bar world to refer to "empties", e.g. when serving tables ("Can I clear your empties?"), at least in my part of the world.

inexorabletash:
Thanks for updating - LGTM to merge

fdwr (approved):

PR #478, opened 2023-11-11
Add support for operations needed for well-known transformers e.g. Segment Anything, Stable Diffusion, etc.

wchao1115:
Issue #375. 

1. Add the following ops: `equal, lesser, greater, greaterOrEqual, lesserOrEqual, not, where, copy, sqrt, erf, expand, gather, layerNormalization, reduceArgMin, reduceArgMax, cast, triangular`
2. Remove `squeeze`
3. Add a `constant` overload that accepts `start, end, step`.
4. Add `shape` method in `MLOperand` for runtime tensor shape query.
5. Add new data type `int64` and `uint64` for `reduceArgMin` and `reduceArgMax`.

Notes:
- The proposed `identity` op is named `copy` to denote explicit action.
- The proposed `elementwiseIf` is named `where` to be consistent with PyTorch and ONNX.
- The proposed `triangularMatrix` is named `triangular` for brevity.
- `squeeze` is removed. Framework seeking `squeeze`, `unsqueeze`, `flatten` should just use `reshape`. See canonical implementations of these ops as `reshape` in the `reshape` section. 
- Instead of the proposed unification of all normalizations under one new normalization operation, a new `layerNormalization` op is added. There are enough subtleties in the way affine parameters (e.g. scale, bias) are expected across batchNorm, instanceNorm, and layerNorm that could become ambiguous under the unified op. Additionally, downstream implementations of these normalizations have already been done in various platforms. The unification at the WebNN level could potentially confuse framework developers and affects currently defined platform-specific optimizations.
- The proposed `fillSequence` op is added as a `constant` overload due to its nature of resource allocation activity. 
- The new sidebar loosens up the crowded *MLGraphBuilder interface* section for better op discoverability. 

Big thanks to @fdwr for the thoughtful and thorough proposal.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/478.html" title="Last updated on Dec 7, 2023, 3:05 AM UTC (e8c9864)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/478/c6ed2ca...e8c9864.html" title="Last updated on Dec 7, 2023, 3:05 AM UTC (e8c9864)">Diff</a>

PR #709, opened 2024-06-18
Write example decompositions as complete functions

inexorabletash:
This makes it clearer how the example decompositions map to the methods they are defining - variables are explicitly named as parameters, or unpacked from the arguments.

Also fix code treating MLOperand's dataType() as a getter instead of a method.

No normative changes.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/709.html" title="Last updated on Jun 25, 2024, 1:33 AM UTC (392ae28)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/709/5c64074...inexorabletash:392ae28.html" title="Last updated on Jun 25, 2024, 1:33 AM UTC (392ae28)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -955,7 +955,7 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
     const context = await navigator.ml.createContext();
     const builder = new MLGraphBuilder(context);
     // 1. Create a computational graph 'C = 0.2 * A + B'.
-    const constant = builder.constant(operandType.dataType, 0.2);
+    const constant = builder.constant(operandType.dataType(), 0.2);
     const A = builder.input('A', operandType);
     const B = builder.input('B', operandType);
     const C = builder.add(builder.mul(A, constant), B);
@@ -1573,7 +1573,7 @@ partial interface MLGraphBuilder {
             builder.sub(input, builder.reshape(mean, shape)),
             builder.sqrt(builder.add(
               builder.reshape(variance, shape),
-              builder.constant(input.dataType, options.epsilon))))),
+              builder.constant(input.dataType(), options.epsilon))))),
         builder.reshape(options.bias, shape));
     }
     </pre>
@@ -1643,16 +1643,17 @@ partial interface MLGraphBuilder {
           return input;
         } else {
           return builder.min(
-            input, builder.constant(input.dataType, options.maxValue));
+            input, builder.constant(input.dataType(), options.maxValue));
         }
       } else {
         if (options.maxValue === undefined) {
           return builder.max(
-            input, builder.constant(input.dataType, options.minValue));
+            input, builder.constant(input.dataType(), options.minValue));
         } else {
           return builder.min(
-            builder.max(input, builder.constant(input.dataType, options.minValue)),
-            builder.constant(input.dataType, options.maxValue));
+            builder.max(
+              input, builder.constant(input.dataType(), options.minValue)),
+            builder.constant(input.dataType(), options.maxValue));
         }
       }
     }
@@ -2557,12 +2558,12 @@ partial interface MLGraphBuilder {
     <pre highlight="js">
     function elu(builder, input, options) {
       return builder.add(
-        builder.max(builder.constant(input.dataType, 0), input),
+        builder.max(builder.constant(input.dataType(), 0), input),
         builder.mul(
-          builder.constant(input.dataType, options.alpha),
+          builder.constant(input.dataType(), options.alpha),
           builder.sub(
-            builder.exp(builder.min(builder.constant(input.dataType, 0), input)),
-            builder.constant(input.dataType, 1))));
+            builder.exp(builder.min(builder.constant(input.dataType(), 0), input)),
+            builder.constant(input.dataType(), 1))));
     }
     </pre>
   </details>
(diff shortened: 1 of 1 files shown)


inexorabletash:
@huningxin & @fdwr - please take a look?

index.bs
```diff
@@ -4289,106 +4333,122 @@ partial interface MLGraphBuilder {
...
+               zero),
+            (options.recurrentBias ?
+               builder.slice(
+                 options.recurrentBias, [2 * hiddenSize], [hiddenSize]) :
+               zero)),
+          builder.add(
+            builder.matmul(
+              input,
+              builder.transpose(builder.slice(
+                weight, [2 * hiddenSize, 0], [hiddenSize, inputSize]))),
```
fdwr, line 4392: This first-parameter-hugging makes it a little murkier the indentation levels when I'm visually trying to correspond the beginning and ending of things (yes, it saves some digital ink, but code isn't really printed on trees anymore). I'll look for some options to configure that... (but resolve me)

fdwr (approved):
Thank you for wrapping these all in named functions - that makes it clearer 🧼😎.

inexorabletash:
FYI I'll look into lint tooling to help enforce this, but I didn't want to block on it. The easiest might be pulling eslint in as another node dependency, if that doesn't seem too bad.

zolkis:
Yes, that sets a nice and clear context to the code snippets.
Whether to enforce it... maybe not, as not every code snippet might be meant to be quoted as a full function.

index.bs
```diff
@@ -1564,16 +1564,18 @@ partial interface MLGraphBuilder {
...
+    function batchNormalization(builder, input, mean, variance, options) {
+      const shape = [1, input.shape()[options.axis], 1, 1];
+      return builder.add(
+        builder.mul(
+          builder.reshape(options.scale, shape),
+          builder.div(
+            builder.sub(input, builder.reshape(mean, shape)),
+            builder.sqrt(builder.add(
+              builder.reshape(variance, shape),
+              builder.constant(input.dataType, options.epsilon))))),
```
huningxin, line 1576: This was a previous issue, `dataType` is a method rather than an attribute. Same for the other occurrences. 
```suggestion
              builder.constant(input.dataType(), options.epsilon))))),
```

index.bs
```diff
@@ -3085,96 +3097,105 @@ partial interface MLGraphBuilder {
...
-      const desc = {
-        dataType: 'float32',
-        dimensions: [numDirections, 1, hiddenSize]
-      };
-      const totalSize = numDirections * hiddenSize;
-      hiddenState = builder.constant(desc, new Float32Array(totalSize).fill(0));
-    }
+    function gru(
+      builder, input, weight, recurrentWeight, steps, hiddenSize, options) {
+      function squeeze(builder, op) {
```
huningxin, line 3102: Should we just use https://www.w3.org/TR/webnn/#emulation-squeeze?

inexorabletash: I could go either way. I like it being standalone, but there's some benefit in avoiding repetition. Linking to the squeeze emulation from the code might be awkward.

index.bs
inexorabletash: Done in 46442ad

index.bs
inexorabletash: Actually... these inline `function squeeze` examples use `op.shape().remove(0)` which was introduced in https://github.com/webmachinelearning/webnn/pull/478/commits/e284f2acc4c8db0db5f2415adc8be7ef188299c6 ... I'm not sure what the `remove()` is intending to do as there is no such method on Array.

inexorabletash: Okay, replaced with a link in 73bca91 - I think this is better.

inexorabletash:
@huningxin take another look, and merge if it looks good? Thanks!

fdwr (approved):

index.bs
```diff
@@ -5950,8 +5945,9 @@ partial interface MLGraphBuilder {
     function where(builder, condition, input, other) {
       const c = builder.clamp(condition, {'minValue': 0, 'maxValue': 1});
       builder.add(
-        builder.mul(input, builder.cast(c, input.dataType())),
-        builder.mul(other, builder.cast(builder.logicalNot(c), other.dataType())));
```
huningxin: The above two lines were good and won't need to change.

index.bs
```diff
@@ -5864,10 +5942,13 @@ partial interface MLGraphBuilder {
...
-    const c = builder.clamp(condition, {'minValue': 0, 'maxValue': 1});
-    builder.add(
-      builder.mul(input, builder.cast(c, input.dataType())),
-      builder.mul(other, builder.cast(builder.logicalNot(c), other.dataType())));
+    function where(builder, condition, input, other) {
+      const c = builder.clamp(condition, {'minValue': 0, 'maxValue': 1});
+      builder.add(
+        builder.mul(input, builder.cast(c, input.dataType()())),
+        builder.mul(
+          other, builder.cast(builder.logicalNot(c), other.dataType()())));
```
fdwr, line 5950:
```suggestion
        builder.mul(input, builder.cast(c, input.dataType())),
        builder.mul(
          other, builder.cast(builder.logicalNot(c), other.dataType())));
```

huningxin (approved):
LGTM, thanks for the fix!

index.bs
inexorabletash: Oops. :) Thanks.

PR #717, opened 2024-07-07
Allow MLGraphBuilder.build() to be called only once

### LINKED ISSUE #567: Can an MLGraphBuilder be reused?
While it doesn't seem impossible for implementations to support building multiple `MLGraph` instances out of a single `MLGraphBuilder` instance (potentially reusing subsets of the graph by choosing different inputs and outputs) it seems like it adds complexity to implementations to support.

Are there any known use cases for `MLGraphBuilder` reuse?

a-sully:
Fixes #567

If `MLGraphBuilder.build()` results in anything other than a `TypeError` - e.g. it successfully resolves with an or `MLGraph` or rejects with a `NotSupportedError` - that `MLGraphBuilder` has now been "built" and all subsequent methods will now reject with an `InvalidStateError`.

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/717.html" title="Last updated on Jul 10, 2024, 6:20 PM UTC (57e3962)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/717/3a916f2...a-sully:57e3962.html" title="Last updated on Jul 10, 2024, 6:20 PM UTC (57e3962)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -656,7 +656,7 @@ The implementation may use views, as above, for intermediate values.
 
 Before the execution, the computation graph that is used to compute one or more specified outputs needs to be converted, compiled, and optimized. The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion. The user agent may also perform these optimizations during graph conversion.
 
-The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. {{MLGraphBuilder/build()}} may be called at most once per {{MLGraphBuilder}}.
+The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. Each {{MLGraphBuilder}} can build at most one {{MLGraph}}.
 
 The {{MLGraph}} underlying implementation will be composed of platform-specific representations of operators and operands which correspond to the {{MLGraphBuilder}}'s [=operators=] and {{MLOperand}}s, but which are not script-visible and may be compositions or decompositions of the graph as constructed by script.
 
@@ -1311,8 +1311,8 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
   </summary>
     1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If |name| is empty, then [=exception/throw=] a {{TypeError}}.
+    1. If any {{MLOperand}}s in [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/inputs=] have a {{MLOperand/[[name]]}} equal to |name|, then [=exception/throw=] a {{TypeError}}.
     1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If any {{MLOperand}}s in [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/inputs=] have the same {{MLOperand/[[name]]}}, then [=exception/throw=] a {{TypeError}}.
     1. *Make graph connections:*
         1. Let |operand| be the result of [=creating an MLOperand=] given [=this=] and |descriptor|.
         1. Set |operand|.{{MLOperand/[[name]]}} to |name|.
--- a/index.bs
+++ b/index.bs
@@ -656,7 +656,7 @@ The implementation may use views, as above, for intermediate values.
 
 Before the execution, the computation graph that is used to compute one or more specified outputs needs to be converted, compiled, and optimized. The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion. The user agent may also perform these optimizations during graph conversion.
 
-The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. The compilation step produces an {{MLGraph}} that represents a compiled graph for optimal execution.
+The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. {{MLGraphBuilder/build()}} may be called at most once per {{MLGraphBuilder}}.
 
 The {{MLGraph}} underlying implementation will be composed of platform-specific representations of operators and operands which correspond to the {{MLGraphBuilder}}'s [=operators=] and {{MLOperand}}s, but which are not script-visible and may be compositions or decompositions of the graph as constructed by script.
 
@@ -1277,6 +1277,9 @@ The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph buil
     : <dfn>\[[context]]</dfn> of type {{MLContext}}
     ::
         The context of type {{MLContext}} associated with this {{MLGraphBuilder}}.
+    : <dfn>\[[hasBuilt]]</dfn> of type {{boolean}}
+    ::
+        Whether {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} has been called. Once built, the {{MLGraphBuilder}} can no longer create [=operators=] or compile {{MLGraph}}s.
   </dl>
 </div>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1395,10 +1404,6 @@ Build a composed graph up to a given output operand into a computational graph a
         1. If |operand| is in [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/inputs=], [=set/append=] |operand| to |inputs|.
         1. [=list/For each=] |input| of |operand|.{{MLOperand/[[operator]]}}'s [=operator/inputs=]:
             1. [=queue/Enqueue=] |input| to |queue|.
```
a-sully, line 1406: Note: this PR does not address https://github.com/webmachinelearning/webnn/issues/567#issuecomment-1996062048 since there didn't appear to be consensus on whether that was desirable (FYI @inexorabletash @fdwr). If folks have strong feelings in either direction then I propose we file a separate issue to discuss; otherwise I propose we leave this as-is for now. WDYT?

a-sully:
@huningxin @fdwr PTAL?

index.bs
```diff
@@ -656,7 +656,7 @@ The implementation may use views, as above, for intermediate values.
 
 Before the execution, the computation graph that is used to compute one or more specified outputs needs to be converted, compiled, and optimized. The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion. The user agent may also perform these optimizations during graph conversion.
 
-The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. The compilation step produces an {{MLGraph}} that represents a compiled graph for optimal execution.
+The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. {{MLGraphBuilder/build()}} may be called at most once per {{MLGraphBuilder}}.
```
inexorabletash, line 659: Avoid "may" (RFC 2119); you could just use "can" here.

index.bs
```diff
@@ -1305,8 +1309,10 @@ Create a named {{MLOperand}} based on a descriptor, that can be used as an input
   <summary>
     The <dfn method for=MLGraphBuilder>input(|name|, |descriptor|)</dfn> method steps are:
   </summary>
+    1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If |name| is empty, then [=exception/throw=] a {{TypeError}}.
     1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If any {{MLOperand}}s in [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/inputs=] have the same {{MLOperand/[[name]]}}, then [=exception/throw=] a {{TypeError}}.
```
inexorabletash, line 1315: nit: I'd move this step up one, so the checks related to `name` are adjacent.

inexorabletash: nit: "have the same name" ... as what? It's pretty obvious in context, but could be made more rigorous as "have name equal to name"

index.bs
inexorabletash: SGTM.

index.bs
```diff
@@ -5558,8 +5609,9 @@ partial interface MLGraphBuilder {
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLGraphBuilde id=softplus-noargs>softplus()</dfn> method steps are:
+    The <dfn method for=MLGraphBuilder id=softplus-noargs>softplus()</dfn> method steps are:
```
inexorabletash, line 5612: Thanks for the fix. It seems like we should be able to "lint" this - e.g. verify that no methods (etc) are defined "for" a scope that isn't defined in IDL, or something

inexorabletash: If you add this near the end of tools/lint.mjs it will catch the error:

```js
// Ensure every method dfn is correctly associated with an interface
const interfaces = new Set(
  root.querySelectorAll('dfn[data-dfn-type=interface]').map(e => e.innerText));
for (const dfn of root.querySelectorAll('dfn[data-dfn-type=method]')) {
  const dfnFor = dfn.getAttribute('data-dfn-for');
  if (!dfnFor || !interfaces.has(dfnFor)) {
    error(`Method definition '${dfn.innerText}' for undefined '${dfnFor}'`);
  }
}
```

a-sully: Created #720

huningxin (approved):
LGTM with a comment

Checked with @Honry  that ORT WebNN EP currently doesn't share constant operands when building sub-graphs for a (merged) model.

index.bs
```diff
@@ -656,7 +656,7 @@ The implementation may use views, as above, for intermediate values.
 
 Before the execution, the computation graph that is used to compute one or more specified outputs needs to be converted, compiled, and optimized. The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion. The user agent may also perform these optimizations during graph conversion.
 
-The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. The compilation step produces an {{MLGraph}} that represents a compiled graph for optimal execution.
+The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in the background without blocking the calling thread, and returns a {{Promise}} that resolves to an {{MLGraph}}. Each {{MLGraphBuilder}} can build at most one {{MLGraph}}.
```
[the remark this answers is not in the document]
a-sully: I'm happy to add that sentence back if you wish. I removed it since it seemed redundant given that we already state elsewhere:

> The [MLGraph](https://www.w3.org/TR/webnn/#mlgraph) interface represents a compiled computational graph

and 

> Before the execution, the computation graph that is used to compute one or more specified outputs needs to be converted, compiled, and optimized. The key purpose of the compilation step is to enable optimizations that span two or more operations, such as operation or loop fusion. The user agent may also perform these optimizations during graph conversion.

index.bs
fdwr: Even with this change, we can still AFAICS create a disconnected graph with a single graphBuilder, right? It's a legitimate graph construct.

image

index.bs
```diff
@@ -5966,6 +6027,8 @@ partial interface MLGraphBuilder {
   <summary>
     The <dfn method for=MLGraphBuilder>where(|condition|, |input|, |other|)</dfn> method steps are:
   </summary>
+    1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
```
fdwr, line 6030: 🫤 Seeing "if hasBuilt is true then throw this kind of error" repeated in the logic of *every* single operator feels unclean. Can we have a common graph builder "operator initialization" function that they all call, and that error check can be the very first (and currently only) step in it, kinda like how we centralized [casting](https://www.w3.org/TR/webnn/#algorithms-casting) and broadcasting functionality? Mind you, every single operator still then needs "1. Perform builder-operator-initialization", but it feels 🧼➕ imo. Opinion?

a-sully: Hmm I agree the current state is annoyingly duplicative, but I couldn't decide on a useful name of a helper method... This check is performed in `MLGraphBuilder.build()` as well all methods which mint an `MLOperand`. Calling a "builder-operator-initialization" algorithm from the `build()` steps is a bit odd (and nothing is being initialized anyways).

Lacking a good name, I figured inlining the one-liner would be more readable ¯\\_(ツ)_/¯

If you can think of a better name then I'm happy to change it :P

fdwr: The pervasive naming challenge... Let me ponder a bit, after some sleep.

inexorabletash: Throwing out another idea: we could maybe use a [Bikeshed text macro](https://speced.github.io/bikeshed/#metadata-text-macro) to reduce the burden on spec authors. We [define one today for EMULATED](https://github.com/webmachinelearning/webnn/blob/fe96a7809e33fc7dd1cf58717743089f559d8e96/index.bs#L39) to avoid repetition. This doesn't reduce the burden on spec readers, though.

index.bs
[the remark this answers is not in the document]
huningxin: > I removed it since it seemed redundant given that we already state elsewhere:

Thanks for the explanation. Then fine with me.

index.bs
a-sully: True, though we'd still have to name the macro :P 

`THROW_IF_BUILT`?

fdwr: No bright insights came to me after sleep or over the weekend 😅. Going to merge now and maybe think of something later.

fdwr (approved):
👍 The builder patterns I've seen generally let you create multiple objects from the builder, with invalidation being a rare case, but I see advantages to both; and it's easier to potentially relax this later if needed than to restrict it later.

PR #720, opened 2024-07-10
Ensure every method dfn is correctly associated with an interface

a-sully:
As suggested by @inexorabletash in https://github.com/webmachinelearning/webnn/pull/717/files/9c9bfa2a5c20f711aad516192088fc6d7b97fb36#r1672817903


huningxin (approved):
LGTM!

PR #723, opened 2024-07-11
Define error handling of MLNamedArrayBufferViews transfer algorithm

### LINKED ISSUE #351: Need to define error handling of MLNamedArrayBufferViews transfer algorithm

inexorabletash:
* Throw if the passed buffer source is detached.
* Explicitly rethrow if transfering throws.

The callers already handle failure.

Fixes #351

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/723.html" title="Last updated on Jul 12, 2024, 6:08 PM UTC (6edc5e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/723/af67a72...inexorabletash:6edc5e8.html" title="Last updated on Jul 12, 2024, 6:08 PM UTC (6edc5e8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -910,7 +910,7 @@ Issue(351): Move the above algorithm into [[WEBIDL]].
     1. Let |transferredViews| be a new {{MLNamedArrayBufferViews}}.
     1. [=map/For each=] |name| → |view| of |views|:
         1. Let |transferredBuffer| be the result of [=ArrayBuffer/transfer|transferring=] |view|'s [=BufferSource/underlying buffer=].
-        1. Assert: The above step never throws an exception.
+        1. [=Assert=]: The above step never throws an exception.
         1. Let |constructor| be the appropriate [=view constructor=] for the type of {{ArrayBufferView}} |view| from |realm|.
         1. Let |elementsNumber| be the result of |view|'s [=BufferSource/byte length=] / |view|'s [=element size=].
         1. Let |transferredView| be [$Construct$](|constructor|, |transferredBuffer|, |view|.\[[ByteOffset]], |elementsNumber|).
--- a/index.bs
+++ b/index.bs
@@ -886,11 +886,31 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 <details open algorithm>
   <summary>
+    A [=buffer source type=] instance |bufferSource| is
+    <dfn for="BufferSource">detachable</dfn> if the following steps return true:
+  </summary>
+    1.  Let |jsArrayBuffer| be the result of [=converted to a JavaScript value|converting=]
+        |bufferSource| to a JavaScript value.
+    1.  If |jsArrayBuffer| has a \[[ViewedArrayBuffer]] internal slot, then set |jsArrayBuffer| to
+        |jsArrayBuffer|.\[[ViewedArrayBuffer]].
+    1.  If [$IsSharedArrayBuffer$](|jsArrayBuffer|) is true, then return false.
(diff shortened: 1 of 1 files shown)


inexorabletash:
@reillyeon can you take an initial look at this? It seems like WebIDL changes have made this trivial, but maybe I'm missing something.

index.bs
```diff
@@ -890,7 +890,9 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
   </summary>
     1. Let |transferredViews| be a new {{MLNamedArrayBufferViews}}.
     1. [=map/For each=] |name| → |view| of |views|:
```
reillyeon, line 892: I think part of the core of #351 is that if this loop aborts early then the state of the buffers in `views` is unclear since some of them might've been transferred. If we can guarantee that  none of these other steps can fail then we could run a loop before this to check that all the buffers are detachable and then not try transferring any of them so that we can assert that an error is not thrown.

reillyeon (changes requested):

index.bs
inexorabletash: Ah, thanks.

inexorabletash:
Okay @reillyeon - now that I've really read the issue, this should match what was proposed there.

My plan is:
* Land this with a local copy of the BufferSource/detachable algorithm.
* File a PR against WebIDL referencing this spec (i.e. indicating it's useful)
* Remove the local copy if/when that merges.

reillyeon (approved):

inexorabletash:
Thanks @reillyeon !

@huningxin & @fdwr - can you please review?

huningxin (approved):
LGTM!

fdwr (approved):
👍

PR #725, opened 2024-07-16
Add security consideration for computation control-flow attacks

### LINKED ISSUE #443: Add security consideration for computation control-flow attack based on weights / constants change
This security issue was raised by @quidity (Thanks Alex!) in WebNN Chromium [CL review](https://chromium-review.googlesource.com/c/chromium/src/+/4664465/comment/1f1f9982_6e41a0fb/) where Alex mentioned:

> it can't be good for the weights to be changed during the computation

> I'm worried about compromised renderers - while JS can't change the weights a compromised renderer can. I'm not convinced that there won't be control flow differences or changes to assumptions based on weights (now or in the future) - for instance I could imagine an optimization that checks if something row is zero then retains that assumption throughout a computation.

This attack should be discussed in WebNN's [Security Considerations](https://www.w3.org/TR/webnn/#security) section.

There are some related discussions in current WebNN spec:
1. The [MLGraphBuilder.constant()](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-constant-method) method algorithm, in particular step 5 & 6, specifies that the content of weights is copied from the user-supplied ArrayBuffer and a platform constant tensor is created with that content. 

> Step 5. Let bytes be the result of invoking [get a copy of the bytes held by the buffer source](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy) given bufferView.

> Step 6. Make a request to the underlying platform to register operand as a tensor constant with bytes as value and store a reference to the corresponding implementation-defined object to operand.[[operand]].

2. There is also a [Graph Initialization](https://www.w3.org/TR/webnn/#api-mlcommandencoder-graph-initialization) stage that does "weight preprocessing"

> "weight preprocessing" where all the constant inputs to the graph are preprocessed and cached at the operating system level for subsequent graph execution calls. 

IIUC, once the graph is fully compiled and initialized, the content of weights is copied to platform constant tensor and the platform constant tensor are preprocessed and cached by operating system. The WebNN implementation won't access the weights in user-supplied ArrayBuffer anymore, and there is not a surface where the JS / comprised renderer can access the platform constant tensor, so the comprised renderer should not be able to change the weights during the computation for such a attack.

/cc @wacky6 @RafaelCintron @wchao1115 @fdwr

inexorabletash:
As noted in #443 if constants aren't actually constant then bad things can happen. Add a note to mention that implementations should mitigate this.

Purely editorial change. Fixes #443

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/725.html" title="Last updated on Jul 17, 2024, 6:53 AM UTC (cb5b8de)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/725/f5fb055...inexorabletash:cb5b8de.html" title="Last updated on Jul 17, 2024, 6:53 AM UTC (cb5b8de)">Diff</a>


inexorabletash:
@huningxin & @fdwr - please take a look at merge if it looks good?

huningxin (approved):
LGTM!

fdwr (approved):
👍 TY.

PR #726, opened 2024-07-16
Clarify the cast() op behavior between different data types

### LINKED ISSUE #489: Clarify the casting behavior from floating-point / signed integers <-> unsigned integers
Opening this issue to track the [discussion](https://github.com/webmachinelearning/webnn/pull/478#discussion_r1405576949) for `cast` operator in PR #478.

(raised by @wacky6 in Chromium [CL-5056249](https://chromium-review.googlesource.com/c/chromium/src/+/5056249/comment/159c35cc_28e8f1e6/) review)

Casting to/from fp <-> int or fp is well understood. But what about the casting behavior from fp / int <-> uint?

Would -1.23 fp32 cast to 0 or 255 uint8?

The spec PR mentions cast is implementation defined, which isn't ideal. We should at least provide what caller should expect.

@fdwr [commented](https://github.com/webmachinelearning/webnn/pull/478#discussion_r1408719971)

> what about the casting behavior from fp / int <-> uint?

That's a trickier case for conformance, casting from a wider range into a narrower range, as different hardware gives differing results. On CPU, using SSE vs the classic FPU could return different results. On GPU, you could get different results depending on whether your GPU supported typed UAV's or native structured UAV's. On NPU, I don't even know yet.

> Would -1.23 fp32 cast to 0 or 255 uint8?

I can say that locally for -1.0f -> uint8, I get 255 on CPU via C++ `static_cast<uint8_t>`, but I get 0 on my GPU (because -1.0f is mapped to int32_t -1, which then clamps to `[0,255]` when written out to the typed UAV (since uint8_t is not a natively supported type within HLSL). Then for -1.0f -> uint16_t, it's a similar story, 0xFFFF on CPU but 0 on GPU. Though, if I tried this on a GPU with `D3D12_FEATURE_DATA_D3D12_OPTIONS4::Native16BitShaderOpsSupported` true, then I might well get 0xFFFF instead.

So, if we require an always consistent answer in the spec for negative float -> uint cases, then we'd need some intermediate casts. Surprisingly though, this issue evidently hasn't come up so far in the DML EP. Want to open an issue for it?

@wchao1115 [mentioned](https://github.com/webmachinelearning/webnn/pull/478#discussion_r1410222259):

Type casting from floating-point to signed/unsigned integer is a complex process, and one without clear industry standard b/c the casting will be both accuracy-lossy and range-dependent. It is generally understood simply as an "undefined" behavior where the outcome may depend on a number of runtime factors including the support in the hardware. This is part of the reason why no other framework attempts to define it concretely to-date and leave part of it as implementation-dependent. The same situation is applied to WebNN.

Another open is if the behavior depends on different hardware internals, would it cause some finger-printing issues?

inexorabletash:
Inspired by the ONNX documentation[1], outline what inference-time casts are well-defined and which are implementation-defined for out-of-range values.

Fixes #489

1: https://onnx.ai/onnx/operators/onnx__Cast.html

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/726.html" title="Last updated on Jul 17, 2024, 4:44 PM UTC (7dc738d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/726/c9e70f6...inexorabletash:7dc738d.html" title="Last updated on Jul 17, 2024, 4:44 PM UTC (7dc738d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1709,9 +1709,11 @@ Casting between {{MLOperandDataType}}s is specified for some cases and [=impleme
     <th class=split>
       <span class=top-right>Target type</span>
       <span class=bottom-left>Input type</span>
+    </th>
     <th>
       {{MLOperandDataType/"float32"}},
       {{MLOperandDataType/"float16"}}
+    </th>
     <th>
       {{MLOperandDataType/"int32"}},
       {{MLOperandDataType/"uint32"}},
@@ -1719,16 +1721,24 @@ Casting between {{MLOperandDataType}}s is specified for some cases and [=impleme
       {{MLOperandDataType/"uint64"}},
       {{MLOperandDataType/"int8"}},
       {{MLOperandDataType/"uint8"}}
+    </th>
+  </tr>
   <tr>
     <th>
       {{MLOperandDataType/"float32"}},
       {{MLOperandDataType/"float16"}}
+    </th>
     <td>
-      <p>If in range, nearest representable value.
-      <p>If out of range, +/-Infinity.
+      If in range, nearest representable value.
+
+      If out of range, +/-Infinity.
+    </td>
     <td>
-      <p>If in range, truncated.
-      <p>If out of range, [=implementation-defined=].
+      If in range, truncated.
+
+      If out of range, [=implementation-defined=].
+    </td>
+  </tr>
   <tr>
     <th>
       {{MLOperandDataType/"int32"}},
@@ -1737,12 +1747,18 @@ Casting between {{MLOperandDataType}}s is specified for some cases and [=impleme
       {{MLOperandDataType/"uint64"}},
       {{MLOperandDataType/"int8"}},
       {{MLOperandDataType/"uint8"}}
+    </th>
     <td>
-      <p>If in range, nearest representable value.
-      <p>If out of range, +/-Infinity.
+      If in range, nearest representable value.
+
+      If out of range, +/-Infinity.
+    </td>
     <td>
-      <p>If in range, same value.
-      <p>If out of range, lowest N bits reinterpreted as target type, assuming two's complement for signed types.
+      If in range, same value.
+
+      If out of range, lowest N bits reinterpreted as target type, assuming two's complement for signed types.
+    </td>
+  </tr>
 </table>
 
 NOTE: For example, casting -1 from {{MLOperandDataType/"int8"}} to {{MLOperandDataType/"uint8"}} is specified to yield 255. But casting -1 from {{MLOperandDataType/"float32"}} to {{MLOperandDataType/"uint8"}} is [=implementation-defined=].
--- a/index.bs
+++ b/index.bs
@@ -200,9 +200,48 @@ th {
 
 th, td {
   border-bottom: 1px solid black;
-  border-collapse: collapse;
+  border-collapse: collapse; /* BUG: This property only applies to TABLE. */
   padding-left: 5px;
   padding-right: 5px;
+}
+
+/*
+ * Grid table format.
+ */
+table.grid {
+  border-collapse: collapse;
+  table-layout: fixed;
+  width: 100%;
+}
+
+table.grid td, table.grid th {
+  border: 1px solid black;
+}
+
+table.grid th {
+  text-align: center;
+}
+
+/* For a table header cell that is split diagonally */
+th.split {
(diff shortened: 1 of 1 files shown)


inexorabletash:
@fdwr & @huningxin - could you please review?

Hopefully I didn't go too far with the table styling.

index.bs
```diff
@@ -200,11 +200,50 @@ th {
...
+  border: 1px solid black;
+}
+
+table.grid th {
+  text-align: center;
+}
+
+/* For a table header cell that is split diagonally */
+th.split {
+  background-image: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' preserveAspectRatio='none' viewBox='0 0 1 1'><line x1='0' y1='0' x2='1' y2='1' stroke='black' vector-effect='non-scaling-stroke'/></svg>");
```
fdwr, line 227: Is all the SVG just for the sake of the single pixel thick stroke? I wonder if there's a pure CSS way to achieve it.

index.bs
```diff
@@ -1649,6 +1688,55 @@ partial interface MLGraphBuilder {
...
+    <th>
+      {{MLOperandDataType/"float32"}},
+      {{MLOperandDataType/"float16"}}
+    <th>
+      {{MLOperandDataType/"int32"}},
+      {{MLOperandDataType/"uint32"}},
+      {{MLOperandDataType/"int64"}},
+      {{MLOperandDataType/"uint64"}},
+      {{MLOperandDataType/"int8"}},
+      {{MLOperandDataType/"uint8"}}
```
fdwr, line 1711: Would explicit tag closure aid tooling (like your linting rules)? I know 'th' and 'tr' are some of those magic self-closing optional ones, but it does look weird to me :b.

fdwr (approved):
👍 Thanks JB for adding clarity.

index.bs
inexorabletash: Yes, just for the line. Simpler approaches welcome!

index.bs
inexorabletash: Happy to add them.

index.bs
fdwr: So a helpful bot told me that we could simply use the following CSS...

```
  vector-effect: non-scaling-stroke;
  stroke-width: 1; /* Set your desired stroke width */
```

...but then LLM's fabricate -_-.

Is it sufficient to just use a *thin* border style? For very high DPI monitors, maybe we *don't* want to enforce a single pixel always.

```
  border-style: solid;
  border-width: thin;
```

huningxin (approved):
LGTM!

index.bs
inexorabletash: Done in 7dc738d

(FWIW, the lint tool's parser seems to handle TR/TD/TH as mutually self-closing, it just doesn't know about DD/DT. But it doesn't hurt to add them, and makes some editors happier.)

index.bs
inexorabletash: What I was using was inspired by https://www.peterkrautzberger.org/0213/ Breaking down what `th.split` is doing in this PR:

```css
th.split {                                                           
  background-image: url("data:image/svg+xml;utf8,...");  /* 👈 image of a diagonal line */
  background-repeat: no-repeat; /* 👈 don't tile, and start at top left */
  background-size: 100% 100%; /* 👈 scale to cover the entire table cell */
  position: relative; /* 👈 make positioned, so child labels can be absolutely positioned within it */
}           
```

And breaking down the SVG:

```xml
<svg xmlns='http://www.w3.org/2000/svg' version='1.1'  👈 boilerplate
  preserveAspectRatio='none'   👈 allow stretching to cover entire table cell
  viewBox='0 0 1 1'> 👈 defines our coordinate space; don't need much
    <line 👈 render a single line
       x1='0' y1='0' x2='1' y2='1' 👈 from top-left to bottom-right
       stroke='black'  👈 default wouldn't be visible
       vector-effect='non-scaling-stroke' 👈 we don't want the stroke to be wider or thinner when the viewbox is stretched
    />
</svg>"
```

So... this seems pretty minimal to me, and I'm not sure how the suggested CSS helps. The stroke width's default of 1px can be overridden, e.g. `stroke-width='5px'` but that doesn't eliminate the need for the any of the above - without the `vector-effect` the image being stretched to cover the table cell's arbitrary size means the stroke would scale strangely.

fdwr: Oops, I didn't realize that was a *diagonal* line, misreading it as vertical. (the full breakdown was not necessary, but thanks) - I could have just copied that snippet into my little [SVG Viewer](https://github.com/fdwr/LunaSvgSampleTest) to see what it produced  :b.

fdwr (approved):

index.bs
fdwr: Appreciated.

PR #728, opened 2024-07-17
Workflow: Add steps to run tools/lint.mjs after spec is built

inexorabletash:
Help catch errors by running our bespoke lint tool after the normal publish/deploy. Ideally, in the future this would be integrated into spec-prod to avoid some duplicate work, and made more modular and robust. But hopefully it's better than nothing!

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -23,3 +23,21 @@ jobs:
           W3C_BUILD_OVERRIDE: |
             TR: https://www.w3.org/TR/webnn/
             status: CRD
+      - name: WebNN - Lint Bikeshed Source and Generated HTML
+        id: webnn-lint
+        shell: bash
+        env:
+          BIKESHED_SOURCE: "index.bs"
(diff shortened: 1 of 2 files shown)


inexorabletash:
@dontcallmedom can you take a look?

My knowledge of github workflows can best be described as "hit it with a hammer", so suggestions welcome.

dontcallmedom (approved):
LGTM - spec-prod uploads its output as github artifacts, so you could use that instead of re-generating the spec https://github.com/w3c/spec-prod/blob/796e4101600225fe24d1e967416e348fefdcc266/action.yml#L121 - but I don't think it's critical before merging this

inexorabletash:
Thanks @dontcallmedom !

@huningxin & @fdwr - can you review / merge if it looks like an okay starting point?

huningxin (approved):
LGTM!

fdwr (approved):
👍 Knowing nothing about GitHub YAML configs, I can't say it's right/wrong, but I can say nothing look fishy to me :b.

PR #729, opened 2024-07-17
Avoid "sequence" in prose, linkify use in types

inexorabletash:
Per the spec processing model, algorithms don't usually deal with a sequence<T> type; the JS value has been converted to an Infra list. So try to avoid using that term in prose, especially as it is confusing for recurrent operators which have temporal sequences as well. But do call out an exception: unpacking unions.

When used in an IDL type description, though, do linkify it.

Fix a few other grammar/link glitches, too.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/729.html" title="Last updated on Jul 18, 2024, 11:49 PM UTC (9ffdda7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/729/19762ed...inexorabletash:9ffdda7.html" title="Last updated on Jul 18, 2024, 11:49 PM UTC (9ffdda7)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -3219,7 +3219,7 @@ partial interface MLGraphBuilder {
         - <dfn>hiddenSize</dfn>: an {{unsigned long}} scalar. The value of the third dimension of the cell output tensor shape. It indicates the number of features in the hidden state.
         - <dfn>options</dfn>: an optional {{MLGruOptions}}. The optional parameters of the operation.
 
-    **Returns:** [=sequence=]<{{MLOperand}}>. The first element of is a 3-D tensor of shape *[numDirections, batchSize, hiddenSize]*, the cell output from the last time step of the network. Additionally, if |options|.{{MLGruOptions/returnSequence}} is set to true, the second element is the 4-D output tensor of shape *[steps, numDirections, batchSize, hiddenSize]* containing every cell outputs from each time step in the temporal sequence.
+    **Returns:** [=sequence=]<{{MLOperand}}>. The first element is a 3-D tensor of shape *[numDirections, batchSize, hiddenSize]*, the cell output from the last time step of the network. Additionally, if |options|.{{MLGruOptions/returnSequence}} is set to true, the second element is the 4-D output tensor of shape *[steps, numDirections, batchSize, hiddenSize]* containing every cell outputs from each time step in the temporal sequence.
 </div>
 
 <details open algorithm>
--- a/index.bs
+++ b/index.bs
@@ -984,7 +984,7 @@ In accordance with the [=ArrayBufferView/write|Web IDL warning=], to prevent the
       - <dfn>inputs</dfn>: an {{MLNamedArrayBufferViews}}. The resources of inputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.
       - <dfn>outputs</dfn>: an {{MLNamedArrayBufferViews}}. The pre-allocated resources of required outputs. Will be [=MLNamedArrayBufferViews/transfer|transferred=] if there are no validation errors.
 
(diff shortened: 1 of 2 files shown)


huningxin (approved):
LGTM!

index.bs
```diff
@@ -3219,7 +3219,7 @@ partial interface MLGraphBuilder {
         - <dfn>hiddenSize</dfn>: an {{unsigned long}} scalar. The value of the third dimension of the cell output tensor shape. It indicates the number of features in the hidden state.
         - <dfn>options</dfn>: an optional {{MLGruOptions}}. The optional parameters of the operation.
 
-    **Returns:** a sequence of {{MLOperand}}. The first element of the sequence is a 3-D tensor of shape *[numDirections, batchSize, hiddenSize]*, the cell output from the last time step of the network. Additionally, if |options|.{{MLGruOptions/returnSequence}} is set to true, the second element is the 4-D output tensor of shape *[steps, numDirections, batchSize, hiddenSize]* containing every cell outputs from each time step in the temporal sequence.
+    **Returns:** [=sequence=]<{{MLOperand}}>. The first element of is a 3-D tensor of shape *[numDirections, batchSize, hiddenSize]*, the cell output from the last time step of the network. Additionally, if |options|.{{MLGruOptions/returnSequence}} is set to true, the second element is the 4-D output tensor of shape *[steps, numDirections, batchSize, hiddenSize]* containing every cell outputs from each time step in the temporal sequence.
```
fdwr, line 3222:
```suggestion
    **Returns:** [=sequence=]<{{MLOperand}}>. The first element is a 3-D tensor of shape *[numDirections, batchSize, hiddenSize]*, the cell output from the last time step of the network. Additionally, if |options|.{{MLGruOptions/returnSequence}} is set to true, the second element is the 4-D output tensor of shape *[steps, numDirections, batchSize, hiddenSize]* containing every cell outputs from each time step in the temporal sequence.
```

index.bs
```diff
@@ -5413,7 +5413,8 @@ partial interface MLGraphBuilder {
         - <dfn>input</dfn>: an {{MLOperand}}. The input 4-D tensor.
         - <dfn>options</dfn>: an optional {{MLResample2dOptions}}. The optional parameters of the operation.
 
-    **Returns:** an {{MLOperand}}. The output 4-D tensor.
+    **Returns:**
+        - an {{MLOperand}}. The output 4-D tensor.
```
fdwr, line 5417: (nit) Same line or next line?

```suggestion
    **Returns:** an {{MLOperand}}. The output 4-D tensor.
```
Otherwise it's inconsistent with pad above:

    **Returns:** an {{MLOperand}}. The padded output tensor. Each dimension of the output tensor can be calculated as follows:

fdwr (approved):

index.bs
inexorabletash: I commented on Slack that about half the spec keeps "Returns: ..." on one line, the other half splits the "Returns" across two lines. If you have a preference for one line I'm happy to put up a PR to do that everywhere.

inexorabletash:
Merge when ready!

fdwr (approved):
👍

index.bs
fdwr: 🤔 Hmm, it's reasonable to put them on the same line currently because all current operators either return a single operand anyway, or they (e.g. split and gru) return multiple operands are homogeneous (basically an array of tensors). If we return multiple heterogenous outputs in the future though, it would make sense to list them as bullet point items on the next line, consistently with the "Arguments" and pending "Data types" sections. Well, I don't have a strong preference currently - I just noticed it because those differing lines were near each other in the CR.

PR #730, opened 2024-07-18
Add outputDataType to argmin/argmax

### LINKED ISSUE #653: Consider changing output type of ArgMax/Argmin to int32, or allow passing output_type
Currently WebNN specifies ArgMax/Min returns [int64](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-argminmax).
- [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS17.reduction.reduce_argmin) supported return type is int32 or uint16. 
- [tflite](https://www.tensorflow.org/mlir/tfl_ops#tflarg_max_tflargmaxop) supports int32, int64.
- [dml](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_argmax_operator_desc) supports int32, int64, uint32, uint64.

Returning int64 can't be emulated on CoreML:
- the [cast](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.elementwise_unary.cast) operator only supports casting to fp16, fp32, i32, bool.
- None of the operations support int64 as input(so the output of argMax/argMin can't be connected to the next layer if it's int64)
- Graph output doesn't support i64. Only fp16, fp32, i32 is supported.

Given int32 is the intersection across these backends, I see two options to make it work:
a. Update argMin/Max to always output int32
b. On the spec level, allow passing a param of `output_type`. And allow probing the supported data types for this param.

philloooo:
fixes #653

Add outputDataType to `MLArgMinMaxOptions` and defaults to `int32`. Given `opSupportLimits` is not added to spec yet, no validation steps is done to `outputDataType` right now.

@fdwr @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/webnn/pull/730.html" title="Last updated on Jul 23, 2024, 12:06 AM UTC (26aa1fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/730/92ff390...philloooo:26aa1fd.html" title="Last updated on Jul 23, 2024, 12:06 AM UTC (26aa1fd)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1517,6 +1517,7 @@ Return the index location of the minimum or maxmium values of all the input valu
 <script type=idl>
 dictionary MLArgMinMaxOptions {
   boolean keepDimensions = false;
+  MLOperandDataType outputDataType = "int32";
 };
 
 partial interface MLGraphBuilder {
@@ -1532,6 +1533,10 @@ partial interface MLGraphBuilder {
     : <dfn>keepDimensions</dfn>
     ::
         If true, retains reduced dimensions with [=list/size=] 1.
+    
+    : <dfn>outputDataType</dfn>
+    ::
+        An {{MLOperandDataType}}. The output data type.
 </dl>
 
 <div dfn-for="MLGraphBuilder/argMin(input, axis, options), MLGraphBuilder/argMax(input, axis, options)" dfn-type=argument>
@@ -1552,7 +1557,7 @@ partial interface MLGraphBuilder {
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |outputShape| be the result of [=MLGraphBuilder/calculating reduction output sizes=] given |input|'s [=MLOperand/shape=], « |axis| », and |options|.{{MLArgMinMaxOptions/keepDimensions}}. If that returns failure, then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be a new {{MLOperandDescriptor}}.
-    1. Set |desc|.{{MLOperandDescriptor/dataType}} to {{MLOperandDataType/"int64"}}.
+    1. Set |desc|.{{MLOperandDescriptor/dataType}} to |options|.{{MLArgMinMaxOptions/outputDataType}}.
     1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
     1. *Make graph connections:*
         1. Let |operator| be an [=operator=] for the |op| operation, given |options|.
--- a/index.bs
+++ b/index.bs
@@ -1545,7 +1545,7 @@ partial interface MLGraphBuilder {
         - <dfn>axis</dfn>: The dimension to reduce. The value must be in the range [0, N-1] where N is the [=MLOperand/rank=] of the input tensor.
         - <dfn>options</dfn>: an optional {{MLArgMinMaxOptions}}. The optional parameters of the operation.
 
-    **Returns:** an {{MLOperand}}. The N-D tensor of the reduced shape. The values must be of type {{MLOperandDataType/"int64"}} in the range [0, N-1] where N is the size of the input dimension specified by axis.
+    **Returns:** an {{MLOperand}}. The N-D tensor of the reduced shape. The values must be of type |options|.{{MLArgMinMaxOptions/outputDataType}} in the range [0, N-1] where N is the size of the input dimension specified by axis.
 </div>
 
 <details open algorithm>
--- a/index.bs
+++ b/index.bs
@@ -1555,6 +1555,7 @@ partial interface MLGraphBuilder {
     1. [=Assert=]: |op| is one of "argMin", "argMax".
     1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1517,6 +1517,7 @@ Return the index location of the minimum or maxmium values of all the input valu
 <script type=idl>
 dictionary MLArgMinMaxOptions {
   boolean keepDimensions = false;
+  MLOperandDataType outputDataType = "int32";
```
fdwr, line 1520: `uint32`? That way:
- it's consistent with WebNN in general, where everything related to axes and dimensions is uint32 (the one notable exception being gather which allows negative indices).
- it's possible to represent the full range of a single 1D tensor (otherwise we lose half the addressable space).
- when going from int64 down to uint32, we're less likely to chop of the upper bit.

Of course, passing `int32` would still be valid for CoreML usage.

fdwr: Let's see...
- DML accepts both uint32 and int32: https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_argmin_operator_desc
- TF evidently only accepts output_type int32 (not uint32) https://www.tensorflow.org/api_docs/python/tf/math/argmin.
- CoreML only returns int32 https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.reduction.reduce_argmin.

Well shoot, I guess it should stay `int32` due to backend limitations, meaning we'll only be able to access half of the addressable space.

fdwr (approved):
👍 Thanks Phillis. I have a thought on what the *default* should be, but otherwise LGTM.

index.bs
philloooo: thanks @fdwr  I see your comments in the [CL](https://chromium-review.googlesource.com/c/chromium/src/+/5692538/comment/062884c1_054aad70/)  too. I will reply it here. 

> should the default parameter of an API favor that of the caller viewpoint (frontends) or the implementations (backends)?

If `int64` is emulatable on CoreML then we could possibly favor values that make it easier for callers. The problem is `int64` is not emulatable on CoreML so it will just fail, breaking browser compatibility as [Reily mentioned](https://chromium-review.googlesource.com/c/chromium/src/+/5692538/comment/062884c1_054aad70/). So I think it's better to default to a value that just works on all implementations?

huningxin (commented):
> **Returns:** an {{MLOperand}}. The N-D tensor of the reduced shape. The values must be of type {{MLOperandDataType/"int64"}} in the range [0, N-1] where N is the size of the input dimension specified by axis.

This prose needs to be updated as well. What would be the range for `int32` output value?

philloooo:
@huningxin  i think it would still be [0, N-1] range?  The dimensions are of type uint32_t  so it could exceed int32_t range, but I don't think in reality we have tensors that big?

this also expose another issue - both CoreML and tflite backends actually use int32_t for tensor dimensions so if we have tensor that needs to use that extra bit from uint32_t they won't work on these backends.  Should we actually consider use int32_t for dimensions?

a-sully:
> I haven't seen any tensor of a model has such big dimension.

While we closed issue #456, this table is still relevant: https://github.com/webmachinelearning/webnn/issues/456#issuecomment-1692533722 (and the max for Core ML is 5D).  The dimensions will always be non-negative and within `MAX_INT`

a-sully:
> While we closed issue #456, this table is still relevant: [#456 (comment)](https://github.com/webmachinelearning/webnn/issues/456#issuecomment-1692533722) (and the max for Core ML is 5D). The dimensions will always be non-negative and within `MAX_INT`

Apologies, that table is referring to max _rank_ while we're talking about the max _dimension_ here

> Yes, then would it be [0, min(N-1, MAX_INT)] range? And should we add a validation step to ensure the size of the dimension being reduced can be indexed by that range?

+1, though we should consider [0, `MAX_INT`] (or even something less than `MAX_INT`). In that case, we could make WebNN's argmin/argmax return just `int32` or `uint32` and backends which natively return other types can losslessly `cast` the result to `int32` (which might be able to be optimized away in many cases)

philloooo:
@huningxin I've created a separate issue #734 for int32 vs uint32.
For this issue for now, I've added a validation step you suggested, which we can remove later when #734 is resolved. Given the validation step, we can keep the return value to be still range of [0, N-1]

index.bs
```diff
@@ -1555,6 +1555,7 @@ partial interface MLGraphBuilder {
     1. [=Assert=]: |op| is one of "argMin", "argMax".
     1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If |input|'s [=MLOperand/shape=][|axis|] is greater than |options|.{{MLArgMinMaxOptions/outputDataType}}'s maximum value, [=exception/throw=] a {{TypeError}}.
```
huningxin, line 1558: Nit: As the index starts from 0, `MAX_INT + 1` size dimension would be addressable by `int32_t` index?

philloooo: that's technically true.. but in reality, the outputDataType usually matches with the dimension's constraint, for CoreML case, the shape[axis] also won't be able to have MAX_INT32 + 1 as it's of type int32_t..

huningxin: > for CoreML case, the shape[axis] also won't be able to have MAX_INT32 + 1 as it's of type int32_t. So I think it's better to leave as <= output_type_max_value?

That makes sense. Thanks for the explanation.

huningxin (approved):
LGTM!

fdwr (approved):
👍

PR #738, opened 2024-07-23
Update dimension valid range to signed integer

### LINKED ISSUE #734: Restrict dimensions size to [0, INT32_MAX]
Currently we use uint32 for dimensions, so the range is [0, UINT32_MAX]. 

Both CoreML and tflite backends actually use int32_t for dimensions, so the range can't exceed `INT32_MAX`, meaning if we pass dimensions that's between INT32_MAX and UINT32_MAX they faill on these backends.

We [changed from int32 to uint32](https://github.com/webmachinelearning/webnn/issues/279) to add the non-negative constraint, but it gaves an extra bit for the dimensions range that's not supported for two of the backends. 

How can we make this better? 

## option 1 - change back to int32
Specifies in spec validation steps that they should be non negative.

## option 2 - keep uint32
Specifies in spec validation steps that they can't exceed INT32_MAX

thoughts? @huningxin @fdwr @inexorabletash

philloooo:
fixes #734

Update `valid dimension` definition to be in the range of `long` and greater than 0.

@fdwr @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/webnn/pull/738.html" title="Last updated on Jul 23, 2024, 7:41 PM UTC (11c994f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/738/dbb9d1c...philloooo:11c994f.html" title="Last updated on Jul 23, 2024, 7:41 PM UTC (11c994f)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1071,7 +1071,7 @@ dictionary MLOperandDescriptor {
 </details>
 
 <p>
-A <dfn>valid dimension</dfn> is an integer greater than zero in the range of {{unsigned long}}. Implementations may impose a smaller upper bound.
+A <dfn>valid dimension</dfn> is an integer greater than zero and in the range of {{long}}. Implementations may impose a smaller upper bound.
 </p>
 
 Issue(391): Should 0-size dimensions be supported?


huningxin (approved):
LGTM!

fdwr:
It occurs to me the existing verbiage "Implementations may impose a smaller upper bound" technically *already* covered this case, since CoreML and TFLite impose a smaller bound (2 billion) than the DML EP (4 billion). Though, I'm content with this restriction, since 2 billion is well more than enough for a single dimension (never saw a model that even came close to that in half a decade for a single tensor).

fdwr (approved):

PR #741, opened 2024-07-24
Clarify reduction with empty axes and scalar input

### LINKED ISSUE #740: Question on reduction with empty axes and scalar input
In [`MLReduceOptions`](https://www.w3.org/TR/webnn/#dictdef-mlreduceoptions)
- If empty, no dimensions are reduced, and the shape of the output tensor is the same as the shape of the input tensor.

Does that mean it should be an no-op (a.k.a `identity`)?

Is that true when you specifies empty axes for scalar input too?

If that's the case, our [WPTs are wrong](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/webnn/resources/test_data/reduce_sum_square.json;l=38?q=reduce_sum_square) - it's clearly still reducing?

inexorabletash:
Improve the description of the reduction ops:

- Give in-line definition for L1/L2.
- Make "input values" slightly clearer.
- Explicitly note that if the input is a scalar, so is the output.
- State how the reduction function is applied for empty axes.
- Standardize on "reduction operation" not "reduce operation".
- Don't style list of reduction ops as non-normative note.

Fixes #740

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/741.html" title="Last updated on Jul 30, 2024, 5:22 PM UTC (f982418)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/741/8734bea...inexorabletash:f982418.html" title="Last updated on Jul 30, 2024, 5:22 PM UTC (f982418)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4975,7 +4975,7 @@ partial interface MLGraphBuilder {
     **Returns:** an {{MLOperand}}. The reduced output tensor. If the input operand is a scalar, the reduction function is applied to the scalar value and the output is also a scalar.
 </div>
 
-<div class="note">
+<div>
     **Reduction types:**
         - *L1*: Compute the <a href="https://mathworld.wolfram.com/L1-Norm.html">L1 norm</a>, the sum of the absolute value of all the input values.
         - *L2*: Compute the <a href="https://mathworld.wolfram.com/L2-Norm.html">L2 norm</a>, the square root of the sum of the square of all the input values.
--- a/index.bs
+++ b/index.bs
@@ -5030,7 +5030,7 @@ partial interface MLGraphBuilder {
 
 <details open>
   <summary>
-    The following reduce algorithms are supported.
+    The following reduction algorithms are supported.
   </summary>
     <div algorithm>
     The <dfn method for=MLGraphBuilder>reduceL1(|input|, |options|)</dfn> method steps are:
--- a/index.bs
+++ b/index.bs
@@ -1115,7 +1115,7 @@ dictionary MLOperandDescriptor {
 </details>
 
 <p>
-A <dfn>valid dimension</dfn> is an integer greater than zero in the range of {{unsigned long}}. Implementations may impose a smaller upper bound.
+A <dfn>valid dimension</dfn> is an integer greater than zero and in the range of {{long}}. Implementations may impose a smaller upper bound.
 </p>
 
 Issue(391): Should 0-size dimensions be supported?
--- a/index.bs
+++ b/index.bs
@@ -5005,16 +5005,16 @@ partial interface MLGraphBuilder {
 
 <div>
     **Reduction types:**
-        - *L1*: Compute the <a href="https://mathworld.wolfram.com/L1-Norm.html">L1 norm</a>, the sum of the absolute value of all the input values.
-        - *L2*: Compute the <a href="https://mathworld.wolfram.com/L2-Norm.html">L2 norm</a>, the square root of the sum of the square of all the input values.
-        - *LogSum*: Compute the log value of the sum of all the input values.
-        - *LogSumExp*: Compute the log value of the sum of the exponent of all the input values.
-        - *Max*: Compute the maximum value of all the input values.
-        - *Mean*: Compute the average value of all the input values.
-        - *Min*: Compute the minimum value of all the input values.
-        - *Product*: Compute the product of all the input values.
-        - *Sum*: Compute the sum of all the input values.
-        - *SumSquare*: Compute the sum of the square of all the input values.
+        - *L1*: Compute the <a href="https://mathworld.wolfram.com/L1-Norm.html">L1 norm</a>, the sum of the absolute value of the input values.
+        - *L2*: Compute the <a href="https://mathworld.wolfram.com/L2-Norm.html">L2 norm</a>, the square root of the sum of the square of the input values.
+        - *LogSum*: Compute the log value of the sum of the input values.
+        - *LogSumExp*: Compute the log value of the sum of the exponent of the input values.
+        - *Max*: Compute the maximum value of the input values.
+        - *Mean*: Compute the average value of the input values.
+        - *Min*: Compute the minimum value of the input values.
+        - *Product*: Compute the product of the input values.
+        - *Sum*: Compute the sum of the input values.
+        - *SumSquare*: Compute the sum of the square of the input values.
 </div>
 
 <details open algorithm>
--- a/index.bs
+++ b/index.bs
@@ -4934,7 +4934,8 @@ partial interface MLGraphBuilder {
 </div>
 
 ### Reduction operations ### {#api-mlgraphbuilder-reduce}
-Reduce the input tensor along all dimensions, or along the axes specified in the {{MLReduceOptions/axes}}  array parameter. For each specified axis, the dimension with that index is reduced, i.e. the resulting tensor will not contain it, unless the {{MLReduceOptions/keepDimensions}} option is specified. The values of the resulting tensor are calculated using the specified reduction function that takes as parameters all the values across the reduced dimension.
+Reduce the input tensor along all dimensions, or along the axes specified in the {{MLReduceOptions/axes}}  array parameter. For each specified axis, the dimension with that index is reduced, i.e. the resulting tensor will not contain it, unless the {{MLReduceOptions/keepDimensions}} option is specified. The values of the resulting tensor are calculated using the specified reduction function that takes as parameters all the input values across the reduced dimensions.
+
 <script type=idl>
 dictionary MLReduceOptions {
   sequence<[EnforceRange] unsigned long> axes;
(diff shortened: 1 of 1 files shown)


inexorabletash:
@philloooo - does this make the behavior clearer?

index.bs
```diff
@@ -4971,21 +4972,21 @@ partial interface MLGraphBuilder {
...
-        - *L2*: Compute the <a href="https://mathworld.wolfram.com/L2-Norm.html">L2 norm</a> of all the input values along the axes.
-        - *LogSum*: Compute the log value of the sum of all the input values along the axes.
-        - *LogSumExp*: Compute the log value of the sum of the exponent of all the input values along the axes.
-        - *Max*: Compute the maximum value of all the input values along the axes.
-        - *Mean*: Compute the average value of all the input values along the axes.
-        - *Min*: Compute the minimum value of all the input values along the axes.
-        - *Product*: Compute the product of all the input values along the axes.
-        - *Sum*: Compute the sum of all the input values along the axes.
-        - *SumSquare*: Compute the sum of the square of all the input values along the axes.
+        - *L1*: Compute the <a href="https://mathworld.wolfram.com/L1-Norm.html">L1 norm</a>, the sum of the absolute value of all the input values.
```
huningxin, line 4980: When we say "all the input values", is it only for "all dimensions are reduced" case?

inexorabletash: No, so _"input values"_ is not clear, I guess. I felt "along the axes" was unclear (since it can be no axes), so I was wanting "input values" to mean all input values under consideration, as specified by the axes parameter. Up top it says _"The values of the resulting tensor are calculated using the specified reduction function that takes as parameters all the input values across the reduced dimensions."_

I'll make the axes definition more explicit.

We could make it clearer that axes are what

inexorabletash: Take a look at 7fbceaa - does that help?

huningxin (approved):
LGTM!

index.bs
huningxin: Yes, that looks much clearer, thanks!

index.bs
```diff
@@ -4959,7 +4960,13 @@ partial interface MLGraphBuilder {
 <dl dfn-type=dict-member dfn-for=MLReduceOptions>
     : <dfn>axes</dfn>
     ::
-        The dimensions to reduce. The values in the list must be in the range [0, N-1] where N is the [=MLOperand/rank=] of the input tensor. If not present, all dimensions are reduced. If empty, no dimensions are reduced, and the shape of the output tensor is the same as the shape of the input tensor.
+        The dimensions to reduce, which also specifies which of the values in the input tensor are used as input values to the reduction function. The values in the list must be in the range [0, N-1] where N is the [=MLOperand/rank=] of the input tensor.
```
fdwr, line 4963:
```suggestion
        The dimensions to reduce, which also specifies which of the values in the input tensor are used with the reduction function. The axes in the list must be in the range [0, N-1] where N is the [=MLOperand/rank=] of the input tensor.
```
Propose avoiding the word "values" for the axis list because "values" was just used in the previous sentence in a very different context, meaning the elements inside the input tensor, whereas values here actually means the axes. Alternately, "The axis values must be..." could work.

Little repetitive: "...which of the *values* in the *input* are used as *input* *values*...".

index.bs
```diff
@@ -4934,7 +4934,8 @@ partial interface MLGraphBuilder {
 </div>
 
 ### Reduction operations ### {#api-mlgraphbuilder-reduce}
-Reduce the input tensor along all dimensions, or along the axes specified in the {{MLReduceOptions/axes}}  array parameter. For each specified axis, the dimension with that index is reduced, i.e. the resulting tensor will not contain it, unless the {{MLReduceOptions/keepDimensions}} option is specified. The values of the resulting tensor are calculated using the specified reduction function that takes as parameters all the values across the reduced dimension.
+Reduce the input tensor along all dimensions, or along the axes specified in the {{MLReduceOptions/axes}}  array parameter. For each specified axis, the dimension with that index is reduced, i.e. the resulting tensor will not contain it, unless the {{MLReduceOptions/keepDimensions}} option is specified. The values of the resulting tensor are calculated using the specified reduction function that takes as parameters all the input values across the reduced dimensions.
```
fdwr, line 4937:
```suggestion
Reduce the input tensor along all dimensions, or along the axes specified in the {{MLReduceOptions/axes}} array parameter. For each specified axis, the dimension with that index is reduced, i.e. the resulting tensor will not contain it, unless the {{MLReduceOptions/keepDimensions}} option is specified. The values of the resulting tensor are calculated using the specified reduction function that takes as parameters all the input values across the reduced dimensions.
```
(nit extra space)

index.bs
```diff
@@ -4971,21 +4978,21 @@ partial interface MLGraphBuilder {
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
         - <dfn>options</dfn>: an optional {{MLReduceOptions}}. The optional parameters of the operation.
 
-    **Returns:** an {{MLOperand}}. The reduced output tensor.
+    **Returns:** an {{MLOperand}}. The reduced output tensor. If the input operand is a scalar, the reduction function is applied to the scalar value and the output is also a scalar.
```
fdwr, line 4981:
```suggestion
    **Returns:** an {{MLOperand}}. The reduced output tensor. If the input operand is a scalar, the reduction function is applied to the scalar value, and the output is also a scalar.
```
Comma before coordinate conjunction around independent clauses reduces [Garden-path sentence misparsing](https://en.wikipedia.org/wiki/Garden-path_sentence).

index.bs
```diff
@@ -4971,21 +4978,21 @@ partial interface MLGraphBuilder {
...
-        - *L2*: Compute the <a href="https://mathworld.wolfram.com/L2-Norm.html">L2 norm</a> of all the input values along the axes.
-        - *LogSum*: Compute the log value of the sum of all the input values along the axes.
-        - *LogSumExp*: Compute the log value of the sum of the exponent of all the input values along the axes.
-        - *Max*: Compute the maximum value of all the input values along the axes.
-        - *Mean*: Compute the average value of all the input values along the axes.
-        - *Min*: Compute the minimum value of all the input values along the axes.
-        - *Product*: Compute the product of all the input values along the axes.
-        - *Sum*: Compute the sum of all the input values along the axes.
-        - *SumSquare*: Compute the sum of the square of all the input values along the axes.
+        - *L1*: Compute the <a href="https://mathworld.wolfram.com/L1-Norm.html">L1 norm</a>, the sum of the absolute value of all the input values.
```
fdwr, line 4986: 👍 I appreciate this additional clarity for L1 and L2. (resolve me)

index.bs
```diff
@@ -4971,21 +4978,21 @@ partial interface MLGraphBuilder {
...
+        - *L2*: Compute the <a href="https://mathworld.wolfram.com/L2-Norm.html">L2 norm</a>, the square root of the sum of the square of all the input values.
+        - *LogSum*: Compute the log value of the sum of all the input values.
+        - *LogSumExp*: Compute the log value of the sum of the exponent of all the input values.
+        - *Max*: Compute the maximum value of all the input values.
+        - *Mean*: Compute the average value of all the input values.
```
fdwr, line 4991: Removing "along the axes" now makes this "Compute the average value of all the input values" sound like it finds the average of *all* values in the input tensor 🤨, but it only finds the average within each related *group* of shared input values (sharing the same axis, of those being reduced). Impetus to remove it?

Proposals:
- Remove "all":
```suggestion
        - *Mean*: Compute the average value of the input values.
```
- Add "related":
```suggestion
        - *Mean*: Compute the average value of all the related input values.
```
- Restore "along the axes":
```suggestion
        - *Mean*: Compute the average value of all the input values along the axes.
```

index.bs
```diff
@@ -5008,7 +5015,7 @@ partial interface MLGraphBuilder {
 
 <details open algorithm>
   <summary>
-    To <dfn for="MLGraphBuilder" data-lt="reduce-op">create reduce operation</dfn> given [=string=] |op|, {{MLOperand}} |input|, {{MLReduceOptions}} |options|, and optional [=/list=] |allowedDataTypes|, run the following steps:
+    To <dfn for="MLGraphBuilder">create reduction operation</dfn> given [=string=] |op|, {{MLOperand}} |input|, {{MLReduceOptions}} |options|, and optional [=/list=] |allowedDataTypes|, run the following steps:
```
fdwr, line 5018: 👍 Yeah, "reduction" sounds more natural in full sentence context, as it matches the other noun-form "binary operation" and noun "pooling".

- The following reduce algorithms are supported.
- The following pooling algorithms are supported.
- The following argMin/argMax algorithms are supported.
- The element-wise binary operation algorithms ... as follows.
- The element-wise logical operation algorithms ... as follows.

(now, that does leave argMin/argMax feeling a little out of place from the rest, but "The following argMinization/argMaximization algorithms" sounds verbose :b).

(resolve me)

fdwr (commented):
Clarification is goodness 👍. Some wording thoughts.

index.bs
inexorabletash: How about "selected input values" ?

The impetus for removing "along the axes" is that if no axes are selected the operation still happens. Plus it's very repetitive.

index.bs
inexorabletash: My intention here was to define (casually, not a full `<dfn>`) the "input values" term, which would then be used later on. And make it explicit that there are values that go through the reduction function, even if there are no axes specified. But yes,"values" here is confusing.

Obviously I wasn't successful. :stuck_out_tongue:

index.bs
inexorabletash: I like removing "all". Let's try that.

inexorabletash: I think more explanation of how reduction works across axes is good, but ISTM that keeping these definitions of the different functions concise is important. Anything more we want to add to text around the axes parameter?

fdwr (approved):
👍

index.bs
fdwr: > I like removing "all". Let's try that. [f982418](https://github.com/webmachinelearning/webnn/commit/f9824188789c2c0653764c1538a0db799978179c)

Fine with me.

> Anything more we want to add to text around the axes parameter?

Maybe later gator, but happy to merge.

Issue #378, opened 2023-04-11
Add samples of shape broadcasting 


huningxin:
This was proposed by @wacky6 in Chromium [CL review of `prelu`](https://chromium-review.googlesource.com/c/chromium/src/+/4396686/comment/f02acaeb_3c2795f2/). Thanks Jiewei!

Some WebNN ops support broadcasting, including [Element-wise binary operations](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-binary) and [`prelu`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-prelu).

Although WebNN spec refers to [General Broadcasting Rules of NumPy](https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html#general-broadcasting-rules), it would be helpful to add some samples of the acceptable combinations in the informative section.

ONNX does a great job by [Broadcasting in ONNX](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md) doc.

a-sully:
I discovered this issue after seeing @inexorabletash's comment here https://github.com/webmachinelearning/webnn/pull/523/files#r1466950307. In the spirit of https://github.com/webmachinelearning/webnn/issues/462, it would be nice to explicitly define shape broadcasting, rather than having it "implementation defined" [as it is currently specced](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-broadcast-shapes)

(apologies if this is a naive question, but) **is there anything preventing us from specifying broadcasting?**

A spec is better than examples, though both would be best :) (ideally with the examples in the form of web platform tests!)

a-sully:
Looking into this a bit more, this was discussed a bit in https://github.com/webmachinelearning/webnn/issues/324. That issue is mostly focused on _which input layouts should be supported_ ("NHWC", "NCHW", both, etc). This is tangential to the question of how broadcasting should work on the various layouts.

From https://github.com/webmachinelearning/webnn/issues/324#issuecomment-1502761644 by @wacky6

> Layout support comes up in MLOperand implementation that allows data shape broadcasting. https://chromium-review.googlesource.com/c/chromium/src/+/4396686/comment/f02acaeb_3c2795f2/
> 
> Supporting both channel-first and channel-last layout will complicate spec steps and implementation because the current numpy broadcast rule implies right-most first broadcast.
> 
> Example: caller wants to apply a per-channel multiplication.
> 
[...]

PR #743, opened 2024-07-26
Make prelu() bidirectionally broadcast, improve broadcast wording

### LINKED ISSUE #590: Consider adopting new broadcasting rules
Reviving [a discussion](https://github.com/webmachinelearning/webnn/pull/534#issuecomment-1912885281) from #534, which defined shape broadcasting but didn't touch on the question of what WebNN's shape broadcasting rules _should_ be

WebNN currently specifies two kinds of broadcasting rules: [unidirectional](https://webmachinelearning.github.io/webnn/#unidirectionally-broadcasting-the-shapes) and [bidirectional](https://webmachinelearning.github.io/webnn/#bidirectionally-broadcasting-the-shapes)

Of the popular ML frameworks, ONNX (which WebNN is largely based on) appears to be an outlier in making a distinction between "unidirectional" and "multidirectional" broadcasting. This distinction is not made by:
- Numpy: https://numpy.org/doc/stable/user/basics.broadcasting.html
- PyTorch: https://pytorch.org/docs/stable/notes/broadcasting.html (which references Numpy)
- Tensorflow/XLA/StableHLO: https://www.tensorflow.org/xla/broadcasting (which also references Numpy, though it has some other restrictions)

The ["unidirectional broadcastable"](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md#unidirectional-broadcasting) constraint of some ONNX ops (e.g. [`prelu()`](https://github.com/onnx/onnx/blob/main/docs/Operators.md#PRelu)) requires workarounds when exporting from other formats to ONNX - like in this example of using TVM to export PyTorch to ONNX: https://github.com/pytorch/pytorch/issues/70570#issuecomment-1034379620.

What should we do?

## Option 1: Adopt [Numpy's broadcasting rules](https://numpy.org/doc/stable/user/basics.broadcasting.html)

Rationale: Numpy's broadcasting rules are a standard across the industry. It seems reasonable to be what we expose to the web

Outcome: "bidirectional broadcasting" will be the only type of broadcasting exposed to the web. The user agent muse ensure that the constraints of the underlying framework  - such as unidirectional broadcasting for ONNX (@fdwr has [suggested](https://github.com/webmachinelearning/webnn/pull/534#issuecomment-1936896645) that this is trivial), and lack of inferred broadcasting specifications for XLA (more on that below) - are satisfied.

## Option 2: Adopt [XLA's broadcasting rules](https://www.tensorflow.org/xla/broadcasting)

Rationale: The [XLA Principles](https://openxla.org/xla/broadcasting#principles) apply to WebNN, too:

> The XLA language is as strict and explicit as possible, avoiding implicit "magical" features. Such features might make some computations slightly easier to define, but at the cost of more assumptions baked into user code that will be difficult to change in the long term. If necessary, implicit magical features can be added in client-level wrappers... With regard to broadcasting, XLA requires explicit broadcasting specifications on operations between arrays of different ranks. This is different from NumPy, which infers the specification when possible.

Outcome: Both "unidirectional broadcasting" and "bidirectional broadcasting" concepts would be removed from the WebNN spec. To facilitate explicit broadcasts, something like [StableHLO's `broadcast_in_dim` op](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#broadcast_in_dim) would need to be added to WebNN

## Option 3: Keep the status quo

Rationale: It's the status quo

Outcome: No action needed regarding the current spec. However, all models ported to WebNN will need to abide by this "unidirectionally broadcastable" constraint which is specific to ONNX

inexorabletash:
- Just say "broadcast" not "broadcasted"; Merriam-Webster says the former is preferable.

- Make prelu() bidirectionally broadcast instead of unidirectionally broadcast, as it's not an obvious unidirectional case.

- Make preamble more consistent across operations in calling out broadcasting and stating the requirements for input tensors.

- Shorten algorithm link, e.g. from "...broadasting the shapes A's shape and B's shape" to just "...broadcasting A's shape and B's shape"

Resolves #590

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -2262,10 +2262,7 @@ partial interface MLGraphBuilder {
 ### Element-wise binary operations ### {#api-mlgraphbuilder-binary}
 Compute the element-wise binary addition, subtraction, multiplication, division, power, maximum and minimum of the two input tensors.
 
-The element-wise binary operations will be broadcasted according to
-[[!numpy-broadcasting-rule]]. The [=MLOperand/rank=] of the output tensor is the maximum
-[=MLOperand/rank=] of the input tensors. For each dimension of the output tensor, its size
-is the maximum size along that dimension of the input tensors.
+The operation will be broadcast according to [[!numpy-broadcasting-rule]]. The input tensors must be [=bidirectionally broadcastable=]. The [=MLOperand/rank=] of the output tensor is the maximum [=MLOperand/rank=] of the input tensors. For each dimension of the output tensor, its size is the maximum size along that dimension of the input tensors.
 
 <script type=idl>
 partial interface MLGraphBuilder {
@@ -2306,7 +2303,7 @@ partial interface MLGraphBuilder {
     1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |a| and |b| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |a|'s [=MLOperand/dataType=] is not equal to |b|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
-    1. Let |outputShape| be the result of [=bidirectionally broadcasting the shapes=] |a|'s [=MLOperand/shape=] and |b|'s [=MLOperand/shape=].
+    1. Let |outputShape| be the result of [=bidirectionally broadcasting=] |a|'s [=MLOperand/shape=] and |b|'s [=MLOperand/shape=].
         1. If that returns failure, then [=exception/throw=] a {{TypeError}}.
     1. Let |descriptor| be the result of [=creating an MLOperandDescriptor=] given |a|'s [=MLOperand/dataType=] and |outputShape|.
     1. *Make graph connections:*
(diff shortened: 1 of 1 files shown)


inexorabletash:
It seemed like this was the only actionable work for #590 

I was pondering adding informative notes to the algorithms e.g. explaining when/where bidirectional/unidirectional broadcasting should be used, but inspiration didn't strike. Suggestions welcome!

inexorabletash:
@a-sully and @fdwr can you take a look?

inexorabletash:
... and if we want more text about broadcasting, I sketched something out in https://github.com/webmachinelearning/webnn/issues/378#issuecomment-2253655000 - I could pull that into this PR if desired.

a-sully:
LGTM 👍  I especially like _s/broadasting the shapes/broadcasting_

index.bs
```diff
@@ -4871,7 +4869,7 @@ partial interface MLGraphBuilder {
 <div dfn-for="MLGraphBuilder/prelu(input, slope)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>slope</dfn>: an {{MLOperand}}. The slope tensor. Its shape is either the same as, or [=unidirectionally broadcastable=] to the shape of input tensor *input*.
+        - <dfn>slope</dfn>: an {{MLOperand}}. The slope tensor. It shape must mbe [=bidirectionally broadcastable=] to the shape of *input*.
```
fdwr, line 4872:
```suggestion
        - <dfn>slope</dfn>: an {{MLOperand}}. The slope tensor. It shape must be [=bidirectionally broadcastable=] to the shape of *input*.
```

index.bs
```diff
@@ -5832,8 +5830,7 @@ partial interface MLGraphBuilder {
 ### where ### {#api-mlgraphbuilder-where}
 Select the values from the trueValue or the falseValue tensor depending on the corresponding values of the condition tensor, where non-zero is true and zero is false. The condition tensor is often the output of one of the element-wise logical operations.
 
-The input tensors must be [=bidirectionally broadcastable=] and will be broadcasted according to [[!numpy-broadcasting-rule]] to the final output shape. The [=MLOperand/rank=] of the output tensor is the maximum [=MLOperand/rank=] of the input tensors.
-For each dimension of the output tensor, its size is the maximum size along that dimension of the input tensors.
+The operation will be broadcast according to [[!numpy-broadcasting-rule]]. The input tensors must be [=bidirectionally broadcastable=]. The [=MLOperand/rank=] of the output tensor is the maximum [=MLOperand/rank=] of the input tensors. For each dimension of the output tensor, its size is the maximum size along that dimension of the input tensors.
```
fdwr, line 5833: "The past tense of broadcast is broadcast". Sigh, English.

fdwr (approved):
👍

fdwr (approved):

huningxin (approved):
LGTM!

Issue #477, opened 2023-11-06
API lacks handling for async ML device errors on the context


bbernhar:
What happens if a WebNN operation dispatched through `MLContext` encounters some internal error which causes the GPU device to get removed? 

I would expect WebNN to provide a spec into how fatal (device) errors are handled so the WebNN developer could respond appropriately. If we want to do more with `MLContext` (ex. create buffers), I believe we'll need a more robust error mechanism like WebGPU [1].

[1] https://www.w3.org/TR/webgpu/#errors-and-debugging

bbernhar:
@anssiko this issue is for non-interop and still needs a follow-up proposal (low priority atm).

zolkis:
Or, should we add plain error event(s) to MLContext?
Was there a specific design reason for the WebGPU type of error handling, i.e. not using error events, but tracking a promise on lost()?

> After a device is [lost](https://www.w3.org/TR/webgpu/#lose-the-device) (described below), errors are no longer surfaced. (...) Additionally, no errors are generated by the device loss itself. Instead, the [GPUDevice](https://www.w3.org/TR/webgpu/#gpudevice).[lost](https://www.w3.org/TR/webgpu/#dom-gpudevice-lost) promise resolves to indicate the device is lost.

inexorabletash:
A promise has slightly better ergonomics because (1) the transition to "lost" only happens once for a device, and (2) it works even if your code runs after the state has changed; you aren't forced to add an event listener immediately.

https://www.w3.org/2001/tag/doc/promises-guide#when-to-use

anssiko:
@bbernhar if you agree with the general direction of the proposal https://github.com/webmachinelearning/webnn/issues/778 please feel free to close this issue to focus our attention.

bbernhar:
@anssiko SGTM to move the discussion there.

PR #744, opened 2024-07-29
Specify destroy() methods and behavior around context loss and error reporting

inexorabletash:
Based on @mingmingtasd's work in the Chromium prototype implementation. 

`MLContext` gains a `destroy()` method and a `lost` promise attribute. Calling `destroy()` releases resources for the context itself and any associated `MLGraph`s and `MLTensor`s, and any outstanding `build()` requests for an associated `MLGraphBuilder` are aborted. The `lost` attribute then resolves, and any future use of the context fails. If the context is lost for other reasons, the same behavior occurs (releasing associated resources), and the `lost` promise provides an implementation-defined message explaining the reason. 

`MLGraph` similarly gains a `destroy()` attribute. When destroyed - either by an explicit call, or because the associated context is lost - then any outstanding `dispatch()` requests using the graph are aborted.

This also modifies the omnipresent "has built" tests on `MLGraphBuilder` methods to be a "can built" test which also checks that the builder's context is not lost. 

For #477

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/744.html" title="Last updated on Jan 14, 2025, 1:18 AM UTC (f59890c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/744/df555e1...inexorabletash:f59890c.html" title="Last updated on Jan 14, 2025, 1:18 AM UTC (f59890c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -986,10 +986,11 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
         1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
     1. If [=validating tensors with descriptors=] given |inputs| and |graph|.{{MLGraph/[[inputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. If [=validating tensors with descriptors=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
-    1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
-        1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
+    1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
 
-            Issue(778): Add a mechanism for reporting errors during graph execution.
+                Issue(778): Add a mechanism for reporting errors during graph execution.
 
     1. Return {{undefined}}.
 </details>
@@ -1068,11 +1069,12 @@ Creates an {{MLTensor}} associated with this {{MLContext}}.
     1. If [=this=] [=MLContext/is lost=], then return [=a new promise=] [=rejected=] with an "{{InvalidStateError}}" {{DOMException}}.
     1. Let |tensor| be the result of [=creating an MLTensor=] given [=this=], and |descriptor|.
     1. Let |promise| be [=a new promise=].
-    1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
-        1. Create |tensor|.{{MLTensor/[[data]]}} given |descriptor| and initialize all bytes to zeros.
-        1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
-        1. Otherwise, [=queue an ML task=] with |global| to [=resolve=] |promise| with |tensor|.
-    1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
+    1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Create |tensor|.{{MLTensor/[[data]]}} given |descriptor| and initialize all bytes to zeros.
+            1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+            1. Otherwise, [=queue an ML task=] with |global| to [=resolve=] |promise| with |tensor|.
+        1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
     1. Return |promise|.
 </details>
 
@@ -1097,11 +1099,12 @@ Reads back the {{MLTensor/[[data]]}} of an {{MLTensor}} from the {{MLContext}}.{
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
-    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
-        1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
-        1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
-        1. Otherwise, [=queue an ML task=] with |global| to [=ArrayBuffer/create=] an {{ArrayBuffer}} |result| given |bytes| and |realm| and then [=resolve=] |promise| with |result|.
-    1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
+            1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+            1. Otherwise, [=queue an ML task=] with |global| to [=ArrayBuffer/create=] an {{ArrayBuffer}} |result| given |bytes| and |realm| and then [=resolve=] |promise| with |result|.
+        1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
     1. Return |promise|.
 </details>
 
@@ -1127,17 +1130,18 @@ Bring-your-own-buffer variant of {{MLContext/readTensor(tensor)}}. Reads back th
     1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If [=validating buffer with descriptor=] given |outputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
-    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
-        1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
-        1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
-        1. Otherwise, [=queue an ML task=] with |global| and the following steps:
-            1. If |outputData| is [=BufferSource/detached=], [=reject=] |promise| with a {{TypeError}}, and abort these steps.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
+            1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+            1. Otherwise, [=queue an ML task=] with |global| and the following steps:
+                1. If |outputData| is [=BufferSource/detached=], [=reject=] |promise| with a {{TypeError}}, and abort these steps.
 
-                Note: [=Validating buffer with descriptor=] above will fail if |outputData| is detached, but it is possible that |outputData| could be detached between that step and this one.
+                    Note: [=Validating buffer with descriptor=] above will fail if |outputData| is detached, but it is possible that |outputData| could be detached between that step and this one.
 
-            1. [=ArrayBuffer/Write=] |bytes| to |outputData|.
-            1. [=Resolve=] |promise| with {{undefined}}.
-    1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
+                1. [=ArrayBuffer/Write=] |bytes| to |outputData|.
+                1. [=Resolve=] |promise| with {{undefined}}.
+        1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{InvalidStateError}}" {{DOMException}}.
     1. Return |promise|.
 </details>
 
@@ -1163,10 +1167,11 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
     1. If [=validating buffer with descriptor=] given |inputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
     1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |tensor|.{{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=].
-    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
-        1. Copy |bytes| to |tensor|.{{MLTensor/[[data]]}}.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Copy |bytes| to |tensor|.{{MLTensor/[[data]]}}.
 
-            Issue(778): Add a mechanism for reporting errors while writing to a tensor.
+                Issue(778): Add a mechanism for reporting errors while writing to a tensor.
 
     1. Return {{undefined}}.
 </details>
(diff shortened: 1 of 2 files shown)


inexorabletash:
Some points for discussion:

* This uses the internal slot `[[lost]]` as both (1) the promise value and (2) how to check if a context "is lost". Maybe an explicit boolean internal slot would be clearer, albeit more verbose?
* In the implementation, the "lost" state is equivalent to the context's mojo remote being connected. This is only checked explicitly when (1) creating an `MLGraphBuilder` and (2) creating an `MLBuffer`. This spec change does the former, but MLBuffer is not in the spec yet so that's not present. I did add an explicit check in `compute()`; I think that'll happen in the implementation indirectly because the dispatch would fail, but an explicit check seemed like a good idea?
* The `MLContext` section of the spec could be reorganized a bit. Separate PR?

@mingmingtasd and @reillyeon could you do an initial review?

index.bs
```diff
@@ -1039,6 +1050,31 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+<dl dfn-type=dict-member dfn-for=MLContextLostInfo>
+    : <dfn>message</dfn>
+    :: An [=implementation-defined=] message providing information about the error that occurred.
+</dl>
+
+<div algorithm>
+The <dfn attribute for=MLContext>lost</dfn> getter steps are to return [=this=]'s {{MLContext/[[lost]]}} {{Promise}}.
+</div>
+
+A {{MLContext}} <dfn for=MLContext>is lost</dfn> if its {{MLContext/[[lost]]}} {{Promise}} is settled.
```
reillyeon, line 1075: Can you find a definition to link for "is settled"?

index.bs
```diff
@@ -1309,6 +1345,7 @@ The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph buil
     The [=new=] <dfn constructor for=MLGraphBuilder lt="MLGraphBuilder(context)">MLGraphBuilder(|context|)</dfn> constructor steps are:
   </summary>
     1. If [=this=]'s [=relevant global object=]'s [=associated Document=] is not [=allowed to use=] the [=webnn-feature|webnn=] feature, then [=exception/throw=] a "{{SecurityError}}" {{DOMException}}.
+    1. If |context| [=MLContext/is lost=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
```
reillyeon, line 1348: For completeness this check should be attached to basically every method, but in particular the ones that require interaction with the hardware such as `build()`, `compute()` and `dispatch()`.

It's an interesting question whether if we implement the optimization @a-sully is working on for `constant()` we should have that method throw if the context is lost or just ignore the invalid state until `build()` is called.

reillyeon (changes requested):

index.bs
inexorabletash: 7456eff is the best I can do.

index.bs
inexorabletash: Added to `build()` in 01cf420; already present in `compute()`, and `dispatch()` isn't in the spec yet.

We maybe want to change the "If this.[[hasBuilt]] is true, then ..." tests to be "If this **can build**, then..." and define **can build** for a builder as ![[has built] && context is not lost as  Thoughts?

reillyeon: Oh, I like that idea!

inexorabletash: Okay, tried it out in d2ec3f4

reillyeon (approved):

mingmingtasd:
Thanks! @inexorabletash @reillyeon 
I am also working on a chromium [CL ](https://chromium-review.googlesource.com/c/chromium/src/+/5737474) to expose `MLContext::destroy` to make a context lost.

mingmingtasd (approved):

reillyeon:
Something this PR doesn't do is specify the behavior around rejecting in-flight asynchronous operations.

inexorabletash:
> Something this PR doesn't do is specify the behavior around rejecting in-flight asynchronous operations.

Thoughts on how to do that and to what level of detail?

* in `compute()`, is that handled by _execute graph_ returning error, or do we need more steps?
* in `build()` it looks like failure other than "does not support a requested feature" isn't covered.

One generic approach for both of those would be to replace: _"Queue an ML task with global to resolve promise with ..."_ with:

> 1. **Queue an ML task** with _global_ to run these steps:
>      1. If _context_ **is lost**, then reject _promise_ with an InvalidStateError.
>      2. Resolve _promise_ with ...

... which I think covers the script-observable behavior, but not that the async steps internally should fail.

fdwr:
@RafaelCintron

reillyeon:
There are two separate considerations: what happens to the promise returned by a method and what happens to an asynchronous operation itself. Operations like `build()`, `compute()` and `dispatch()` aren't abortable (or at least, the specification should not require that they are aborted, only that whether or not they are aborted is not visible to script). We can however be explicit that the promises returned by any methods on an object are synchronously rejected when the `destroy()` method is called or the context is lost. We could either specify this exactly the way the Chromium implementation works, by having each object hold a list of all promises and reject them, or we could add a note to the "in parallel" steps which says "when the context is lost or _this_ is destroyed, reject _promise_ and potentially abort these steps". The definition of "potentially abort" is the tricky one because it has to consider cases like destroying a single buffer that is part of a larger set of pending operations which should still be able to complete.

huningxin:
@inexorabletash 

> * in `compute()`, is that handled by _execute graph_ returning error, or do we need more steps?

I suppose the following step of _execute graph_ could handle the device lost error:

> Issue a compute request to graph.[[implementation]] given name and inputResources and wait for completion.

>>    If that returns an error, then return an "OperationError" DOMException.

A question is if the error is device lost, should it run the "context-lost" steps? Like the Chromium prototype does in [`HandleComputationFailure()`](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/dml/graph_impl_dml.cc;l=5820;bpv=1;bpt=0). Or we can just assume the "context-lost" steps are triggered by "When an MLContext context is no longer available to fulfill requests" asynchronously?

> * in `build()` it looks like failure other than "does not support a requested feature" isn't covered.

Or we could just say "If that returns an error,  then queue an ML task with global to reject promise with an 'OperationError' DOMException" similar to `compute()`?

> One generic approach for both of those would be to replace: _"Queue an ML task with global to resolve promise with ..."_ with:
> 
> > 1. **Queue an ML task** with _global_ to run these steps:
> >    
> >    1. If _context_ **is lost**, then reject _promise_ with an InvalidStateError.
> >    2. Resolve _promise_ with ...

The new steps may not run because these steps may already be aborted (by "abort these steps") if previous steps fail .

bbernhar:
> We can however be explicit that the promises returned by any methods on an object are synchronously rejected when the destroy() method is called or the context is lost.

+1. `MLContext.destroy()` steps could be spec to the following:

Script timeline:
1. Immediately ALL reject pending promises made off |this| context.
2. Issue steps for |this| context on the device/queue timeline.

Device/queue timeline:
1. Wait for async operations on the device to complete.
2. Then Lose |this| context.

Note: impl. is always free to abort pending async ops immediately (and release buffers).

inexorabletash:
> A question is if the error is device lost, should it run the "context-lost" steps? Like the Chromium prototype does in [`HandleComputationFailure()`](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/dml/graph_impl_dml.cc;l=5820;bpv=1;bpt=0). Or we can just assume the "context-lost" steps are triggered by "When an MLContext context is no longer available to fulfill requests" asynchronously?

I think this would be script observable - i.e. what order do the `lost` and `compute()` promises settle. So we should spec it explicitly.

mingmingtasd:
> The definition of "potentially abort" is the tricky one because it has to consider cases like destroying a single buffer that is part of a larger set of pending operations which should still be able to complete.

@reillyeon @huningxin @bbernhar  I think the DirectML backend [here ](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/dml/command_recorder.cc;l=128;bpv=0;bpt=0?q=command_recorder&ss=chromium%2Fchromium%2Fsrc) has considered this,  `MLBuffer` can be destroyed prior to `Compute()`  and `Dispatch` but its resource will be kept alive anyway until all the pending GPU work done. For example, if you destroy a `MLBuffer` used by `Dispatch` before the `Dispatch` is done, `Dispatch` can still complete but you can't continue to call `ReadBuffer` to read back the results. It seems OK and as expected?

reillyeon:
I think we've settled on the idea that destroying an `MLBuffer` will only make `readBuffer()` fail. Pending compute tasks will still complete. 

Destroying an `MLContext` (or context loss) is the equivalent of calling `destroy()` on all the builders, graphs and buffers created by the context. We haven't yet decided the semantics of destroying a graph but it will probably similarly allow pending compute tasks to complete.

mingmingtasd:
> I think we've settled on the idea that destroying an `MLBuffer` will only make `readBuffer()` fail. Pending compute tasks will still complete.
> 
> Destroying an `MLContext` (or context loss) is the equivalent of calling `destroy()` on all the builders, graphs and buffers created by the context. We haven't yet decided the semantics of destroying a graph but it will probably similarly allow pending compute tasks to complete.

I have submitted a CL to expose `MLGraph::destroy` : https://chromium-review.googlesource.com/c/chromium/src/+/5799069

inexorabletash:
(not actually ready for review, but I wanted to force an updated preview to be generated)

index.bs
```diff
@@ -964,6 +977,8 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
   <summary>
     The <dfn method for=MLContext>dispatch(|graph|, |inputs|, |outputs|)</dfn> method steps are:
   </summary>
+    1. If |graph|.{{MLGraph/[[context]]}} is not [=this=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
```
reillyeon, line 980: The other places where we throw an error for this condition use `TypeError`.

index.bs
```diff
@@ -1222,13 +1239,56 @@ dictionary MLSingleInputSupportLimits {
...
+The {{MLContext/destroy()}} method can be called to release all resources associated with the context. Any scheduled compute requests will fail.
+
+<details open algorithm>
+<summary>
+    The <dfn method for=MLContext>destroy()</dfn> method steps are:
+</summary>
+    1. If [=this=] [=MLContext/is lost=], then abort these steps.
+    1. For each {{MLGraph}} |graph| where |graph|.{{MLTensor/[[context]]}} equals [=this=]:
+        1. Run the {{MLGraph/destroy()}} method steps for |graph|.
+    1. Run the steps to [=context lost|lose=] [=this=].
```
reillyeon, line 1253: Any reason to do this before destroying the tensors?

index.bs
```diff
@@ -1222,13 +1239,56 @@ dictionary MLSingleInputSupportLimits {
...
+
+<details open algorithm>
+<summary>
+To <dfn lt="context lost">lose</dfn> the {{MLContext}} |context|, run these steps:
+</summary>
+    1. Let |global| be |context|'s [=relevant global object=].
+    1. [=Queue an ML task=] with |global| to run these steps:
+        1. Let |info| be a new {{MLContextLostInfo}}.
+        1. Set |info|.{{MLContextLostInfo/message}} to an [=implementation-defined=] value.
+        1. [=Resolve=] |context|.{{MLContext/[[lost]]}} with |info|.
```
reillyeon, line 1270: These steps should also cause graphs and tensors to be destroyed.

index.bs
```diff
@@ -1712,16 +1795,18 @@ Build a composed graph up to a given output operand into a computational graph a
...
     1. Let |promise| be [=a new promise=].
-    1. Run the following steps [=in parallel=]:
+    1. Run the following steps [=in parallel=], but [=/abort when=] |graph|.{{MLGraph/[[context]]}} [=MLContext/is lost=]:
         1. Let |graphImpl| be the result of converting [=this=]'s [=MLGraphBuilder/graph=] with |operands|, |operators|, |inputs|, and |outputs|'s [=map/values=] into an [=implementation-defined=] format which can be interpreted by the underlying platform.
-            1. If the underlying platform does not support a requested feature, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{OperationError}}" {{DOMException}}, and abort these steps.
+        1. If the previous step failed, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{OperationError}}" {{DOMException}}, and abort these steps.
         1. Set |graph|.{{MLGraph/[[implementation]]}} to |graphImpl|.
         1. [=Queue an ML task=] with |global| to [=resolve=] |promise| with |graph|.
+    1. [=/If aborted=], then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{OperationError}}" {{DOMException}}, and abort these steps.
```
reillyeon, line 1809: Should we use an `InvalidStateError` here to match the behavior when this fails synchronously?

reillyeon (changes requested):

index.bs
inexorabletash: I was following https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ml/ml_context.cc;l=150

@mingmingtasd - do you recall if this ordering was intentional or arbitrary?

index.bs
inexorabletash: Agreed - I should move the steps from destroy() here.

index.bs
inexorabletash: No strong opinion from me; I initially had this logic merged with the preceding substep ("if the previous step failed... Operation error") and kept the type when splitting it out.

index.bs
mingmingtasd: Yes, we'll destroy all the MLTensors and MLGraphs under the MLContext, but `Run the steps to [=context lost|lose=] [=this=]` this step seems unnecessary.

index.bs
```diff
@@ -972,6 +987,7 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
     1. If [=validating tensors with descriptors=] given |inputs| and |graph|.{{MLGraph/[[inputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. If [=validating tensors with descriptors=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. If |graph|.{{MLGraph/[[context]]}} is no longer available to fulfill requests, then run the [=context lost=] steps for it.
```
reillyeon, line 990: Use "abort when" here as well?

reillyeon (changes requested):

index.bs
inexorabletash: Done. 

The formulation I'm using of "Enqueue... which abort when... " is a shortcut for:

1. Enqueue the following steps to ...
     1. Run these steps, but abort when ...
         1. do it
         2. do more of it
     1. If aborted, ....
 
Should I expand it out?

reillyeon: The expanded form is probably better, because right now the indentation of the "if aborted" steps doesn't make it look like they are also supposed to be running in parallel.

inexorabletash: ... and then we need to figure out how to represent "lost" over on the context's timeline - right now the concept sits on the ML task runner.

index.bs
```diff
@@ -1080,16 +1097,17 @@ Reads back the {{MLTensor/[[data]]}} of an {{MLTensor}} from the {{MLContext}}.{
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
-    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
```
huningxin, line 1100: Is `|tensor|.{{MLGraph/[[context]]}}` a typo (an old issue)? Should it be `|tensor|.{{MLTensor/[[context]]}}` or just `[=this=].{{MLContext/[[timeline]]}}`? If it is a typo, the line 1096 also needs to be fixed. For example,
```suggestion
    1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
```

huningxin: Another question is whether it should check context lost earlier? Like `createTensor` steps do, e.g.

```
1. If [=this=] [=MLContext/is lost=], then return [=a new promise=] [=rejected=] with an "{{InvalidStateError}}" {{DOMException}}.
```

index.bs
```diff
@@ -1109,7 +1127,7 @@ Bring-your-own-buffer variant of {{MLContext/readTensor(tensor)}}. Reads back th
     1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If [=validating buffer with descriptor=] given |outputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
-    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}, which [=/abort when=] [=this=] [=MLContext/is lost=]:
```
huningxin, line 1130: Same here

index.bs
inexorabletash: Given that the context loss will destroy every associated tensor, I don't think that is needed. I'm not opposed to adding it explicitly though.

index.bs
```diff
@@ -1249,9 +1320,26 @@ interface MLGraph {};
...
+### {{MLGraph/destroy()}}  ### {#api-mlgraph-destroy}
+
+The {{MLGraph/destroy()}} method can be called to release all resources associated with the graph.
+
+<details open algorithm>
+<summary>
+    The <dfn method for=MLGraph>destroy()</dfn> method steps are:
+</summary>
+    1. If [=this=].{{MLGraph/[[isDestroyed]]}} is true, then abort these steps.
+    1. Set [=this=].{{MLGraph/[[isDestroyed]]}} to true.
```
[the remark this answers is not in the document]
inexorabletash: Added in 0fc4151dc670f65dff4d92ac1af59cc6e6e31a30 but it's a very generic note, it doesn't mention the timeline. Should it?

index.bs
```diff
@@ -1222,13 +1241,65 @@ dictionary MLSingleInputSupportLimits {
     :: {{MLSupportLimits}} for output operand.
 </dl>
 
+### {{MLContext/destroy()}}  ### {#api-mlcontext-destroy}
+
+The {{MLContext/destroy()}} method can be called to release all resources associated with the context. Any scheduled compute requests will fail.
```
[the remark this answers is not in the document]
inexorabletash: I softened the text in 0fc4151dc670f65dff4d92ac1af59cc6e6e31a30 because "scheduled" seemed unnecessary. I added tensor requests (oops, forgot creation though).

And added to the `writeTensor` steps; not sure how I missed that.

index.bs
inexorabletash: For the "type confusion" I added a lint tool test and caught several more places. It's not perfect, but it is progress!

index.bs
inexorabletash: Fixed in 0fc4151dc670f65dff4d92ac1af59cc6e6e31a30 and added lint check.

index.bs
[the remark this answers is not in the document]
inexorabletash: Added "creation" in 95c3b26

index.bs
inexorabletash: Done in 88e8d58 - I left "is lost" alone for now though.

index.bs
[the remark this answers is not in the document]
reillyeon: We can be a bit more explicit and queue a task on the context timeline to "mark resources owned by this graph as freeable".

index.bs
```diff
@@ -1104,21 +1125,23 @@ Bring-your-own-buffer variant of {{MLContext/readTensor(tensor)}}. Reads back th
...
-
-                Note: [=Validating buffer with descriptor=] above will fail if |outputData| is detached, but it is possible that |outputData| could be detached between that step and this one.
-
-            1. [=ArrayBuffer/Write=] |bytes| to |outputData|.
-            1. [=Resolve=] |promise| with {{undefined}}.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
+            1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+            1. Otherwise, [=queue an ML task=] with |global| and the following steps:
```
reillyeon, line 1137:
```suggestion
            1. Otherwise, [=queue an ML task=] with |global| to run these steps:
```

reillyeon (approved):

huningxin (approved):
LGTM!

fdwr (approved):
Thanks. I scanned it, and it looks fine (no close scrutiny given Reilly and Ningxin already looked). I'll delay merging a bit though because Rafael wanted to look at it tomorrow. ⏳

RafaelCintron (approved):
LGTM.  @inexorabletash , thank you very much for putting this together.

mwyrzykowski:
Also looks good from WebKit's side @inexorabletash - informal approval from discussion in today's call

PR #747, opened 2024-07-30
Introduce MLConstantOperand

### LINKED ISSUE #668: Do we need an `MLConstantOperand`?
The `constant()` operator is special. Constants are effectively inputs that are known at the time of graph compilation (i.e. `build()`). 

You may ask, why do we need a separate method when we could just pass this constant data as an `input()`?

Most discussions I've seen so far about `constant()` (https://github.com/webmachinelearning/webnn/issues/614#issuecomment-2019116877, [this thread about the since-abandoned "fill sequence" overload](https://chromium-review.googlesource.com/c/chromium/src/+/5378104/comment/c4535e9f_314c5790/), etc) have been concerned with _optimizing performance_. There are a number of compile-time optimizations a backend may perform if it knows that some data is constant.

Our experience with CoreML has given us another reason:

***The backend requires that a parameter to an operator must be constant***

Take `conv2d()`, for example. It's defined for CoreML [here](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.conv.conv). The `bias` parameter is defined as:

- `bias: const tensor<[C_out],T>`

Meanwhile, WebNN allows this to be any `MLOperand`, as long as it's of the appropriate shape and dtype:

```
dictionary MLConv2dOptions {
  // ...
  MLOperand bias;
  // ...
};
```

This appears to be a straightforward plumbing through of [DML's interface](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_convolution_operator_desc), which does not require the `BiasTensor` to be a constant. Neither does [the corresponding operator for TFLite](https://www.tensorflow.org/mlir/tfl_ops#tflconv_2d_tflconv2dop). From what I can tell, this seems to be because these frameworks don't have a way to express that some input tensor must be const. The options are either to pass the parameter as the framework's generic representation of a Tensor - which would in practice always(?) be created from a `constant()` - or to pass the parameter as a 1D array directly. If the parameters may be large (and perhaps unbounded), the former is the more appropriate choice.

To get a sense for whether this is a reasonable hypothesis, I've inspected of all† uses of the affected operators in the [WebNN Samples repo](https://github.com/webmachinelearning/webnn-samples/) repo:

| `operator.param`   |  Usage in [WebNN Samples](https://github.com/webmachinelearning/webnn-samples/) |
| ------------------ | ------------------- |
| `batchNormalization.mean` | Constant only |
| `batchNormalization.variance` | Constant only |
| `batchNormalization.scale` | Constant only |
| `batchNormalization.bias` | Constant only |
| `conv2d.bias` | Constant only |
| `convTranspose2d.bias` | Not used |
| `gru.weight` | Constant only |
| `gru.recurrentWeight` | Constant only |
| `gru.bias` | Constant only |
| `gru.recurrentBias` | Constant only |
| `instanceNormalization.scale` | Constant only†† |
| `instanceNormalization.bias` | Constant only†† |
| `layerNormalization.scale` | Not used |
| `layerNormalization.bias` | Not used |
| `lstm.weight` | Not used |
| `lstm.recurrentWeight` | Not used |
| `lstm.bias` | Not used |
| `lstm.recurrentBias` | Not used |
| `lstm.peepholeWeight` | Not used |
| `prelu.slope` | Not used |

†This list only includes WebNN operators which trivially map to CoreML operators. WebNN operators which need to be in terms of other CoreML operators will be subject to the restrictions of those respective CoreML operators. For example, CoreML doesn't have operators for `gruCell` or `lstmCell`, so these operators will need to be implemented in terms of `gru` and `lstm`, respectively. These operators will in turn need many of their parameters to be const, as well

††[One caller](https://github.com/webmachinelearning/webnn-samples/blob/e5bdbab3dd12fe015c3c65c86d3f1a5bf2a73aba/style_transfer/fast_style_transfer_net.js#L31) of passes the result of a `reshape`... but that's only because the sample was written before `constant()` took an `MLOperandDescriptor`. The `reshape` is just assigning dimensions to a `constant()`. Nowadays we'd just pass the `constant()` directly

Remarkably, _every single instance_ where one of these params is used in the WebNN Samples, it was created from a `constant()`. Cool!

Of course, this is not close to a comprehensive list of all models hope to run with WebNN. That being said, **if there are no significant known use cases for passing any of these parameters as non-constant tensors - _if their non-constness is simply a limitation in the framework and there are no useful reasons to pass non-const tensors_ - I think there's a reasonable argument that WebNN should require these parameters to be constants**. @fwdr could you perhaps provide some more color here? :)

It seems that we have the following options to support each of these operators on CoreML:

1. Require that *`operator.param`* must be a constant `MLOperand` (my tentative preference)
   - Note that it's much easier to _relax_ these restrictions than to try to impose them later
   - This can be done two ways:
     1. Create an `MLConstantOperand` interface which extends `MLOperand`, and specify that *`param`* takes an `MLConstantOperand`
     2. Specify that `MLOperand` has a "kind", [as the Chromium implementation already does](https://crsrc.org/c/third_party/blink/renderer/modules/ml/webnn/ml_operand.h;drc=53c1d06fee5a3e98e7f265419f6f84df2c2c2d62;l=112), and throw a `TypeError` if not a "constant" kind. This may be confusing to developers
1. Make *`operator.param`* a `sequence<MLNumber>`
   - This doesn't make much sense for anything other than 1-D tensors
1. Decide that this is a quirk of CoreML and fail if *`operator.param`* is not a constant only on CoreML
   - Currently this failure happens during `build()` on Chromium, though we could conceivably make this a synchronous check on the respective builder method, especially if we have defined procedures for querying for backend-specific support (see #463)

Thoughts?

a-sully:
Fixes #668

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/747.html" title="Last updated on Jul 30, 2024, 10:44 PM UTC (919a280)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/747/4697a0d...a-sully:919a280.html" title="Last updated on Jul 30, 2024, 10:44 PM UTC (919a280)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1233,6 +1233,38 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
 To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |builder| and {{MLOperand}} |operand|, return true if |operand|.{{MLOperand/[[builder]]}} is |builder|, and false otherwise.
 </p>
 
+## {{MLConstantOperand}} interface ## {#api-mlconstantoperand}
+
+An {{MLConstantOperand}} is an {{MLOperand}} which represents constant weight data. Its {{MLConstantOperand/[[data]]}} is statically known during graph construction time, which allows for optimizations such as weight preprocessing.
+
+<script type=idl>
+[SecureContext, Exposed=(Window, DedicatedWorker)]
+interface MLConstantOperand : MLOperand {};
+</script>
+
+<div class=internal-slots>
+{{MLConstantOperand}} has the following internal slots:
+  <dl dfn-type=attribute dfn-for="MLConstantOperand">
+    : <dfn>\[[data]]</dfn> of type [=byte sequence=]
+    ::
+        The bytes associated with this {{MLConstantOperand}}.
+  </dl>
+</div>
+
+### Creating an {{MLConstantOperand}} ### {#api-mlconstantoperand-create}
+The {{MLConstantOperand}} objects are created by the {{MLGraphBuilder/constant(descriptor, bufferView)}} and {{MLGraphBuilder/constant(type, value)}} methods of {{MLGraphBuilder}}, internally using the following algorithm.
+
+<details open algorithm>
+  <summary>
+    To <dfn>create an MLConstantOperand</dfn> given {{MLGraphBuilder}} |builder|, {{MLOperandDescriptor}} |desc|, and [=byte sequence=] |bytes| run the following steps:
+  </summary>
+    1. Let |constantOperand| be a new {{MLConstantOperand}}.
+    1. Set |constantOperand|.{{MLOperand/[[builder]]}} to |builder|.
+    1. Set |constantOperand|.{{MLOperand/[[descriptor]]}} to |desc|.
+    1. Set |constantOperand|.{{MLConstantOperand/[[data]]}} to |bytes|.
+    1. Return |constantOperand|.
+</details>
+
 #### {{MLNumber}} #### {#api-mlnumber-typedef}
 
 <dfn typedef>MLNumber</dfn> is used when specifying the type of a numeric option for an {{MLOperand}} which can be of any {{MLOperandDataType}}, including both 64-bit integer types ({{MLOperandDataType/"uint64"}} and {{MLOperandDataType/"int64"}}) and 32-bit floating point ({{MLOperandDataType/"float32"}}). Implementations process the value according to the corresponding {{MLOperandDataType}}. For example, if {{MLGraphBuilder/clamp(input, options)}} is called with an {{MLOperand}} with [=MLOperand/dataType=] {{MLOperandDataType/"uint32"}}, the {{MLNumber}} parameters are explicitly [=cast=] to {{unsigned long}}.
@@ -1287,10 +1319,10 @@ interface MLGraphBuilder {
   MLOperand input(USVString name, MLOperandDescriptor descriptor);
 
   // Create an operand for a graph constant.
-  MLOperand constant(MLOperandDescriptor descriptor, ArrayBufferView bufferView);
+  MLConstantOperand constant(MLOperandDescriptor descriptor, ArrayBufferView bufferView);
 
   // Create a scalar operand from the specified number of the specified type.
-  MLOperand constant(MLOperandDataType type, MLNumber value);
+  MLConstantOperand constant(MLOperandDataType type, MLNumber value);
 
   // Compile the graph up to the specified output operands asynchronously.
   Promise<MLGraph> build(MLNamedOperands outputs);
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1406,10 +1439,14 @@ Data truncation will occur when the specified value exceeds the range of the spe
   </summary>
     1. If [=this=].{{MLGraphBuilder/[[hasBuilt]]}} is true, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. Set |value| to the result of [=casting=] |value| to |type|.
+    1. Let |bytes| be a [=byte sequence=] represented by |value|.
```
inexorabletash, line 1442: nit: "representing"

index.bs
```diff
@@ -1406,10 +1439,14 @@ Data truncation will occur when the specified value exceeds the range of the spe
...
+
+      ISSUE(TODO): Specify this in terms of https://tc39.es/ecma262/multipage/structured-data.html#sec-numerictorawbytes.
```
inexorabletash, line 1444: nit: Make this a link (bikeshed didn't do it automagically?!?)

index.bs
```diff
@@ -1233,6 +1233,38 @@ The {{MLOperand}} objects are created by the methods of {{MLGraphBuilder}}, inte
...
+
+<script type=idl>
+[SecureContext, Exposed=(Window, DedicatedWorker)]
+interface MLConstantOperand : MLOperand {};
+</script>
+
+<div class=internal-slots>
+{{MLConstantOperand}} has the following internal slots:
+  <dl dfn-type=attribute dfn-for="MLConstantOperand">
+    : <dfn>\[[data]]</dfn> of type [=byte sequence=]
```
inexorabletash, line 1248: Just thinking about how future spec text would specify the actual operation on tensor data... should this just be a list of numbers instead of a byte sequence? Otherwise every access to a tensor would have to invoke an algorithm to get the numbers, and that would need to reference "decoding" the byte sequence. Sure, that can be wrapped in helpers, but is there any benefit to holding the bytes here vs. just a list of numbers?

index.bs
inexorabletash: If we want to keep this (see previous note about bytes), I think referencing the ECMAScript algorithm is fine. Just write `[$NumericToRawBytes$](|type|, |value|, true)` - it's true that the types don't quite align between the algorithms but it's really really really obvious that "int8" in our enum is the same as INT8 in https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarray-objects

index.bs
a-sully: Good thinking. I agree that sounds nice :)

This is making me realize that I've never before had to specify anything touching the contents of an `ArrayBufferView`... For the `constant(descriptor, bufferView)` method, how do I get a `[/list]` of numbers out of the `bufferView`?

index.bs
a-sully: Used `[$NumericToRawBytes$]` in the latest patch to show what that looks like, but this could soon be moot

index.bs
a-sully: No longer relevant

index.bs
inexorabletash: I'm not finding an example and I can't think of a place outside ECMAScript itself that would need to. You may have to write it yourself, using [GetValueFromBuffer](https://tc39.es/ecma262/#sec-getvaluefrombuffer) which [get a copy of the bytes](https://webidl.spec.whatwg.org/#dfn-get-buffer-source-copy) does.

a-sully:
Closing due to https://github.com/webmachinelearning/webnn/issues/668#issuecomment-2384262749

Issue #624, opened 2024-03-27
Need clarify the usage of axes=[0,1] for resample2d


BruceDai:
This issue was raised as @fdwr's below [comment](https://chromium-review.googlesource.com/c/chromium/src/+/5382118/comment/57f65e31_76bf34f6/) for reviewing Chromium [CL-5382118](https://chromium-review.googlesource.com/c/chromium/src/+/5382118). 

> `axes=[0,1]` is an odd one, and I wonder the history behind it (decided before I joined). Some other ML libraries simply treat axes as numbers (no associated semantics to any particular dimensions), but WebNN limits them to `[0,1]`, `[1,2]`, `[2,3]`; and while the `axes=[1,2]` and `axes=[2,3]` uses are obvious (for NHWC and NCHW tensors respectively), the `[0,1]` scenario makes me go hmm, because if the semantic interpretation of the input tensor is **NH**WC, then it would mean resampling the batch and height (but not the width 🙃), which wouldn't make any sense; whereas if the semantic layout of the input was **NC**HW, then it would mean simultaneously resampling the batch dimension (useful for replication) and expanding the channel count, which *can* make sense. However, if resampling NC in **NC**HW is useful, then resampling NC in **N**HW**C** should be deemed useful too, but `axes=[0,3]` is currently disallowed by the spec. 🤔

/cc @fdwr @huningxin @a-sully, PTAL, thanks.

inexorabletash:
Maybe asking the same question in a different way:

If [`resample2d()`](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-resample2d-method) is only for use with tensors of NHWC and NCHW layouts, then should it take an [`MLInputOperandLayout`](https://www.w3.org/TR/webnn/#enumdef-mlinputoperandlayout) param instead of an [`axes`](https://www.w3.org/TR/webnn/#dom-mlresample2doptions-axes) param? 

If not, what are the other use cases, and how can we support them via [`axes`](https://www.w3.org/TR/webnn/#dom-mlresample2doptions-axes)?

inexorabletash:
I guess [0,1] makes sense if you're just dealing directly with 2D data. But you could do a trivial reshape() to NHWC or NCHW with N=1 C=1 and back if needed.

philloooo:
Signal boosting this one.  Changing to take [MLInputOperandLayout](https://www.w3.org/TR/webnn/#enumdef-mlinputoperandlayout) seems to make sense to me? 

[...]

PR #752, opened 2024-08-20
Allow arbitrary axes for resample2d

philloooo:
Fixs #624

This allow any arbitrary axes as long as they are valid dimensions for resample2d. Remove restriction on it to be the spacial dimensions.

[Chromium prototype](https://chromium-review.googlesource.com/c/chromium/src/+/5798093)

@fdwr @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/webnn/pull/752.html" title="Last updated on Aug 22, 2024, 5:39 PM UTC (471cc12)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/752/d7d0363...philloooo:471cc12.html" title="Last updated on Aug 22, 2024, 5:39 PM UTC (471cc12)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -5202,7 +5202,7 @@ partial interface MLGraphBuilder {
 </div>
 
 ### resample2d ### {#api-mlgraphbuilder-resample2d-method}
-Resample the tensor values from the source to the destination spatial dimensions according to the scaling factors.
+Resample the tensor values from the source to the destination dimensions according to the axes and scaling factors.
 <script type=idl>
 enum MLInterpolationMode {
   "nearest-neighbor",
@@ -5237,18 +5237,18 @@ partial interface MLGraphBuilder {
     : <dfn>scales</dfn>
     ::
         A list of length 2.
-        Specifies the scaling factor in each spatial dimensions of the input: *[scaleHeight, scaleWidth]*.
+        Specifies the scaling factor in each dimensions of the input in {{MLResample2dOptions/axes}} : *[scaleForFirstAxis, scaleForSecondAxis]*.
         The default value is [1.0, 1.0].
 
     : <dfn>sizes</dfn>
     ::
         A list of length 2.
-        Specifies the target sizes for each spatial dimensions of the input: *[sizeHeight, sizeWidth]*. When the target sizes are specified, the {{MLResample2dOptions/scales}} argument is ignored, since the scaling factor values are derived from the target sizes of each spatial dimension of the input.
+        Specifies the target sizes for each dimensions of the input in {{MLResample2dOptions/axes}} : *[sizeForFirstAxis, ForSecondAxis]*. When the target sizes are specified, the {{MLResample2dOptions/scales}} argument is ignored, since the scaling factor values are derived from the target sizes of the input.
 
     : <dfn>axes</dfn>
     ::
         A list of length 2.
-        Specifies the two consecutive dimensions of the input tensor to which the interpolation algorithm applies. The valid values are [0, 1], [1, 2] or [2, 3].
+        Specifies the two dimensions of the input tensor to which the interpolation algorithm applies.
         The default value is [2, 3].
 </dl>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -5237,30 +5237,30 @@ partial interface MLGraphBuilder {
     : <dfn>scales</dfn>
     ::
         A list of length 2.
-        Specifies the scaling factor in each spatial dimensions of the input: *[scaleHeight, scaleWidth]*.
+        Specifies the scaling factor in each dimensions of the input in {{MLResample2dOptions/axes}} : *[scaleForFirstAxis, scaleForSecondAxis]*.
```
fdwr, line 5240: Plurality agreement (existing).
```suggestion
        Specifies the scaling factor in each dimension of the input in {{MLResample2dOptions/axes}} : *[scaleForFirstAxis, scaleForSecondAxis]*.
```

Slight wording suggestion to avoid "in" twice in a row. That also matches the wording below "Specifies the target sizes for each dimension..." more closely:
```suggestion
        Specifies the scaling factor for each input dimension from {{MLResample2dOptions/axes}} : *[scaleForFirstAxis, scaleForSecondAxis]*.
```

huningxin (approved):
LGTM!

index.bs
```diff
@@ -5237,30 +5237,30 @@ partial interface MLGraphBuilder {
...
         The default value is [1.0, 1.0].
 
     : <dfn>sizes</dfn>
     ::
         A list of length 2.
-        Specifies the target sizes for each spatial dimensions of the input: *[sizeHeight, sizeWidth]*. When the target sizes are specified, the {{MLResample2dOptions/scales}} argument is ignored, since the scaling factor values are derived from the target sizes of each spatial dimension of the input.
+        Specifies the target sizes for each dimensions of the input in {{MLResample2dOptions/axes}} : *[sizeForFirstAxis, ForSecondAxis]*. When the target sizes are specified, the {{MLResample2dOptions/scales}} argument is ignored, since the scaling factor values are derived from the target sizes of the input.
```
fdwr, line 5246:
```suggestion
        Specifies the target sizes for each dimension of the input from {{MLResample2dOptions/axes}} : *[sizeForFirstAxis, ForSecondAxis]*. When the target sizes are specified, the {{MLResample2dOptions/scales}} argument is ignored, since the scaling factor values are derived from the target sizes of the input.
```

fdwr, line 5246: I don't think we need to *explicitly* mention it, because it's implicit in the definition of `*[sizeForFirstAxis, ForSecondAxis]*` that for cases like `axes = [3, 0]` and `sizes = [640, 3]`, it would mean axis 3 is size 640, and axis 0 is size 3 (that is`output.dimensions[axes[0]]` = sizes[0] and `output.dimensions[axes[1]] = sizes[1]`); but it would be prudent to include test case for it.

index.bs
```diff
@@ -5237,30 +5237,30 @@ partial interface MLGraphBuilder {
...
   <summary>
-    To <dfn for="MLGraphBuilder">check resample options</dfn> given |options|, run the following steps:
+    To <dfn for="MLGraphBuilder">check resample options</dfn> given |options| and |input|, run the following steps:
   </summary>
     1. If |options|.{{MLResample2dOptions/scales}} does not [=map/exist=], set it to the [=/list=] « 1.0, 1.0 ».
     1. Otherwise, if any of its values is not greater than 0, or if its [=list/size=] is not 2, return false.
     1. If |options|.{{MLResample2dOptions/sizes}} [=map/exists=], and if its size is not 2, or if any of its values is not greater than 0, return false.
     1. If |options|.{{MLResample2dOptions/axes}} does not [=map/exists=], set it to the [=/list=] « 2, 3 ».
-    1. Otherwise, if its value is not one of « 0, 1», « 1, 2», « 2, 3 », return false.
+    1. Otherwise, if |options|.{{MLResample2dOptions/axes}} contains duplicate values, or if any of its elements is not in [=the range=] 0 to |input|'s [=MLOperand/rank=], exclusive, then return false.
```
fdwr, line 5263: The elegant aspect I like about this simple check (rather than a hard-coded list of permutations) is that it's more consistent with most other operators, and it also easily permits (in the future) us to relax resample2d to accept pure 2D or 3D rank input, so the caller doesn't need to add extra reshape calls just to prepend leading 1's for the sake of 4Dness. Granted, 2 of 3 current backends require 4Dness anyway ([TF](https://www.tensorflow.org/api_docs/python/tf/image/resize), [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.image_resizing.resample), [DML](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_resample_operator_desc))

fdwr (approved):
👍 Thank you Phillis.

index.bs
fdwr: Btw, I don't think we need to *explicitly* mention in documentation (because it's implicit in the definition of `*[sizeForFirstAxis, ForSecondAxis]*)` that for cases like `axes = [3, 0]` and `sizes = [640, 3]`, it means axis 3 is size 640, and axis 0 is size 3 (that is`output.dimensions[axes[0]] = sizes[0]` and `output.dimensions[axes[1]] = sizes[1]`); but it will be prudent to include a test case for it, where the axes have high axes before low axes. ⏳

fdwr (approved):
👍 LGTM. TY.

PR #770, opened 2024-10-23
Remove pool2d MLRoundingType - Simplify the operand layout support of conv2d and pooling 2d operations

fdwr:
Rather than have both an output rounding field *and* spatial output dimensions, be more explicit in WebNN where the caller computes this ahead of time.

For https://github.com/webmachinelearning/webnn/issues/324

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4575,10 +4575,6 @@ partial interface MLGraphBuilder {
 ### Pooling operations ### {#api-mlgraphbuilder-pool2d}
 Compute a pooling operation across all the elements within the moving window over the input tensor.
 <script type=idl>
-enum MLRoundingType {
-  "floor",
-  "ceil"
-};
 
 dictionary MLPool2dOptions {
   sequence<[EnforceRange] unsigned long> windowDimensions;
@@ -4586,7 +4582,6 @@ dictionary MLPool2dOptions {
   sequence<[EnforceRange] unsigned long> strides;
   sequence<[EnforceRange] unsigned long> dilations;
   MLInputOperandLayout layout = "nchw";
-  MLRoundingType roundingType = "floor";
   sequence<[EnforceRange] unsigned long> outputSizes;
 };
 
@@ -4638,10 +4633,14 @@ partial interface MLGraphBuilder {
 
     : <dfn>outputSizes</dfn>
     ::
-        A list of length 2.
-        Specifies the sizes of the two spacial dimensions of the output tensor. When the output sizes are explicitly specified, the {{MLPool2dOptions/roundingType}} is ignored.
+        A list of length 2: *[outputHeight, outputWidth]*
+        Specifies the sizes of the two spatial dimensions of the output tensor.
 
-        If not specified, the output sizes are automatically computed.
+        The spatial dimensions of the output tensor can be calculated as follows:
+
+        *output size = ((input size - filter size + beginning padding + ending padding) / stride) + 1*
+
+        Then the caller either applies a floor or ceiling depending on whether partial window results are desired.
 </dl>
 
 <div>
@@ -4652,13 +4651,8 @@ partial interface MLGraphBuilder {
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor that contains the
     result of the reduction. The logical shape is interpreted according to the
-    value of *layout*. More specifically, if the *options.roundingType* is {{MLRoundingType/"floor"}}, the spatial dimensions of the output tensor can be calculated as follow:
-
-    *output size = floor(1 + (input size - filter size + beginning padding + ending padding) / stride)*
-
-    or if *options.roundingType* is {{MLRoundingType/"ceil"}}:
-
-    *output size = ceil(1 + (input size - filter size + beginning padding + ending padding) / stride)*
+    value of *layout*, taking the batch and channel count from the input with
+    the spatial sizes from *outputSizes*.
 </div>
 
 <div class="note">
@@ -4671,7 +4665,7 @@ partial interface MLGraphBuilder {
 
 <details open algorithm>
   <summary>
-    To <dfn for=MLGraphBuilder>calculate pool2d output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, {{MLRoundingType}} |roundingType|, [=/list=] of 2 unsigned integers |windowDimensions|, [=/list=] of 4 unsigned integers |padding|, [=/list=] of 2 unsigned integers |strides|, [=/list=] of 2 unsigned integers |dilations|, and optional [=/list=] of 2 unsigned integers |outputSizes|, perform these steps. They return a [=/list=] of 4 unsigned integers.
+    To <dfn for=MLGraphBuilder>calculate pool2d output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, [=/list=] of 2 unsigned integers |windowDimensions|, [=/list=] of 4 unsigned integers |padding|, [=/list=] of 2 unsigned integers |strides|, [=/list=] of 2 unsigned integers |dilations|, and optional [=/list=] of 2 unsigned integers |outputSizes|, perform these steps. They return a [=/list=] of 4 unsigned integers.
   </summary>
     1. Switch on |layout|:
         <dl class=switch>
@@ -4688,24 +4682,8 @@ partial interface MLGraphBuilder {
                 1. Let |inputWidth| be |inputShape|[2].
                 1. Let |channels| be |inputShape|[3].
         </dl>
-    1. If |outputSizes| is not given, then:
-        1. Let |outputHeight| be |outputSizes|[0].
-        1. Let |outputWidth| be |outputSizes|[1].
-    1. Otherwise:
-        1. Let |outputSizes| be the result of [=MLGraphBuilder/calculating conv2d output sizes=] given |inputHeight|, |inputWidth|, |windowDimensions|[0], |windowDimensions|[1], |padding|, |strides|, and |dilations|.
-        1. Let |outputHeight| be |outputSizes|[0].
-        1. Let |outputWidth| be |outputSizes|[1].
-        1. Switch on |roundingType|:
-            <dl class=switch>
-                : {{MLRoundingType/"floor"}}
-                ::
-                    1. Set |outputWidth| to floor(|outputWidth|).
-                    1. Set |outputHeight| to floor(|outputHeight|).
-                : {{MLRoundingType/"ceil"}}
-                ::
-                    1. Set |outputWidth| to ceiling(|outputWidth|).
-                    1. Set |outputHeight| to ceiling(|outputHeight|).
-            </dl>
+    1. Let |outputHeight| be |outputSizes|[0].
+    1. Let |outputWidth| be |outputSizes|[1].
     1. Switch on |layout|:
         <dl class=switch>
             : {{MLInputOperandLayout/"nchw"}}
@@ -4737,7 +4715,7 @@ partial interface MLGraphBuilder {
     1. If |options|.{{MLPool2dOptions/dilations}}'s [=list/size=] is not 2, then [=exception/throw=] a {{TypeError}}.
     1. If any value in |options|.{{MLPool2dOptions/dilations}} is not greater than 0, then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be a copy of |input|.{{MLOperand/[[descriptor]]}}.
-    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating pool2d output sizes=] given |options|.{{MLPool2dOptions/layout}}, |input|'s [=MLOperand/shape=], |options|.{{MLPool2dOptions/roundingType}}, |options|.{{MLPool2dOptions/windowDimensions}}, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, |options|.{{MLPool2dOptions/dilations}}, and |options|.{{MLPool2dOptions/outputSizes}} (if it [=map/exists=]).
+    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating pool2d output sizes=] given |options|.{{MLPool2dOptions/layout}}, |input|'s [=MLOperand/shape=], |options|.{{MLPool2dOptions/windowDimensions}}, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, |options|.{{MLPool2dOptions/dilations}}, and |options|.{{MLPool2dOptions/outputSizes}} (if it [=map/exists=]).
     1. If any [=list/item=] in |outputShape| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
     1. Set |desc|.{{MLOperandDescriptor/dimensions}} to |outputShape|.
     1. *Make graph connections:*


index.bs
```diff
@@ -4688,24 +4682,8 @@ partial interface MLGraphBuilder {
                 1. Let |inputWidth| be |inputShape|[2].
                 1. Let |channels| be |inputShape|[3].
         </dl>
-    1. If |outputSizes| is not given, then:
```
huningxin, line 5479: `outputSizes` is optional, we should keep this checking. I feel the existing code has an issue, it should use `outputSizes` if it is given?

index.bs
```diff
@@ -4688,24 +4682,8 @@ partial interface MLGraphBuilder {
...
-        1. Let |outputHeight| be |outputSizes|[0].
-        1. Let |outputWidth| be |outputSizes|[1].
-    1. Otherwise:
-        1. Let |outputSizes| be the result of [=MLGraphBuilder/calculating conv2d output sizes=] given |inputHeight|, |inputWidth|, |windowDimensions|[0], |windowDimensions|[1], |padding|, |strides|, and |dilations|.
-        1. Let |outputHeight| be |outputSizes|[0].
-        1. Let |outputWidth| be |outputSizes|[1].
-        1. Switch on |roundingType|:
```
huningxin, line 5486: As `roundingType` is removed, what's the default rounding type if `outputSizes` is not present?

index.bs
fdwr: Yeah, the previous `if` looked like a typo (so that should be fixed in any case).

Well, with this change, `outputSizes` would be required (caller is explicit now). Do you think this simplification to be problematic? 🤔

huningxin: Then `outputSizes` are not an option anymore, should it be moved to another parameter? Like

```js
 MLOperand averagePool2d(MLOperand input, sequence<[EnforceRange] unsigned long> outputSizes, optional MLPool2dOptions options = {});
```

index.bs
```diff
@@ -5476,24 +5466,8 @@ partial dictionary MLOpSupportLimits {
...
-                : {{MLRoundingType/"floor"}}
-                ::
-                    1. Set |outputWidth| to floor(|outputWidth|).
-                    1. Set |outputHeight| to floor(|outputHeight|).
-                : {{MLRoundingType/"ceil"}}
-                ::
-                    1. Set |outputWidth| to ceiling(|outputWidth|).
-                    1. Set |outputHeight| to ceiling(|outputHeight|).
-            </dl>
+    1. Let |outputHeight| be |outputSizes|[0].
```
inexorabletash, line 5469: FYI, [Infra recently grew new syntax](https://infra.spec.whatwg.org/#example-list-multiple-assignment) to streamline this:

1. Let « _outputHeight_, _outputWidth_ » be _outputSizes_.

No change needed here, I'll do PR for a bulk pass when I'm back at work (soon!)

anssiko: @inexorabletash, I'm so glad to hear you're back soon. 💪

On behalf of the group, welcome back!

index.bs
a-sully: https://w3ctag.github.io/design-principles/#prefer-dictionaries suggests:

> You should also consider accepting mandatory parameters through a dictionary, if it would make the API more readable, especially when they are of primitive types.

...but it also says:

> The dictionary itself should be an optional argument, so that if the author is happy with all of the default options, they can avoid passing an extra argument.

I'd err on the side of putting required arguments into the dictionary - and making the dictionary itself required - because:
- It's more readable, since primitive types are now described (as the Design Principles suggest)
  - `averagePool2d(input, {outputSizes: [2, 3]})` > `averagePool2d(input, [2, 3])`
- It's more readable, since it's otherwise not obvious to the caller why some configuration options belong in the options dictionary while others don't
  - `averagePool2d(input, {outputSizes: [2, 3], layout: "nchw"})` > `averagePool2d(input, [2, 3], {layout: "nchw"})`
- It's more extensible, since the options dict can be changed without changing the method signature (though making a non-required dictionary field required would still be a breaking change)

index.bs
huningxin: `outputSizes` would be required, after this change. Resolving this issue.

index.bs
huningxin: > I'd err on the side of putting required arguments into the dictionary - and making the dictionary itself required

SGTM. Thanks for the explanation, @a-sully !

fdwr: Will do.

Also I just reread this bit "When the output sizes are explicitly specified, the output padding values in [outputPadding](https://www.w3.org/TR/webnn/#dom-mlconvtranspose2doptions-outputpadding) are ignored" which implies that since outputSizes is required now, that outputPadding is always ignored, meaning we can delete it.

index.bs
```diff
@@ -5459,7 +5449,7 @@ partial dictionary MLOpSupportLimits {
 
 <details open algorithm>
   <summary>
-    To <dfn for=MLGraphBuilder>calculate pool2d output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, {{MLRoundingType}} |roundingType|, [=/list=] of 2 unsigned integers |windowDimensions|, [=/list=] of 4 unsigned integers |padding|, [=/list=] of 2 unsigned integers |strides|, [=/list=] of 2 unsigned integers |dilations|, and optional [=/list=] of 2 unsigned integers |outputSizes|, perform these steps. They return a [=/list=] of 4 unsigned integers.
+    To <dfn for=MLGraphBuilder>calculate pool2d output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, [=/list=] of 2 unsigned integers |windowDimensions|, [=/list=] of 4 unsigned integers |padding|, [=/list=] of 2 unsigned integers |strides|, [=/list=] of 2 unsigned integers |dilations|, and optional [=/list=] of 2 unsigned integers |outputSizes|, perform these steps. They return a [=/list=] of 4 unsigned integers.
```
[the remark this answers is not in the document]
fdwr: Thanks Anssi. 🙏

index.bs
fdwr: > and making the dictionary itself required

Hmm, the following didn't work (bikeshed errors):

```
  MLOperand averagePool2d(MLOperand input, required MLPool2dOptions options);
```

index.bs
```diff
@@ -5395,7 +5385,7 @@ partial dictionary MLOpSupportLimits {
 
 <details open algorithm>
   <summary>
-    To <dfn for=MLGraphBuilder>calculate pool2d output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, {{MLRoundingType}} |roundingType|, [=/list=] of 2 unsigned integers |windowDimensions|, [=/list=] of 4 unsigned integers |padding|, [=/list=] of 2 unsigned integers |strides|, [=/list=] of 2 unsigned integers |dilations|, and optional [=/list=] of 2 unsigned integers |outputSizes|, perform these steps. They return a [=/list=] of 4 unsigned integers.
+    To <dfn for=MLGraphBuilder>calculate pool2d output sizes</dfn> given {{MLInputOperandLayout}} |layout|, [=/list=] of 4 unsigned integers |inputShape|, [=/list=] of 2 unsigned integers <var ignore>windowDimensions</var>, [=/list=] of 4 unsigned integers <var ignore>padding</var>, [=/list=] of 2 unsigned integers <var ignore>strides</var>, [=/list=] of 2 unsigned integers <var ignore>dilations</var>, and optional [=/list=] of 2 unsigned integers |outputSizes|, perform these steps. They return a [=/list=] of 4 unsigned integers.
```
huningxin, line 5388: If `windowDimensions`, `padding`, `strides` and `dilations` parameters are not used in the algorithm steps, they probably can be removed.

`outputSizes` is not optional.

index.bs
```diff
@@ -5404,21 +5394,7 @@ partial dictionary MLOpSupportLimits {
...
-                : {{MLRoundingType/"floor"}}
-                ::
-                    1. Set |outputWidth| to floor(|outputWidth|).
-                    1. Set |outputHeight| to floor(|outputHeight|).
-                : {{MLRoundingType/"ceil"}}
-                ::
-                    1. Set |outputWidth| to ceiling(|outputWidth|).
-                    1. Set |outputHeight| to ceiling(|outputHeight|).
-            </dl>
+    1. Let « |outputHeight|, |outputWidth| » be |outputSizes|.
```
huningxin, line 5397: Should we add validation steps for user-supplied `outputSizes`? Chromium prototype does that: https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/public/cpp/graph_validation_utils.cc;l=1179;drc=60130be50f68c89220d8ace62c87155025267a2c;bpv=1;bpt=1

If we add `outputSizes` validation steps, `windowDimensions`, `padding`, `strides` and `dilations` parameters would be required.

index.bs
inexorabletash: Since this algorithm is only invoked in one place (the "create pooling operations" steps) can we just move it inline?

index.bs
```diff
@@ -5451,7 +5427,7 @@ partial dictionary MLOpSupportLimits {
     1. If |options|.{{MLPool2dOptions/dilations}}'s [=list/size=] is not 2, then [=exception/throw=] a {{TypeError}}.
     1. If any value in |options|.{{MLPool2dOptions/dilations}} is not greater than 0, then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be a copy of |input|.{{MLOperand/[[descriptor]]}}.
-    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating pool2d output sizes=] given |options|.{{MLPool2dOptions/layout}}, |input|'s [=MLOperand/shape=], |options|.{{MLPool2dOptions/roundingType}}, |options|.{{MLPool2dOptions/windowDimensions}}, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, |options|.{{MLPool2dOptions/dilations}}, and |options|.{{MLPool2dOptions/outputSizes}} (if it [=map/exists=]).
+    1. Let |outputShape| be the result of [=MLGraphBuilder/calculating pool2d output sizes=] given |options|.{{MLPool2dOptions/layout}}, |input|'s [=MLOperand/shape=], |options|.{{MLPool2dOptions/windowDimensions}}, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, |options|.{{MLPool2dOptions/dilations}}, and |options|.{{MLPool2dOptions/outputSizes}} (if it [=map/exists=]).
```
jlewis-austin, line 5430: The outputSizes validation above in 13.2 doesn't appear to be correct--shouldn't it be similar to this logic?

inexorabletash:
This PR needs a redo since there's been a lot of spec churn since Nov 2024. But one thing I noticed in the current spec text:

> If options.outputSizes exists, or if options.padding does not exist, set options.padding to the list « 0, 0, 0, 0 ».

That implies that if outputSizes becomes a required operation, then any passed padding will always be ignored. Flaw in the current text? I think the first clause "if options.outputSizes exists" is bogus today, and should be removed.

index.bs
fdwr: Already done by Joshua.

```
1. If neither |outputHeight| equals floor( |calculatedOutputHeight| ) and |outputWidth| equals floor( |calculatedOutputWidth| ), nor |outputHeight| equals ceil( |calculatedOutputHeight| ) and |outputWidth| equals ceil( |calculatedOutputWidth| ), then [=exception/throw=] a {{TypeError}}.
```

index.bs
fdwr: I think this is now resolved by Joshua?

index.bs
fdwr: Joshua resolved this in a separate CR. Thanks.

index.bs
fdwr: Updated.

fdwr:
I'm confused by the [build errors](https://github.com/webmachinelearning/webnn/actions/runs/16358575625/job/46222103485?pr=770), as I didn't touch any of those lines, and `[=string=]` seems unrelated 🙃:

```
      LINE 1713:36: Multiple possible 'string' dfn refs.
      Arbitrarily chose https://infra.spec.whatwg.org/#string
      To auto-select one of the following refs, insert one of these lines into a <pre class=link-defaults> block:
      spec:infra; type:dfn; text:string
      spec:rdf12-concepts; type:dfn; text:string
      [=string=]
```

dontcallmedom:
the build error is indeed unrelated to the particular change, but linked to the [definition set bikeshed uses for auto-linking](https://github.com/w3c/webref/issues/1619).

the simplest short term solution is to add `spec:infra; type:dfn; text:string` in the existing `<pre class="link-defaults">` on L50.

fdwr:
@dontcallmedom Thanks, that helped. I encountered a few other random errors (some which were unrelated to my change 🙃), and so I had to extend the linter to display line numbers to save my sanity:

```text
index.html:14723: Unexpected normative reference to [WEBNN]
```

huningxin:
From another perspective, would dropping `roundingType` prevent us from supporting dynamic shape (specifying the concrete input shape at the inference time) in the future? On native ML API side, both [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.pool.avg_pool) and [ONNX](https://onnx.ai/onnx/operators/onnx__AveragePool.html) supports `ceil_mode` which is same as [PyTorch](https://docs.pytorch.org/docs/stable/generated/torch.nn.AvgPool2d.html)’s ceil mode.

fdwr:
> From another perspective, would dropping roundingType prevent us from supporting dynamic shape (specifying the concrete input shape at the inference time) in the future?

**Ningxin**: We could always add it back, but yes, if free/dynamic/variable dimensions per https://github.com/webmachinelearning/webnn/issues/883 proceeds, then we'd probably want the output shape rounding direction (and then https://issues.chromium.org/issues/421421343 needs completing). In any case, mind if I rename the field to something more elucidating than `roundingType`, because it has nothing to do with generic rounding of the pooling operator (and no relation to the `roundEven` operator), but *specifically* to output *shape* computation. So maybe `MLPool2dOptions::outputShapeRounding`.

huningxin:
> `MLPool2dOptions::outputShapeRounding`.

SGTM

fdwr:
>> MLPool2dOptions::outputShapeRounding.

> SGTM

@huningxin: Any changes desired?

index.bs
```diff
@@ -7651,16 +7656,16 @@ partial dictionary MLOpSupportLimits {
...
     ::
         A list of length 2: *[outputHeight, outputWidth]*
         Specifies the sizes of the two spatial dimensions of the output tensor.
-
-        The spatial dimensions of the output tensor can be calculated for a single dimension via:
-
-        *output size = ((input size - filter size + beginning padding + ending padding) / stride) + 1*
-
-        Then the caller either applies a floor or ceiling depending on whether partial window results are desired.
+        When the output sizes are explicitly specified, the {{MLPool2dOptions/roundingType}} is ignored.
```
huningxin: Use {{MLPool2dOptions/outputShapeRounding}}?
```suggestion
        When the output sizes are explicitly specified, the {{MLPool2dOptions/outputShapeRounding}} is ignored.
```

index.bs
```diff
@@ -7671,7 +7676,13 @@ partial dictionary MLOpSupportLimits {
 
     **Returns:** an {{MLOperand}}. The output 4-D tensor that contains the
     result of the reduction. The logical shape is interpreted according to the
-    value of {{MLPool2dOptions/layout}}.
+    value of {{MLPool2dOptions/layout}}. More specifically, if the {{MLPool2dOptions/roundingType}} is {{MLRoundingType/"floor"}}, the spatial dimensions for a single dimension of the output tensor can be calculated as follows:
```
huningxin: Same here
```suggestion
    value of {{MLPool2dOptions/layout}}. More specifically, if the {{MLPool2dOptions/outputShapeRounding}} is {{MLRoundingType/"floor"}}, the spatial dimensions for a single dimension of the output tensor can be calculated as follows:
```

index.bs
```diff
@@ -7747,8 +7758,22 @@ partial dictionary MLOpSupportLimits {
     1. *Calculate the output shape:*
         1. Let « |windowHeight|, |windowWidth| » be |options|.{{MLPool2dOptions/windowDimensions}}.
         1. Let « |calculatedOutputHeight|, |calculatedOutputWidth| » be the result of [=MLGraphBuilder/calculating conv2d output sizes=] given |inputHeight|, |inputWidth|, |windowHeight|, |windowWidth|, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, and |options|.{{MLPool2dOptions/dilations}}.
+        1. If |options|.{{MLPool2dOptions/outputSizes}} [=map/exists=], then:
         1. Let « |outputHeight|, |outputWidth| » be |options|.{{MLPool2dOptions/outputSizes}}.
         1. If neither |outputHeight| equals floor( |calculatedOutputHeight| ) and |outputWidth| equals floor( |calculatedOutputWidth| ), nor |outputHeight| equals ceil( |calculatedOutputHeight| ) and |outputWidth| equals ceil( |calculatedOutputWidth| ), then [=exception/throw=] a {{TypeError}}.
+        1. Otherwise:
+            1. Let « |outputHeight|, |outputWidth| » be « |calculatedOutputHeight|, |calculatedOutputWidth| ».
+            1. Switch on |options|.{{MLPool2dOptions/roundingType}}:
```
huningxin: Same here
```suggestion
            1. Switch on |options|.{{MLPool2dOptions/outputShapeRounding}}:
```

index.bs
```diff
@@ -7747,8 +7758,22 @@ partial dictionary MLOpSupportLimits {
     1. *Calculate the output shape:*
         1. Let « |windowHeight|, |windowWidth| » be |options|.{{MLPool2dOptions/windowDimensions}}.
         1. Let « |calculatedOutputHeight|, |calculatedOutputWidth| » be the result of [=MLGraphBuilder/calculating conv2d output sizes=] given |inputHeight|, |inputWidth|, |windowHeight|, |windowWidth|, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, and |options|.{{MLPool2dOptions/dilations}}.
+        1. If |options|.{{MLPool2dOptions/outputSizes}} [=map/exists=], then:
         1. Let « |outputHeight|, |outputWidth| » be |options|.{{MLPool2dOptions/outputSizes}}.
```
huningxin: Should this line and the following line be sub-steps of "If..., then" step?

fdwr: Yeah, should indent. 👉👉👉👉

index.bs
fdwr: Derp, thanks. I'm surprised Bikeshed didn't complain about this.

index.bs
```diff
@@ -7759,11 +7759,11 @@ partial dictionary MLOpSupportLimits {
         1. Let « |windowHeight|, |windowWidth| » be |options|.{{MLPool2dOptions/windowDimensions}}.
         1. Let « |calculatedOutputHeight|, |calculatedOutputWidth| » be the result of [=MLGraphBuilder/calculating conv2d output sizes=] given |inputHeight|, |inputWidth|, |windowHeight|, |windowWidth|, |options|.{{MLPool2dOptions/padding}}, |options|.{{MLPool2dOptions/strides}}, and |options|.{{MLPool2dOptions/dilations}}.
         1. If |options|.{{MLPool2dOptions/outputSizes}} [=map/exists=], then:
-        1. Let « |outputHeight|, |outputWidth| » be |options|.{{MLPool2dOptions/outputSizes}}.
+            1. Let « |outputHeight|, |outputWidth| » be |options|.{{MLPool2dOptions/outputSizes}}.
         1. If neither |outputHeight| equals floor( |calculatedOutputHeight| ) and |outputWidth| equals floor( |calculatedOutputWidth| ), nor |outputHeight| equals ceil( |calculatedOutputHeight| ) and |outputWidth| equals ceil( |calculatedOutputWidth| ), then [=exception/throw=] a {{TypeError}}.
```
huningxin: And this line should also indent.

fdwr: True. I was thinking it would just fall through safely, but it's unnecessary to check this condition because we already know that `calculating conv2d output sizes` computing legal sizes.

PR #774, opened 2024-10-29
Convert MLOperand methods into readonly attributes

### LINKED ISSUE #666: Reconsider `MLOperand` methods

a-sully:
Fixes #666

Implements the changes proposed in https://github.com/webmachinelearning/webnn/issues/666#issuecomment-2342019337 while simplifying the declaration of these getters to use the pattern suggested in https://webidl.spec.whatwg.org/#idl-frozen-array

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/774.html" title="Last updated on Oct 29, 2024, 8:32 PM UTC (a8b0b3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/774/9287c0b...a-sully:a8b0b3d.html" title="Last updated on Oct 29, 2024, 8:32 PM UTC (a8b0b3d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1229,8 +1229,8 @@ For instance, an {{MLOperand}} may represent a constant feeding to an operation 
 <script type=idl>
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface MLOperand {
-  MLOperandDataType dataType();
-  sequence<unsigned long> shape();
+  readonly attribute MLOperandDataType dataType;
+  readonly attribute FrozenArray<unsigned long> shape;
 };
 
 dictionary MLOperatorOptions {
@@ -1315,8 +1315,7 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
 
 Issue(whatwg/webidl#1388): Support for unions of {{bigint}} and [=numeric types=] is new in [[WEBIDL]], and implementation support is also limited. Prototype implementations are encouraged to provide feedback for this approach.
 
-### {{MLOperand/dataType()}} ### {#api-mloperand-datatype}
-Returns the data type of the {{MLOperand}}.
+### {{MLOperand/dataType}} attribute ### {#api-mloperand-datatype}
 
 <div>
     **Returns:** an {{MLOperandDataType}}. The data type of the operand.
@@ -1324,13 +1323,12 @@ Returns the data type of the {{MLOperand}}.
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLOperand>dataType()</dfn> method steps are:
+    The <dfn attribute for=MLOperand>dataType</dfn> method steps are:
   </summary>
     1. Return [=this=]'s [=MLOperand/dataType=].
 </details>
 
-### {{MLOperand/shape()}} ### {#api-mloperand-shape}
-Returns the shape of the {{MLOperand}}.
+### {{MLOperand/shape}} attribute ### {#api-mloperand-shape}
 
 <div>
     **Returns:** [=sequence=]<{{unsigned long}}>. The shape of the operand.
@@ -1338,7 +1336,7 @@ Returns the shape of the {{MLOperand}}.
 
 <details open algorithm>
   <summary>
-    The <dfn method for=MLOperand>shape()</dfn> method steps are:
+    The <dfn attribute for=MLOperand>shape</dfn> method steps are:
   </summary>
     1. Return [=this=]'s [=MLOperand/shape=].
 </details>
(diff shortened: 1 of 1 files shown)


a-sully:
@fdwr this PR follows up on our recent thread on #666. I expect to migrate the Chromium implementation in the same way as the `dimensions` -> `shape` refactoring. PTAL?

a-sully:
> I expect to migrate the Chromium implementation in the same way as the `dimensions` -> `shape` refactoring. PTAL?

Hmm unfortunately this style of migration may not be possible in this case, since the Blink bindings code doesn't support interfaces which have a method and an attribute of the same name. See the crashes in https://crrev.com/c/5975719

We can have getters or methods, but not both at the same time... so a migration period is not possible :/ 

Would you prefer to make this change now or later?

fdwr:
> We can have getters or methods, but not both at the same time... so a migration period is not possible :/
> 
> Would you prefer to make this change now or later?

Well I'd like to frontload any breaking changes sooner than later, which smoothens the path to origin trial. However, if we can't mitigate this in any way on the Chromium end, can we closely synchronize the changes by having the corresponding update in samples already ready before completing the Chromium change?

@ibelem : For the [ORT-based samples](https://aka.ms/webnn), maybe we could update the sample logic to map a given Chromium version to a correspondingly compatible build of ORT, so there is implicitly a window of migration in the sample logic, even if Chromium breaks?

huningxin (approved):
LGTM

/cc @Honry @BruceDai for required changes in webnn-samples, webnn-baseline and tests.

fdwr (approved):
👍

Issue #778, opened 2024-11-05
Proposal: Report non-fatal errors from the WebNN timeline


a-sully:
## The Problem (see #477)

Our current method for surfacing `dispatch()` errors is to "lose" the `MLContext`. As I mentioned in https://github.com/webmachinelearning/webnn/pull/754#discussion_r1747441955 I don't think it makes sense for this to be the only option for surfacing errors from `dispatch()`:

> I don't think we can assume every failed `dispatch()` results in a lost `MLContext`, especially considering platforms where an `MLContext` is not so closely tied to a single GPU

Losing the `MLContext` is a very heavy-handed failure mode. In the current Chromium implementation (which, to be fair, is an implementation detail), this means killing the GPU process, which impacts the system well beyond the scope of WebNN. I don't think the `MLContext` is always the right blast radius for a `dispatch()` error.

There is also no way whatsoever to surface an error from `writeTensor()`!

## State of the World

Here are examples of how I've observed `dispatch()` fail in the current Chromium implementation:

1. The process executing the graph OOMs and crashes
   - blowing away the `MLContext` may indeed be the only option
2. Some resource allocation while executing the graph fails and aborts _gracefully_. Some thoughts on how to react:
    - it may be reasonable to blow away the entire `MLContext` e.g. if you assume an OOM is imminent,
    - it may be reasonable to just blow away the `MLGraph` e.g. if you suspect the graph may always be too large to ever execute successfully. This would free up some memory and possibly prevent an otherwise imminent OOM
    - it may be reasonable to assume this issue is transitory
3. Graph execution fails due to a runtime error _inherent in running the compiled graph in the current environment_, meaning that executing this graph _will always fail_. Ideally this type of failure would be surfaced earlier during `MLGraphBuilder.build()`, but unfortunately this is not always the case. **This is currently the most common failure mode for Chromium's CoreML backend**. Some thoughts on how to react:
    - blowing away the entire `MLContext` is not a useful option
    - it may be reasonable to blow away the `MLGraph`, especially if you're confident it will never execute successfully
    - it's possible - though it seems unlikely - this issue is transitory
    - you may not know whether you're actually in case 4
4. Graph execution fails due to a runtime error _caused by the specific graph inputs and outputs_. From what I can tell, this is always(?) due to issues with the user agent implementation. For example, TFLite does not support negative or OOB indices for `gather` ops (see #486), and so Chromium's TFLite backend must address this, which it does not (yet). Some thoughts on how to react:
[...]

PR #785, opened 2024-11-14
Editorial: Link to error handling proposal from the MLTensor explainer

a-sully:
Update the explainer to point to #778 instead of #477, which is now closed


huningxin (approved):
LGTM!

Issue #529, opened 2024-01-25
Specify WebNN timelines


a-sully:
The spec mentions some timelines (a "parallel timeline", "a GPU timeline", "a different timeline", "the offloaded timeline", etc...) but these timelines are not described anywhere. Meanwhile, https://github.com/webmachinelearning/webnn/issues/482 mentions a "device timeline" and "content timeline" (and, at the time of writing, there have been early discussions about whether an `MLQueue` is needed - which may or may not require a "queue timeline", as well)

These timelines should be clearly defined, including:
- their behavior on implementations that involve multiple devices and timelines (see https://github.com/webmachinelearning/webnn/issues/350)
- their interactions with [WebGPU's timelines](https://www.w3.org/TR/webgpu/#programming-model-timelines)

bbernhar:
From what I gather, we have at-least 3 timelines:

1. Content timeline. For JavaScript execution.
2. Context timeline. For any device or queue operation issued by the UA.
3. Timeline-agnostic. Catch all for when other timelines are not relevant.

WebNN is similar to WebGL's programming model in this aspect, `WebGLRenderingContext` is akin to `MLContext` in design, neither make the underlying queue or device visible to the web developer. WebGL does not interop with WebGPU but if it had, I suspect the timelines couldn't be 1:1 with WebGPU.

zolkis:
Is it correct to say that the current standard prose on [parallelism](https://html.spec.whatwg.org/#parallelism) is enough to capture timelines? 

> To run steps in parallel means those steps are to be run, one after another, at the same time as other logic in the standard (e.g., at the same time as the [event loop](https://html.spec.whatwg.org/#event-loop)). This standard does not define the precise mechanism by which this is achieved, be it time-sharing cooperative multitasking, fibers, threads, processes, using different hyperthreads, cores, CPUs, machines, etc. By contrast, an operation that is to run immediately must interrupt the currently running task, run itself, and then resume the previously running task.

Do we need to define a specialized term for timelines?

If we do, we should also define the relationships vs context, graph etc.:
- does a context support/encapsulate/control multiple timelines?
- can a graph be executed on multiple timelines / multiple contexts?

From the app script point of view, what is the minimal differentiation of terms we need to do?

bbernhar:
> Is it correct to say that the current standard prose on [parallelism](https://html.spec.whatwg.org/#parallelism) is enough to capture timelines?

[...]

PR #787, opened 2024-11-15
Specify MLTensor

a-sully:
This PR specifies key methods related to `MLTensor`, while hand-waving over much of the juicy details, such as how to actually specify the `MLContext`'s timeline (#529)

My hope is that this is a mostly-straightforward improvement which is "good enough" to allow us to remove `MLContext.compute()` as well as providing a more concrete starting point for more detailed discussions about e.g. specifying timelines in follow-up PRs

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/787.html" title="Last updated on Nov 22, 2024, 7:55 PM UTC (493ede3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/787/d4b4e9d...a-sully:493ede3.html" title="Last updated on Nov 22, 2024, 7:55 PM UTC (493ede3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -861,7 +861,7 @@ The {{MLContext}} interface represents a global state of neural network compute 
 
 <script type=idl>
 typedef record<USVString, ArrayBufferView> MLNamedArrayBufferViews;
-typedef record<DOMString, MLTensor> MLNamedTensors;
+typedef record<USVString, MLTensor> MLNamedTensors;
 
 dictionary MLComputeResult {
   MLNamedArrayBufferViews inputs;
@@ -870,6 +870,7 @@ dictionary MLComputeResult {
 
 [SecureContext, Exposed=(Window, DedicatedWorker)]
 interface MLContext {
+  // ISSUE(791): compute() will soon be removed in favor of dispatch().
   Promise<MLComputeResult> compute(
       MLGraph graph, MLNamedArrayBufferViews inputs, MLNamedArrayBufferViews outputs);
   undefined dispatch(MLGraph graph, MLNamedTensors inputs, MLNamedTensors outputs);  
@@ -930,7 +931,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
     To <dfn>validate buffer with descriptor</dfn> given {{AllowSharedBufferSource}} |bufferSource| and {{MLOperandDescriptor}} |descriptor|, run the following steps:
   </summary>
     1. If |bufferSource|'s [=BufferSource/byte length=] is not equal to |descriptor|'s [=MLOperandDescriptor/byte length=] return false.
-    1. Switch on |bufferSource|:
+    1. Switch on the type of |bufferSource|:
         <dl class=switch>
             : {{ArrayBuffer}}
             :: Return true.
@@ -972,7 +973,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
         1. If the byte length of |outputTensor| is not equal to |outputDesc|'s [=MLOperandDescriptor/byte length=], then return a {{TypeError}}.
         1. If |outputTensor|'s [=element type=] doesn't match |outputValue|'s [=element type=], then return a {{TypeError}}.
         1. Request the underlying implementation of |graph| to set the values of elements in |outputValue| to the values of elements in |outputTensor|.
-    1. Return undefined.
+    1. Return {{undefined}}.
 </details>
 
 ### {{MLNamedArrayBufferViews}} transfer algorithm ### {#mlnamedarraybufferviews-transfer-alg}
@@ -996,7 +997,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 ### {{MLContext/compute()}}  ### {#api-mlcontext-compute}
 
-ISSUE: {{MLContext/compute()}} will be deprecated and removed in favor of <code>[dispatch()](https://github.com/webmachinelearning/webnn/blob/main/mltensor-explainer.md#compute-vs-dispatch)</code>.
+ISSUE(791): {{MLContext/compute()}} will be deprecated and removed in favor of <code>[dispatch()](https://github.com/webmachinelearning/webnn/blob/main/mltensor-explainer.md#compute-vs-dispatch)</code>.
 
 Asynchronously carries out the computational workload of a compiled graph {{MLGraph}} on a separate timeline, either on a worker thread for the CPU execution, or on a GPU/NPU timeline for submitting a workload onto the command queue. The asynchronous nature of this call avoids blocking the calling thread while the computation for result is ongoing. This method of execution requires an {{MLContext}} created with {{MLContextOptions}}. Otherwise, it [=exception/throws=] an "{{OperationError}}" {{DOMException}}.
 
@@ -1104,16 +1105,16 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
     1. Let |allTensors| be a [=/list=] of {{MLTensor}}s consisting of |inputs|'s [=map/values=] [=list/extended=] by |outputs|'s [=map/values=].
     1. If |allTensors| contains any duplicate [=list/items=], then [=exception/throw=] a {{TypeError}}.
     1. [=list/For each=] |tensor| of |allTensors|:
-      1. If |tensor|.{{MLTensor/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
-      1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
-    1. If [=validating tensors with descriptors=] given |inputs| and |graph|. {{MLGraph/[[inputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If [=validating tensors with descriptors=] given |outputs| and |graph|. {{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
+        1. If |tensor|.{{MLTensor/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
+        1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating tensors with descriptors=] given |inputs| and |graph|.{{MLGraph/[[inputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating tensors with descriptors=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
 
             Issue(778): Add a mechanism for reporting errors during graph execution.
 
-    1. Return undefined.
+    1. Return {{undefined}}.
 </details>
 
 #### Examples #### {#api-mlcontext-dispatch-examples}
@@ -1254,7 +1255,7 @@ Bring-your-own-buffer variant of {{MLContext/readTensor(tensor)}}. Reads back th
 
                 Note: [=Validating buffer with descriptor=] above will fail if |outputData| is detached, but it's possible |outputData| may detach between then and now.
 
-            1. [=ArrayBuffer/write=] |bytes| to |outputData|.
+            1. [=ArrayBuffer/Write=] |bytes| to |outputData|.
             1. [=Resolve=] |promise| with {{undefined}}.
     1. Return |promise|.
 </details>
@@ -1286,10 +1287,10 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
 
             Issue(778): Add a mechanism for reporting errors while writing to a tensor.
 
-    1. Return undefined.
+    1. Return {{undefined}}.
 </details>
 
-Note: similar to `dispatch()`, `writeTensor()` itself provides no signal that the write has completed. To inspect the contents of a tensor, callers should await the results of reading back the tensor.
+Note: Similar to `dispatch()`, `writeTensor()` itself provides no signal that the write has completed. To inspect the contents of a tensor, callers should await the results of reading back the tensor.
 
 ### {{MLContext/opSupportLimits()}}  ### {#api-mlcontext-opsupportlimits}
 The {{MLContext/opSupportLimits()}} exposes level of support that differs across implementations at operator level. Consumers of the WebNN API are encouraged to probe feature support level by using {{MLContext/opSupportLimits()}} to determine the optimal model architecture to be deployed for each target platform.
@@ -1667,7 +1668,7 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
     1. Set [=this=].{{MLTensor/[[isDestroyed]]}} to true.
     1. Enqueue the following steps to [=this=].{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Release [=this=].{{MLTensor/[[data]]}}.
-    1. Return undefined.
+    1. Return {{undefined}}.
 </details>
 
 Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor once all previously submitted operations using it are complete.
--- a/index.bs
+++ b/index.bs
@@ -935,7 +935,7 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 <details open algorithm>
   <summary>
-    To <dfn>validate tensors with descriptors</dfn> given an {{MLNamedTensors}} |namedTensors| with [=record=]&lt;{{USVString}}, {{MLOperandDescriptor}}&gt |namedDescriptors|:
+    To <dfn>validate tensors with descriptors</dfn> given an {{MLNamedTensors}} |namedTensors| with [=record=]&lt;{{USVString}}, {{MLOperandDescriptor}}&gt; |namedDescriptors|:
   </summary>
     1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
--- a/index.bs
+++ b/index.bs
@@ -1215,7 +1215,7 @@ Reads back the {{MLTensor/[[data]]}} of an {{MLTensor}} from the {{MLContext}}.{
     1. Let |realm| be [=this=]'s [=relevant realm=].
     1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
-    1. If |tensor|'s {{MLTensor/readable}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
     1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
@@ -1243,7 +1243,7 @@ Bring-your-own-buffer variant of {{MLContext/readTensor(tensor)}}. Reads back th
     1. Let |global| be [=this=]'s [=relevant global object=].
     1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
-    1. If |tensor|'s {{MLTensor/readable}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If [=validating buffer with descriptor=] given |outputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
     1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
@@ -1277,7 +1277,7 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
   </summary>
     1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
-    1. If |tensor|'s {{MLTensor/writable}} returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/writable}} is false, then [=exception/throw=] a {{TypeError}}.
     1. If [=validating buffer with descriptor=] given |inputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
     1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |tensor|.{{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=].
--- a/index.bs
+++ b/index.bs
@@ -927,10 +927,18 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 <details open algorithm>
   <summary>
-    To <dfn>validate buffer with descriptor</dfn> given {{ArrayBufferView}} |bufferView| and {{MLOperandDescriptor}} |descriptor|, run the following steps:
+    To <dfn>validate buffer with descriptor</dfn> given {{AllowSharedBufferSource}} |bufferSource| and {{MLOperandDescriptor}} |descriptor|, run the following steps:
   </summary>
-    1. If |bufferView|'s [=element type=] does not match to |descriptor|.{{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility), return false.
-    1. If |bufferView|.\[[ByteLength]] is not equal to |descriptor|'s [=MLOperandDescriptor/byte length=], return false.
+    1. If |bufferSource|'s [=BufferSource/byte length=] is not equal to |descriptor|'s [=MLOperandDescriptor/byte length=] return false.
+    1. Switch on |bufferSource|:
+        <dl class=switch>
+            : {{ArrayBuffer}}
+            :: Return true.
+            : {{SharedArrayBuffer}}
+            :: Return true.
+            : {{ArrayBufferView}}
+            :: If |bufferSource|'s [=element type=] matches |descriptor|'s {{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility) return true, otherwise return false.
+        </dl>
 </details>
 
 <details open algorithm>
@@ -1087,7 +1095,7 @@ Schedules the computational workload of a compiled {{MLGraph}} on the {{MLContex
     **Returns:** {{undefined}}.
 </div>
 
-Note: `dispatch()` itself provides no signal that graph execution has completed. Rather, callers should await the results of reading back the output tensors. See [[#api-mlcontext-dispatch-examples]] below.
+Note: similar to `dispatch()`, `writeTensor()` itself provides no signal that the write has completed. To inspect the contents of a tensor, callers should await the results of reading back the tensor.
 
 <details open algorithm>
   <summary>
@@ -1103,7 +1111,7 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
     1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
 
-        Issue(778): Add a mechanism for reporting errors during graph execution.
+            Issue(778): Add a mechanism for reporting errors during graph execution.
 
     1. Return undefined.
 </details>
@@ -1132,13 +1140,13 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
     const [inputTensorA, inputTensorB, outputTensorC] =
         await Promise.all([
           context.createTensor({
-            dataType: 'float32', shape: [2, 2], writable: true
+            dataType: A.dataType, shape: A.shape, writable: true
           }),
           context.createTensor({
-            dataType: 'float32', shape: [2, 2], writable: true
+            dataType: B.dataType, shape: B.shape, writable: true
           }),
           context.createTensor({
-            dataType: 'float32', shape: [2, 2], readable: true
+            dataType: C.dataType, shape: C.shape, readable: true
           })
         ]);
 
@@ -1207,7 +1215,7 @@ Reads back the {{MLTensor/[[data]]}} of an {{MLTensor}} from the {{MLContext}}.{
     1. Let |realm| be [=this=]'s [=relevant realm=].
     1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
-    1. If |tensor|.{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/readable}}"] is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/readable}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
     1. Let |promise| be [=a new promise=].
     1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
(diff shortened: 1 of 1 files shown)


a-sully:
There's one remaining trivial bikeshed error which I intend to fix tomorrow. Please don't let that stop you from reviewing this PR :)

Thanks!

index.bs
```diff
@@ -1322,6 +1541,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+
+    : <dfn>\[[descriptor]]</dfn> of type {{MLTensorDescriptor}}
+    ::
+        The {{MLTensor}}'s descriptor.
+
+    : <dfn>\[[isDestroyed]]</dfn> of type {{boolean}}
+    ::
+        Whether {{MLTensor}}.{{MLTensor/destroy()}} has been called. Once destroyed, the {{MLTensor}} can no longer be used.
+
+    : <dfn>\[[data]]</dfn> of an [=implementation-defined=] type
```
a-sully, line 1596: I considered making this a _Data Block_ but I haven't fully thought through the implications of that yet, so... I went for maximal hand-waviness for now

https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-data-blocks

index.bs
```diff
@@ -1049,6 +1074,203 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+      1. If |tensor|'s {{MLTensor/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
+      1. If |tensor|'s {{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating tensors with descriptors=] given |inputs| and |graph|'s {{MLGraph/[[inputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating tensors with descriptors=] given |outputs| and |graph|'s {{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
+    1. Enqueue the following steps to |graph|'s {{MLGraph/[[context]]}}'s {{MLContext/[[timeline]]}}:
+        1. Issue a compute request to |graph|'s {{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
+
+        Issue(778): Add a mechanism for reporting errors during graph execution.
+
+    1. Return {{undefined}}.
```
inexorabletash, line 1106: Nit: per the processing model, this can just be undefined (not the WebIDL type `undefined`). Not a big deal, and I also don't think specs are consistent.

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+    ::
+        The {{MLTensor}}'s descriptor.
+
+    : <dfn>\[[isDestroyed]]</dfn> of type {{boolean}}
+    ::
+        Whether {{MLTensor}}.{{MLTensor/destroy()}} has been called. Once destroyed, the {{MLTensor}} can no longer be used.
+
+    : <dfn>\[[data]]</dfn> of an [=implementation-defined=] type
+    ::
+        The bytes backing the {{MLTensor}}. This data may only be modified from the {{MLTensor/[[context]]}}'s {{MLContext/[[timeline]]}}.
```
inexorabletash, line 1601: Reading is similarly restricted to the timeline, isn't it?

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+<div class=internal-slots>
+{{MLTensor}} has the following internal slots:
+  <dl dfn-type=attribute dfn-for="MLTensor">
+    : <dfn>\[[context]]</dfn> of type {{MLContext}}
+    ::
+        The {{MLTensor}}'s associated context.
+
+    : <dfn>\[[descriptor]]</dfn> of type {{MLTensorDescriptor}}
+    ::
+        The {{MLTensor}}'s descriptor.
```
inexorabletash, line 1593: Is it worth mentioning that this is immutable? We don't do that for many other slots/properties, so not required.

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+
+The <dfn attribute for=MLTensor>shape</dfn> [=getter steps=] are to return [=this=]'s [=MLTensor/shape=].
+
+The <dfn attribute for=MLTensor>readable</dfn> [=getter steps=] are to return [=this=]'s [=MLTensor/readable=].
+
+The <dfn attribute for=MLTensor>writable</dfn> [=getter steps=] are to return [=this=]'s [=MLTensor/writable=].
+
+### Creating an {{MLTensor}} ### {#api-mltensor-create}
+
+An {{MLTensor}} is created with by its associated {{MLContext}}. Note that creating an {{MLTensor}} does not include initializing its {{MLTensor/[[data]]}}. This {{MLTensor/[[data]]}} should be initialized soon afterwards.
```
inexorabletash, line 1623: Who are this "note" and "should" sentences aimed at? Since this is an internal algorithm, as written I think these comments are only useful for spec authors. Just drop them? 

also, typo: "with by"

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+  <summary>
+    The <dfn method for=MLTensor>destroy()</dfn> method steps are:
+  </summary>
+    1. Set [=this=]'s {{MLTensor/[[isDestroyed]]}} to true.
+    1. Enqueue the following steps to [=this=]'s {{MLTensor/[[context]]}}'s {{MLContext/[[timeline]]}}:
+        1. Free [=this=]'s {{MLTensor/[[data]]}}.
+    1. Return {{undefined}}.
+</details>
+
+Note: Since no further operations can be enqueued using this tensor, implementations can free resource allocations associated with this tensor.
```
inexorabletash, line 1654: Maybe "can free any additional resource allocations", since the steps are pretty explicit about freeing what seems to be the primary resource.

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+    1. Let |tensor| be a new {{MLTensor}}.
+    1. Set |tensor|'s {{MLTensor/[[context]]}} to |context|.
+    1. Set |tensor|'s {{MLTensor/[[descriptor]]}} to |descriptor|.
+    1. Set |tensor|'s {{MLTensor/[[isDestroyed]]}} to false.
+    1. Return |tensor|.
+</details>
+
+### {{MLTensor/destroy()}} ### {#api-mltensor-destroy}
+
+Destroys the {{MLTensor}}. This method is idempotent.
```
inexorabletash, line 1638: Maybe: "Frees the resources associated with the MLTensor." ? Just because "destroy" might imply that the object itself is deleted, which it isn't.

(Now that these sorts of methods are proliferating in the platform, I'm guessing standard spec text will emerge. Apologies if you've already researched this!)

index.bs
```diff
@@ -1049,6 +1074,203 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    The <dfn method for=MLContext>readTensor(|tensor|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. If |tensor|'s {{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. Let |promise| be [=a new promise=].
+    1. Enqueue the following steps to |tensor|'s {{MLGraph/[[context]]}}'s {{MLContext/[[timeline]]}}:
+        1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|'s {{MLTensor/[[data]]}}.
```
inexorabletash, line 1211: Can this fail? Maybe that's part of #778

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+    ::
+        The bytes backing the {{MLTensor}}. This data may only be modified from the {{MLTensor/[[context]]}}'s {{MLContext/[[timeline]]}}.
+  </dl>
+</div>
+
+An {{MLTensor}}'s <dfn for=MLTensor>dataType</dfn> is its {{MLTensor/[[descriptor]]}}'s {{MLOperandDescriptor/dataType}}.
+
+An {{MLTensor}}'s <dfn for=MLTensor>shape</dfn> is its {{MLTensor/[[descriptor]]}}'s {{MLOperandDescriptor/shape}}.
+
+An {{MLTensor}} <dfn for=MLTensor>readable</dfn> if its {{MLTensor/[[descriptor]]}}'s {{MLTensorDescriptor/readable}} is true.
```
inexorabletash, line 1609: nit: "An MLTensor is readable..." (missing "is") - and decide if you want "is" part of the definition or not.

index.bs
```diff
@@ -1322,6 +1544,115 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+
+An {{MLTensor}} <dfn for=MLTensor>writable</dfn> if its {{MLTensor/[[descriptor]]}}'s {{MLTensorDescriptor/writable}} is true.
```
inexorabletash, line 1611: nit: "An MLTensor is writable..." (ditto)

index.bs
```diff
@@ -1049,6 +1074,203 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>readTensor(|tensor|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. If |tensor|'s {{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
```
inexorabletash, line 1208: The definition for MLTensor/readable has minor issues (see below) but I think this should read "If tensor is readable ..." or "If tensor.[[descriptor]].["readable"] ..."

index.bs
```diff
@@ -1049,6 +1074,203 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+</div>
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>readTensor(|tensor|, |outputData|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. If |tensor|'s {{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
```
inexorabletash, line 1235: Same as previous "readable" issue

index.bs
```diff
@@ -1049,6 +1074,203 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    **Returns:** {{undefined}}.
+</div>
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>writeTensor(|tensor|, |inputData|)</dfn> method steps are:
+  </summary>
+    1. If |tensor|'s {{MLGraph/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|'s {{MLTensor/writable}} is false, then [=exception/throw=] a {{TypeError}}.
```
inexorabletash, line 1262: Same as previous "readable" issue

index.bs
a-sully: Resolving since we can continue discussion in the duplicate thread

index.bs
a-sully: Removed these definitions since I think they cause more confusion than good... WDYT?

index.bs
a-sully: This is now "If |tensor|.{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/readable}}"] is false..."

index.bs
a-sully: removed definition

index.bs
a-sully: added "accessed"

index.bs
a-sully: I'll leave this for a follow-up since `MLOperand` could use this treatment, too

index.bs
a-sully: It can. Not really part of #778, though, since `readTensor()` is already async and can already return errors when reading back the tensor data

Anyways, added a failure step

index.bs
```diff
@@ -1049,6 +1074,203 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+
+Schedules the computational workload of a compiled {{MLGraph}} on the {{MLContext}}'s {{MLContext/[[timeline]]}}.
+
+<div dfn-for="MLContext/dispatch(graph, inputs, outputs)" dfn-type=argument>
+    **Arguments:**
+      - <dfn>graph</dfn>: an {{MLGraph}}. The computational graph to be executed.
+      - <dfn>inputs</dfn>: an {{MLNamedTensors}}. The inputs to the computational graph.
+      - <dfn>outputs</dfn>: an {{MLNamedTensors}}. The outputs to the computational graph.
+
+    **Returns:** {{undefined}}.
```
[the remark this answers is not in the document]
a-sully: Good idea. Added a note about this

a-sully:
Apologies for the accidental force-push 🤦 

Thanks for the feedback @inexorabletash! I believe I've addressed all your comments

index.bs
inexorabletash: I thought about suggesting that too. :) Removing SGTM

index.bs
inexorabletash: Oops, sorry... I forgot we don't use map syntax for dictionaries [we use dotted property syntax for those too](https://github.com/webmachinelearning/webnn/blob/main/docs/SpecCodingConventions.md#dictionary-members), so it should be:

"If tensor.[[descriptor]].readable ..."

index.bs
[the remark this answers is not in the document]
inexorabletash: LGTM!

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+  <summary>
+    The <dfn method for=MLContext>readTensor(|tensor|, |outputData|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/readable}}"] is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If [=validating buffer with descriptor=] given |outputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. Let |promise| be [=a new promise=].
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}'s {{MLContext/[[timeline]]}}:
```
inexorabletash, line 1241: This is still using "foo's [[internal slot]]" for the second part, when we more consistently use dotted notation for internal slots. But I admit "tensor.[[context]].[[timeline]]" is a little unwieldy too.

index.bs
inexorabletash: SGTM!

inexorabletash (commented):
Thanks! Most updates LGTM.

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>readTensor(|tensor|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/readable}}"] is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
```
huningxin, line 1210: Could the `readable` getter be referenced by `{{MLTensor/readable}}`? Such as
```suggestion
    1. If |tensor|'s {{MLTensor/readable}} is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
```

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+</div>
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>readTensor(|tensor|, |outputData|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/readable}}"] is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
```
huningxin, line 1238: Same here

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    1. If [=validating buffer with descriptor=] given |outputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. Let |promise| be [=a new promise=].
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}'s {{MLContext/[[timeline]]}}:
+        1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
+        1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+        1. Otherwise, [=queue an ML task=] with |global| to [=ArrayBuffer/write=] |bytes| to |outputData| and then [=resolve=] |promise| with {{undefined}}.
```
huningxin, line 1244: The `outputData` might be detached before this step gets executed. Should we throw an exception? It seems arraybuffer-write steps just assert arraybuffer's byte length is greater than bytes' length.

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
+    1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |tensor|.{{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=].
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Copy |bytes| to |tensor|.{{MLTensor/[[data]]}}.
+
+        Issue(778): Add a mechanism for reporting errors while writing to a tensor.
+
+    1. Return undefined.
+</details>
+
```
huningxin, line 1277: Is it necessary to also add a note mentioning there is no signal that the write operation has completed.

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    **Returns:** {{undefined}}.
+</div>
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>writeTensor(|tensor|, |inputData|)</dfn> method steps are:
+  </summary>
+    1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/writable}}"] is false, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 1266: Same here

huningxin (commented):
Thanks @a-sully !

index.bs
```diff
@@ -1049,6 +1074,207 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>readTensor(|tensor|, |outputData|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If |tensor|{{MLTensor/[[descriptor]]}}["{{MLTensorDescriptor/readable}}"] is false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
+    1. If [=validating buffer with descriptor=] given |outputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then return [=a new promise=] [=rejected=] with a {{TypeError}}.
```
huningxin, line 1239: ["validate buffer with descriptor"](https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/787.html#validate-buffer-with-descriptor) algorithm steps take an `ArrayBufferView` and check its element type according to descriptor data type. Should we change it to support `AllowSharedBufferSource`?

index.bs
a-sully: Just to clarify, are you referring to the case where the buffer is detached between the call to the "validating buffer with descriptor" algorithm above (which checks byte lengths are equal) and here?

Seems reasonable to me. I've updated the algorithm and added a note describing this case. @reillyeon do you know off the top of your head whether this is currently handled in the Chromium implementation?

index.bs
a-sully: SGTM, added a note similar to that of `dispatch()`

index.bs
```diff
@@ -1322,6 +1548,111 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+  <summary>
+    The <dfn method for=MLTensor>destroy()</dfn> method steps are:
+  </summary>
+    1. Set [=this=].{{MLTensor/[[isDestroyed]]}} to true.
+    1. Enqueue the following steps to [=this=].{{MLTensor/[[context]]}}'s {{MLContext/[[timeline]]}}:
+        1. Free [=this=].{{MLTensor/[[data]]}}.
+    1. Return undefined.
+</details>
+
+Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor.
```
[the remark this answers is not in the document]
a-sully: I don't think that's necessary, especially since resources will be freed on _both the context and content timelines_. Please feel free to unresolve if you disagree

index.bs
reillyeon: Chromium handles this by throwing a `TypeError`. It provides a specific message that the buffer was detached, but it's also semantically equivalent to the original size check which asserted that the contents of the tensor would fit in the buffer, since a detached buffer has zero length.

index.bs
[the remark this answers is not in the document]
huningxin: It seems not clear to me what the additional resources are. If they are allocated on GPU/NPU device, for example, the staging buffers for uploading/reading-back, should they be released on context's timeline?

index.bs
```diff
@@ -1049,6 +1082,216 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+<div dfn-for="MLContext/dispatch(graph, inputs, outputs)" dfn-type=argument>
+    **Arguments:**
+      - <dfn>graph</dfn>: an {{MLGraph}}. The computational graph to be executed.
+      - <dfn>inputs</dfn>: an {{MLNamedTensors}}. The inputs to the computational graph.
+      - <dfn>outputs</dfn>: an {{MLNamedTensors}}. The outputs of the computational graph.
+
+    **Returns:** {{undefined}}.
+</div>
+
+Note: similar to `dispatch()`, `writeTensor()` itself provides no signal that the write has completed. To inspect the contents of a tensor, callers should await the results of reading back the tensor.
```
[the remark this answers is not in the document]
a-sully: Doh, that's a silly error. Fixing in the next patch

index.bs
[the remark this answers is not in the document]
a-sully: Right, but the resources used to implement an `MLTensor` (such as staging buffers) are an implementation detail, and once the tensor can no longer be used, it shouldn't matter what timeline the resources are released on (or if they're released at all) since it's not observable

This is a non-normative note anyways so I don't think we need to be dictating timelines here?

index.bs
inexorabletash: Algorithm steps generally shouldn't be accessing getters, since script can overwrite the getter. Specs do get a bit sloppy here, but referencing the internal state is preferred.

a-sully: Ack. Fixed in [2c9ddfb](https://github.com/webmachinelearning/webnn/pull/787/commits/2c9ddfb7c0808095cbfcd298c9186d2123c1f9ac)

huningxin: @inexorabletash thanks for the clarification! @a-sully , apologize for my misunderstanding.

index.bs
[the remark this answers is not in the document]
huningxin: > once the tensor can no longer be used, it shouldn't matter what timeline the resources are released on (or if they're released at all) since it's not observable

The already enqueued tensor write/read operations may still need to access the staging buffers.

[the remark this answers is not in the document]
a-sully: I'm hesitant to get too in the weeds in a non-normative note. From the perspective of the spec, what's important is that this `MLTensor` can no longer be used, so UAs are can release its associated resources as they see fit. Of course some of those resources might be bound up in ongoing operations, but from the perspective of the spec that's an implementation detail we trust the UA to figure out without running into UAFs.

FWIW this text was lifted directly from [the `GPUBuffer.destroy()` method of the WebGPU spec](https://www.w3.org/TR/webgpu/#dom-gpubuffer-destroy), which this method is modeled after 😛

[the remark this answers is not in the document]
a-sully: > I feel "release" may be a better term

Agreed, updated the algorithm

> BTW, I found [GPUBuffer Destruction](https://www.w3.org/TR/webgpu/#buffer-destruction) has a note for GPU resource:
> 
> > Note: This allows the user agent to reclaim the GPU memory associated with the GPUBuffer once all previously submitted operations using it are complete.

Ah yes, good call out. I've amended the note here to reflect that... WDYT?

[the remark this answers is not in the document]
huningxin: Looks good! Thanks for your patience.

huningxin (approved):
LGTM!

huningxin:
@inexorabletash @reillyeon Any further comments? Also add @fdwr , PTAL.

index.bs
```diff
@@ -861,6 +861,7 @@ The {{MLContext}} interface represents a global state of neural network compute
 
 <script type=idl>
 typedef record<USVString, ArrayBufferView> MLNamedArrayBufferViews;
+typedef record<DOMString, MLTensor> MLNamedTensors;
```
inexorabletash, line 864: Can we use `USVString` instead?

(I'm guessing this is from long iteration on the MLBuffer explainer, which predated the spec adopting USVString in 3a916f26feed501467f37110db5754e8de73151a)

index.bs
```diff
@@ -871,7 +872,15 @@ dictionary MLComputeResult {
 interface MLContext {
   Promise<MLComputeResult> compute(
```
inexorabletash, line 873: Maybe introduce a comment here that `compute()` is going away. There's a note later, but some of us (maybe just me?) fixate on the IDL.

index.bs
```diff
@@ -913,10 +927,29 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
 
 <details open algorithm>
   <summary>
-    To <dfn>validate buffer with descriptor</dfn> given {{ArrayBufferView}} |bufferView| and {{MLOperandDescriptor}} |descriptor|, run the following steps:
+    To <dfn>validate buffer with descriptor</dfn> given {{AllowSharedBufferSource}} |bufferSource| and {{MLOperandDescriptor}} |descriptor|, run the following steps:
+  </summary>
+    1. If |bufferSource|'s [=BufferSource/byte length=] is not equal to |descriptor|'s [=MLOperandDescriptor/byte length=] return false.
+    1. Switch on |bufferSource|:
```
inexorabletash, line 933: Maybe "Switch on the type of ..." 

(Usually "switch" is operating on a value. https://garykac.github.io/procspec/#switch would even suggest that the cases be phrases, e.g. "If buffer source is a ..." but I think that's overkill here.)

index.bs
```diff
@@ -913,10 +927,29 @@ When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with t
...
+<details open algorithm>
+  <summary>
+    To <dfn>validate tensors with descriptors</dfn> given an {{MLNamedTensors}} |namedTensors| with [=record=]&lt;{{USVString}}, {{MLOperandDescriptor}}&gt; |namedDescriptors|:
   </summary>
-    1. If |bufferView|'s [=element type=] does not match to |descriptor|.{{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility), return false.
-    1. If |bufferView|.\[[ByteLength]] is not equal to |descriptor|'s [=MLOperandDescriptor/byte length=], return false.
+    1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
+    1. [=map/For each=] |name| → |tensor| of |namedTensors|:
+        1. If |namedDescriptors|[|name|] does not [=map/exist=], then return false.
+        1. If |tensor|.{{MLTensor/[[descriptor]]}} is not equal to |namedDescriptors|[|name|], then return false.
```
inexorabletash, line 951: I think we can defer to a follow-up, but this is assuming an equality for maps. Ideally we'd either have a definition for that (https://garykac.github.io/procspec/#conditions) or equate the specific values.

This is not new to this addition - we use list equality all over the place, and that's not rigorously defined.

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+Note: `dispatch()` itself provides no signal that graph execution has completed. Rather, callers should await the results of reading back the output tensors. See [[#api-mlcontext-dispatch-examples]] below.
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>dispatch(|graph|, |inputs|, |outputs|)</dfn> method steps are:
+  </summary>
+    1. Let |allTensors| be a [=/list=] of {{MLTensor}}s consisting of |inputs|'s [=map/values=] [=list/extended=] by |outputs|'s [=map/values=].
+    1. If |allTensors| contains any duplicate [=list/items=], then [=exception/throw=] a {{TypeError}}.
+    1. [=list/For each=] |tensor| of |allTensors|:
+      1. If |tensor|.{{MLTensor/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
```
inexorabletash, line 1107: Not indented enough - doesn't render as a sub-list. Bikeshed's markdown needs 4 spaces.

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+      1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating tensors with descriptors=] given |inputs| and |graph|. {{MLGraph/[[inputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
```
inexorabletash, line 1109: Remove extra space after `|graph|.`

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    1. If [=validating tensors with descriptors=] given |outputs| and |graph|. {{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
```
inexorabletash, line 1110: Remove extra space after `|graph|.`

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>createTensor(|descriptor|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. Let |tensor| be the result of [=creating an MLTensor=] given [=this=], and |descriptor|.
+    1. Let |promise| be [=a new promise=].
+    1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}:
+        1. Create |tensor|.{{MLTensor/[[data]]}} given |descriptor| and initialize all bytes to zeros.
+        1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
```
inexorabletash, line 1194: Future: I think we are supposed to ensure the exception is created with the same _realm_ as the original caller, but the existing spec text isn't consistent about doing that for parallel steps that throw.

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    1. Let |promise| be [=a new promise=].
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Let |bytes| be a [=/byte sequence=] containing a copy of |tensor|.{{MLTensor/[[data]]}}.
+        1. If that fails, then [=queue an ML task=] with |global| to [=reject=] |promise| with an "{{UnknownError}}" {{DOMException}}, and abort these steps.
+        1. Otherwise, [=queue an ML task=] with |global| and the following steps:
+            1. If |outputData| is [=BufferSource/detached=], [=reject=] |promise| with a {{TypeError}}, and abort these steps.
+
+                Note: [=Validating buffer with descriptor=] above will fail if |outputData| is detached, but it's possible |outputData| may detach between then and now.
+
+            1. [=ArrayBuffer/write=] |bytes| to |outputData|.
```
inexorabletash, line 1257: `Write` should be capitalized. (Don't worry, Bikeshed will link it correctly.)

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+            1. [=Resolve=] |promise| with {{undefined}}.
```
inexorabletash, line 1258: Don't need to use `{{}}` around undefined, like the "Return undefined" usage (but as long as we're consistent I'm happy)

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>writeTensor(|tensor|, |inputData|)</dfn> method steps are:
+  </summary>
+    1. If |tensor|.{{MLGraph/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/writable}} is false, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating buffer with descriptor=] given |inputData| and |tensor|.{{MLTensor/[[descriptor]]}} returns false, then [=exception/throw=] a {{TypeError}}.
+    1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
+    1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |tensor|.{{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=].
```
inexorabletash, line 1283: Do we have to deal with growable SABs, i.e. that might asynchronously change size? IDL says they are a "shared buffer of a fixed number of bytes" but ECMAScript supports `grow()` on them? I haven't dug in, but maybe this is a potential TOC/TOU issue and we need to snapshot the initial size?

index.bs
```diff
@@ -1049,6 +1082,215 @@ Note: Invocations of {{MLContext/compute()}} will fail if any of the {{MLContext
...
+    1. Enqueue the following steps to |tensor|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Copy |bytes| to |tensor|.{{MLTensor/[[data]]}}.
+
+            Issue(778): Add a mechanism for reporting errors while writing to a tensor.
+
+    1. Return undefined.
+</details>
+
+Note: similar to `dispatch()`, `writeTensor()` itself provides no signal that the write has completed. To inspect the contents of a tensor, callers should await the results of reading back the tensor.
```
inexorabletash, line 1292: Nit: Capitalize "Similar"

index.bs
inexorabletash: Oh good - thank you for doing my research for me!

inexorabletash (approved):
LGTM! (I can't mark comment threads as resolved)

index.bs
a-sully: Filed #793

index.bs
```diff
@@ -1325,6 +1568,111 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
 
+## {{MLTensorDescriptor}} dictionary ## {#api-mltensordescriptor}
+
+An {{MLTensorDescriptor}} describes the characteristics and capabilities of an {{MLTensor}}.
+
+<script type=idl>
+dictionary MLTensorDescriptor : MLOperandDescriptor {
+  boolean readable = false;
+  boolean writable = false;
+};
```
fdwr, line 1579: There was also a `boolean importableToWebGPU = false;` flag before. Is that deferred until later?

index.bs
```diff
@@ -1325,6 +1568,111 @@ To <dfn for="MLGraphBuilder">validate operand</dfn> given {{MLGraphBuilder}} |bu
...
+    1. Let |tensor| be a new {{MLTensor}}.
+    1. Set |tensor|.{{MLTensor/[[context]]}} to |context|.
+    1. Set |tensor|.{{MLTensor/[[descriptor]]}} to |descriptor|.
+    1. Set |tensor|.{{MLTensor/[[isDestroyed]]}} to false.
+    1. Return |tensor|.
+</details>
+
+### {{MLTensor/destroy()}} ### {#api-mltensor-destroy}
+
+Releases the resources associated with the {{MLTensor}}. This method is idempotent.
```
fdwr, line 1658: `This method is idempotent`

Hmm, what does it mean in this context? `idempotent` to me means yielding the same value (e.g. x * 1 or x + 0). Do you mean that calling `destroy` multiple times is valid and does not throw an error the 2nd/3rd/4th time?

fdwr (commented):
Minor comments, else LGTM. Thanks for including in the spec. @RafaelCintron, any thoughts?

index.bs
fdwr: Okie dokie.

fdwr (approved):

PR #786, opened 2024-11-14
Remove descriptions of outdated MLContext requirements

a-sully:
These statements are no longer accurate:

- an `MLContext` created with the `"gpu"``MLDeviceType` does in fact accept inputs which are not `ArrayBufferView`s
- `compute()` does not throw if the `MLContext` is not created with `MLContextOptions`

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/786.html" title="Last updated on Nov 14, 2024, 10:48 PM UTC (b1c62df)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/786/9a11f60...a-sully:b1c62df.html" title="Last updated on Nov 14, 2024, 10:48 PM UTC (b1c62df)">Diff</a>

Issue #749, opened 2024-08-08
MLContextOptions.deviceType seems unnecessary outside of conformance testing


mwyrzykowski:
WebKit has a few objections to MLContextOptions.deviceType:

### MLContextOptions.deviceType is currently unimplementable currently via CoreML
CoreML is the only public framework on macOS for running computations on the NPU / ANE. The reason as pointed out https://github.com/webmachinelearning/webnn/issues/623#issuecomment-2025767956 and possibly elsewhere, CoreML allows you to specify MLComputeUnitsCPUOnly, MLComputeUnitsCPUAndGPU, MLComputeUnitsAll, and MLComputeUnitsCPUAndNeuralEngine, but not MLComputeUnitsNeuralEngine only for instance. Even assuming we could request CoreML to add support for only the ANE, that support would be limited to a future version of macOS and therefore could only be supported on currently unreleased versions of macOS.

Also note, MLComputeUnitsCPUAndNeuralEngine, is only available on macOS 13.0 and later and I'm not sure if some browser vendors want to support earlier versions of macOS.

### MLContextOptions.deviceType, as currently specified, would lead to additional fragmentation, works on some devices, fails on others
Under [Section 4](https://www.w3.org/TR/webnn/#privacy) we see the wording:
> if this type cannot be satisfied, an "[OperationError](https://webidl.spec.whatwg.org/#operationerror)" [DOMException](https://webidl.spec.whatwg.org/#idl-DOMException) is thrown,

It would be preferable to fallback instead of failing, otherwise a website which specifies NPU and runs on a variety of NPUs, but maybe not the NPU of a certain device, the website would not run if the author does not handle fallback. This seems unnecessary, when the browser could fallback to the GPU or CPU themselves.

### The browser has better insight into workloads than the website author
It was pointed out in https://github.com/webmachinelearning/webnn/pull/322 that a website author may want their models to not run on the GPU so they don't impact WebGPU rendering performance. While a good idea in theory, the browser implementation and native frameworks have much greater insight into the graphics chip utilization and the most appropriate resource to run a model on without impacting target frameworks.

[...]

PR #795, opened 2024-11-27
Remove the MLContext.compute() method

### LINKED ISSUE #791: Remove `MLContext.compute()`
This method has been superseded by `MLContext.dispatch()`, which is being specified in #787. Once that PR merges, we should remove `compute()` from the spec

Note that `compute()` [has already been removed from the Chromium implementation](https://crbug.com/331351967)

a-sully:
Fixes #791. Made possible by #787 which specified `dispatch()`

The changes proposed in #786 have also been rolled into this PR, since they can't really be decoupled from `compute()` removal at this point

I've tried to keep this PR minimal and focused on removing algorithms and IDL definitions which are no longer relevant. The spec still contains lots of normative text which was written with `compute()` in mind which doesn't map as well to `dispatch()` - specifically text which contains assumptions about how graph execution relates to compute devices. I've lightly updated some of this text where necessary to match the `dispatch()` paradigm, but my intention is to avoid opening the can of worms that is #749 in this PR :)

***
<a href="https://pr-preview.s3.amazonaws.com/a-sully/webnn/pull/795.html" title="Last updated on Nov 28, 2024, 5:02 AM UTC (efdc1e8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/795/dd5cbdc...a-sully:efdc1e8.html" title="Last updated on Nov 28, 2024, 5:02 AM UTC (efdc1e8)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -624,7 +624,7 @@ In order to not allow an attacker to target a specific implementation that may c
 
 Issue: Hinting partially mitigates the concern. Investigate additional mitigations.
 
-The API design minimizes the attack surface for the compiled computational graph. The {{MLGraphBuilder}} interface that hosts the various operations is a data definition API and as such doesn't execute anything, only constructs data. What follows, is that the potential for an attack is limited to when binding the data to the graph before executing it by invoking the {{MLContext}}.{{MLContext/compute()}} method. This enables implementers to focus on hardening the {{MLContext}}.{{MLContext/compute()}} method. For example, by making sure it honors the boundary of data and fails appropriately when the bounds are not respected.
+The API design minimizes the attack surface for the compiled computational graph. The {{MLGraphBuilder}} interface that hosts the various operations is a data definition API and as such doesn't execute anything, only constructs data. What follows, is that the potential for an attack is limited to when binding the data to the graph before executing it by invoking the {{MLContext}}.{{MLContext/dispatch()}} method. This enables implementers to focus on hardening the {{MLContext}}.{{MLContext/dispatch()}} method. For example, by making sure it honors the boundary of data and fails appropriately when the bounds are not respected.
 
 Purpose-built Web APIs for measuring high-resolution time mitigate against timing attacks using techniques such as resolution reduction, adding jitter, detection of abuse and API call throttling [[hr-time-3]]. The practical deployment of WebNN implementations are likely to bring enough jitter to make timing attacks impractical (e.g. because they would use IPC) but implementers are advised to consider and test their implementations against timing attacks.
 
@@ -694,7 +694,7 @@ A key part of the {{MLGraphBuilder}} interface are methods such as {{MLGraphBuil
 
 An [=operator=] has a <dfn for=operator>label</dfn>, a string which may be included in diagnostics such as [=exception=] messages. When an [=operator=] is created its [=operator/label=] is initialized in an [=implementation-defined=] manner and may include the passed {{MLOperatorOptions/label}}.
 
-Note: Implementations are encouraged to use the {{MLOperatorOptions/label}} provided by developers to enhance error messages and improve debuggability, including both synchronous errors during graph construction and for errors that occur during asynchronous {{MLGraphBuilder/build()}} or {{MLContext/compute()}} operations.
+Note: Implementations are encouraged to use the {{MLOperatorOptions/label}} provided by developers to enhance error messages and improve debuggability, including both synchronous errors during graph construction and for errors that occur during asynchronous {{MLGraphBuilder/build()}} or {{MLContext/dispatch()}} operations.
 
 At inference time, every {{MLOperand}} will be bound to a tensor (the actual data), which are essentially multidimensional arrays. The representation of the tensors is implementation dependent, but it typically includes the array data stored in some buffer (memory) and some metadata describing the array data (such as its shape).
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8488,33 +8341,6 @@ Compile the graph up to the output operand.
...
-    const outputs = {
-      'output': outputBuffer
-    };
-    const result = await context.compute(graph, inputs, outputs);
-
-    console.log('Output value: ' + result.outputs.output);
-    // Output value: 2.25,2.25,2.25,2.25,2.25,2.25,2.25,2.25
-  </pre>
-</details>
-</div>
```
a-sully, line 8516: I considered re-creating this example with `dispatch()`... but it would end up looking a whole lot like https://webmachinelearning.github.io/webnn/#api-mlcontext-dispatch-examples. Is that still useful? Open to suggestions

a-sully:
@huningxin @fdwr PTAL?

inexorabletash (approved):
LGTM

This does make me realize that as part of #529 we'll likely need to define queuing for operations on the timeline that share tensors, to avoid racy behavior. With ArrayBuffers transferred away synchronously during `compute()`, there wasn't the possibility for multiple `compute()`s to compete for access to the buffer or script to read/write the contents.

index.bs
```diff
@@ -694,7 +694,7 @@ A key part of the {{MLGraphBuilder}} interface are methods such as {{MLGraphBuil
 
 An [=operator=] has a <dfn for=operator>label</dfn>, a string which may be included in diagnostics such as [=exception=] messages. When an [=operator=] is created its [=operator/label=] is initialized in an [=implementation-defined=] manner and may include the passed {{MLOperatorOptions/label}}.
 
-Note: Implementations are encouraged to use the {{MLOperatorOptions/label}} provided by developers to enhance error messages and improve debuggability, including both synchronous errors during graph construction and for errors that occur during asynchronous {{MLGraphBuilder/build()}} or {{MLContext/compute()}} operations.
+Note: Implementations are encouraged to use the {{MLOperatorOptions/label}} provided by developers to enhance error messages and improve debuggability, including both synchronous errors during graph construction and for errors that occur during asynchronous {{MLGraphBuilder/build()}} or {{MLContext/dispatch()}} operations.
```
huningxin, line 697: {{MLContext/dispatch()}} doesn't report asynchronous errors, should we mention {{MLContext/readTensor()}} instead, or link to issue #778?

index.bs
```diff
@@ -711,7 +711,7 @@ The {{MLGraphBuilder}}.{{MLGraphBuilder/build()}} method compiles the graph in t
...
+Once the {{MLGraph}} is constructed, the {{MLContext}}.{{MLContext/dispatch()}} method performs the execution of the graph asynchronously either on a parallel timeline in a separate worker thread for the CPU execution or on a GPU timeline in a GPU command queue. This method returns immediately without blocking the calling thread while the actual execution is offloaded to a different timeline. The caller supplies the input values using {{MLNamedTensors}}, binding the input {{MLOperand}}s to their values. The caller also supplies {{MLNamedTensors}} for output {{MLOperand}}s which will contain the result of graph execution, if successful, which may be read back to script using the {{MLContext}}.{{MLContext/readTensor(tensor)}} method. This type of execution supports both the CPU and GPU device.
```
huningxin, line 714: > This type of execution supports both the CPU and GPU device.

This is an old issue, but we probably should mention NPU as well? (feel free to resolve)

index.bs
huningxin: It seems to be duplicate. Should we also remove other examples of getting context and building graph in https://webmachinelearning.github.io/webnn/#examples? They also seem to be duplicated by dispatch example.

a-sully: Hmm. Yeah examples 6 and 8 (which I assume are the ones you're referring to) don't seem very useful. Example 7 ("The following code implements the graph....") seems pretty useful, though, especially since it gives a visual of how the operands relate to each other

Tentatively removed examples 6 and 8 in the latest patch b94b334

index.bs
a-sully: Updated the text and linked to #778. WDYT?

index.bs
a-sully: Added NPU to the list

index.bs
huningxin: > Example 7 ("The following code implements the graph....") seems pretty useful, though, especially since it gives a visual of how the operands relate to each other

Agreed.

huningxin: But keeping example 7 alone would make it incomplete, could we add `const context = await navigator.ml.createContext();` before `const builder = new MLGraphBuilder(context);`?

huningxin (approved):
LGTM with a comment to example, thanks!

index.bs
a-sully: Done in efdc1e8

huningxin (approved):
LGTM, thanks!

index.bs
fdwr: I was going to point this out too. Examples 6, 7, 8, 9 are all part of a series, and so deleting just parts of it make it incomplete, but I suppose so long as https://webmachinelearning.github.io/webnn/#api-mlcontext-dispatch-examples exists, then we at least have a whole example in the spec still.

fdwr (approved):
👍 LGTM Mr. Sully.

PR #796, opened 2024-12-03
Complain about RFC2119 terms, and fix usage

inexorabletash:
A local change I had that needed dusting off. It's probably worth reviewing/merging, but not urgently.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/796.html" title="Last updated on Dec 4, 2024, 12:49 AM UTC (8d8d43a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/796/4e80205...inexorabletash:8d8d43a.html" title="Last updated on Dec 4, 2024, 12:49 AM UTC (8d8d43a)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -21,6 +21,7 @@ Markup Shorthands: css no
 Logo: https://webmachinelearning.github.io/webmachinelearning-logo.png
 Deadline: 2023-10-01
 Assume Explicit For: yes
+Complain About: accidental-2119 yes
 Status Text: <p>
   Since the <a href="https://www.w3.org/TR/2023/CR-webnn-20230330/">initial Candidate Recommendation Snapshot</a> the Working Group has gathered further <a href="https://webmachinelearning.github.io/webnn-status/">implementation experience</a> and added new operations and data types needed for well-known <a href="https://github.com/webmachinelearning/webnn/issues/375">transformers to support generative AI use cases</a>. In addition, informed by this implementation experience, the group removed <code>MLCommandEncoder</code>, support for synchronous execution, and higher-level operations that can be expressed in terms of lower-level primitives in a performant manner. The group has also updated the specification to use modern authoring conventions to improve interoperability and precision of normative definitions.
   The group is developing a new feature, a <a href="https://github.com/webmachinelearning/webnn/issues/482">backend-agnostic storage type</a>, to improve performance and interoperability between the WebNN, WebGPU APIs and purpose-built hardware for ML and expects to republish this document as a Candidate Recommendation Snapshot when ready for implementation.
@@ -401,7 +402,7 @@ This section illustrates application-level use cases for neural network
 inference hardware acceleration. All applications in those use cases can be
 built on top of pre-trained deep neural network (DNN) [[models]].
 
-Note: Please be aware that some of the use cases described here, are by their very nature, privacy-invasive. Developers who are planning to use the API for such use cases should ensure that the API is being used to benefit users, for purposes that users understand, and approve. They should apply the Ethical Principles for Web Machine Learning [[webmachinelearning-ethics]] and implement appropriate privacy risk mitigations such as transparency, data minimisation, and users controls.
+Note: Please be aware that some of the use cases described here, are by their very nature, privacy-invasive. Developers who are planning to use the API for such use cases <span class=allow-2119>should</span> ensure that the API is being used to benefit users, for purposes that users understand, and approve. They <span class=allow-2119>should</span> apply the Ethical Principles for Web Machine Learning [[webmachinelearning-ethics]] and implement appropriate privacy risk mitigations such as transparency, data minimisation, and users controls.
 
 ### Person Detection ### {#usecase-person-detection}
 
@@ -630,6 +631,10 @@ Purpose-built Web APIs for measuring high-resolution time mitigate against timin
 
 ## Guidelines for new operations ## {#security-new-ops}
 
+*This section is non-normative.*
+
+<div class=informative>
+
 To ensure operations defined in this specification are shaped in a way they can be implemented securely, this section includes guidelines on how operations are expected to be defined to reduce potential for implementation problems. These guidelines are expected to evolve over time to align with industry best practices:
 
 - Prefer simplicity of arguments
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1446,7 +1453,7 @@ dictionary MLTensorDescriptor : MLOperandDescriptor {
 
 The {{MLTensor}} interface represents a tensor which may be used as an input or output to an {{MLGraph}}. The memory backing an {{MLTensor}} should be allocated in an [=implementation-defined=] fashion according to the requirements of the {{MLContext}} and the {{MLTensorDescriptor}} used to create it. Operations involving the {{MLTensor/[[data]]}} of an {{MLTensor}} occur on the {{MLContext/[[timeline]]}} of its associated {{MLContext}}. 
 
-Note: The [=implementation-defined=] requirements of how an {{MLTensor}} is allocated may include constraints such as that the memory is allocated with a particular byte alignment or in a particular memory pool.
+The [=implementation-defined=] requirements of how an {{MLTensor}} is allocated may include constraints such as that the memory is allocated with a particular byte alignment or in a particular memory pool.
```
inexorabletash, line 1456: I changed this from an informative note to a normative statement since it seems to exactly be in the spirit of normative RFC2119's "may".

inexorabletash:
@huningxin & @fdwr - could you please review? But this is low priority!

zolkis:
Looks good. Ugh. Spec editing is becoming super standardized. Almost programming. Might be good one day for automating implementations by AI. :)

index.bs
```diff
@@ -951,7 +958,7 @@ Schedules the computational workload of a compiled {{MLGraph}} on the {{MLContex
     **Returns:** {{undefined}}.
 </div>
 
-Note: `dispatch()` itself provides no signal that graph execution has completed. Rather, callers should await the results of reading back the output tensors. See [[#api-mlcontext-dispatch-examples]] below.
+Note: `dispatch()` itself provides no signal that graph execution has completed. Rather, callers can `await` the results of reading back the output tensors. See [[#api-mlcontext-dispatch-examples]] below.
```
fdwr, line 961: 🤔 Shouldn't this `should` still be a `should`? `Can` sounds overly permissive, as you really wouldn't want to read the output tensor data before it is actually available to be read (that is, read the `ArrayBuffer` before `mlContext.readTensor` finished).

index.bs
```diff
@@ -2676,7 +2683,7 @@ partial dictionary MLOpSupportLimits {
             interpreted according to the value of *options*.{{MLConvTranspose2dOptions/filterLayout}} and {{MLConvTranspose2dOptions/groups}}.
         - <dfn>options</dfn>: an optional {{MLConvTranspose2dOptions}}.
 
-    **Returns:** an {{MLOperand}}. The output 4-D tensor that contains the transposed convolution result. The output shape is interpreted according to the *options*.{{MLConvTranspose2dOptions/inputLayout}} value. More specifically, unless the *options*.{{MLConvTranspose2dOptions/outputSizes}} values are explicitly specified, the *options*.{{MLConvTranspose2dOptions/outputPadding}} may be needed to compute the spatial dimension values of the output tensor as follows:
+    **Returns:** an {{MLOperand}}. The output 4-D tensor that contains the transposed convolution result. The output shape is interpreted according to the *options*.{{MLConvTranspose2dOptions/inputLayout}} value. More specifically, unless the *options*.{{MLConvTranspose2dOptions/outputSizes}} values are explicitly specified, the *options*.{{MLConvTranspose2dOptions/outputPadding}} is be needed to compute the spatial dimension values of the output tensor as follows:
```
fdwr, line 2686: `is be` -> `is`

index.bs
```diff
@@ -4462,7 +4469,7 @@ partial dictionary MLOpSupportLimits {
     1. If |hiddenSize| * 6 is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
         <details class=note>
           <summary>Why |hiddenSize| * 6 ?</summary>
-          Some underlying platforms operate on a single bias tensor which is a concatenation of {{MLGruCellOptions/bias}} and {{MLGruCellOptions/recurrentBias}}. Therefore, 3 * |hiddenSize| + 3 * |hiddenSize| must also be a [=valid dimension=].
+          Some underlying platforms operate on a single bias tensor which is a concatenation of {{MLGruCellOptions/bias}} and {{MLGruCellOptions/recurrentBias}}. Therefore, 3 * |hiddenSize| + 3 * |hiddenSize| also needs to be a [=valid dimension=].
```
fdwr, line 4472: Was `must` inappropriate here? It is a spec requirement that `3 * |hiddenSize| + 3 * |hiddenSize|` is a valid dimension.

fdwr (approved):
2 q's and 1 typo, else LGTM.

index.bs
inexorabletash: For RFC2119 conformance, the magic word "MUST" in specs is an implementation requirement. That is, this is something that the browser implementation of WebNN MUST do. This informative note is not stating an implementation requirement, so it's not appropriate. So thinking through who the note is for...

(1) a potential script author; but then it would need to be phrased as "you MUST pass X where X * 6 is a valid dimensions"; but then it is *not* an implementation requirement, so can't use the magic word MUST, or
(2) an implementer trying to write a conforming implementation; but then it would need to be phrased as "if X * 6 is not a valid dimension, then the implementation MUST throw an exception" - which *is* an implementation requirement... but already stated in the algorithm, because algorithms in modern web specs are implicitly all MUST requirements.

Ultimately, this note isn't really _either_ those those; it's explaining the "why" of the text to a potential implementer (or proxy, like a reviewer), and IMHO that's not a normative requirement. So... the best thing to do is just avoid the word.

index.bs
inexorabletash: Fixed in 8d8d43a8c52afec374f7915ed337cd285970186d

index.bs
fdwr: > often leads to some linguistic contortions

Indeed... 😉

index.bs
fdwr: 🤔 Increasingly I'm wondering whether this "RFC2119" is more detrimental to reader clarity than helpful o_o... 🤷

fdwr (approved):
👍

huningxin (approved):
LGTM!

PR #806, opened 2025-01-21
Editorial: Fix glitch with otherwise clause in pool op creation

inexorabletash:
A pair of steps in "create pooling operation" had an "If" condition with multiple clauses, and a subsequent "Otherwise" step that needs to apply if only the first clause is true. Rework the lines to correct the logic, and also align with subsequent argument validation.

Also, put "Otherwise, ..." steps on their own line consistently.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/806.html" title="Last updated on Jan 26, 2025, 8:14 AM UTC (4d5898d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/806/17a44e2...inexorabletash:4d5898d.html" title="Last updated on Jan 26, 2025, 8:14 AM UTC (4d5898d)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -6552,7 +6552,7 @@ partial dictionary MLOpSupportLimits {
     1. If |allowedDataTypes| is given and it does not [=list/contain=] |input|'s [=MLOperand/dataType=], then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/rank=] is not 4, then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLPool2dOptions/windowDimensions}} does not [=map/exist=], set |options|.{{MLPool2dOptions/windowDimensions}} to the height and width dimensions of the shape of |input|.
-    1. If |options|.{{MLPool2dOptions/windowDimensions}}'s not 2, then [=exception/throw=] a {{TypeError}}.
+    1. If |options|.{{MLPool2dOptions/windowDimensions}}'s [=list/size=] is not 2, then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLPool2dOptions/outputSizes}} [=map/exists=], or if |options|.{{MLPool2dOptions/padding}} does not [=map/exist=], set |options|.{{MLPool2dOptions/padding}} to the [=/list=] « 0, 0, 0, 0 ».
     1. If |options|.{{MLPool2dOptions/padding}}'s [=list/size=] is not 4, then [=exception/throw=] a {{TypeError}}.
     1. If |options|.{{MLPool2dOptions/strides}} does not [=map/exist=], set |options|.{{MLPool2dOptions/strides}} to the [=/list=] « 1, 1 ».
--- a/index.bs
+++ b/index.bs
@@ -833,8 +833,10 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
     1. Otherwise:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/default=]".
         1. Set |context|.{{MLContext/[[lost]]}} to [=a new promise=].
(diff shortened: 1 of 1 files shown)


inexorabletash:
@huningxin @fdwr - super minor fix, can you please take a look?

But while I'm here... the step _"set options.windowDimensions to the height and width dimensions of the shape of input."_ could be made more rigorous by actually giving the dimensions. This would be most easily done by pulling "[calculate pool2d output sizes](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-calculate-pool2d-output-sizes)" steps inline so that _inputHeight_ and _inputWidth_ are calculated earlier given _options.layout_. I didn't want to do that without asking for opinions first. Thoughts?

fdwr (approved):
👍

fdwr:
> But while I'm here... the step _"set options.windowDimensions to the height and width dimensions of the shape of input."_ could be made more rigorous by actually giving the dimensions. This would be most easily done by pulling "[calculate pool2d output sizes](https://webmachinelearning.github.io/webnn/#mlgraphbuilder-calculate-pool2d-output-sizes)" steps inline so that _inputHeight_ and _inputWidth_ are calculated earlier given _options.layout_. I didn't want to do that without asking for opinions first. Thoughts?

I'm unsure why "calculate pool2d output sizes" is separate from the main algorithm, since "calculate pool2d output sizes" doesn't appear to be commonly shared anywhere else, but maybe it's that way because these others are also in independent sections...

- calculate reduction output sizes
- calculate conv2d output sizes
- calculate convtranspose2d output sizes
- calculate matmul output sizes
- calculate padding output sizes

...and if we're keeping the output calculation separate in the other parts, then consistency is nice. Maybe we can *refer* to it rather than move inline?

inexorabletash:
> 
> I'm unsure why "calculate pool2d output sizes" is separate from the main algorithm, since it doesn't appear to be commonly shared anywhere else, but maybe it's that way because these others are also in independent sections...
> 
(snip)
> 
> ...and if we're keeping the output calculation separate in the other operators, then consistency is nice. 

I did a quick poll of my fellow Googlers a while back, and they were in favor of inlining algorithms that are used in only one place. I think with the _Calculate the output shape:_ (etc) subsections in the algorithms we maintain readability

> Maybe we can _refer_ to it rather than move inline? 

Yeah, I think the intent is pretty obvious right now.  We could also pull just the "Switch on layout..." step out and pass _batches_, _channels_, _inputHeight_, and _inputWidth_ in, rather than _inputShape_. But at that point there's not much left and the description of the algorithm inputs is almost as long as the algorithm itself.

huningxin:
> But while I'm here... the step _"set options.windowDimensions to the height and width dimensions of the shape of input."_ could be made more rigorous by actually giving the dimensions.

+1

> > Maybe we can _refer_ to it rather than move inline?
> 
> Yeah, I think the intent is pretty obvious right now. We could also pull just the "Switch on layout..." step out and pass _batches_, _channels_, _inputHeight_, and _inputWidth_ in, rather than _inputShape_. But at that point there's not much left and the description of the algorithm inputs is almost as long as the algorithm itself.

I prefer to inline the "calculate pool2d output sizes" steps, because this algorithm internally calls "calculating conv2d output sizes" steps. By inline it, "pool2d" steps can be consistent with "conv2d" steps.

huningxin (approved):
LGTM!

huningxin:
@inexorabletash , is this PR ready to merge? "calculate pool2d output sizes" change could be done by a separate CL, WDYT?

inexorabletash:
> @inexorabletash , is this PR ready to merge? "calculate pool2d output sizes" change could be done by a separate CL, WDYT?

Agreed, I'd prefer to merge this and do a follow-on. Thanks for fixing my glitch, too!

Issue #350, opened 2023-02-17
Need to understand how WebNN supports implementation that involves multiple devices and timelines


wchao1115:
During the conversation on WebNN and WebGPU interop session, at the WebGPU F2F meeting in San Francisco, Myles Maxfield made an observation about the fact that WebNN `MLContext` is a single-device context. He raised the awareness that Apple's CoreML can distribute a workload within a single ML graph across multiple devices including the ANE, and said that it would be nice  if WebNN has a provision to support that situation as well. 

I think this feedback is related to a discussion in the working group of late about ways to create a notion of a *default* MLContext (when there is no explicit device associated with the context -- either a CPU device or a WebGPU device) that is more open to the implementation. This default context could effectively be whatever the implementer wants it to be including a multi-device context in the case of the Apple's WebKit implementation over CoreML.

plehegar:
See also [Device Selection Explainer](https://github.com/webmachinelearning/webnn/blob/main/device-selection-explainer.md) and https://github.com/webmachinelearning/webnn/pull/809

Issue #302, opened 2022-12-01
API simplification: context types, context options, createContext()


zolkis:
Lifted from #298 for brevity.

### Proposal
Provide a single context type as a mapped descriptor for the combination of resources used in the context, e.g. a valid combination of device(s). (Somewhat analogous to the [adapter](https://www.w3.org/TR/webgpu/#adapters) plus device(s) concept in Web GPU.)
```javascript
enum MLContextType {  
  "cpu",   // script-controlled context
  "gpu"   // script-controlled context
  "webgpu",  // managed by the user agent
  // later other context types may be defined, even using multiple devices, e.g. "cpu+npu" etc.
  // Note: in fact all these context types could be separate interface classes as well...
};

enum MLPowerPreference {  // a hint
  "default",
  "high-performance",
  "low-power"
};

dictionary MLContextOptions {  // not a hint
  MLContextType contextType = "cpu";
  MLPowerPreference powerPreference = "default";
  GPUDevice? gpuDevice = null;
};

[SecureContext, Exposed=(Window, DedicatedWorker)]
interface ML {
  Promise<MLContext> createContext(optional MLContextOptions options);

  [Exposed=(DedicatedWorker)]
  MLContext createContextSync(optional MLContextOptions options);

  // Internal slots
  // [[boolean managed]]  // `true` if the user agent controls the context (not really needed)
  // [[MLContextType contextType]] 
  // [[MLPowerPreference powerPreference]]
  // [[implementation]] // perhaps "adapter" would be better

  // further methods (and eventually properties) will follow
};

```
### Rationale for change
- Including WebGPU into context options would allow a single definition of `createContext()`, 
- handle the current context types ("default" and "webgpu") and also the current device types ("cpu" and "gpu") and possibly other (and combined) device types in the future. That way, it would simplify and minimize algorithms.
- This would simplify the issues mentioned in 
 https://github.com/webmachinelearning/webnn/pull/257, 
 https://github.com/webmachinelearning/webnn/pull/255, 
 https://github.com/webmachinelearning/webnn/pull/162.

Related to #303.

anssiko:
This is considered a v2 feature per https://www.w3.org/2022/12/15-webmachinelearning-minutes.html

zolkis:
Following the proposals from #322 , adapted the proposal. Web IDL:
```js
enum MLContextType {  
  "cpu",   // script-controlled context
  "webgpu",  // managed by the user agent
[...]

PR #809, opened 2025-01-27
Remove MLDeviceType

### LINKED ISSUE #749: MLContextOptions.deviceType seems unnecessary outside of conformance testing

zolkis:
Resolves #749. 
Addresses #302, #350.
See also the [device selection explainer](https://github.com/webmachinelearning/webnn/blob/main/device-selection-explainer.md).

This is a minimal change for now. It may require changing prose or adding more details / instructions in related sections/algorithms.

***
<a href="https://pr-preview.s3.amazonaws.com/zolkis/webnn/pull/809.html" title="Last updated on Feb 6, 2025, 2:58 PM UTC (f4b8402)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/809/4acd8e4...zolkis:f4b8402.html" title="Last updated on Feb 6, 2025, 2:58 PM UTC (f4b8402)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -669,11 +669,11 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring; 
 
 The WebGPU API identifies <a href="https://gpuweb.github.io/gpuweb/#privacy-machine-artifacts">machine-specific artifacts</a> as a privacy consideration. Similarly, the WebNN API's compute unit scheduling may under certain circumstances introduce a fingerprint. However, similarly to WebGPU, such fingerprints are identical across most or all of the devices of each vendor, mitigating the concern. Furthermore, software implementations can be used to further eliminate such artifacts.
 
-The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. An {{MLDeviceType}} normatively indicates the kind of device and is one of: {{MLDeviceType/"cpu"}}, {{MLDeviceType/"gpu"}}, {{MLDeviceType/"npu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. An {{MLPowerPreference}} indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
+The WebNN API defines developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the underlying execution device for the workload. An {{MLPowerPreference}} indicates preference as related to the desired low power consumption or high performance, is considered a hint only and as such does not increase entropy of the fingerprint.
 
 Issue(623): {{MLContextOptions}} is under active development, and the design is expected to change, informed by further implementation experience and new use cases from the wider web community.
 
-If a future version of this specification introduces support for a new {{MLDeviceType}} that can only support a subset of {{MLOperandDataType}}s, that could introduce a new fingerprint.
+If a future version of this specification introduces support for a new {{MLContextOptions}} member for supporting only a subset of {{MLOperandDataType}}s, that could introduce a new fingerprint.
 
 In general, implementers of this API are expected to apply <a href="https://gpuweb.github.io/gpuweb/#privacy-considerations">WebGPU Privacy Considerations</a> to their implementations where applicable.
 
@@ -729,7 +729,13 @@ An {{MLContext}} interface represents a global state of neural network execution
 
 In a situation when a GPU context executes a graph with a constant or an input in the system memory as an {{ArrayBufferView}}, the input content is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of the graph execution. This data upload and download cycles will only occur whenever the execution device requires the data to be copied out of and back into the system memory, such as in the case of the GPU. It doesn't occur when the device is a CPU device. Additionally, the result of the graph execution is in a known layout format. While the execution may be optimized for a native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert the content back to a known layout format at the end of the graph in order to maintain the expected behavior from the caller's perspective.
 
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's {{MLPowerPreference}} and {{MLDeviceType}} options.
+When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently only the {{MLPowerPreference}} option.
+
+<div class="note">
+Depending on the underlying platform, the user agent <span class=allow-2119>may</span> select different combinations of CPU, NPU and GPU devices.
+</div>
+
+For a history and rationale of this design, please see the <a href="https://github.com/webmachinelearning/webnn/blob/master/device-selection-explainer.md">device selection explainer</a>.
 
 ## Task Source ## {#programming-model-task-source}
 
@@ -764,12 +770,6 @@ WorkerNavigator includes NavigatorML;
 
 ## {{ML}} interface ## {#api-ml}
 <script type=idl>
-enum MLDeviceType {
-  "cpu",
-  "gpu",
-  "npu"
-};
-
 enum MLPowerPreference {
   "default",
   "high-performance",
@@ -777,7 +777,6 @@ enum MLPowerPreference {
 };
 
 dictionary MLContextOptions {
-  MLDeviceType deviceType = "cpu";
   MLPowerPreference powerPreference = "default";
 };
 
@@ -791,16 +790,6 @@ interface ML {
 ### {{MLContextOptions}} ### {#api-mlcontextoptions}
 
 Issue(623): {{MLContextOptions}} is under active development, and the design is expected to change, informed by further implementation experience and new use cases from the wider web community. The Working Group is considering additional API controls to allow the definition of a fallback device, multiple devices in a preferred order, or an exclusion of a specific device. Other considerations under discussion include error handling, ultimate fallback, and quantized operators. Feedback is welcome on any of these design considerations from web developers, library authors, OS and hardware vendors, and other stakeholders via GitHub:
-
-The <dfn dfn-for=MLContextOptions dfn-type=dict-member>deviceType</dfn> option is an <dfn dfn-type=enum>MLDeviceType</dfn> and indicates the application's preference for the kind of device used for the context. It is one of the following:
-<dl dfn-for="MLDeviceType">
-<dt>"<dfn enum-value>cpu</dfn>"</dt>
-<dd>Provides the broadest compatibility and usability across all client devices with varying degrees of performance.</dd>
-<dt>"<dfn enum-value>gpu</dfn>"</dt>
-<dd>Provides the broadest range of achievable performance across graphics hardware platforms from consumer devices to professional workstations. The underlying platform implementation may fall back to other devices for certain operators and parts of the graph.</dd>
-<dt>"<dfn enum-value>npu</dfn>"</dt>
-<dd>Provides power efficiency for sustained workloads across hardware platforms with purpose-built accelerators. The underlying platform implementation may fall back to other devices for certain operators and parts of the graph.</dd>
-</dl>
 
 The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> option is an <dfn dfn-type=enum>MLPowerPreference</dfn> and indicates the application's preference as related to power consumption. It is one of the following:
 <dl dfn-for="MLPowerPreference">
@@ -828,16 +817,13 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
     1. Let |context| be a new {{MLContext}} in |realm|.
     1. If |options| is a {{GPUDevice}} object:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/webgpu=]".
-        1. Set |context|.{{MLContext/[[deviceType]]}} to {{MLDeviceType/"gpu"}}.
         1. Set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
     1. Otherwise:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/default=]".
         1. Set |context|.{{MLContext/[[lost]]}} to [=a new promise=] in |realm|.
-        1. If |options|["{{MLContextOptions/deviceType}}"] [=map/exists=], then set |context|.{{MLContext/[[deviceType]]}} to |options|["{{MLContextOptions/deviceType}}"].
-        1. Otherwise, set |context|.{{MLContext/[[deviceType]]}} to {{MLDeviceType/"cpu"}}.
         1. If |options|["{{MLContextOptions/powerPreference}}"] [=map/exists=], then set |context|.{{MLContext/[[powerPreference]]}} to |options|["{{MLContextOptions/powerPreference}}"].
         1. Otherwise, set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
-    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}}, |context|.{{MLContext/[[deviceType]]}} and |context|.{{MLContext/[[powerPreference]]}}, return failure.
+    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}} and |context|.{{MLContext/[[powerPreference]]}}, return failure.
     1. Return |context|.
 </details>
 
@@ -870,7 +856,7 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
 </details>
 
 ## {{MLContext}} interface ## {#api-mlcontext}
-The {{MLContext}} interface represents a global state of neural network compute workload and execution processes. Each {{MLContext}} object has associated [=context type=], {{MLDeviceType}} and {{MLPowerPreference}}.
+The {{MLContext}} interface represents a global state of neural network compute workload and execution processes. Each {{MLContext}} object has associated [=context type=] and {{MLPowerPreference}}.
 
 <script type=idl>
 typedef record<USVString, MLTensor> MLNamedTensors;
@@ -904,9 +890,6 @@ interface MLContext {
     : <dfn>\[[contextType]]</dfn> of type [=context type=].
     ::
         The {{MLContext}}'s [=context type=].
-    : <dfn>\[[deviceType]]</dfn> of type {{MLDeviceType}}.
-    ::
-        The {{MLContext}}'s {{MLDeviceType}}.
     : <dfn>\[[powerPreference]]</dfn> of type {{MLPowerPreference}}.
     ::
         The {{MLContext}}'s {{MLPowerPreference}}.
@@ -928,10 +911,6 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
 <dt>"<dfn>webgpu</dfn>"</dt>
 <dd>Context created from WebGPU device.</dd>
 </dl>
-
-<div class="note">
-When the {{MLContext/[[contextType]]}} is set to [=context type/default=] with the {{MLContextOptions}}.{{MLContextOptions/deviceType}} set to {{MLDeviceType/"gpu"}}, the user agent is responsible for creating an internal GPU device that operates within the context and is capable of ML workload submission on behalf of the calling application.
-</div>
 
 <details open algorithm>
   <summary>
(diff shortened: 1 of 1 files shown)


inexorabletash:
@zolkis can you use the [keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) "Resolves #749" in the PR description? This creates a visible link between the issue and PR.

Also... would you consider #302 addressed by this PR?

index.bs
```diff
@@ -729,7 +729,9 @@ An {{MLContext}} interface represents a global state of neural network execution
...
 
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's {{MLPowerPreference}} and {{MLDeviceType}} options.
+When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently the {{MLPowerPreference}} option. Depending on the underlying platform, the user agents may select different devices or combination of CPU, NPU and GPU devices.
```
inexorabletash, line 732:
```suggestion
When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently only the {{MLPowerPreference}} option. Depending on the underlying platform, the user agent may select different combinations of CPU, NPU and GPU devices.
```

index.bs
```diff
@@ -729,7 +729,9 @@ An {{MLContext}} interface represents a global state of neural network execution
...
+
+For a history and rationale of this design, please see the <a href="https://github.com/webmachinelearning/webnn/blob/master/device-selection-explainer.md">device selection explainer</a>.
```
inexorabletash, line 734: Consider using markdown rather than HTML syntax; I think in this case it's going to be the same, but avoiding raw HTML lets Bikeshed make smarter decisions about rendering.

inexorabletash (commented):
Overall LGTM, lots of nitpicks.

index.bs
```diff
@@ -828,16 +813,13 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
...
         1. Set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
     1. Otherwise:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/default=]".
         1. Set |context|.{{MLContext/[[lost]]}} to [=a new promise=].
-        1. If |options|["{{MLContextOptions/deviceType}}"] [=map/exists=], then set |context|.{{MLContext/[[deviceType]]}} to |options|["{{MLContextOptions/deviceType}}"].
-        1. Otherwise, set |context|.{{MLContext/[[deviceType]]}} to {{MLDeviceType/"cpu"}}.
         1. If |options|["{{MLContextOptions/powerPreference}}"] [=map/exists=], then set |context|.{{MLContext/[[powerPreference]]}} to |options|["{{MLContextOptions/powerPreference}}"].
         1. Otherwise, set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
-    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}}, |context|.{{MLContext/[[deviceType]]}} and |context|.{{MLContext/[[powerPreference]]}}, return failure.
+    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}} and |context|.{{MLContext/[[powerPreference]]}}, return failure.
```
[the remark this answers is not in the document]
zolkis: Makes sense. I just wanted a minimal change as the first approach.

IMHO we should also remove the "default" power pref, like in [WebGPU](https://www.w3.org/TR/webgpu/#enumdef-gpupowerpreference). In the algorithm above, we would not even need to set it. But I thought that would be a separate PR... together with eventually reworking the **_[[context type]]_** ("default" context to generic (or auto) and "webgpu", clarifying whether and how it can change.

index.bs
zolkis: I used HTML because the spec uses that in 2 other places.

index.bs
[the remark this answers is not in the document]
inexorabletash: Separate PR is fine with me!

index.bs
inexorabletash: That's fine. We can decide separately and update conventions doc/lint tool.

anssiko:
@zolkis I believe this PR also addresses #350 by making `MLContext` device agnostic.

index.bs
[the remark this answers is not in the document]
inexorabletash: (feel free to resolve this conversation; I don't have that button)

index.bs
inexorabletash: (feel free to resolve this conversation; I don't have that button)

index.bs
```diff
@@ -764,20 +766,13 @@ WorkerNavigator includes NavigatorML;
 
 ## {{ML}} interface ## {#api-ml}
 <script type=idl>
-enum MLDeviceType {
```
fdwr, line 767: > An {{MLDeviceType}} normatively indicates the kind of device and is one of: {{MLDeviceType/"cpu"}}, {{MLDeviceType/"gpu"}}, {{MLDeviceType/"npu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown

This afternoon we were talking with {@inexorabletash, @reillyeon, @RafaelCintron, Phillis} about the likely need for a caller to know whether a particular device is supported or not, because an app may want to (if say GPU is not supported) use a different more performant fallback than for WebNN to silently fall back to CPU. For example, if GPU was unavailable (even though you preferred high performance), then it might be faster to execute the model with WebGPU shaders than WebNN CPU, or it might be *okay* to use CPU, but the app could load a different model that's more CPU-friendly, if it knew that was the case.

This need is the reverse of what we currently have, querying the context for what kind of devices it actually supports rather than prescribing a preferred device type. Now, preferred device type prescription may return someday as a weaker hint (or list of hints) as we learn more about usage/needs, but for either case prescription or post-query, having device enums could be useful. 🤔

Thoughts? Do we delete `MLDeviceType` for now and then resurrect it after deciding how to query which kinds of devices a context can support? Do keep it but remove the context creation parameter?

zolkis: The question is whether the `"high-performance"' `MLPowerPreference` provides enough hint to implementations to decide which fallback to choose. 

In what use cases we need to hand the decision over to clients of the API? If we have such cases, we will still need a kind of explicit device selection and error handling, with clearly specified fallback scenarios. 

About querying for capabilities of the underlying platform, there was this [proposal](https://github.com/webmachinelearning/webnn/issues/749#issuecomment-2429821928), also figuring in the device selection explainer. We might also need to consider other mappings than combinations of `cpu`, `gpu`, `npu`, as we had issues handling these.

I suggest we open a separate issue for this and address it later. Meanwhile, removing the current device types and making contexts device agnostic will give us means to test various ideas in the implementations and decide with a fresh approach what we need to bubble up and expose at the API level.

huningxin: @fdwr 
> For example, if GPU was unavailable (even though you preferred high performance), then it might be faster to execute the model with WebGPU shaders than WebNN CPU

I am wondering whether this fallback to WebGPU shaders could be handled by implementation rather than app? I mean WebGPU shaders could be an implementation option of executing WebNN graphs on GPU device, for example WebNN may be able to use [ONNX Runtime Native WebGPU execution provider](https://github.com/microsoft/onnxruntime/issues/22077).

> or it might be okay to use CPU, but the app could load a different model that's more CPU-friendly, if it knew that was the case.

This sounds like an interesting new use case.

@zolkis 
> I suggest we open a separate issue for this and address it later.

+1

zolkis: >Thoughts? Do we delete `MLDeviceType` for now and then resurrect it after deciding how to query which kinds of devices a context can support? Do keep it but remove the context creation parameter?

I've been thinking what about changing `MLDeviceType` to a `MLDevicePreference`, only as an option when creating a context, not as a straight property of the context. But when considering all use case ramifications and device combinations to handle, it might still lead to full explicit device selection, with standardized fallback scenarios and/or error handling - back to the problems we wanted to solve, coming from the heterogeneity of those devices and their combinations across platforms.

Querying capabilities might be a better solution, but it's complex to do it in a future-proof way (e.g., which set of capabilities?). I think this is similar to regular resource negotiation, for which the existing protocols use a number of patterns - most of which would have privacy issues in Web APIs, and are also quite complex. 

The best we could do is likely app hints on context creation and eventually build. Power/performance options are good because they connect back to the needs that created the different AI accelerators to start with. But we could elaborate more on what e.g. `"high-performance"` or `"low-power"` means in various scenarios, and when/how we report errors, so that the app could change to a model that matches better the underlying system.

We could also extend the context options with new ones, e.g. expressing minimal requirements from the apps side, like compute (TOPS), memory/model size, quantization etc.

In the end, I see it as a positive challenge to remove device types for now, and work with device-agnostic contexts for a while, -- might gather more experience on what implementations could adapt for, and what would be the best way to go forward. In the meantime, we should work on querying capabilities ([option 3](https://github.com/webmachinelearning/webnn/blob/main/device-selection-explainer.md#considered-alternatives) from the explainer).

zolkis: I will update the device selection explainer with this new use case / option, and also lift the arguments from #302 into the explainer.

anssiko:
@fdwr commented:
>Thoughts? Do we delete MLDeviceType for now and then resurrect it after deciding how to query which kinds of devices a context can support? Do keep it but remove the context creation parameter?

Since enums are not web-exposed, we could leave the `MLDeviceType` enum in place (for later reuse/repurpose) but remove it from the `MLContextOptions` dictionary. In addition could update [this note](https://www.w3.org/TR/webnn/#issue-a5abcae7) to add that a mechanism to query devices a context can support is being considered.

I created a new issue for discussing such a query mechanism: https://github.com/webmachinelearning/webnn/issues/815

>or it might be okay to use CPU, but the app could load a different model that's more CPU-friendly, if it knew that was the case.

@zolkis I think this should be added to the device selection explainer as a new use case.

mwyrzykowski (approved):

index.bs
```diff
@@ -729,7 +729,9 @@ An {{MLContext}} interface represents a global state of neural network execution
...
 
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's {{MLPowerPreference}} and {{MLDeviceType}} options.
+When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently only the {{MLPowerPreference}} option. Depending on the underlying platform, the user agent may select different combinations of CPU, NPU and GPU devices.
```
mwyrzykowski, line 732: I think the wording
> Depending on the underlying platform, the user agent may select different combinations of CPU, NPU and GPU devices.
is superfluous in this case as it is an implementation detail for the UA

index.bs
fdwr: I'm okay to remove it for now. Thanks for updating the explainer and listening.

index.bs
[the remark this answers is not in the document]
fdwr: > I don't have that button

@anssiko - 🤔 Do you know how enable the 'Resolve conversation" button for Joshua? I feel a commenter should be able to resolve their own comment at least.

fdwr (approved):

index.bs
inexorabletash: What do you think about keeping the text, but making it a non-normative note?

zolkis: Yes, it is implementation detail, and superfluous. A note might indeed work, but if you think it's not actually needed, I'd rather remove it.

index.bs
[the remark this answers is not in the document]
anssiko: I've looked at this, and many people on the internet have had the same request, [yet still](https://github.blog/changelog/2018-09-19-conversation-resolution-permissions/) "only the pull request author or users with write access to the repository have the ability to resolve the conversation". What follows is that that button (along with write access) comes with an editor hat.

index.bs
```diff
@@ -669,11 +669,11 @@ Unlike WebGPU, this API does not intrinsically support custom shader authoring;
...
 
-The WebNN API defines two developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. An {{MLDeviceType}} normatively indicates the kind of device and is one of: {{MLDeviceType/"cpu"}}, {{MLDeviceType/"gpu"}}, {{MLDeviceType/"npu"}}. If this type cannot be satisfied, an "{{OperationError}}" {{DOMException}} is thrown, thus this type can in some cases add two bits of entropy to the fingerprint. An {{MLPowerPreference}} indicates preference as related to the power consumption and is considered a hint only and as such does not increase entropy of the fingerprint.
+The WebNN API defines developer-settable preferences to help inform [[#programming-model-device-selection]] and allow the implementation to better select the most appropriate underlying execution device for the workload. An {{MLPowerPreference}} indicates a preference as related to the desired low power consumption or high performance; this is considered a hint only and as such does not increase entropy of the fingerprint.
```
huningxin, line 672: The current text "select the most appropriate underlying execution device for the workload" sounds like a context can only support one device. But we also want to ["allow supporting multiple devices with one context."](https://github.com/webmachinelearning/webnn/blob/main/device-selection-explainer.md#proposed-minimum-viable-solution)?

How about just say "better select the underlying execution device for the workload"? That  aligns with the text "the user agent selects and creates the underlying execution device by taking into account these options" at line 732.

index.bs
huningxin: > What do you think about keeping the text, but making it a non-normative note?

+1

index.bs
huningxin, line 822: Not new, but I am wondering whether the power preference is considered only a hint if the step returns a failure because the power preference is not supported by the user agent.

index.bs
```diff
@@ -828,16 +813,13 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
     1. Let |context| be a new {{MLContext}} object with |realm|.
     1. If |options| is a {{GPUDevice}} object:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/webgpu=]".
-        1. Set |context|.{{MLContext/[[deviceType]]}} to {{MLDeviceType/"gpu"}}.
         1. Set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
     1. Otherwise:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/default=]".
         1. Set |context|.{{MLContext/[[lost]]}} to [=a new promise=].
```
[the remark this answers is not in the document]
zolkis: Looks fixed after rebasing.

index.bs

index.bs

zolkis:
Apologies for the forced pushes. Squashed these simple changes into one commit, but was interleaved with others from the main, so it needed rebase, then reorder/merge, then squash/merge. Now it should be clean to merge.

index.bs
```diff
@@ -729,7 +729,13 @@ An {{MLContext}} interface represents a global state of neural network execution
...
 
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account the application's {{MLPowerPreference}} and {{MLDeviceType}} options.
+When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently only the {{MLPowerPreference}} option.
+
+<div class="note">
```
anssiko, line 734: To address https://github.com/webmachinelearning/webnn/pull/809#discussion_r1941790380 we could fold this text into the non-normative note:

```suggestion
<div class="note">
When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently only the {{MLPowerPreference}} option.

```

index.bs
```diff
@@ -828,16 +817,13 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
...
         1. Set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
     1. Otherwise:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/default=]".
         1. Set |context|.{{MLContext/[[lost]]}} to [=a new promise=] in |realm|.
-        1. If |options|["{{MLContextOptions/deviceType}}"] [=map/exists=], then set |context|.{{MLContext/[[deviceType]]}} to |options|["{{MLContextOptions/deviceType}}"].
-        1. Otherwise, set |context|.{{MLContext/[[deviceType]]}} to {{MLDeviceType/"cpu"}}.
         1. If |options|["{{MLContextOptions/powerPreference}}"] [=map/exists=], then set |context|.{{MLContext/[[powerPreference]]}} to |options|["{{MLContextOptions/powerPreference}}"].
         1. Otherwise, set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
-    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}}, |context|.{{MLContext/[[deviceType]]}} and |context|.{{MLContext/[[powerPreference]]}}, return failure.
+    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}} and |context|.{{MLContext/[[powerPreference]]}}, return failure.
```
anssiko, line 826: To stay consistent with this statement in the privacy considerations:
>An MLPowerPreference indicates preference as related to the desired low power consumption or high performance, is considered a hint only and as such does not increase entropy of the fingerprint.

Perhaps we could simply:

```suggestion
    1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}}, return failure.
```

If this is more complicated, then a separate issue makes sense.

anssiko (approved):

index.bs
zolkis: Well, it still provides an optional bailout for implementations if anything is wrong. It's true it should not happen solely based on `powerPreference`, since it's supposed to be only a _hint_. So I guess this suggestion will work.

index.bs
anssiko: Proposed fix https://github.com/webmachinelearning/webnn/pull/809/commits/fee790b2750a0ef1ffd3db3732828e9eb9942a9b

fdwr: That makes sense, as it's merely a preference.

huningxin (approved):
LGTM!

anssiko:
Per our discussion on [WebML WG Teleconference – 13 February 2025](https://www.w3.org/2025/02/13-webmachinelearning-minutes.html#faa0), this PR is ready to merge. Thank you everyone!

PR #827, opened 2025-03-11
Lint: Add additional checks and improve HTML parsing

inexorabletash:
New checks:

- Validate that constraint table links from steps are reasonable.
- Validate that constraint tables use links not styles for params.

Improved checks:

- Make duplicate word detection case insensitive and match at start/end of lines.

Internal changes:

- Massage HTML to allow node-html-parser to handle tables.
- Improve output when matching at line boundaries.
- Comment checks with "Generic" or "WebNN".

### DIFF
--- a/tools/lint.mjs
+++ b/tools/lint.mjs
@@ -72,7 +72,7 @@ log('massaging HTML...');
   const re = new RegExp(
     '(<(' + tags.join('|') + ')\\b[^>]*>)' +
       '(.*?)' +
-      '(?=<(:?' + tags.join('|') + '|/(' + containers.join('|') + '))\\b)',
+      '(?=<(' + tags.join('|') + '|/(' + containers.join('|') + '))\\b)',
     'sg');
   file = file.replaceAll(
     re, (_, opener, tag, content) => `${opener}${content}</${tag}>`);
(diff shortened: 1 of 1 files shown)


inexorabletash:
A couple of additional lint checks inspired by gotchas we worked through in the "wave3" reviews.

The annoying part of this change was dealing with node-html-parser limitations. Fun fact: without the workarounds, it just drops the contents of tables entirely from the DOM it produces. _"Why aren't my lint checks working!?!?! Oh...."_ :roll_eyes:

inexorabletash:
@fdwr & @huningxin - please take a look?

inexorabletash:
> So I got lost in parts (e.g. `'(?=<(:?' + tags.join('|')...`) 🤷‍♂️, but then I was already lost in that same part before. 😅

Yeah, regular expressions are pretty much write-only. :) I like [Perl's extended regex syntax](https://perldoc.perl.org/perlre#/x-and-/xx) which lets you write regexes spanning multiple lines and including comments. Ah well.

inexorabletash:
Although now that I'm looking at it, `(:?` should be `(?:` so hmmm...

inexorabletash:
> Although now that I'm looking at it, `(:?` should be `(?:` so hmmm...

Turned out to be harmless, fixed in 08fa6fe

fdwr (approved):

huningxin (approved):
LGTM!

PR #828, opened 2025-03-12
Add rankRange to opSupportLimits()

### LINKED ISSUE #456: Define the maximum number of operand dimensions (maximum rank)

inexorabletash:
Add rankRange to opSupportLimits() by renaming MLSupportLimits to MLDataTypeLimits (still used for inputs, constants, and output operands), and deriving MLTensorLimits (used for tensor operands) which adds the new member as a MLRankRange dictionary.

This also adds a note about an operand's allowed data types and ranks, indicating the implementations may extend or impose limits. 

In addition, maxTensorByteLength is added to the MLOpSupportLimits dictionary. 

Based on the work in Chromium by Reilly Grant, Wei Wang, and Junwei Fu.

Resolves #456

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/828.html" title="Last updated on Apr 3, 2025, 11:37 PM UTC (e023048)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/828/5daf9d7...inexorabletash:e023048.html" title="Last updated on Apr 3, 2025, 11:37 PM UTC (e023048)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1169,15 +1169,18 @@ The {{MLOpSupportLimits}} has the following top level members, aside from these,
 <script type="idl">
 dictionary MLOpSupportLimits {
   MLInputOperandLayout preferredInputLayout;
-  MLSupportLimits input;
-  MLSupportLimits constant;
-  MLSupportLimits output;
+  [EnforceRange] unsigned long long maxTensorByteLength;
+  MLDataTypeLimits input;
+  MLDataTypeLimits constant;
+  MLDataTypeLimits output;
 };
 </script>
 
 <dl dfn-type=dict-member dfn-for=MLOpSupportLimits>
     : <dfn>preferredInputLayout</dfn>
     :: Preferred input layout for layout dependent operators like {{MLGraphBuilder/conv2d()}}.
+    : <dfn>maxTensorByteLength</dfn>
+    :: The maximum supported length of tensors, in bytes.
     : <dfn>input</dfn>
     :: Support limits for input {{MLOperand}}s for an {{MLGraph}}.
     : <dfn>constant</dfn>
@@ -1185,48 +1188,73 @@ dictionary MLOpSupportLimits {
     : <dfn>output</dfn>
     :: Support limits for output {{MLOperand}}s for an {{MLGraph}}.
 </dl>
-#### {{MLSupportLimits}} dictionary #### {#api-mlcontext-supportlimits-dictionary}
+#### {{MLDataTypeLimits}} dictionary #### {#api-mlcontext-datatypelimits-dictionary}
 <script type="idl">
-dictionary MLSupportLimits {
+dictionary MLDataTypeLimits {
   sequence<MLOperandDataType> dataTypes;
 };
 </script>
-<dl dfn-type=dict-member dfn-for=MLSupportLimits>
+<dl dfn-type=dict-member dfn-for=MLDataTypeLimits>
     : <dfn>dataTypes</dfn>
     :: Supported data types.
+</dl>
+
+#### {{MLRankRange}} dictionary #### {#api-mlcontext-rankrange-dictionary}
+<script type="idl">
+dictionary MLRankRange {
+  unsigned long min;
+  unsigned long max;
+};
+</script>
+<dl dfn-type=dict-member dfn-for=MLRankRange>
+    : <dfn>min</dfn>
+    :: Minimum supported rank.
+    : <dfn>max</dfn>
+    :: Maximum supported rank.
+</dl>
+
+#### {{MLTensorLimits}} dictionary #### {#api-mlcontext-tensorlimits-dictionary}
+<script type="idl">
+dictionary MLTensorLimits : MLDataTypeLimits {
+  MLRankRange rankRange;
+};
+</script>
+<dl dfn-type=dict-member dfn-for=MLTensorLimits>
+    : <dfn>rankRange</dfn>
+    :: Minimum and maximum supported ranks.
 </dl>
 
 #### {{MLBinarySupportLimits}} dictionary #### {#api-mlcontext-binarysupportlimits-dictionary}
 <script type="idl">
 dictionary MLBinarySupportLimits {
-  MLSupportLimits a;
-  MLSupportLimits b;
-  MLSupportLimits output;
+  MLTensorLimits a;
+  MLTensorLimits b;
+  MLDataTypeLimits output;
 };
 </script>
 
 <dl dfn-type=dict-member dfn-for=MLBinarySupportLimits>
     : <dfn>a</dfn>
-    :: {{MLSupportLimits}} for a operand.
+    :: {{MLTensorLimits}} for a operand.
     : <dfn>b</dfn>
-    :: {{MLSupportLimits}} for b operand.
+    :: {{MLTensorLimits}} for b operand.
     : <dfn>output</dfn>
-    :: {{MLSupportLimits}} for output operand.
+    :: {{MLDataTypeLimits}} for output operand.
 </dl>
 
 #### {{MLSingleInputSupportLimits}} dictionary #### {#api-mlcontext-singleinputsupportlimits-dictionary}
 <script type="idl">
 dictionary MLSingleInputSupportLimits {
-  MLSupportLimits input;
-  MLSupportLimits output;
+  MLTensorLimits input;
+  MLDataTypeLimits output;
 };
 </script>
 
 <dl dfn-type=dict-member dfn-for=MLSingleInputSupportLimits>
     : <dfn>input</dfn>
-    :: {{MLSupportLimits}} for input operand.
+    :: {{MLTensorLimits}} for input operand.
     : <dfn>output</dfn>
-    :: {{MLSupportLimits}} for output operand.
+    :: {{MLDataTypeLimits}} for output operand.
 </dl>
 
 ### {{MLContext/destroy()}}  ### {#api-mlcontext-destroy}
@@ -1466,7 +1494,13 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
 
 If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}.
 
+Implementations may support additional data types for operands. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLDataTypeLimits/dataTypes}} value of the corresponding member for the operation.
+
+Issue: Can conforming implementations support *fewer* data types for an operation than what is listed for their [=allowed data types=]? In other words, is the list of allowed types only a recommendation?
+
 If an operation requires input operands with a particular [=MLOperand/rank=], the <dfn>allowed ranks</dfn> for each of the operation's input operands, including both positional arguments and options, are given as an explicit rank (e.g. 1), or <dfn lt="any rank">N</dfn> to allow any dimensionality, or the <dfn lt="same rank as">same as</dfn> another operand. More specific constraints are common, such as when an input operand's shape must be [=/unidirectionally broadcastable=] to or [=/bidirectionally broadcastable=] with another input operand; in these cases, the [=/allowed ranks=] are listed as a range, with specific validation given as steps in the operation.
+
+Implementations may (and usually do) impose an upper bound on the [=MLOperand/rank=] of operands. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/rankRange}}.{{MLRankRange/max}} value of the corresponding member for the operation.
 
 {{MLOperatorOptions}} has the following members:
 <dl dfn-type=dict-member dfn-for=MLOperatorOptions>
@@ -1950,12 +1984,12 @@ partial interface MLGraphBuilder {
 };
 
 dictionary MLBatchNormalizationSupportLimits {
-  MLSupportLimits input;
-  MLSupportLimits mean;
-  MLSupportLimits variance;
-  MLSupportLimits scale;
-  MLSupportLimits bias;
-  MLSupportLimits output;
+  MLTensorLimits input;
+  MLTensorLimits mean;
+  MLTensorLimits variance;
+  MLTensorLimits scale;
+  MLTensorLimits bias;
+  MLDataTypeLimits output;
 };
 
 partial dictionary MLOpSupportLimits {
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1466,8 +1494,14 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
 
 If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}.
 
+Implementations may support additional data types for operands. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLDataTypeLimits/dataTypes}} value of the corresponding member for the operation.
+
+Issue: Can conforming implementations support *fewer* data types for an operation than what is listed for their [=allowed data types=]? In other words, is the list of allowed types only a recommendation?
```
reillyeon, line 1499: As currently implemented in Chromium "allowed data types" is an upper bound, not a lower bound. If a platform supports more than what is defined in the spec the extra data types are not exposed.

There was discussion at TPAC 2024 about defining "required data types" for operators as well, which would be a required subset and be the only supported types unless the site explicitly opted into performing the feature detection using `opSupportLimits()` to check whether additional data types were supported.

index.bs
```diff
@@ -1466,8 +1494,14 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
...
+
 If an operation requires input operands with a particular [=MLOperand/rank=], the <dfn>allowed ranks</dfn> for each of the operation's input operands, including both positional arguments and options, are given as an explicit rank (e.g. 1), or <dfn lt="any rank">N</dfn> to allow any dimensionality, or the <dfn lt="same rank as">same as</dfn> another operand. More specific constraints are common, such as when an input operand's shape must be [=/unidirectionally broadcastable=] to or [=/bidirectionally broadcastable=] with another input operand; in these cases, the [=/allowed ranks=] are listed as a range, with specific validation given as steps in the operation.
 
+Implementations may (and usually do) impose an upper bound on the [=MLOperand/rank=] of operands. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/rankRange}}.{{MLRankRange/max}} value of the corresponding member for the operation.
```
reillyeon, line 1503: And lower bound?

inexorabletash: A lower bound higher than what is specified as the minimum seems odd to me. Do we anticipate that happening?

index.bs
inexorabletash: Okay. So possible phrasing is:

> Implementations may support fewer data types for operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLDataTypeLimits/dataTypes}} value of the corresponding member for the operation.
> 
> Issue: Can we define the subset of data types that are required for each operator?

Thoughts? Anywhere else we need to sprinkle "may" and "should" ?

index.bs
philloooo: e.g. [prelu on CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.activation.prelu) that requires minimum rank of 3? For this case I did emulate instead of exposing such rank range, but for some cases we probably prefer to not emulate but just expose such rankRange.

index.bs
reillyeon: That's enough "may" and "should" for now.

index.bs
reillyeon: I think about it more as the implementation supporting a narrower range than supported by the spec, which could include both a higher lower bound and a lower upper bound. I agree that it is more likely that an implementation is constrained on the high end. The most common example on the low end is not supporting scalars, which we can polyfill easily enough that it doesn't come up in practice.

It makes things easier for frameworks if both min and max are provided by `opSupportLimits()` even if the min value in practice is mostly defined by the spec and not overridden by the implementation.

inexorabletash: Done in 7328146

index.bs
inexorabletash: Done in 7328146

inexorabletash:
Ready for review, I think. Please take a look?

index.bs
```diff
@@ -1174,64 +1174,92 @@ The {{MLOpSupportLimits}} has the following top level members, aside from these,
...
+
 #### {{MLBinarySupportLimits}} dictionary #### {#api-mlcontext-binarysupportlimits-dictionary}
 <script type="idl">
 dictionary MLBinarySupportLimits {
-  MLSupportLimits a;
-  MLSupportLimits b;
-  MLSupportLimits output;
+  MLTensorLimits a;
+  MLTensorLimits b;
+  MLDataTypeLimits output;
```
fdwr, line 1237: Isn't the output rank also informative?

```javascript
  MLTensorLimits output;
```

If we say that the reason the output rank *isn't useful* is because the operator will always produce a valid output rank given valid inputs (and thus it doesn't matter anyway), then should we also presume that an operator will always produce a valid output data type (barring `cast` and `argMin`/`argMax`) given valid inputs, and thus the output data type is not necessary either? I'm not actually arguing to remove `output` data type, but rather pointing out the odd asymmetry, which justifies output rank's existence. If output rank matters for no other operator, it at least matters for `reshape` (where the output shape is not inferred from the input tensors).

index.bs
```diff
@@ -1174,64 +1174,92 @@ The {{MLOpSupportLimits}} has the following top level members, aside from these,
...
+    :: Minimum supported rank.
+    : <dfn>max</dfn>
+    :: Maximum supported rank.
+</dl>
+
+#### {{MLTensorLimits}} dictionary #### {#api-mlcontext-tensorlimits-dictionary}
+<script type="idl">
+dictionary MLTensorLimits : MLDataTypeLimits {
+  MLRankRange rankRange;
+};
```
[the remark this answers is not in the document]
fdwr: (forgive me if this is the current Chromium implementation with this CR attempting to formalize the implementation, and I missed it during code reviews earlier)

index.bs
```diff
@@ -1471,8 +1499,14 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
...
+
+Issue: Should we specify the subset of data types that must be supported for each operator?
+
 If an operation requires input operands with a particular [=MLOperand/rank=], the <dfn>allowed ranks</dfn> for each of the operation's input operands, including both positional arguments and options, are given as an explicit rank (e.g. 1), or <dfn lt="any rank">N</dfn> to allow any dimensionality, or the <dfn lt="same rank as">same as</dfn> another operand. More specific constraints are common, such as when an input operand's shape must be [=/unidirectionally broadcastable=] to or [=/bidirectionally broadcastable=] with another input operand; in these cases, the [=/allowed ranks=] are listed as a range, with specific validation given as steps in the operation.
 
+Implementations may impose a more restricted lower bound and/or upper bound on the [=MLOperand/rank=] of operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/rankRange}}.{{MLRankRange/min}} and {{MLTensorLimits/rankRange}}.{{MLRankRange/max}} values of the corresponding member for the operation.
```
fdwr, line 1508: > a more restricted lower bound

(a realization) Hmm, that's definitely true for the *upper bound*, but I'm trying to think of any case where a backend's *lower bound* would be more restricted that WebNN's? If such a case exists, then I'd really prefer the Chromium backend simply prepend a few 1's to the shape. e.g. Say hypothetically TFLite/LiteRT only supported elementwise `add` from 2D to 6D (that would be silly, but say so), and you wanted to pass a 1D tensor with 42 elements in it, in which case the Chromium backend just massages that to shape description to `[1,42]` before forwarding onward to TFLite/LiteRT, invisible to the WebNN caller (because *being visible* to the caller provides no performance or functionality benefits here).

We can keep this wording, because maybe there will be a meaningful case in the future, but generally hopefully not. (resolve me)

fdwr (commented):
A few thoughts good sir. 🔍

index.bs
huningxin: > it at least matters for `reshape` (where the output shape is not inferred from the input tensors).

+1, also `expand` whose output rank is defined the size of `newShape`. Reduction ops may also be concerned, say if the platform doesn't support scalar output (rank 0), then reduce all dimensions with `keepDimensions = false` would not be supported.

index.bs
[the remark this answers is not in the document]
inexorabletash: This is the current Chromium implementation but that doesn't mean we shouldn't discuss. But I'll tag in @reillyeon to respond here.

index.bs
inexorabletash: Thoughts @reillyeon ?

index.bs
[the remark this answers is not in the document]
reillyeon: Inheritance simply made this easier to implement in Chromium. Dictionary type names and inheritance aren't visible to script so whatever we do here is just to make the spec clearer to implementers. I have no problem with defining `MLTensorLimits` and `MLDataTypeLimits` as separate types.

index.bs
reillyeon: I agree that since the output type and shape are always a function of the inputs and options we don't need to specify what's supported here explicitly. This does mean that we need a different way of expressing constraints like the data types that can be passed to `cast()` which can be implied by the supported output data types.

index.bs
reillyeon: I'm hesitant to reduce expressivity of the API here, but I agree it would be simpler if we could only have a `maxRank` property and that polyfiling support for lower ranks is simple enough to require implementations to do.

That said, internal to the implementation it's nice that we have this defined as a range because it's easier to check that the rank is between two values that comes from the same place rather than checking the min against a set of spec-defined values and the max against a set of backend-defined values. Similarly, for an ONNX EP or TFLite delegate it might be easier not to have to hard-code the set of minimum values from the spec instead of getting them as values from `opSupportLimits()` since they have to check that for the maximum value anyways.

index.bs
inexorabletash: How about:
* Removing `output` members from all of the support limits dictionaries.
* Adding this dictionary type, which would be used for `cast()`, `argMin()`, `argMax()`

```webidl
dictionary MLSingleInputWithOutputDataTypeSupportLimits {
  MLTensorLimits input;
  sequence<MLOperandDataType> outputDataTypes;
}
```

reillyeon: `argMin()` and `argMax()` call this argument `outputDataType` while `cast()` calls it `type`. I'd rather have the `opSupportLimits()` structure use the argument names. Could we rename both arguments to `dataType`?

index.bs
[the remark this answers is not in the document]
fdwr: > I have no problem with defining MLTensorLimits and MLDataTypeLimits as separate types.

Bundling the data type list and rank range under `MLTensorLimits` makes sense. It's just the inheritance aspect that felt funky. What about:

```webidl
typedef sequence<MLOperandDataType> MLDataTypeList;

dictionary MLTensorLimits {
  MLDataTypeLimits dataTypes;
  MLRankRange rankRange;
};
```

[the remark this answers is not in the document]
reillyeon: That's fine by me.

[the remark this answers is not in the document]
inexorabletash: addressed in e023048

index.bs
inexorabletash: Re: renaming arguments - `outputDataType` in the dicts for `argMin()`/`argMax()` would be a breaking change (is ORT using this yet?) but that shouldn't hold us back right now. But I personally like the self-documenting nature of including "output" for someone reading a code snippet; if it were just `dataType` the reader might wonder why that wasn't implicit. Changing `cast()`'s argument name is a non-breaking change since it's not web-visible. 

Re: the extremes of either including ranks in `output`, or only including output data types when it's an input argument - any further opinions? I lean towards keeping the API... lean. But I acknowledge I'm a weirdo.

reillyeon: I agree with keeping the API lean and just documenting valid input parameters.

fdwr (approved):
👍

inexorabletash:
Thanks @fdwr ! This should still match the Chromium implementation at least as far as what's web-observable, even if the IDL differs somewhat. 

The discussion around whether or how to represent output tensors is still unresolved; two options:

* Discuss it here (pro: already active thread)
* Merge this as-is, file a new issue (pro: incremental progress)

Preferences?

fdwr:
> The discussion around whether or how to represent output tensors is still unresolved; two options:
> 
> * Discuss it here (pro: already active thread)
> * Merge this as-is, file a new issue (pro: incremental progress)
> 
> Preferences?

Incremental progress?

index.bs
huningxin: > Re: the extremes of either including ranks in `output`, or only including output data types when it's an input argument - any further opinions?

I slightly prefer to have ranks in `output`. It basically represents output row in each op's constraints table.

huningxin (approved):
LGTM!

index.bs
inexorabletash: Added #835 to capture some of this discussion.

inexorabletash:
I opened #835 to capture the output tensor discussion. So... good to merge this as-is?

index.bs
philloooo: I also prefer having ranks for output, the `reshape` example that @fdwr bring up is a valid one that makes [CoreML backend output rank validation](https://source.corp.google.com/h/webium-private/prototype/+/main:services/webnn/coreml/graph_builder_coreml.cc;l=5144;drc=a6b7e04eeeef31a788fd9188dbb76de8d37ee708) return an async error instead of type error right now.

fdwr: Resolving this comment which is now captured by: https://github.com/webmachinelearning/webnn/issues/835

index.bs
```diff
@@ -1184,64 +1184,97 @@ The {{MLOpSupportLimits}} has the following top level members, aside from these,
 <script type="idl">
 dictionary MLOpSupportLimits {
   MLInputOperandLayout preferredInputLayout;
-  MLSupportLimits input;
-  MLSupportLimits constant;
-  MLSupportLimits output;
+  [EnforceRange] unsigned long long maxTensorByteLength;
+  MLDataTypeLimits input;
+  MLDataTypeLimits constant;
+  MLDataTypeLimits output;
```
fdwr, line 1190: (already merged, but looking back over it again) Would it make more sense for these to be `MLTensorLimits` to capture any maximum ranks [imposed by the API](https://github.com/webmachinelearning/webnn/issues/456#issuecomment-1692533722)? For example, DML has a global maximum rank of 8D (regardless of operator), and CoreML (Phillis can correct me if incorrect) has a limit of 5D.

PR #829, opened 2025-03-13
Allow Uint8Array in constant() etc. as a generic ArrayBufferView type

### LINKED ISSUE #825: Always allow Uint8Array in constant() as a generic ArrayBuffer slice type
`constant()` currently allows either the specific typed-array type for a data type (e.g. `"float32"` is mapped to `Float32Array`) as well as `ArrayBuffer`. It would be nice to also allow `Uint8Array` as a generic way of representing a slice of an `ArrayBuffer`. The reasons for this are:

1. For data types where we don't have a specific typed-array type (as was the case for `"float16"` before `Float16Array` was added to ECMAScript) always allowing `Uint8Array` provides a backwards-compatible migration path when such a type is introduced.
2. For code calling WebNN from WASM, it's annoying to have to build these specific types out of what is usually a generic `Uint8Array` view over the WASM memory. For example this `switch` statement in the ONNX WebNN EP, which I notice introduces an unnecessary copy of the buffer data in order to reduce the complexity of the code: https://github.com/microsoft/onnxruntime/blob/ccf8fdd9ea8b0df0eb4b092323479ce8a47be161/js/web/lib/wasm/jsep/backend-webnn.ts#L312

inexorabletash:
As noted in the issue, it is convenient allow Uint8Array as a generic way of representing a slice of an ArrayBuffer, instead of always requireing either an ArrayBuffer to be passed (which may involve copying), or constructing a view type corresponding to the dataType.

Resolves #825

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/829.html" title="Last updated on Mar 13, 2025, 9:56 PM UTC (1668590)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/829/3dd9216...inexorabletash:1668590.html" title="Last updated on Mar 13, 2025, 9:56 PM UTC (1668590)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -42,7 +42,6 @@ Text Macro: EMULATED generically emulated from the usage of other operations as 
 urlPrefix: https://tc39.es/ecma262/; spec: ECMA-262
     type: dfn
         text: element size; url: table-the-typedarray-constructors
-        text: element type; url: table-the-typedarray-constructors
         text: view constructor; url: table-the-typedarray-constructors
         text: equally close values; url: sec-ecmascript-language-types-number-type
         text: settled; for: Promise; url: sec-promise-objects
@@ -924,8 +923,14 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
             : {{SharedArrayBuffer}}
             :: Return true.
             : {{ArrayBufferView}}
-            :: If |bufferSource|'s [=element type=] matches |descriptor|'s {{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility) return true, otherwise return false.
+            ::
+                1. If |bufferSource| is a {{Uint8Array}} object, then return true.
+                1. If |bufferSource| matches |descriptor|'s {{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility), then return true.
+                1. Return false.
         </dl>
+
+    Note: Using {{Uint8Array}} regardless of the |descriptor|'s {{MLOperandDescriptor/dataType}} is supported as a generic way of representing a slice of an {{ArrayBuffer}}, for example part of a {{WebAssembly}}.{{Memory}} instance. Developers are encouraged to use more specific view types when authoring WebNN code for readability and maintainability.
+
 </details>
 
 <details open algorithm>


inexorabletash:
The "element type" link seemed unnecessary, so I dropped it. Wording improvements welcome!

@reillyeon can you take an initial look?

reillyeon (approved):

inexorabletash:
@huningxin @fdwr - please review? Your call on whether we want any more discussion on this - I saw you both gave :+1: on the issue.

index.bs
```diff
@@ -924,8 +923,14 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
...
-            :: If |bufferSource|'s [=element type=] matches |descriptor|'s {{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility) return true, otherwise return false.
+            ::
+                1. If |bufferSource| is a {{Uint8Array}} object, then return true.
+                1. If |bufferSource| matches |descriptor|'s {{MLOperandDescriptor/dataType}} according to [this table](#appendices-mloperanddatatype-arraybufferview-compatibility), then return true.
+                1. Return false.
         </dl>
+
+    Note: Using {{Uint8Array}} regardless of the |descriptor|'s {{MLOperandDescriptor/dataType}} is supported as a generic way of representing a slice of an {{ArrayBuffer}}, for example part of a {{WebAssembly}}.{{Memory}} instance. Developers are encouraged to use more specific view types when authoring WebNN code for readability and maintainability.
+
 </details>
```
fdwr, line 934: Is Float16Array at stage 4 now? If so, should we update this text too:

> Float16Array is at ECMA Stage 3 signaling its design is finished. Implementers wanting to enable this type ahead native implementations can emulate the type by passing raw bits via Uint16Array. [Issue webnn#373]

fdwr: Never mind - it still has 3 checkboxes for stage 4 completion: https://github.com/tc39/proposal-float16array/issues/7

fdwr (approved):
> I saw you both gave 👍 on the issue.

Yep, and here's another 👍🙂.

huningxin (approved):
LGTM!

PR #830, opened 2025-03-19
Allow tensors for graph constants.

### LINKED ISSUE #760: Support building graphs from `MLTensor` containing constants
Demonstrate how `MLTensor` can be used to help web developers manage constant data (e.g., trained weights) on-device.

Dependent PRs
* `MLConstantOperand`: https://github.com/webmachinelearning/webnn/issues/668#issue-2272156183
* `MLTensor`: https://github.com/webmachinelearning/webnn/pull/754

## Motivation
* Allow constant data to be uploaded directly to the device, which is a capability that Execution Providers (EPs) leverage to prevent out-of-memory (OOM) errors ([ORT example](https://github.com/microsoft/onnxruntime/pull/21634)).
* Re-use constant buffers in system memory between graphs, particularly for encoder-decoder models like Whisper.

## Design
 `MLTensor` containing constant data will be associated upon creating the `MLOperand`. At build(), the constant data will be forwarded into the device. The original constant data (ie. `ArrayBuffer` input or uploaded device data held by `MLTensor`) can be discarded immediately after createConstant() succeeds.

## Example JS
```js
// Upload constant data directly to device
constantTensor = ctx.createConstant(...); // immutable

builder1 = new MLGraphBuilder(ctx);
constantOp1 = builder1.constant(constantTensor);
constantOp2 = builder2.constant(constantTensor);
// ...
graph1 = await builder1.build(...);
graph2 = await builder2.build(...);

// Optional: free-up system memory
constantTensor.destroy();
```

## Proposed IDL
```idl

interface MLConstantTensor : MLTensor {};

partial interface MLContext {
    Promise<MLConstantTensor> createConstant(MLOperandDataType dataType, ArrayBufferView sourceData);
};

    MLOperand constant(MLConstantTensor tensor);
};
```

Edits:
* 9/16: Added `MLOperandDescriptor` as required by `MLOperand`
* 9/18: Added constant-initializer to createTensor()
* 9/19: Reuse input(..) via constant usage flag
* 1/29: Have new tensor type passed to constant()

bbernhar:
As noted in the issue, sharing weights between graphs is a effective way to enable reuse of device buffers, allowing web developers to manage weight data in graphs similarly to other input or output buffers.

@fdwr @huningxin @reillyeon 

Based on the work in Chromium by Austin Sullivan.

Resolves #760

***
<a href="https://pr-preview.s3.amazonaws.com/bbernhar/webnn/pull/830.html" title="Last updated on Apr 21, 2025, 9:22 PM UTC (8b972ff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/830/69bb88d...bbernhar:8b972ff.html" title="Last updated on Apr 21, 2025, 9:22 PM UTC (8b972ff)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1493,6 +1493,10 @@ typedef (bigint or unrestricted double) MLNumber;
     : <dfn>\[[operator]]</dfn> of type [=operator=]
     ::
         Reference to {{MLOperand}}'s corresponding [=operator=].
+  
+    : <dfn>\[[constantTensor]]</dfn> of type {{MLConstantTensor}}
+    ::
+        The {{MLOperand}}'s tensor (only for constant operands).
   </dl>
 </div>
 
@@ -1842,6 +1846,7 @@ Create a constant {{MLOperand}} of the specified data type and shape that contai
     1. If [=this=] [=MLGraphBuilder/can not build=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. *Make graph connections:*
         1. Let |operand| be the result of [=creating an MLOperand=] given [=this=] and |tensor|.{{MLTensor/[[descriptor]]}}.
+        1. Set |operand|.{{MLOperand/[[constantTensor]]}} to |tensor|.
         1. Add |operand| to [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/constants=] with |tensor| as value.
     1. Return |operand|.
 </details>
@@ -1893,6 +1898,7 @@ Build a composed graph up to a given output operand into a computational graph a
         1. If |name| is empty, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
         1. If [=MLGraphBuilder/validating operand=] given [=this=] and |operand| returns false, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
         1. If |operand| is in [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/inputs=] or [=computational graph/constants=], then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
+        1. If |operand|.{{MLOperand/[[constantTensor]]}} exists and |operand|.{{MLOperand/[[constantTensor]]}}.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}. 
     1. Let |operands| be a new empty [=/set=].
     1. Let |operators| be a new empty [=/set=].
     1. Let |inputs| be a new empty [=/set=].
--- a/index.bs
+++ b/index.bs
@@ -1620,6 +1620,7 @@ interface MLTensor {
     : <dfn>\[[data]]</dfn> of an [=implementation-defined=] type
     ::
         The bytes backing the {{MLTensor}}. This data may only be accessed or modified from the {{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}.
+
     : <dfn>\[[isConstant]]</dfn> of type {{boolean}}
     ::
         Whether the {{MLTensor}} was created by [=create a constant MLTensor=].
--- a/index.bs
+++ b/index.bs
@@ -1683,7 +1683,7 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
 
 Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor once all previously submitted operations using it are complete.
 
-### Creating a constant {{MLTensor}} ### {#api-mlconstanttensor-create}
+### Creating a constant {{MLTensor}} ### {#api-mlconstant-tensor-create}
 
 A constant {{MLTensor}} is created by its associated {{MLContext}}.
 
--- a/index.bs
+++ b/index.bs
@@ -941,7 +941,7 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
   </summary>
     1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
-        1. If |tensor|.{{MLTensor/[[isConstant]]}} is true then return false.
+        1. If |tensor|.{{MLTensor/[[isConstant]]}} is true, then return false.
         1. If |namedDescriptors|[|name|] does not [=map/exist=], then return false.
         1. If |tensor|.{{MLTensor/[[descriptor]]}} is not [=MLOperandDescriptor/equal=] to |namedDescriptors|[|name|], then return false.
     1. Return true.
--- a/index.bs
+++ b/index.bs
@@ -869,7 +869,7 @@ interface MLContext {
   undefined dispatch(MLGraph graph, MLNamedTensors inputs, MLNamedTensors outputs);
 
   Promise<MLTensor> createTensor(MLTensorDescriptor descriptor);
-  Promise<MLConstantTensor> createConstantTensor(
+  Promise<MLTensor> createConstantTensor(
     MLOperandDescriptor descriptor, AllowSharedBufferSource inputData);
 
   Promise<ArrayBuffer> readTensor(MLTensor tensor);
@@ -941,7 +941,7 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
   </summary>
     1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
-        1. If |tensor| is a {{MLConstantTensor}} then return false.
+        1. If |tensor|.{{MLTensor/[[isConstant]]}} is true then return false.
         1. If |namedDescriptors|[|name|] does not [=map/exist=], then return false.
         1. If |tensor|.{{MLTensor/[[descriptor]]}} is not [=MLOperandDescriptor/equal=] to |namedDescriptors|[|name|], then return false.
     1. Return true.
@@ -1070,14 +1070,14 @@ Creates an {{MLTensor}} associated with this {{MLContext}}.
 
 ### {{MLContext/createConstantTensor()}}  ### {#api-mlcontext-createconstanttensor}
 
-Creates an {{MLConstantTensor}} associated with this {{MLContext}}.
+Creates a constant {{MLTensor}} associated with this {{MLContext}}.
 
 <div dfn-for="MLContext/createConstantTensor(descriptor, inputData)" dfn-type=argument>
     **Arguments:**
       - <dfn>descriptor</dfn>: an {{MLOperandDescriptor}}.
       - <dfn>inputData</dfn>: an {{AllowSharedBufferSource}}. The buffer whose bytes will be written into the tensor.
 
-    **Returns:** {{Promise}}<{{MLConstantTensor}}>.
+    **Returns:** {{Promise}}<{{MLTensor}}>.
 </div>
 
 <details open algorithm>
@@ -1091,7 +1091,7 @@ Creates an {{MLConstantTensor}} associated with this {{MLContext}}.
     1. If [=validating buffer with descriptor=] given |inputData| and |descriptor| returns false, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
     1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
     1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |descriptor|'s [=MLOperandDescriptor/byte length=].
-    1. Let |tensor| be the result of [=creating an MLConstantTensor=] given [=this=], and |descriptor|.
+    1. Let |tensor| be the result of [=creating a constant MLTensor=] given [=this=], and |descriptor|.
     1. Let |promise| be [=a new promise=] in |realm|.
     1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}:
         1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
@@ -1494,7 +1494,7 @@ typedef (bigint or unrestricted double) MLNumber;
     ::
         Reference to {{MLOperand}}'s corresponding [=operator=].
   
-    : <dfn>\[[constantTensor]]</dfn> of type {{MLConstantTensor}}
+    : <dfn>\[[constantTensor]]</dfn> of type {{MLTensor}}
     ::
         The {{MLOperand}}'s tensor (only for constant operands).
   </dl>
@@ -1596,6 +1596,7 @@ interface MLTensor {
   readonly attribute FrozenArray<unsigned long> shape;
   readonly attribute boolean readable;
   readonly attribute boolean writable;
+  readonly attribute boolean constant;
 
   undefined destroy();
 };
@@ -1619,6 +1620,9 @@ interface MLTensor {
     : <dfn>\[[data]]</dfn> of an [=implementation-defined=] type
     ::
         The bytes backing the {{MLTensor}}. This data may only be accessed or modified from the {{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}.
+    : <dfn>\[[isConstant]]</dfn> of type {{boolean}}
+    ::
+        Whether the {{MLTensor}} was created by [=create a constant MLTensor=].
   </dl>
 </div>
 
@@ -1634,6 +1638,8 @@ The <dfn attribute for=MLTensor>readable</dfn> [=getter steps=] are to return [=
 
 The <dfn attribute for=MLTensor>writable</dfn> [=getter steps=] are to return [=this=].{{MLTensor/[[descriptor]]}}.{{MLTensorDescriptor/writable}}.
 
+The <dfn attribute for=MLTensor>constant</dfn> [=getter steps=] are to return [=this=]'s {{MLTensor/[[isConstant]]}}.
+
 ### Creating an {{MLTensor}} ### {#api-mltensor-create}
 
 An {{MLTensor}} is created by its associated {{MLContext}}.
@@ -1647,6 +1653,7 @@ An {{MLTensor}} is created by its associated {{MLContext}}.
     1. Set |tensor|.{{MLTensor/[[context]]}} to |context|.
     1. Set |tensor|.{{MLTensor/[[descriptor]]}} to |descriptor|.
     1. Set |tensor|.{{MLTensor/[[isDestroyed]]}} to false.
+    1. Set |tensor|.{{MLTensor/[[isConstant]]}} to false.
     1. Return |tensor|.
 </details>
 
@@ -1670,36 +1677,25 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
 
 Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor once all previously submitted operations using it are complete.
 
-## {{MLConstantTensor}} interface ## {#api-mlconstanttensor}
+### Creating a constant {{MLTensor}} ### {#api-mlconstanttensor-create}
 
-The {{MLConstantTensor}} interface represents a tensor which may be used as a constant to an {{MLGraph}}. The memory backing an {{MLConstantTensor}} should be allocated in an [=implementation-defined=] fashion according to the requirements of the {{MLContext}} and the {{MLOperandDescriptor}} used to create it. Operations involving the {{MLTensor/[[data]]}} of an {{MLConstantTensor}} occur on the {{MLContext/[[timeline]]}} of its associated {{MLContext}}.
-
-The [=implementation-defined=] requirements of how an {{MLConstantTensor}} is allocated may include constraints such as that the memory is allocated with a particular byte alignment or in a particular memory pool.
-
-<script type=idl>
-[SecureContext, Exposed=(Window, Worker)]
-interface MLConstantTensor : MLTensor {
-};
-</script>
-
-### Creating an {{MLConstantTensor}} ### {#api-mlconstanttensor-create}
-
-An {{MLConstantTensor}} is created by its associated {{MLContext}}.
+A constant {{MLTensor}} is created by its associated {{MLContext}}.
 
 <details open algorithm>
   <summary>
-    To <dfn>create an MLConstantTensor</dfn> given {{MLContext}} |context|, {{MLOperandDescriptor}} |inputDescriptor|, run the following steps:
+    To <dfn>create a constant MLTensor</dfn> given {{MLContext}} |context|, {{MLOperandDescriptor}} |inputDescriptor|, run the following steps:
   </summary>
     1. Let |realm| be |context|'s [=relevant realm=].
-    1. Let |tensor| be a new {{MLConstantTensor}} in |realm|.
+    1. Let |tensor| be a new {{MLTensor}} in |realm|.
     1. Set |tensor|.{{MLTensor/[[context]]}} to |context|.
     1. Let |tensorDescriptor| be a new {{MLTensorDescriptor}}.
-    1. Set |tensorDescriptor|.{{MLTensorDescriptor/readable}} to true.
+    1. Set |tensorDescriptor|.{{MLTensorDescriptor/readable}} to false.
     1. Set |tensorDescriptor|.{{MLTensorDescriptor/writable}} to false.
     1. Set |tensorDescriptor|.{{MLOperandDescriptor/dataType}} to |inputDescriptor|.{{MLOperandDescriptor/dataType}}.
     1. Set |tensorDescriptor|.{{MLOperandDescriptor/shape}} to |inputDescriptor|.{{MLOperandDescriptor/shape}}.
     1. Set |tensor|.{{MLTensor/[[descriptor]]}} to |tensorDescriptor|.
     1. Set |tensor|.{{MLTensor/[[isDestroyed]]}} to false.
+    1. Set |tensor|.{{MLTensor/[[isConstant]]}} to true.
     1. Return |tensor|.
 </details>
 
@@ -1725,8 +1721,8 @@ interface MLGraphBuilder {
   // Create a scalar operand from the specified number of the specified type.
   MLOperand constant(MLOperandDataType type, MLNumber value);
 
-  // Create an operand from a specified tensor.
-  MLOperand constant(MLConstantTensor tensor);
+  // Create an operand from a specified constant tensor.
+  MLOperand constant(MLTensor tensor);
 
   // Compile the graph up to the specified output operands asynchronously.
   Promise<MLGraph> build(MLNamedOperands outputs);
@@ -1833,7 +1829,7 @@ Create a constant {{MLOperand}} of the specified data type and shape that contai
 
 <div dfn-for="MLGraphBuilder/constant(tensor)" dfn-type=argument>
     **Arguments:**
-        - <dfn>tensor</dfn>: an {{MLConstantTensor}}. The tensor containing the initialized data.
+        - <dfn>tensor</dfn>: an {{MLTensor}}. The constant tensor containing the initialized data.
     **Returns:** an {{MLOperand}}. The constant output tensor.
 </div>
 
@@ -1843,6 +1839,7 @@ Create a constant {{MLOperand}} of the specified data type and shape that contai
   </summary>
     1. If |tensor|.{{MLTensor/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
     1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[isConstant]]}} is false, then [=exception/throw=] a {{TypeError}}.
     1. If [=this=] [=MLGraphBuilder/can not build=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. *Make graph connections:*
         1. Let |operand| be the result of [=creating an MLOperand=] given [=this=] and |tensor|.{{MLTensor/[[descriptor]]}}.
--- a/index.bs
+++ b/index.bs
@@ -869,6 +869,8 @@ interface MLContext {
   undefined dispatch(MLGraph graph, MLNamedTensors inputs, MLNamedTensors outputs);
 
   Promise<MLTensor> createTensor(MLTensorDescriptor descriptor);
+  Promise<MLConstantTensor> createConstantTensor(
+    MLOperandDescriptor descriptor, AllowSharedBufferSource inputData);
 
   Promise<ArrayBuffer> readTensor(MLTensor tensor);
   Promise<undefined> readTensor(MLTensor tensor, AllowSharedBufferSource outputData);
@@ -939,6 +941,7 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
   </summary>
     1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
+        1. If |tensor| is a {{MLConstantTensor}} then return false.
         1. If |namedDescriptors|[|name|] does not [=map/exist=], then return false.
         1. If |tensor|.{{MLTensor/[[descriptor]]}} is not [=MLOperandDescriptor/equal=] to |namedDescriptors|[|name|], then return false.
     1. Return true.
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1627,6 +1664,90 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
 
+## {{MLConstantTensor}} interface ## {#api-mlconstanttensor}
+
+The {{MLConstantTensor}} interface represents a tensor which may be used as a constant to an {{MLGraph}}. The memory backing an {{MLConstantTensor}} should be allocated in an [=implementation-defined=] fashion according to the requirements of the {{MLContext}} and the {{MLOperandDescriptor}} used to create it. Operations involving the {{MLConstantTensor/[[data]]}} of an {{MLConstantTensor}} occur on the {{MLContext/[[timeline]]}} of its associated {{MLContext}}.
+
+The [=implementation-defined=] requirements of how an {{MLConstantTensor}} is allocated may include constraints such as that the memory is allocated with a particular byte alignment or in a particular memory pool.
+
+<script type=idl>
+[SecureContext, Exposed=(Window, Worker)]
+interface MLConstantTensor : MLTensor {
```
inexorabletash, line 1675: Since `MLConstantTensor` inherits from `MLTensor`, a lot of the added spec text seems unnecessary - the internal slots should be inherited, definitions like "An MLConstantTensor's dataType..." aren't needed, etc.

Like in C++, only the behavior differences need to be called out (like virtual method overrides), and like in C++ this might require refactors in the "base class". 

Also, presumably `writeTensor()` (and `readTensor()` ?)  should throw if an `MLConstantTensor` is passed.

index.bs
```diff
@@ -1627,6 +1664,90 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
 
 Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor once all previously submitted operations using it are complete.
 
+## {{MLConstantTensor}} interface ## {#api-mlconstanttensor}
+
+The {{MLConstantTensor}} interface represents a tensor which may be used as a constant to an {{MLGraph}}. The memory backing an {{MLConstantTensor}} should be allocated in an [=implementation-defined=] fashion according to the requirements of the {{MLContext}} and the {{MLOperandDescriptor}} used to create it. Operations involving the {{MLConstantTensor/[[data]]}} of an {{MLConstantTensor}} occur on the {{MLContext/[[timeline]]}} of its associated {{MLContext}}.
```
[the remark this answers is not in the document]
inexorabletash: There's nothing currently in the spec precisely defining a timeline; most web specs simply use "in parallel" to indicate another process/thread but possibly just another "sequence" on the same thread, allowing implementation flexibility.

So "timeline" here shouldn't imply GPU (or "non-CPU"), just that it must be treated as if on a separate sequence/thread/process and can't be accessed synchronously from script.

index.bs
```diff
@@ -1627,6 +1664,90 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
+
+<details open algorithm>
+  <summary>
+    To <dfn>create an MLConstantTensor</dfn> given {{MLContext}} |context|, {{MLOperandDescriptor}} |descriptor|, and {{AllowSharedBufferSource}} |inputData|, run the following steps:
+  </summary>
+    1. Let |realm| be |context|'s [=relevant realm=].
+    1. Let |tensor| be a new {{MLConstantTensor}} in |realm|.
+    1. Set |tensor|.{{MLConstantTensor/[[context]]}} to |context|.
+    1. Set |tensor|.{{MLConstantTensor/[[descriptor]]}} to |descriptor|.
+    1. Set |tensor|.{{MLConstantTensor/[[data]]}} to |inputData|.
```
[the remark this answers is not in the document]
bbernhar: Removed.

index.bs
```diff
@@ -1627,6 +1664,90 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
+    1. Set |tensor|.{{MLConstantTensor/[[isDestroyed]]}} to false.
+    1. Set |tensor|.{{MLConstantTensor/[[descriptor]]}}.{{MLTensorDescriptor/readable}} to true.
```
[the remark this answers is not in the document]
bbernhar: The `descriptor` should be a `MLTensorDescriptor` with static attributes.

index.bs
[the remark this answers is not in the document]
bbernhar: Removed.

index.bs
```diff
@@ -941,6 +942,7 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
         1. If |namedDescriptors|[|name|] does not [=map/exist=], then return false.
         1. If |tensor|.{{MLTensor/[[descriptor]]}} is not [=MLOperandDescriptor/equal=] to |namedDescriptors|[|name|], then return false.
+        1. If |tensor| is a {{MLConstantTensor}} then return false.
```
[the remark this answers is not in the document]
bbernhar: SGTM.

index.bs
```diff
@@ -1065,6 +1067,41 @@ Creates an {{MLTensor}} associated with this {{MLContext}}.
...
+    1. If [=this=] [=MLContext/is lost=], then return [=a new promise=] in |realm| [=rejected=] with an "{{InvalidStateError}}" {{DOMException}}.
+    1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating buffer with descriptor=] given |inputData| and |descriptor| returns false, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
+    1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
+    1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |descriptor|'s [=MLOperandDescriptor/byte length=].
+    1. Let |tensor| be the result of [=creating an MLConstantTensor=] given [=this=], and |descriptor|.
+    1. Let |promise| be [=a new promise=] in |realm|.
+    1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Create |tensor|.{{MLConstantTensor/[[data]]}} given |descriptor| and initialize all bytes to zeros.
```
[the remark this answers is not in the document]
bbernhar: Removed.

index.bs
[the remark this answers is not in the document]
bbernhar: > Does this a normative requirement?

It is required that the order of operations in the context is executed in the same sequence as they are called; however, it is not mandatory for tensors to be backed by on-device storage—this decision is left to the implementation.

index.bs
bbernhar: >Since MLConstantTensor inherits from MLTensor, a lot of the added spec text seems unnecessary

SGTM.

> Also, presumably writeTensor() (and readTensor() ?) should throw if an MLConstantTensor is passed.

It is valid to read a constant tensor but not write - assigned static usages would throw.

index.bs
```diff
@@ -1627,6 +1664,41 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
+An {{MLConstantTensor}} is created by its associated {{MLContext}}.
+
+<details open algorithm>
+  <summary>
+    To <dfn>create an MLConstantTensor</dfn> given {{MLContext}} |context|, {{MLOperandDescriptor}} |descriptor|, run the following steps:
+  </summary>
+    1. Let |realm| be |context|'s [=relevant realm=].
+    1. Let |tensor| be a new {{MLConstantTensor}} in |realm|.
+    1. Set |tensor|.{{MLTensor/[[context]]}} to |context|.
+    1. Set |tensor|.{{MLTensor/[[descriptor]]}} to |descriptor|.
```
inexorabletash, line 1695: The `descriptor` variable here is an `MLOperandDescriptor` but the **[[descriptor]]** internal slot is a `MLTensorDescriptor`

index.bs
```diff
@@ -1627,6 +1664,41 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
+The [=implementation-defined=] requirements of how an {{MLConstantTensor}} is allocated may include constraints such as that the memory is allocated with a particular byte alignment or in a particular memory pool.
+
+<script type=idl>
+[SecureContext, Exposed=(Window, Worker)]
+interface MLConstantTensor : MLTensor {
+};
+</script>
+
+
+The <dfn attribute for=MLConstantTensor>readable</dfn> [=getter steps=] are to return true.
```
inexorabletash, line 1680: `MLConstantTensor` doesn't define `readable` or `writable` attributes, just inherits them, so we wouldn't normally spec it this way. You might be able to get away with defining overrides on `MLConstantTensor` but I'm not sure what would happen if script invokes the `MLTensor` getter functions with an `MLConstantTensor` "this" object. 

Since `MLTensor`'s definitions of `readable` and `writable` are to read the properties out of the **[[descriptor]]** internal slot's dict, as part of addressing setting that slot (see other comment) you can just set them to true/false during the tensor creation.

bbernhar: Okay. Seems we can remove this as well then.

index.bs
bbernhar: Done. 

Relabeled the passed descriptor to `inputDescriptor` to keep it unambiguous.

index.bs
```diff
@@ -869,6 +869,7 @@ interface MLContext {
   undefined dispatch(MLGraph graph, MLNamedTensors inputs, MLNamedTensors outputs);
 
   Promise<MLTensor> createTensor(MLTensorDescriptor descriptor);
+  Promise<MLConstantTensor> createConstantTensor(MLOperandDescriptor descriptor, AllowSharedBufferSource inputData);
```
inexorabletash, line 872: nitpick: Wrap this to ~80 columns or it'll cause horizontal scrollbars to appear. The lint tool will flag it.

index.bs
```diff
@@ -1627,6 +1664,39 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
+An {{MLConstantTensor}} is created by its associated {{MLContext}}.
+
+<details open algorithm>
+  <summary>
+    To <dfn>create an MLConstantTensor</dfn> given {{MLContext}} |context|, {{MLOperandDescriptor}} |inputDescriptor|, run the following steps:
+  </summary>
+    1. Let |realm| be |context|'s [=relevant realm=].
+    1. Let |tensor| be a new {{MLConstantTensor}} in |realm|.
+    1. Set |tensor|.{{MLTensor/[[context]]}} to |context|.
+    1. Let |descriptor| be a new {{MLTensorDescriptor}}.
```
inexorabletash, line 1690: There's nothing wrong with this, but the lint tool will flag subsequent lines because it wants a variable named `descriptor` to be an `MLOperandDescriptor` in subsequent steps. (Without that lint rule it's easy to link to the wrong thing.) If you rename this `tensorDescriptor` the lint tool will not complain.

inexorabletash: Also, a different issue: the lint tool will flag this line with `Let ... be a new ...` as requiring a "realm". This is a false positive since dictionaries are just maps and don't need a realm. To quiet that, we'll need to update the lint tool here:

https://github.com/webmachinelearning/webnn/blob/4c8c75e200f8930f30d1b10c8dbecb8e6c953eeb/tools/lint.mjs#L409

Just add `"MLTensorDescriptor"` to the list of dictionary types. Sorry about that! (I have a to-do item to parse the spec's IDL and determine what is a dictionary automagically, but that hasn't happened yet.)

index.bs
```diff
@@ -1058,7 +1060,42 @@ Creates an {{MLTensor}} associated with this {{MLContext}}.
...
+    1. If [=this=] [=MLContext/is lost=], then return [=a new promise=] in |realm| [=rejected=] with an "{{InvalidStateError}}" {{DOMException}}.
+    1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
+    1. If [=validating buffer with descriptor=] given |inputData| and |descriptor| returns false, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
+    1. Let |bytes| be the result of [=getting a copy of the bytes held by the buffer source=] given |inputData|.
+    1. [=Assert=]: |bytes|'s [=byte sequence/length=] is equal to |descriptor|'s [=MLOperandDescriptor/byte length=].
+    1. Let |tensor| be the result of [=creating an MLConstantTensor=] given [=this=], and |descriptor|.
+    1. Let |promise| be [=a new promise=] in |realm|.
+    1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Create |tensor|.{{MLTensor/[[data]]}} given |descriptor|.
```
inexorabletash, line 1097: Nitpicky, but: the `createTensor()` steps are written as if this "Create tensor.[[data]] given descriptor." steps can fail. Do we want to move (or copy) the following "If that fails..." step to just below this one?

(The `writeTensor()` steps have an issue about whether copying the bytes in can fail.)

inexorabletash (changes requested):
Overall LGTM, just some annoying nitpicks. I ran the lint tool locally over the PR to avoid surprises later.

index.bs
```diff
@@ -1058,7 +1060,42 @@ Creates an {{MLTensor}} associated with this {{MLContext}}.
     1. Let |promise| be [=a new promise=] in |realm|.
     1. Enqueue the following steps to [=this=].{{MLContext/[[timeline]]}}:
         1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
-            1. Create |tensor|.{{MLTensor/[[data]]}} given |descriptor| and initialize all bytes to zeros.
+            1. Create |tensor|.{{MLTensor/[[data]]}} given |descriptor|.
```
huningxin, line 1063: I suppose "initialize all bytes to zeros" should be kept for `CreateTensor()`.

index.bs
```diff
@@ -1627,6 +1664,39 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
 
 Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor once all previously submitted operations using it are complete.
 
+## {{MLConstantTensor}} interface ## {#api-mlconstanttensor}
+
+The {{MLConstantTensor}} interface represents a tensor which may be used as a constant to an {{MLGraph}}. The memory backing an {{MLConstantTensor}} should be allocated in an [=implementation-defined=] fashion according to the requirements of the {{MLContext}} and the {{MLOperandDescriptor}} used to create it. Operations involving the {{MLTensor/[[data]]}} of an {{MLConstantTensor}} occur on the {{MLContext/[[timeline]]}} of its associated {{MLContext}}.
```
huningxin, line 1669: Following on discussion (which is on outdated change): https://github.com/webmachinelearning/webnn/pull/830#discussion_r2004938741.

Given operations on MLConstantTensor.data should occur on the MLContext.timeline, the [MLGraphBuilder.build()](https://pr-preview.s3.amazonaws.com/bbernhar/webnn/pull/830.html#dom-mlgraphbuilder-build) steps may need to reflect that, because it now may access the MLConstantTensor.data.

bbernhar: Yes, good point. Not 100% sure on the formal language: seems "in parallel" should become "enqueue...". PTAL again.

index.bs
bbernhar: Fixed

index.bs
bbernhar: I suppose we can copy as `createConstantTensor()` could reject on write failure now.

index.bs
bbernhar: Done.

index.bs
bbernhar: @inexorabletash Do I revert? Looks like lint.mjs has been updated.

inexorabletash: Yes! Sorry about that... I wasn't sure if I could get a more intelligent change in; it went much more quickly than I expected! @bbernhar

anssiko:
(FYI - I enabled the CI workflow for this PR and all the tests passed, including the newly added lint rules from https://github.com/webmachinelearning/webnn/pull/831)

index.bs
```diff
@@ -1817,9 +1914,10 @@ Build a composed graph up to a given output operand into a computational graph a
         1. Set |graph|.{{MLGraph/[[outputDescriptors]]}}[|name|] to |operand|.{{MLOperand/[[descriptor]]}}.
     1. Set [=this=].{{MLGraphBuilder/[[hasBuilt]]}} to true.
     1. Let |promise| be [=a new promise=] in |realm|.
-    1. Run the following steps [=in parallel=]:
+    1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Run these steps, but [=/abort when=] |graph|.{{MLGraph/[[context]]}} [=MLContext/is lost=]:
             1. Let |graphImpl| be the result of converting [=this=]'s [=MLGraphBuilder/graph=] with |operands|, |operators|, |inputs|, and |outputs|'s [=map/values=] into an [=implementation-defined=] format which can be interpreted by the underlying platform.
+                1. During this step, for each |operand| in |operands|, if |operand| was created using {{MLGraphBuilder/constant(tensor)|constant()}} with a given {{MLConstantTensor}} |tensor|, the {{MLTensor/[[data]]}} of |tensor| may be accessed on the {{MLContext/[[timeline]]}}.
```
inexorabletash, line 1920: Note that "may" here in normative has the special RFC 2119 meaning, i.e. that it's a optional rather than a requirement for implementations, which is probably not what is intended.

Also, since these steps are enqueued to the context's timeline, it's implicit that all of this work is running on that timeline, so I don't think this step is necessary.

bbernhar: Removed.

inexorabletash (approved):
LGTM! Thank you for your patience.

index.bs
```diff
@@ -1065,6 +1068,42 @@ Creates an {{MLTensor}} associated with this {{MLContext}}.
...
+</div>
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLContext>createConstantTensor(|descriptor|, |inputData|)</dfn> method steps are:
+  </summary>
+    1. Let |global| be [=this=]'s [=relevant global object=].
+    1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. If [=this=] [=MLContext/is lost=], then return [=a new promise=] in |realm| [=rejected=] with an "{{InvalidStateError}}" {{DOMException}}.
+    1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
```
reillyeon, line 1090: Why do the steps above and below return rejected promises but this step throws an exception?

index.bs
```diff
@@ -1627,6 +1666,39 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
...
 
+## {{MLConstantTensor}} interface ## {#api-mlconstanttensor}
+
+The {{MLConstantTensor}} interface represents a tensor which may be used as a constant to an {{MLGraph}}. The memory backing an {{MLConstantTensor}} should be allocated in an [=implementation-defined=] fashion according to the requirements of the {{MLContext}} and the {{MLOperandDescriptor}} used to create it. Operations involving the {{MLTensor/[[data]]}} of an {{MLConstantTensor}} occur on the {{MLContext/[[timeline]]}} of its associated {{MLContext}}.
+
+The [=implementation-defined=] requirements of how an {{MLConstantTensor}} is allocated may include constraints such as that the memory is allocated with a particular byte alignment or in a particular memory pool.
+
+<script type=idl>
+[SecureContext, Exposed=(Window, Worker)]
+interface MLConstantTensor : MLTensor {
```
reillyeon, line 1677: Since an `MLConstantTensor` cannot be used with `dispatch()` or `writeTensor()` (and using it with `readTensor()`  is questionably useful) does it make sense for it to inherit from `MLTensor`?

bbernhar: Thanks for taking a look, Reilly.

Aside from the restricted usage, `MLConstantTensor` and `MLTensor` are conceptually identical, sharing the exact same properties and methods. Redefining them seemed unnecessary given this overlap, but if we now believe that TAG or others might not approve of the IDL, I'm open to revisiting it.

index.bs
bbernhar: Fixed.

index.bs
reillyeon: I agree they are conceptually identical, but if they can't actually be used with the same APIs then it will be confusing to developers. It also adds complexity to the specification because every API that accepts an `MLTensor` has to have a check which blocks `MLConstantTensor`.

I think we can get the same benefits of reducing repetition in algorithms by defining two mixin interfaces:

```webidl
interface mixin MLDisposableMixin {
  undefined destroy();
};

interface mixin MLOperandMixin {
  readonly attribute MLOperandDataType dataType;
  readonly attribute FrozenArray<unsigned long> shape;
};
```

`MLOperand`, `MLTensor` and `MLConstantTensor` would include `MLOperandMixin`. We could rename `MLConstantTensor` to simply `MLConstant`.

`MLContext`, `MLGraph`, `MLGraphBuilder`, `MLTensor` and `MLConstantTensor` would include `MLDisposableMixin`, which would also define an internal slot for `[[isContextDestroyed]]`.

bbernhar: As spec'd, `readTensor()` accepts an `MLConstantTensor`, allowing it to be used like any other `MLTensor`. The existing tensor APIs (i.e., readTensor or writeTensor) are already restricted by usage, so there is really only one new check here—for dispatch. I also don't foresee a situation where we will need more than `constant()` to use them.

I agree mixins could help reduce the broader overlap between operands and tensors (and ML objects) and I would be happy to create a separate proposal to refactor them, as a follow-up.

Is what has been spec'd here workable for now?

reillyeon: Is there a need for `readTensor()` to work for `MLConstantTensor`?

I'm okay with the specification as written as long as the refactoring work makes `MLConstantTensor` no longer inherit from `MLTensor`.

bbernhar: > Is there a need for readTensor() to work for MLConstantTensor?

Yes. We use it for debugging, where we get NaNs or 0s back and can't determine whether the contents were written correctly or if the data pulled from the ONNX file was incorrect. If Ort decides to discard the original JS/WASM copy, using this becomes even more difficult. Coming from WebGPU, which gives a read-only usage to buffers but can read-back, (e.g. uniform), it would be confusing.

reillyeon: Isn't that sort of debugging better handled by browser-side tools which let you inspect the WebNN graph or dump it in the platform-specific format? I think I'm having trouble imagining the debugging scenario you have. Is this something you expect web developers (or framework developers) to need to do or is it something we as browser engineers have to do to debug our own implementations?

fdwr: > Since an MLConstantTensor cannot be used with dispatch() or writeTensor()

Hmm, I get why a readonly tensor would be invalid with `writeTensor` (just as a normal `MLTensor` with `writeable` false would also be invalid, and so a constant tensor being unuseable with `writeTensor` is a moot point), but why couldn't we make it valid with `dispatch` (even if it's not its main intention)?

🤔 Actually, does it even need to be its own interface, rather than just an `MLTensor` that has `writeable` permanently false and a property of constness? It seems silly to duplicate all these properties when they are conceptually the same thing, and it makes calling code tougher to pass these around. Constness is a property of the tensor just as much as `dataType` and `shape` are, but it doesn't feel special enough to warrant its own completely separate interface name (adding to cognitive load of additional vocabulary in the API) which adds no additional methods or properties. If anything, I could see just an additional `readonly attribute boolean constant` that's true when created as a constant tensor (and false for a dynamic tensor).

> It also adds complexity to the specification because every API that accepts an MLTensor has to have a check which blocks MLConstantTensor.

Yeah, I wouldn't want that complexity either, but if the scope is just `writeTensor` (and everywhere else that accepts a dynamic tensor also accepts an even more-constant-tensor), then I'd be quite comfortable.

> Isn't that sort of debugging better handled by browser-side tools

It's helpful to read back for display on the webpage itself. e.g. This little test app that shows the values (where I wouldn't care whether the tensor was created as constant, nor *would* I want to care since it's not relevant to the task of debugging them):

image

index.bs
```diff
@@ -1649,6 +1721,9 @@ interface MLGraphBuilder {
   // Create a scalar operand from the specified number of the specified type.
   MLOperand constant(MLOperandDataType type, MLNumber value);
 
+  // Create an operand from a specified tensor.
+  MLOperand constant(MLConstantTensor tensor);
```
fdwr, line 1725: Ah, so I see one place where having a named `MLConstantTensor` specifically makes the function parameter requirement clearer. (resolve me)

fdwr (approved):
👍

reillyeon:
Thank you for the discussion here. I'm starting to come around to the idea of constness being just another property of an `MLTensor` like `readable` or `writable` and the debugging story is compelling. This does make me wonder as Dwayne did about whether it's worthwhile to have a separate interface name. After all, we have similar requirements on `readTensor()` and `writeTensor()` for the `readable` and `writable` properties without creating additional types.

index.bs
bbernhar: @reillyeon 

> Isn't that sort of debugging better handled by browser-side tools which let you inspect the WebNN graph or dump it in the platform-specific format?

Dumping NPU or GPU buffers from the browser is difficult (or likely not possible) without a native tool, which is no better than exposing the existing capability through `readTensor()`.

@fdwr 

> but Bryan, why couldn't we make it valid with dispatch (even if it's not its main intention)?

It could be made valid, I would agree.

reillyeon: > Dumping NPU or GPU buffers from the browser is difficult (or likely not possible) without a native tool[.]

This is true, but I don't want to constrain implementations by requiring that they be able to read back NPU or GPU buffers unless there's a compelling reason to support it. It's also important to remember that the data read back by `readTensor()` might not even be the NPU or GPU buffer used for inference because compiling the model may repack the tensor into an implementation-defined format.

index.bs
```diff
@@ -939,6 +941,7 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
   </summary>
     1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
+        1. If |tensor| is a {{MLConstantTensor}} then return false.
```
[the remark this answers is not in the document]
bbernhar: Yes. I see this being allowed for inputs.

index.bs
```diff
@@ -869,6 +869,8 @@ interface MLContext {
   undefined dispatch(MLGraph graph, MLNamedTensors inputs, MLNamedTensors outputs);
 
   Promise<MLTensor> createTensor(MLTensorDescriptor descriptor);
+  Promise<MLConstantTensor> createConstantTensor(
+    MLOperandDescriptor descriptor, AllowSharedBufferSource inputData);
```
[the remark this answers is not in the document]
bbernhar: > Is there any perf advantage to being able to specify the readability of the tensor, and being able to say it's not readable?

Not that I know of.

index.bs
bbernhar: So you're saying that if implementations can only store weights in some implementation-defined format, it would be difficult to support `readTensor()`? Is this an existing problem? Could you please elaborate?

reillyeon: This is an existing problem with at least DirectML and Core ML:

**DirectML:** When a constant buffer is bound the `DML_TENSOR_FLAG_OWNED_BY_DML` flag is set. From MSDN: "The effect of this flag is that DirectML makes a copy of the tensor data during initialization of an operator, storing it in the persistent resource. This allows DirectML to perform reformatting of the tensor data into other, more efficient forms."

**Core ML:** Constants are written into a weights file as part of a `.mlmodel` package which is then compiled into a `.mlmodelc` package. The `.mlmodel` package is deleted to save space. The format of the `.mlmodelc` package is undocumented and so the weights cannot be recovered from it.

This means that in both cases we must retain our own original copy of the constant tensor in order to implement `readTensor()` for the lifetime of the `MLContext`.

fdwr: So then a constant tensor may imply that *both* `readable` and `writeable` are both false.

index.bs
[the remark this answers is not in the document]
fdwr: Reilly's comment is interesting, that it may not always be *possible* to read it back (short of making a separate copy). 🤔

index.bs
bbernhar: Yes. But note, DirectML still retains an original copy (ex. source buffer) to avoid re-creating them between builds, not specific to `readBuffer()`.

I would be OK making `readable = false`.

huningxin: @reillyeon 

> This means that in both cases we must retain our own original copy of the constant tensor in order to implement `readTensor()` for the lifetime of the `MLContext`.

I suppose even if not to implement `readTensor()`, we may still need to retain the original copy of the constant tensor, because they may be used for building another graph.

@fdwr 

> Well, at least that plays nicely with the existing flags conceptually (so any failure to work with `readTensor` or a `dispatch` input is a natural consequence that `readable == false` rather than it being a special case check).

[`readable`](https://www.w3.org/TR/webnn/#dom-mltensordescriptor-readable) only determines whether the tensor’s contents can be read via `readTensor()`.

[`dispatch()` steps](https://www.w3.org/TR/webnn/#api-mlcontext-dispatch) don't check whether an input tensor is readable or writable.

index.bs
```diff
@@ -1817,7 +1914,7 @@ Build a composed graph up to a given output operand into a computational graph a
         1. Set |graph|.{{MLGraph/[[outputDescriptors]]}}[|name|] to |operand|.{{MLOperand/[[descriptor]]}}.
     1. Set [=this=].{{MLGraphBuilder/[[hasBuilt]]}} to true.
     1. Let |promise| be [=a new promise=] in |realm|.
-    1. Run the following steps [=in parallel=]:
+    1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
```
huningxin, line 1917: Because now a constant operand value may be a constant tensor, the `build()` steps should check whether a constant tensor is destroyed or not before running the following steps in the context timeline.

index.bs
```diff
@@ -1817,7 +1914,7 @@ Build a composed graph up to a given output operand into a computational graph a
...
         1. Run these steps, but [=/abort when=] |graph|.{{MLGraph/[[context]]}} [=MLContext/is lost=]:
             1. Let |graphImpl| be the result of converting [=this=]'s [=MLGraphBuilder/graph=] with |operands|, |operators|, |inputs|, and |outputs|'s [=map/values=] into an [=implementation-defined=] format which can be interpreted by the underlying platform.
```
huningxin, line 1919: It's not clear to me whether the `graphImpl` will make a copy of the content tensor data or reuse it for inference. If it is the latter, the `dispatch()` steps may need to check whether a constant tensor is destroyed or not.

bbernhar: As proposed, `dispatch()` cannot accept a constant tensor so we should be good for now.

index.bs
bbernhar: Good idea. I assume we'll need a separate slot to track. PTAL again.

index.bs
reillyeon: > [W]e may still need to retain the original copy of the constant tensor, because they may be used for building another graph.

My hope is that by making it unreadable we can let the underlying implementation repack the data and potentially avoid keeping multiple copies. I don't think that's possible with any of the frameworks we've prototyped with but it seems plausible.

bbernhar: > My hope is that by making it unreadable we can let the underlying implementation repack the data and potentially avoid keeping multiple copies.

I believe this proposal still achieves that—it does not read or share the packed weight data. Only the original constant data, which was already readable, is retained so that it can be managed by `MLTensor` instead of an ArrayBuffer. This eliminates one copy for DirectML and prevents additional copies for CoreML, assuming the web developer does not retain the source buffer.

reillyeon: Ideally we wouldn't even need to retain the original constant data and could tell the underlying framework "please use that constant that was repacked for this other graph". The framework could then decide whether it was more efficient to keep two copies or compromise between the two graphs for how the data is laid out in memory.

bbernhar: > The framework could then decide whether it was more efficient to keep two copies or compromise between the two graphs for how the data is laid out in memory.

We could encounter 'works on my device' complaints, as implementation-retained copies of the original data still contribute to the framework's overall memory usage. It would be safer, I think, to assume that the original constant data could be retained rather than expecting it cannot, since we know that not all backends support sharing packed data between graph builds (e.g., DML).

reillyeon: I agree that it is safer for frameworks to assume that the data might be retained. I'm only advocating for not specifying the API in a way that implementations must retain it.

index.bs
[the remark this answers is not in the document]
fdwr: > Thank you for the discussion here. I'm starting to come around to the idea of constness being just another property of an MLTensor like readable or writable and the debugging story is compelling. This does make me wonder as Dwayne did about whether it's worthwhile to have a separate interface name. After all, we have similar requirements on readTensor() and writeTensor() for the readable and writable properties without creating additional types.

How do you feel about:

```diff
[SecureContext, Exposed=(Window, Worker)]
interface MLTensor {
  readonly attribute MLOperandDataType dataType;
  readonly attribute FrozenArray<unsigned long> shape;
  readonly attribute boolean readable;
  readonly attribute boolean writable;
+ readonly attribute boolean constant; // True when called via createConstantTensor.

  undefined destroy();
};
```

index.bs
huningxin: I didn't mean user pass constant tensor as dispatch input. I meant, for inference, the compiled graph may still use the constant tensor data that is passed at graph build time.

bbernhar: > I meant, for inference, the compiled graph may still use the constant tensor data that is passed at graph build time.

It is legal to destroy a "constant tensor" after build so the compiled graph must keep its own copy. Should we clarify this behavior as non-normative?

reillyeon: There should be a normative description of how different orderings of calls to `builder.build()` and `constant_tensor.destroy()` should behave, similar to what we've done for using an `MLTensor` with `dispatch()`.

bbernhar: @reillyeon We already have a normative description in `build()` stating that a constant tensor cannot be destroyed. Since it's an MLTensor, it can be destroyed at any point afterward. We could clarify—perhaps in a note—that the graph maintains its own copy, but I'm not sure what other spec changes are needed.

reillyeon: I think that might be enough, because the build process occurs as a single set of synchronous steps on the context timeline we don't need to worry about the tensor being destroyed while those steps are running since tensor destruction also happens on the context timeline.

bbernhar: @huningxin Can we close?

index.bs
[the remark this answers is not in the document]
bbernhar: @huningxin @reillyeon @fdwr 

Are we now in consensus that `MLTensor` may be "constant" instead of having its own type, and that it would be created as `readable = false`? This should address Reilly's opens. Since read-back isn't as critical, I'm fine with that.

👍= okay with this

index.bs
huningxin: @bbernhar 

> It is legal to destroy a "constant tensor" after build so the compiled graph must keep its own copy. Should we clarify this behavior as non-normative?

I suppose we should make this behavior to be normative at the build step. e.g. [constant(descriptor, buffer)](https://www.w3.org/TR/webnn/#dom-mlgraphbuilder-constant) copies the bytes of the buffer source.

index.bs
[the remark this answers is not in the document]
bbernhar: @huningxin Still need your 👍

index.bs
reillyeon: We already have to resolve this question for `constant(buffer)`, where the normatively defined behavior is that it must be indistinguishable from making a copy of the constant at the moment `constant()` was called. Implementations don't have to actually make a copy, they could use reference counting, copy-on-write or any number of fancier options as long as it is indistinguishable from the specified behavior.

index.bs
bbernhar: Resolving due to https://github.com/webmachinelearning/webnn/pull/830#discussion_r2045387429.

index.bs
```diff
@@ -1619,6 +1620,9 @@ interface MLTensor {
     : <dfn>\[[data]]</dfn> of an [=implementation-defined=] type
     ::
         The bytes backing the {{MLTensor}}. This data may only be accessed or modified from the {{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}.
+    : <dfn>\[[isConstant]]</dfn> of type {{boolean}}
```
fdwr, line 1623:
```suggestion

    : <dfn>\[[isConstant]]</dfn> of type {{boolean}}
```
nit - blank line before `dfn`, consistent with lines above.

index.bs
```diff
@@ -939,6 +941,7 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
   </summary>
     1. If |namedTensors|'s [=map/size=] is not equal to |namedDescriptors|'s [=map/size=], then return false.
     1. [=map/For each=] |name| → |tensor| of |namedTensors|:
+        1. If |tensor|.{{MLTensor/[[isConstant]]}} is true then return false.
```
fdwr, line 944:
```suggestion
        1. If |tensor|.{{MLTensor/[[isConstant]]}} is true, then return false.
```
Comma before `then`.

fdwr (approved):
👍

index.bs
```diff
@@ -1749,6 +1825,30 @@ Create a constant {{MLOperand}} of the specified data type and shape that contai
...
+    **Arguments:**
+        - <dfn>tensor</dfn>: an {{MLTensor}}. The constant tensor containing the initialized data.
+    **Returns:** an {{MLOperand}}. The constant output tensor.
+</div>
+
+<details open algorithm>
+  <summary>
+    The <dfn method for=MLGraphBuilder>constant(|tensor|)</dfn> method steps are:
+  </summary>
+    1. If |tensor|.{{MLTensor/[[context]]}} is not [=this=], then [=exception/throw=] a {{TypeError}}.
```
reillyeon, line 1841:
```suggestion
    1. If |tensor|.{{MLTensor/[[context]]}} is not [=this=].{{MLGraphBuilder/[[context]]}}, then [=exception/throw=] a {{TypeError}}.
```

reillyeon (approved):

index.bs
```diff
@@ -1627,6 +1678,28 @@ Releases the resources associated with the {{MLTensor}}. This method is idempote
 
 Note: Since no further operations can be enqueued using this tensor, implementations can free any additional resource allocations associated with this tensor once all previously submitted operations using it are complete.
 
+### Creating a constant {{MLTensor}} ### {#api-mlconstanttensor-create}
```
inexorabletash, line 1681: This ID references the obsolete "MLConstantTensor" idea. The non-constant algorithm uses "api-mltensor-create" so how about:

```suggestion
### Creating a constant {{MLTensor}} ### {#api-constant-mltensor-create}
```

inexorabletash (approved):
One tiny nit

index.bs
huningxin: Thanks for the explanation. It would be helpful by adding a non-normative note explaining the constant tensor can be destroyed after build, implementations need to ensure that won't change the built MLGraph (MLGraph is immutable) and break the dispatch.

bbernhar: Done.

bbernhar:
@fdwr Feel free to merge when ready. Thanks reviewers for the help.

fdwr (approved):
Looks fine to me. Deferring to Ningxin for final review, approval, and merging.

huningxin (approved):
LGTM

PR #831, opened 2025-03-21
Tooling: Use WebIDL2 parser when linting to identify dictionaries

inexorabletash:
Dictionary types need special treatment in a lint rule verifying that realms are used correctly. This requires knowing which types are dictionaries vs. interfaces. This was previously hard-coded, which is a maintenance burden.

Improve this by pulling in the webidl2.js library and using it to parse (and validate!) the spec's WebIDL, which gives us a list of dictionary types.

### DIFF
--- a/tools/lint.mjs
+++ b/tools/lint.mjs
@@ -21,6 +21,7 @@
 'use strict';
 import fs from 'node:fs/promises';
 import {parse} from 'node-html-parser';
+import * as idl from 'webidl2';
 
 // --------------------------------------------------
 // Process options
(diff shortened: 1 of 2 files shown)


inexorabletash:
@huningxin and/or @anssiko - could you review and merge if it looks good? 

This will help with landing #830 - see [this note](https://github.com/webmachinelearning/webnn/pull/830#discussion_r2006543524)

anssiko (approved):
LGTM! I’ll merge so you get to put this into production with #830

PR #834, opened 2025-04-03
Editorial: Miscellaneous cleanups

inexorabletash:
- Don't explicitly "return undefined" from methods; it's implied. The exception would be methods that conditionally return undefined or another value, but WebNN doesn't have any of those. Added lint rule.

- Avoid "not greater than 0"; just "not 0" (if unsigned), or "less than or equal to 0" (if signed). Added lint rule.

- "If" must be followed by ", then". Added lint rule.

- "Otherwise" must be followed by "," or ":". Added lint rule.

- More consistently refer to a list's items instead of values, etc.

- Linkfy references to a list's size.

- Ensure algorithms don't both throw and reject.

### DIFF
--- a/tools/lint.mjs
+++ b/tools/lint.mjs
@@ -223,6 +223,9 @@ for (const match of text.matchAll(/ not the same as /g)) {
 for (const match of text.matchAll(/\bthe \S+ argument\b/g)) {
   error(`Drop 'the' and 'argument': ${format(match)}`);
 }
+for (const match of text.matchAll(/\bnot greater\b/g)) {
+  error(`Prefer "less" to "not greater" (etc):  ${format(match)}`);
+}
 
 // [WebNN] Look for incorrect use of shape for an MLOperandDescriptor
 for (const match of source.matchAll(/(\|\w*desc\w*\|)'s \[=MLOperand\/shape=\]/ig)) {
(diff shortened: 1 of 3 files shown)


inexorabletash:
@huningxin and @fdwr - could you please review?

The _"not greater than 0"_ wording change may be controversial; there's an argument to be made in favor of that phrasing, so I'm happy to change.

tools/lint.mjs
```diff
@@ -303,6 +306,33 @@ for (const v of root.querySelectorAll('var').filter(v => !algorithmVars.has(v)))
...
+  const terms = {
+    throw: 'exception',
+    resolve: 'promise',
+    resolved: 'promise',
+    reject: 'promise',
+    rejected: 'promise',
+  };
+  const other = {
+    exception: 'promise',
+    promise: 'exception',
```
fdwr, line 321: At first I thought these were backwards, but that's the intent.

fdwr (approved):
👍

tools/lint.mjs
```diff
@@ -223,6 +223,9 @@ for (const match of text.matchAll(/ not the same as /g)) {
 for (const match of text.matchAll(/\bthe \S+ argument\b/g)) {
   error(`Drop 'the' and 'argument': ${format(match)}`);
 }
+for (const match of text.matchAll(/\bnot greater\b/g)) {
+  error(`Prefer "less" to "not greater" (etc):  ${format(match)}`);
```
huningxin, line 227: Nit: "less or equal" for completeness?

huningxin (approved):
LGTM with a nit

tools/lint.mjs
inexorabletash: Done in 503e985

inexorabletash:
Thanks, editors! Please hit the "squash and merge" button when you can.

anssiko:
@inexorabletash thanks! Happened to be here, so picked up this one.

PR #843, opened 2025-04-29
Restrict padding options to better match backend limits

### LINKED ISSUE #377: Need clarify constraint of 'beginningPadding' and 'endingPadding' for pad operation in "reflection" and "symmetric" mode
When I implemented `pad` operation for WebNN-Baseline https://github.com/webmachinelearning/webnn-baseline/pull/47, I found there would be missing constraints of 'beginningPadding' and 'endingPadding' in "reflection" and "symmetric" mode, I propose to add constraints likes below table.  @huningxin @wchao1115 @fdwr PTAL, thanks.

|padding\mode|reflection|symmetric|
|--|--|--|
|beginningPadding| beginningPadding[d] < input size of dimension d | beginningPadding[d] <= input size of dimension d|
|endingPadding| endingPadding[d] < input size of dimension d | endingPadding[d] <= input size of dimension d|

Reference: 

1. [DML_PADDING_OPERATOR_DESC](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_padding_operator_desc)
2. [tf.mirrorPad](https://js.tensorflow.org/api/latest/#mirrorPad)

### LINKED ISSUE #739: Limited support for pad on CoreML backend
`pad` is [implemented](https://chromium-review.googlesource.com/c/chromium/src/+/5727184) on CoreML backend but with lots of constraints.
On CoreML (See [doc](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.tensor_operation.pad)), three modes are supported - constant, reflect, replicate(maps to `edge` mode).

1. symmetric mode is not supported (On tflite backend, `edge` mode is not supported.)
2. padding for more than the last two dimensions only supports 'constant' mode. (not documented, but it errors out with this message). One of the webnn samples - [Fast style transfer](https://webmachinelearning.github.io/webnn-samples/style_transfer/) need to use >2D padding for reflect mode.
3. If mode is “reflect” then beginning and ending paddings can be at most input size-1 (same as tensorflow).
4. If mode is “replicate” (aka `edge`) then beginning and ending paddings can be at most input size.

For the third constraint, we can probably update the spec to add such constraint? #377 

For the others, my first question is - @fdwr @huningxin  are there reasonable emulations for these unsupported modes and rank limits?

Further questions are - @mwyrzykowski do you think if CoreML can improve pad support with:
- Add support for `symmetric` mode
- Support padding for more than two dimensions for none `constant` mode.
- Remove the padding value constraints for 'replicate' modes since it doesn't make much sense, it only uses the edge values to replicate.

If none of the restrictions can be emulated / nor supported by CoreML, we will need to either:
a. Add these constraints to WebNN API.
b. Expose the modes limits through `opSupportLimits`. The rank constraints are tricky to expose - as it varies by mode.
c. Just let them fail on CoreML with async error messages.. - Not ideal for browser compatibility.

inexorabletash:
Per discussion in the linked issues and working group, drop "symmetric" padding since it is not widely supported by backends, isn't used by notable models, and can be emulated. Also for non-constant padding, backends typically impose restrictions on the amount of padding to reasonable values based on the input size, so incorporate those limits too.

Resolves #377. Resolves #739.

***
<a href="https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/843.html" title="Last updated on Apr 29, 2025, 11:31 PM UTC (9129a11)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/843/b152d46...inexorabletash:9129a11.html" title="Last updated on Apr 29, 2025, 11:31 PM UTC (9129a11)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7338,6 +7338,7 @@ partial dictionary MLOpSupportLimits {
             <dl class=switch>
                 : {{MLPaddingMode/"constant"}}
                 :: Do nothing.
+                
                 : {{MLPaddingMode/"edge"}}
                 ::
                     1. If |beginningPadding|[|index|] is greater than |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
--- a/index.bs
+++ b/index.bs
@@ -7253,8 +7253,7 @@ Inflate the tensor with constant or mirrored values on the edges.
 enum MLPaddingMode {
   "constant",
   "edge",
-  "reflection",
-  "symmetric"
+  "reflection"
 };
 
 dictionary MLPadOptions : MLOperatorOptions {
@@ -7334,7 +7333,23 @@ partial dictionary MLOpSupportLimits {
     1. If |beginningPadding|'s [=list/size=] and |endingPadding|'s [=list/size=] are not both equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be a copy of |input|.{{MLOperand/[[descriptor]]}}.
     1. Let |outputShape| be a copy of |input|'s [=MLOperand/shape=].
-    1. [=list/For each=] |index| in [=the range=] 0 to |outputShape|'s [=MLOperand/rank=], exclusive:
+    1. [=list/For each=] |index| in [=the exclusive range|the range=] 0 to |outputShape|'s [=MLOperand/rank=], exclusive:
+        1. Switch on |options|.{{MLPadOptions/mode}}:
+            <dl class=switch>
+                : {{MLPaddingMode/"constant"}}
+                :: Do nothing.
+                : {{MLPaddingMode/"edge"}}
+                ::
+                    1. If |beginningPadding|[|index|] is greater than |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
+                    1. If |endingPadding|[|index|] is greater than |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
+
+                    Issue: Restricting {{MLPaddingMode/"edge"}} was not discussed fully with the WG. Should we?
+
+                : {{MLPaddingMode/"reflection"}}
+                ::
+                    1. If |beginningPadding|[|index|] is greater than or equal to |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
+                    1. If |endingPadding|[|index|] is greater than or equal to |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
+            </dl>
         1. Add to |outputShape|[|index|] the value of |beginningPadding|[|index|].
         1. Add to |outputShape|[|index|] the value of |endingPadding|[|index|].
     1. If any [=list/item=] in |outputShape| is not a [=valid dimension=], then [=exception/throw=] a {{TypeError}}.
@@ -7352,7 +7367,7 @@ partial dictionary MLOpSupportLimits {
 <div class="example">
 <details open>
   <summary>
-    Examples for constant, edge, reflection and symmetric padding:
+    Examples for constant, edge, and reflection padding:
   </summary>
   <pre highlight="js">
     // input: [[1,2,3], [4,5,6]]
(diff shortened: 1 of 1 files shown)


inexorabletash:
Notes:

-  This introduces restrictions for "edge" which as noted by @huningxin [here](https://github.com/webmachinelearning/webnn/issues/739#issuecomment-2834127226) wasn't discussed in the WG. Thoughts?
- This does not touch the prose chunk of the [pad spec](https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/843.html#api-mlgraphbuilder-pad). Do we want a table like the one @BruceDai sketched out in the [initial issue](https://github.com/webmachinelearning/webnn/issues/377#issue-1661725334), add text to the "Arguments" definitions of [beginningPadding](https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/843.html#dom-mlgraphbuilder-pad-input-beginningpadding-endingpadding-options-beginningpadding)/[endingPadding](https://pr-preview.s3.amazonaws.com/inexorabletash/webnn/pull/843.html#dom-mlgraphbuilder-pad-input-beginningpadding-endingpadding-options-endingpadding), or ...?

index.bs
```diff
@@ -7334,7 +7333,23 @@ partial dictionary MLOpSupportLimits {
...
+        1. Switch on |options|.{{MLPadOptions/mode}}:
+            <dl class=switch>
+                : {{MLPaddingMode/"constant"}}
+                :: Do nothing.
+                : {{MLPaddingMode/"edge"}}
+                ::
+                    1. If |beginningPadding|[|index|] is greater than |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
+                    1. If |endingPadding|[|index|] is greater than |outputShape|[|index|], then [=exception/throw=] a {{TypeError}}.
+
+                    Issue: Restricting {{MLPaddingMode/"edge"}} was not discussed fully with the WG. Should we?
```
fdwr, line 7346: > Should we?

Well if CoreML was the one odd out of the lot, I'd recommend just emulating those cases where padding size was larger (it seems an odd limitation anyway, unlike with reflection where I understand the implementation reasoning) rather than complicating the spec with each mode having its own valid extents, where fill has none, edge replication has `<=`, and edge-exclusive reflection (subtly different from edge replication) has `<` limits 😶. I see that ORT (using CPU and DML EP) and DirectML have no size limitations for any padding mode {constant fill, edge replication, reflection¹, wrap}. Then TF doesn't have even edge replication at all, requiring emulation anyway. So it's not a clear choice. 🤔

¹ there is no *size-related* error in ORT, but evidently some execution providers don't fill in all the values beyond the reflected size, for which I'll need to file a bug. 🐞

fdwr (approved):
Thank you Joshua.

If we need to restore the edge-inclusive reflection¹, then we can easily construct it from two pad calls, just inflating by 1 replicated tensel and then using edge-exclusive reflection: `pad(pad(input, [1,1,1,1...], "replicate"), paddingMinusOne, "reflect")`.

¹ I'm starting to refer to it as this, since "reflection" and "symmetric" alone are less clear, as I always forget which "reflection" is edge-inclusive (standard behavior for textures in graphics API's) vs which reflection is edge-exclusive (the more common one in ML).

index.bs
```diff
@@ -7334,7 +7333,23 @@ partial dictionary MLOpSupportLimits {
     1. If |beginningPadding|'s [=list/size=] and |endingPadding|'s [=list/size=] are not both equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be a copy of |input|.{{MLOperand/[[descriptor]]}}.
     1. Let |outputShape| be a copy of |input|'s [=MLOperand/shape=].
-    1. [=list/For each=] |index| in [=the range=] 0 to |outputShape|'s [=MLOperand/rank=], exclusive:
+    1. [=list/For each=] |index| in [=the exclusive range|the range=] 0 to |outputShape|'s [=MLOperand/rank=], exclusive:
+        1. Switch on |options|.{{MLPadOptions/mode}}:
+            <dl class=switch>
+                : {{MLPaddingMode/"constant"}}
+                :: Do nothing.
```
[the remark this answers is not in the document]
inexorabletash: I'll also add that I wasn't sure how to write a "no-op" entry in the switch statement. https://fetch.spec.whatwg.org/ uses "Do nothing." so that's the precedent I followed.

fdwr (approved):
👍

index.bs
huningxin: As discussed in https://github.com/webmachinelearning/webnn/issues/739#issuecomment-2840913060, the only one edge pad of all transformers.js models has the beginning padding size (2) larger than input size (1).

fdwr (approved):
Removing edge replication size restriction LGTM.

huningxin (approved):
LGTM!

PR #847, opened 2025-05-14
Update WebNN architecture diagram

### LINKED ISSUE #840: Clarify WebNN explainer and update diagram
From [the image from the explainer](https://github.com/webmachinelearning/webnn/blob/main/content/webnn_arch.png) (which is also pasted at the bottom) I understand that the API would be a set of new commands to the user under `navigator.ml` to create neural networks. And the browsers will implement the backends and delegate ops to them.

> web browsers may implement the WebNN API using native machine learning API available in the operating system.

On the [polyfill proof of concept](https://github.com/webmachinelearning/webnn-polyfill) you basically add Tensorflow into the browser, if I understand correctly.

Two questions I have are:

1. Will the backend be tensorflow (TF), delegating to 'cpu','webgl', 'webgpu'? From the polyfill, it does not seem to be a proof of concept for your "native backends" but just for those workarounds found by TF.
2. I suspect that this isn't the case, since TF does not delegate to Native APIs. But then, the native APIs you include there are not always installed, for example, I do not have OPENVINO installed on Linux, unless it's under a name I didn't recognise. So, the image below, it seems that one can not run the application, even though TF would (by the CPU, or WEBGL, or even WEBGPU backend in some browsers.) What happens here then? 
3. Just a side comment but I did paste your explainef sample-code in the Dev Console, after enabling the flags and restarting the browser, and it fails with the error below:

```
VM55:1 Uncaught TypeError: Failed to construct 'MLGraphBuilder': parameter 1 is not of type 'MLContext'.  at <anonymous>:1:17
```

Image

anssiko:
Fix #840


explainer.md
```diff
@@ -22,7 +22,52 @@ The design process of the [Web Neural Network API](https://www.w3.org/TR/webnn/)
 
 With emerging ML innovations in both software and hardware ecosystem, one of the main challenges for the web is to bridge this software and hardware development and bring together a solution that scales across hardware platforms and works with any framework for web-based machine learning experiences. We propose the WebNN API as an abstraction for neural networks in the web browsers.
 
-![WebNN architecture](content/webnn_arch.png)
```
inexorabletash, line 25: Delete the PNG resource or keep it?

explainer.md
```diff
@@ -22,7 +22,52 @@ The design process of the [Web Neural Network API](https://www.w3.org/TR/webnn/)
...
+```mermaid
+%%{ init : { "theme" : "neutral" }}%%
+flowchart TD
+
+  style Models stroke-width:0, fill:#fff
```
inexorabletash, line 29: With the hard coded white fill the page looks awkward but not unreadable in Dark Mode:

image

I'm not familiar enough with mermaid to know if there's an easy fix.

inexorabletash (approved):
LGTM with a couple observations.

explainer.md
reillyeon: In theory you should be able to inject some CSS which applies to the whole background. As is I'm concerned this will be confusing.

explainer.md
anssiko: I removed this. We can generate a new one when settled so folks who may reference this file won't get 404.

https://github.com/webmachinelearning/webnn/pull/847/commits/99b42f5700a82f13ac0981fe03a53e3cab7da3fe

explainer.md
anssiko: It looks like there's one combo (`"look" : "handDrawn", "theme" : "neo-dark"`) that produces a rather readable flowchart in dark mode. Based on Mermaid issue discussion, dark mode support has limitations. If someone has played with this more, please share your tips.

https://github.com/webmachinelearning/webnn/pull/847/commits/97bf5cf5364461d8d1b4b9a2a2b4f03afd70b02a

anssiko:
And here's a [preview](https://github.com/webmachinelearning/webnn/blob/arch-diagram/explainer.md) for testing how this renders.

reillyeon (approved):

huningxin (approved):
LGTM with a nit

Issue #815, opened 2025-02-04
Post-context creation query mechanism for supported devices


anssiko:
[ Spun off from PR #809 review [comment](https://github.com/webmachinelearning/webnn/pull/809#discussion_r1936856070) by @fdwr and a discussion with @inexorabletash, @reillyeon, @RafaelCintron, @philloooo -- thanks! ]

(Emphasis mine:)
>This afternoon we were talking with {@inexorabletash, @reillyeon, @RafaelCintron, Phillis} about the likely need for a caller to know whether a particular device is supported or not, because an app may want to (if say GPU is not supported) use a different more performant fallback than for WebNN to silently fall back to CPU. For example, if GPU was unavailable (even though you preferred high performance), then it might be faster to execute the model with WebGPU shaders than WebNN CPU, or it might be okay to use CPU, but the app could load a different model that's more CPU-friendly, if it knew that was the case.
>
>**This need is the reverse of what we currently have, querying the context for what kind of devices it actually supports rather than prescribing a preferred device type.** Now, preferred device type prescription may return someday as a weaker hint (or list of hints, or list of antihints...) as we learn more about usage/needs, but for either case prescription or post-query, having device enums could be useful. 🤔

To kick off this discussion, I'll start with a simple and naive API further abstracted from the `GPUDevice.features` setlike interface:

```js

// only MLPowerPreference supported at context creation time
const context = await navigator.ml.createContext({ powerPreference: 'high-performance' });

// query the newly created context
if (context.devices.has('cpu') && context.devices.has('npu')) {
   // context supports both CPU and NPU, choose the model accordingly
   // how the work is split across CPU and NPU is an implementation detail
} else if (!context.devices.has('gpu')) {
   //  no GPU available, perhaps try WebGPU?
} else {
   // ...
}
```

To help tease out more informed proposals, I'd like folks to test this imaginary API against the following:
- Address [key use cases](https://github.com/webmachinelearning/webnn/blob/main/device-selection-explainer.md#key-use-cases-and-requirements)
- [Design based on user needs, not the underlying capability](https://www.w3.org/TR/design-principles/#usecase-oriented-apis)
- Implementability / [support the full range of devices and platforms](https://www.w3.org/TR/design-principles/#devices-platforms)
- [Use care when exposing APIs for selecting or enumerating devices](https://www.w3.org/TR/design-principles/#device-enumeration)
- Is extensible and [future-proof](https://www.w3.org/TR/design-principles/#naming-future-proofing)
[...]

PR #854, opened 2025-05-30
define graph.devices

### LINKED ISSUE #836: Get devices used for a graph
Hi! This is deriving from discussion here https://github.com/webmachinelearning/webnn/issues/815#issuecomment-2638389869 .  
Eventually which devices are used can only be known after a model is provided. It's a different problem from the context level device query mechanism so I decided to create a separate issue.

I've tested with CoreML's [MLComputePlan](https://developer.apple.com/documentation/coreml/mlcomputeplan-85vdw) which can give you op level device selections after a model is compiled. (Before calling `dispatch`).
Tflite also gives which delegate gets used for each op, after a graph is compiled.

So for WebNN, we can attach such information to MLGraph which represents a compiled graph. 
The first high level information we can attach is a list of `devices` that will be used to execute the graph, since sometimes a graph could be executed using multiple devices.
```
const graph = await builder.build({'C': C});
console.log(graph.devices) // ['cpu', 'npu']
```

We could further expose op level device selection like a map:
```
const graph = await builder.build({'C': C});
console.log(graph.devices) // ['cpu', 'npu']

console.log(graph.deviceForOperations)

{
  "add_1": "cpu",
  "conv2d_2": "npu"
}
```
The only thing is we would need identifiers for each op, we could: 1. auto generate op identifiers using op name + auto incrementing index. 2. Use `label` and only return ops that have labels defined.

Maybe just the high level `graph.devices` is good enough for app developers to make decisions. But I want to provide what's possible with current backends right now.

philloooo:
fixes #836
hi, I've implemented this API on Chromium for both [CoreML backend](https://chromium-review.googlesource.com/c/chromium/src/+/6465656) and [DML backend](https://chromium-review.googlesource.com/c/chromium/src/+/6558473) and  [Tflite backend](https://chromium-review.googlesource.com/c/chromium/src/+/6580116). So I am putting up the spec PR for discussion. 

This re-introduces the MLDeviceType but in the context of a compiled graph. The use case shifts from letting users specify device preference to providing information to developers on where the graph would get run.

So I'd like to hear people's opinion on the naming for 'npu' under this context.

@fdwr @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/webnn/pull/854.html" title="Last updated on Jun 3, 2025, 5:28 PM UTC (8202cfc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/854/ed9b56c...philloooo:8202cfc.html" title="Last updated on Jun 3, 2025, 5:28 PM UTC (8202cfc)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4116,7 +4116,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/dequantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input.
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else the behavior is [=implementation-defined=] (e.g. correct results, incorrect results, or compilation failure).
         - <dfn>zeroPoint</dfn>: an {{MLOperand}}. The zero point tensor to subtract from each input value. It has the same [=MLOperand/shape=] as the scale.
         - <dfn>options</dfn>: an {{MLOperatorOptions}}. Specifies the optional parameters of the operation.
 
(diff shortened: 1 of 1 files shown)


huningxin:
Thanks @philloooo for Chromium prototyping and making this PR! It looks good to me.

>  [DML backend](https://chromium-review.googlesource.com/c/chromium/src/+/6558473), 

As discussed in [WebML WG Teleconference – 22 May 2025](https://www.w3.org/2025/05/22-webmachinelearning-minutes.html#1839), Windows ML was announced as an evolution of DirectML. It would be good to know the implementability of Windows ML (ONNX Runtime).

zolkis:
> So I'd like to hear people's opinion on the naming for 'npu' under this context.

For the record, that has been strongly contested for the overloaded meaning of it in #815, which led to the proposal to use "gpu" or "gpu-like" labels.

However, I think we could experimentally use the _name_ "npu" here -- when we note that it might mean many different HW implementations with wildly different capabilities. However, NPU is gaining more and more the meaning of being power efficient,  low-throughput, moderately low-latency acceleration. We could diversify that naming later.

We should record (as developer use cases) what (programming) benefits could a client get from knowing that a node in a compiled graph is executing e.g. on NPU. As elaborated in this [comment](https://github.com/webmachinelearning/webnn/issues/815#issuecomment-2758753704), it might reflect a past decision made on a temporary condition (i.e. there could be hidden parameters we don't record). 

Anyway, I think this might be a good way to gather data/experience on how typical compiled graphs will look like, which will help improving the spec on the device selection (hints, for now, but later also capabilities ), so this proposal LGTM.

mwyrzykowski (changes requested):
It would be great to see a use case as its not obvious why this is necessary to add to the API

handellm:
@mwyrzykowski in case you didn't see https://github.com/webmachinelearning/webnn/issues/836#issuecomment-3456696119 I added a usecase there.

mwyrzykowski (changes requested):
The use case in https://github.com/webmachinelearning/webnn/issues/836#issuecomment-3456696119 does not require expressing which device a graph is running on. 

A UA or the underlying OS may choose to execute on different processors at any time and attempting to expose this information is at best unreliable and dependent on any number of other processes simultaneously executing on the system.

handellm:
@mwyrzykowski An example: say we have denoising running with graph.devices() equalling ['cpu'], and background blur running with graph.devices() equalling ['gpu']. If blur misbehaves, will it help if adaptation selects to disable denoising? (no)

My thinking is the adaptation would be able to choose based on processing impacting the set of devices used by the misbehaving processing type.

I think "unreliable + dependent on other processes" would be okay as long as the info is fresh, because adaptation happens in conjunction so chances are the decisions will be accurate-ish.

PR #858, opened 2025-06-05
Add `isNaN` and `isInfinite` operators

### LINKED ISSUE #811: Behavior when there are NaNs in argmin/max inputs
Hi!
I have a question re: how backends handle NaNs in the inputs.

I found that for [argmin/argmax](https://www.w3.org/TR/webnn/#api-mlgraphbuilder-argminmax) when there are NaNs in the input, Core ML most of the time returns a wrong min/max index pointing to a number that's not the min/max. On tflite/dml though, they ignore the NaNs and return the correct index pointing to the max/min number.
This might be a bug, or may be working as expected on Core ML? @mwyrzykowski

On the WebNN side, do we think 
a. it's ok to expect the backends behave strangely when there are NaNs? ( if we consider NaNs are unexpected invalid data)
b. or we expect consistent handling across backends for inputs with NaNs ( and spell them out in the spec)? 

@fdwr @huningxin might have some thoughts?

fdwr:
Fixes #811. Since NaN and infinity are closely related concepts, I'm introducing both logical unary operators together for completeness.

```diff
partial interface MLGraphBuilder
{
    ...
+   MLOperand isNaN(MLOperand a, optional MLOperatorOptions options = {})
+   MLOperand isInfinite(MLOperand a, optional MLOperatorOptions options = {})
    ...
};
```

### Is not-a-number:
- numpy.isnan
- tf.math.is_nan
- torch.isnan
- onnx.ai.IsNaN
- DML_OPERATOR_ELEMENT_WISE_IS_NAN
- mil.ops.defs.iOS15.elementwise_binary.not_equal(a, a)

### Is infinite:
- numpy.isinf
- tfm.optimization.math.isinf
- torch.isinf
- onnx.ai.IsInf
- DML_OPERATOR_ELEMENT_WISE_IS_INFINITY
- mil.ops.defs.iOS15.elementwise_unary.abs(..) &&
  mil.ops.defs.iOS15.elementwise_binary.equal(.., Infinity)

## Possible question answers:

- Why name it "`isInfinite`" rather than "`isinf`" like Python API's? This isn't Python, it's closer kin to Javascript's `isFinite` function and `Infinite` constant, and general API naming guidelines favor whole words for readability.
- Why not support specific tests for positive infinity or negative infinity like [ONNX](https://onnx.ai/onnx/operators/onnx__IsInf.html)? I'm not opposed to it, but I'm unsure the use case, most backends don't support it, and one could easily clamp the values before calling `isInfinite` to emulate it.
- Do we really need these functions? No - technically they can be emulated via existing logical operators, but backends may have more optimized forms of them, and their existence leads to better discoverability. One compromise would be to instead document a decomposition in the specification (similar to `flatten` and `unsqueeze`) in the [Operator Emulation](https://webmachinelearning.github.io/webnn/#emulation) section.

Here are some example emulations:

```
function isNaN(value:Number): Boolean
{
    return value != value;
    // low level equivalent: return value.reinterpretAs(uint32) & 0x7FFFFFFF) > 0x7F800000;
}

function isNotNaN(value:Number): Boolean
{
    return value == value;
}

function isNaNOrInfinity(value:Number): Boolean
{
    return (value * 0) != 0;
}

function isNotNaNOrInfinity(value:Number): Boolean
{
    return (value * 0) == 0;
}
```

***
<a href="https://pr-preview.s3.amazonaws.com/fdwr/webnn/pull/858.html" title="Last updated on Aug 9, 2025, 5:23 AM UTC (ddf7cdf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/858/093be92...fdwr:ddf7cdf.html" title="Last updated on Aug 9, 2025, 5:23 AM UTC (ddf7cdf)">Diff</a>


fdwr:
FYI @philloooo.

index.bs
```diff
@@ -3380,10 +3382,12 @@ partial dictionary MLOpSupportLimits {
   MLBinarySupportLimits logicalAnd;
   MLBinarySupportLimits logicalOr;
   MLBinarySupportLimits logicalXor;
+  MLBinarySupportLimits isNaN;
```
huningxin: Use `MLLogicalNotSupportLimits`? Or you may want to rename to something general.

fdwr: Doh yes, like unary logicalNot. `MLLogicalNotSupportLimits` makes sense to me since it does a test and returns a Boolean value, very similar to the others.

index.bs
```diff
@@ -3473,7 +3483,7 @@ Although operations {{MLGraphBuilder/greaterOrEqual()}} and {{MLGraphBuilder/les
   <summary>
     To <dfn for="MLGraphBuilder" data-lt="element-wise-logical-op">create an element-wise logical operation</dfn> given [=string=] |op|, {{MLOperand}} |a|, an optional {{MLOperand}} |b|, and {{MLOperatorOptions}} |options|, run the following steps:
   </summary>
-    1. [=Assert=]: |op| is one of "equal", "notEqual", "greater", "greaterOrEqual", "lesser", "lesserOrEqual", "logicalNot", "logicalAnd", "logicalOr", "logicalXor".
+    1. [=Assert=]: |op| is one of "equal", "notEqual", "greater", "greaterOrEqual", "lesser", "lesserOrEqual", "logicalNot", "logicalAnd", "logicalOr", "logicalXor", "isNaN", "isInfinite".
```
huningxin, line 3668: I suppose `isNan` and `isInfinite` should only support "float32" and "float16" data types. You may want to add step to validate that.

fdwr: True. The alternative is that we take all data types, and for any non-float data types (int32, uint8...), WebNN just returns a tensor of all false 0's, which makes the operator more cleanly generic, but then I don't know any ML library that accepts integers to isNaN anyway, making it unlikely to reach WebNN.

Hmm, I don't like what we currently have where *most* operators visible show their data types in a table, but *some* operators instead bury data types inside algorithmic steps. Maybe I can cleanly promote this up into the existing table instead "Constraints for element-wise logical options"... ⏳

huningxin: > but then I don't know any ML library that accepts integers to isNaN anyway, making it unlikely to reach WebNN.

Right. And the native framework may also reject integers for isNan and isInfinite, such as ONNX Runtime: [IsNaN](https://onnx.ai/onnx/operators/onnx__IsNaN.html), [IsInf](https://onnx.ai/onnx/operators/onnx__IsInf.html)

> Hmm, I don't like what we currently have where most operators visible show their data types in a table, but some operators instead bury data types inside algorithmic steps. Maybe I can cleanly promote this up into the existing table instead "Constraints for element-wise logical options"

That's true. The latter was introduced for op categories with the intention of minimum change, see discussion at original PR: https://github.com/webmachinelearning/webnn/pull/657#issuecomment-2471267357. 

For this PR, I'd recommend to keep the change minimum with two options:

Option 1: Add a step into the "create an element-wise logical operation" algorithm, similar to validation step for "logicalNot":

> 1. If op is one of "isNan" and "isIfinite, then:
>    1. If a ’s dataType is neither "float32" nor "float16 , then throw a TypeError .

Option 2: Add an optional allowedDataTyeps argument into "create an element-wise logical operation" algorithm (similar to "create an element-wise unary operation").

And add a step:

> If allowedDataTypes is given and it does not contain input ’s dataType , then throw a TypeError .

With that, we can call "create an element-wise logical operation" with « "float32" , "float16" » for "isNan" and "IsInfinite". The validation step of `logicalNot` can be removed, we can call "create an element-wise logical operation" with « "uint8" » instead.

fdwr: Went with shorter one for now.

Issue #807, opened 2025-01-22
Caching mechanism for MLGraph


anssiko:
[ Spun off from issue #780 ]

>@reillyeon mentioned to me about introducing a caching mechanism for MLGraph (e.g. save it for later use and avoid repeated graph compilation). Said mechanism might help here.

I'd like to discuss this proposal from @reillyeon a bit more. I believe the group's current working assumption is that graph compilation could take a long time and caching would improve the user experience on subsequent visits to the same site (cross-origin is a harder, separate problem). Depending on the size of the model, underlying implementation and other factors, this could be a significant performance and UX improvement.

@reillyeon have you thought about this more since you came up with the idea? Known implementation blockers?

@bbernhar what could we learn from the WebGPU compilation caches for shaders, pipelines? I see some [toggles in Dawn code](https://source.chromium.org/chromium/chromium/src/+/main:third_party/dawn/src/dawn/native/Toggles.cpp) to control caching and you've done work in this space e.g. in https://issues.chromium.org/issues/41479574 suggesting you might have insights to share.

Also paging @huningxin @fdwr and @RafaelCintron for thoughts. Interested in all insights in the spirit of brainstorming.

A few additional questions:

Do we foresee the caching of compiled graphs to be purely an implementation detail? 

Privacy impact? We already discuss caching-related timing attack vectors in [privacy considerations](https://www.w3.org/TR/webnn/#privacy) and reference the [WebGPU compilation cache considerations](https://gpuweb.github.io/gpuweb/#privacy-user-agent-state). Depending on which way we go, might want to revise these considerations.

reillyeon:
My current thought is that this would be an explicit API:

```js
partial interface MLContext {
  Promise<sequence<DOMString>> listGraphs();
  Promise<MLGraph> loadGraph(DOMString key);
  Promise<undefined> saveGraph(DOMString key, MLGraph graph);
  undefined deleteGraph(DOMString key);
};
```

GPU shader caching is implicit, however the difference is that a shader program is a small input and so it's easy for the site to regenerate the shader so the browser can hash it to compare with the cache. ML models on the other hand are large because of the weights. Loading all the weights just to discover that a cached version of the model is available would be a waste of time and resources.

inexorabletash:
Implicitly, the cache/keys would be partitioned by the origin, same as other storage APIs, which would mitigate privacy concerns. Clearing storage would wipe this cache too. 

[...]

PR #862, opened 2025-06-06
Add WebNN MLGraph Cache Explainer

anssiko:
This synthesis is based on WG discussion in #807, Chromium Design Doc and prototype implementation findings.

With credits to @reillyeon for the API proposal, @shiyi9801 for the prototype implementation, @huningxin for the Chromium Design Doc, and to the entire WG for active issue discussion.

The aim was to try keep this as a ~1-pager, so we can iterate on this more easily.

***
<a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/pull/862.html" title="Last updated on Jun 17, 2025, 9:43 AM UTC (1e1bc1c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/862/61d4d5a...1e1bc1c.html" title="Last updated on Jun 17, 2025, 9:43 AM UTC (1e1bc1c)">Diff</a>

### DIFF
--- a/cache-explainer.md
+++ b/cache-explainer.md
@@ -1,0 +1,104 @@
+# WebNN MLGraph Cache Explainer
+
+## Authors
+
+WebML Working Group participants
+
+## Participate
+
+- https://github.com/webmachinelearning/webnn/issues/807
+
+## Table of contents
+
+1. [Introduction](#introduction)
+1. [Goals](#goals)
+1. [Non-goals](#non-goals)
+1. [User research](#user-research)
+1. [Use cases](#use-cases)
+1. [Proposed API](#proposed-api)
+1. [Considered alternatives](#considered-alternatives)
+1. [Related work](#related-work)
+1. [Privacy and security considerations](#privacy-and-security-considerations)
+1. [References](#references)
+
+## Introduction
+
+The WebNN API enables web applications to perform ML model inference by constructing a graph representation of the model ([`MLGraphBuilder`](https://www.w3.org/TR/webnn/#mlgraphbuilder)), compiling it into a native format ([`MLGraph`](https://www.w3.org/TR/webnn/#mlgraph)), and executing it via [`MLContext.dispatch()`](https://www.w3.org/TR/webnn/#api-mlcontext-dispatch). However, compiling large models on certain devices, such as NPUs, can be time-consuming. To address this, we propose an explicit API for caching compiled graphs, allowing web applications to save and reuse them, thereby reducing the overhead of repeated compilation.
+
+This proposal documents ongoing discussions in the W3C WebML Working Group and builds on existing mechanisms in frameworks like ONNX Runtime.
+
+## Goals
+
+- Provide a mechanism for web applications to save and load compiled `MLGraph` objects.
+- Reduce the time required for repeated ML model inference by avoiding redundant graph compilation.
+- Ensure compatibility with existing WebNN API constructs and workflows.
+
+## Non-goals
+
+- This proposal does not aim to define a universal format for graph serialization across all frameworks.
+- It does not address caching mechanisms for non-WebNN APIs or other types of computational graphs.
+- Cross-origin model sharing is out of scope.
+
+## User research
+
+[If any user research has been conducted to inform the design choices presented,
+discuss the process and findings.
+We strongly encourage that API designers consider conducting user research to
+verify that their designs meet user needs and iterate on them,
+though we understand this is not always feasible.]
+
+## Use cases
+
+### Reduce time to first inference
+
+A web application performing real-time image recognition can save the compiled graph after the first inference. If the page is reloaded, subsequent inferences reuse the cached graph, significantly reducing latency by avoiding both the model redownload and recompilation steps.
+
+## Proposed API
+
+```webidl
+partial interface MLContext {
+  Promise<sequence<DOMString>> listGraphs();
+  Promise<MLGraph> loadGraph(DOMString key);
+  Promise<undefined> saveGraph(DOMString key, MLGraph graph);
+  undefined deleteGraph(DOMString key);
+};
+```
+
+1. **`listGraphs()`**: Returns a list of keys for all cached graphs.
+2. **`loadGraph(key)`**: Loads a cached graph associated with the given key.
+3. **`saveGraph(key, graph)`**: Saves the provided graph under the specified key.
+4. **`deleteGraph(key)`**: Deletes the cached graph associated with the given key.
+
+## Considered alternatives
+
+### Explicit vs implicit API
+
+>GPU shader caching is implicit, however the difference is that a shader program is a small input and so it's easy for the site to regenerate the shader so the browser can hash it to compare with the cache. ML models on the other hand are large because of the weights. Loading all the weights just to discover that a cached version of the model is available would be a waste of time and resources. (via [comment](https://github.com/webmachinelearning/webnn/issues/807#issuecomment-2608135598))
+
+## Related work
+
+### ONNX Runtime EPContext
+
+ONNX Runtime introduced the `EPContext` mechanism to encapsulate compiled blobs into ONNX models. This approach inspired the WebNN caching proposal but is tailored to ONNX-specific workflows.
+
+### WebGPU shader cache
+
+The WebGPU API employs a shader caching mechanism. While similar in concept, it is designed for GPU shaders rather than ML model graphs.
+
+## Privacy and security considerations
+
+### Storage partitioning
+
+To prevent cross-origin data leakage, cached graphs must be partitioned per origin. This ensures that a graph saved by one website cannot be accessed by another.
+
+### Implementation-specific sandbox constraints
+
+Chromium's sandboxing model restricts file system access for GPU processes. Any implementation must comply with these constraints to ensure security.
+
+## References
+
+- [WebNN API Specification](https://github.com/webmachinelearning/webnn)
+- [ONNX Runtime EPContext Design](https://onnxruntime.ai/docs/execution-providers/EP-Context-Design.html#onnxruntime-ep-context-cache-feature-design)
+- [OpenVINO Model Caching Overview](https://docs.openvino.ai/2024/openvino-workflow/running-inference/optimize-inference/optimizing-latency/model-caching-overview.html)
+- [Chromium Sandbox Design](https://chromium.googlesource.com/chromium/src/+/main/docs/design/sandbox.md)
+- [WebGPU Shader Cache](https://docs.google.com/document/d/1CtgsUWTBe6pVEDq3ZksSEc_6eSAqvHZ-h0_zoPu21po/edit?tab=t.0#heading=h.fshi85nj57x0)
--- a/cache-explainer.md
+++ b/cache-explainer.md
@@ -79,8 +79,8 @@ A JS ML framework, such as ONNX Runtime Web, may need to know the input and outp
 
 ```webidl
 partial interface MLGraph {
-  record<USVString, MLOperandDescriptor> inputDescriptors;
-  record<USVString, MLOperandDescriptor> outputDescriptors;
+  record<USVString, MLOperandDescriptor> inputs;
+  record<USVString, MLOperandDescriptor> outputs;
 };
 ```
 
@@ -88,9 +88,9 @@ partial interface MLGraph {
 
 ### Combined build and save
 
-A separate `saveGraph` API might introduce overhead on some native ML frameworks, such as ONNX Runtime, because its implementation may need to hold the source model in the memory and recompile the source model when user code calls `saveGraph`.
+A separate `saveGraph()` API might introduce overhead on some native ML frameworks, such as ONNX Runtime, because its implementation may need to hold the source model in the memory and recompile the source model when user code calls `saveGraph()`.
 
-An alternative consideration is to have a `buildAndSave` method. The implementation can just compile the graph once and drop the source model after the compilation.
+An alternative consideration is to have a `buildAndSave()` method. The implementation can just compile the graph once and drop the source model after the compilation.
 
 ```webidl
 partial interface MLGraphBuilder {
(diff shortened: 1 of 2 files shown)


cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+1. [Use cases](#use-cases)
+1. [Proposed API](#proposed-api)
+1. [Considered alternatives](#considered-alternatives)
+1. [Related work](#related-work)
+1. [Privacy and security considerations](#privacy-and-security-considerations)
+1. [References](#references)
+
+## Introduction
+
+The WebNN API enables web applications to perform ML model inference by constructing a graph representation of the model ([`MLGraphBuilder`](https://www.w3.org/TR/webnn/#mlgraphbuilder)), compiling it into a native format ([`MLGraph`](https://www.w3.org/TR/webnn/#mlgraph)), and executing it via [`MLContext.dispatch()`](https://www.w3.org/TR/webnn/#api-mlcontext-dispatch). However, compiling large models on certain devices, such as NPUs, can be time-consuming. To address this, we propose an explicit API for caching compiled graphs, allowing web applications to save and reuse them, thereby reducing the overhead of repeated compilation.
```
reillyeon: I would say "compiling large models _for_ certain devices" as well as mention that this can be particularly difficult since compilation must happen on potentially slower end-user devices rather than ahead-of-time so that the site doesn't have to provide device-specific binary blobs.

cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+
+[If any user research has been conducted to inform the design choices presented,
+discuss the process and findings.
+We strongly encourage that API designers consider conducting user research to
+verify that their designs meet user needs and iterate on them,
+though we understand this is not always feasible.]
+
+## Use cases
+
+### Reduce time to first inference
```
reillyeon: Somewhat misleading title, since the very first inference will still be blocked by model compilation but first inference on reload will be much faster.

cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+1. **`listGraphs()`**: Returns a list of keys for all cached graphs.
+2. **`loadGraph(key)`**: Loads a cached graph associated with the given key.
+3. **`saveGraph(key, graph)`**: Saves the provided graph under the specified key.
+4. **`deleteGraph(key)`**: Deletes the cached graph associated with the given key.
+
+## Considered alternatives
+
+### Explicit vs implicit API
+
+>GPU shader caching is implicit, however the difference is that a shader program is a small input and so it's easy for the site to regenerate the shader so the browser can hash it to compare with the cache. ML models on the other hand are large because of the weights. Loading all the weights just to discover that a cached version of the model is available would be a waste of time and resources. (via [comment](https://github.com/webmachinelearning/webnn/issues/807#issuecomment-2608135598))
```
reillyeon: Note as well that the model can't be compiled without the weights because the implementation may perform device-specific constant folding and memory layout optimizations.

cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+partial interface MLContext {
+  Promise<sequence<DOMString>> listGraphs();
+  Promise<MLGraph> loadGraph(DOMString key);
+  Promise<undefined> saveGraph(DOMString key, MLGraph graph);
+  undefined deleteGraph(DOMString key);
+};
+```
+
+1. **`listGraphs()`**: Returns a list of keys for all cached graphs.
+2. **`loadGraph(key)`**: Loads a cached graph associated with the given key.
```
reillyeon, line 68: Note that a graph may be evicted from the cache due to storage pressure or browser/platform updates which render previously compiled graphs invalid. Developers should consider the level of durability to be somewhere between IndexedDB and the HTTP cache. I'm not sure if there is a specific concept from the storage specs we could reference here.

cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Privacy and security considerations
+
+### Storage partitioning
+
+To prevent cross-origin data leakage, cached graphs must be partitioned per origin. This ensures that a graph saved by one website cannot be accessed by another.
+
+### Implementation-specific sandbox constraints
+
+Chromium's sandboxing model restricts file system access for GPU processes. Any implementation must comply with these constraints to ensure security.
```
reillyeon: I might make this more generic by saying something like, "for security reasons, model compilation and inference will typically happen in sandboxed processes. This will introduce implementation challenges and care must be taken in how the caching mechanism allows data to be read from and written to disk."

reillyeon (changes requested):

cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+
+A web application performing real-time image recognition can save the compiled graph after the first inference. If the page is reloaded, subsequent inferences reuse the cached graph, significantly reducing latency by avoiding both the model redownload and recompilation steps.
+
+## Proposed API
+
+```webidl
+partial interface MLContext {
+  Promise<sequence<DOMString>> listGraphs();
+  Promise<MLGraph> loadGraph(DOMString key);
+  Promise<undefined> saveGraph(DOMString key, MLGraph graph);
```
huningxin, line 62: A separate `saveGraph` API might introduce overhead on some native ML frameworks, such as ONNX Runtime, because its implementation may need to hold the source model in the memory and recompile the source model when user code calls `saveGraph`.

An alternative consideration is to have a `buildAndSave` method. The implementation can just compile the graph once and drop the source model after the compilation.
```webidl
  Promise<MLGraph> buildAndSave(MLNamedOperands outputs, DOMString key);
};
```

/cc @RafaelCintron

cache-explainer.md
```diff
@@ -0,0 +1,104 @@
...
+### Reduce time to first inference
+
+A web application performing real-time image recognition can save the compiled graph after the first inference. If the page is reloaded, subsequent inferences reuse the cached graph, significantly reducing latency by avoiding both the model redownload and recompilation steps.
+
+## Proposed API
+
+```webidl
+partial interface MLContext {
+  Promise<sequence<DOMString>> listGraphs();
+  Promise<MLGraph> loadGraph(DOMString key);
```
huningxin, line 61: A JS ML framework, such as ONNX Runtime Web, may need to know the input and output operands info (name, shape and data type) to construct input and output tensors for an inference session. The input and output operands info is known if users pass the source model, e.g. ONNX model. With model cache, user may only pass the model key, the framework needs to fetch the input and output operands info from an `MLGraph`.  It would be necessary to expose the `inputDescriptors` and `outputDescriptors` internal slots of `MLGraph` interface.

```webidl
partial interface MLGraph {
  record<USVString, MLOperandDescriptor> inputDescriptors;
  record<USVString, MLOperandDescriptor> outputDescriptors;
};
```

/cc @Honry

anssiko:
@reillyeon @huningxin thanks for your review.

I pushed an update to address your comments: https://github.com/webmachinelearning/webnn/pull/862/commits/20ed41475d6cd99ce793ebf76d19baa7275e419c

I took most of your comments as is, as it is easy to do further editorial tweaks in this format.

One question:
I placed `buildAndSave` in considered alternatives for now. Do we have use cases that support keeping both the `saveGraph` and `buildAndSave`, or should we just keep the latter? Asking in the spirit of minimization of the proposed new API surface.

cache-explainer.md
reillyeon: I think this would definitely be useful for debugging (though I'd name the attributes simply `inputs` and `outputs`) but at least based on my experience so far integrating WebNN into TFLite I don't think this will be necessary as more than a sanity check that the cached model is the one the framework expected to load as the framework already knows what it named the input and output operands previously and their types.

cache-explainer.md
reillyeon: I would expect that a compliant implementation of `build()` could save the compiled model into a temporary file which is deleted unless `saveGraph()` is called later and this doesn't have any downsides over an implementation which keeps the model in memory. In fact it would have a lower impact on system resource consumption because a file-backed model can be easily evicted from memory if there is memory pressure.

I'd like to better understand the ONNX runtime constraints. Is the issue that a model can be made ready for inference faster if the browser knows it doesn't need to be saved and so you want to make that advantage available to developers when they don't intend to use the caching API?

cache-explainer.md
```diff
@@ -0,0 +1,133 @@
...
+
+```webidl
+partial interface MLGraphBuilder {
+  Promise<MLGraph> buildAndSave(MLNamedOperands outputs, DOMString key);
+};
+```
+
+### Explicit vs implicit API
+
+>GPU shader caching is implicit, however the difference is that a shader program is a small input and so it's easy for the site to regenerate the shader so the browser can hash it to compare with the cache. ML models on the other hand are large because of the weights. Loading all the weights just to discover that a cached version of the model is available would be a waste of time and resources. (via [comment](https://github.com/webmachinelearning/webnn/issues/807#issuecomment-2608135598))
```
reillyeon: Remove the ">" here.

cache-explainer.md
huningxin: @reillyeon 

> I would expect that a compliant implementation of build() could save the compiled model into a temporary file which is deleted unless saveGraph() is called later and this doesn't have any downsides over an implementation which keeps the model in memory.

This sounds good. Because the compiled model is saved earlier, when saveGraph() is called, the implementation may just need to move the temporary compiled model file rather than recompile the source model anymore.

> Is the issue that a model can be made ready for inference faster if the browser knows it doesn't need to be saved and so you want to make that advantage available to developers when they don't intend to use the caching API?

AFAIK, for an ONNX Runtime implementation, just compiling a model for inference ([CreateSessionFromModel](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L5731)) should be faster. To implement model cache, the implementation needs to call ([CompileModel](https://github.com/microsoft/onnxruntime/blob/main/include/onnxruntime/core/session/onnxruntime_c_api.h#L5731)) which will re-compile the model and save the compiled model to disk. If the implementation always saves the compiled model to disk at the `build()`, we can consider compiling and saving the model first, then always creating the inference session from the compiled model temporary file.

cache-explainer.md
huningxin: @reillyeon 

> though I'd name the attributes simply inputs and outputs

SGTM.

> as the framework already knows what it named the input and output operands previously and their types.

This might not always be the case. @Honry is experimenting ONNX Runtime Web with WebNN model cache support. In that prototype, users may just pass the model cache key and have the framework to create an inference session from it. Framework doesn't know the input and output operands info in this scenario.

cache-explainer.md
anssiko: Fixed in https://github.com/webmachinelearning/webnn/pull/862/commits/1e1bc1c23991d2169f8c59d7c9de20dd6454b63e

cache-explainer.md
anssiko: Fixed in https://github.com/webmachinelearning/webnn/pull/862/commits/20ed41475d6cd99ce793ebf76d19baa7275e419c

cache-explainer.md
anssiko: Added a "A note on persistence" [sic] https://github.com/webmachinelearning/webnn/pull/862/commits/20ed41475d6cd99ce793ebf76d19baa7275e419c

cache-explainer.md
anssiko: Fixed in https://github.com/webmachinelearning/webnn/pull/862/commits/20ed41475d6cd99ce793ebf76d19baa7275e419c

cache-explainer.md
anssiko: Improved the title with a rename "Reduce time to first inference _on reload_" in https://github.com/webmachinelearning/webnn/pull/862/commits/20ed41475d6cd99ce793ebf76d19baa7275e419c

cache-explainer.md
anssiko: Fixed in https://github.com/webmachinelearning/webnn/pull/862/commits/20ed41475d6cd99ce793ebf76d19baa7275e419c

anssiko:
Thanks for your feedback, @reillyeon @huningxin.

I resolved the conversations that I felt had been settled with an obvious fix.

The remaining open conversations welcome further feedback.

cache-explainer.md
reillyeon: I'm not familiar with the ONNX Runtime API so maybe the method names are misleading me but doesn't the model need to be compiled regardless of whether the results of compilation will be saved or not?

It seems like the main optimization from not writing out the model to disk is that it saves time by not serializing and then deserializing the graph. In the other frameworks we've prototyped with (TFLite and Core ML) there is no option to load a model that isn't in a serialized form.

index.bs
```diff
@@ -3413,7 +3413,7 @@ partial dictionary MLOpSupportLimits {
   </tr>
   <tr>
     <td>*output*</td>
-    <td>[=/same type as|same as=] {{a}}</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
```
reillyeon, line 3416: Looks like some unrelated changes made their way into your branch.

reillyeon (approved):

index.bs
anssiko: Fixed by https://github.com/webmachinelearning/webnn/pull/862/commits/8567f6d8c66e788a47bf97c92f38e0a499478e5b

anssiko:
@huningxin it looks like the remaining opens in this PR await @Honry's ONNX Runtime Web with WebNN model cache support experiment.

Would it make sense to merge this PR now and refine when we get the results? I want to allow for adequate time for that experiment and not rush it. A follow-up PR would be good for that.

huningxin (approved):
LGTM!

huningxin:
@anssiko 

> Would it make sense to merge this PR now and refine when we get the results? I want to allow for adequate time for that experiment and not rush it. A follow-up PR would be good for that.

SGTM, thanks for putting this PR together!

Honry:
> @huningxin it looks like the remaining opens in this PR await @Honry's ONNX Runtime Web with WebNN model cache support experiment.

@anssiko, sounds good.

PR #859, opened 2025-06-05
Add `roundEven` operator

fdwr:
- Fixes #817.
- Addresses incomplete `quantizeLinear` emulation step.

```diff
partial interface MLGraphBuilder
{
    ...
+   MLOperand roundEven(MLOperand input, optional MLOperatorOptions options = {});
    ...
};
```

## Mappings to backend functions:

`RHE` **r**ound **h**alves to nearest **e**ven:
- [DML ROUND](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_element_wise_round_operator_desc) with [DML_ROUNDING_MODE_HALVES_TO_NEAREST_EVEN](https://learn.microsoft.com/en-us/windows/win32/api/directml/ne-directml-dml_rounding_mode)
- [tfl.round (TFL::RoundOp)](https://www.tensorflow.org/mlir/tfl_ops#tflround_tflroundop) 🤔 \*It's not clearly documented for TFLite, but I'm presuming TFL follows TF in using RHE.
- [ONNX Round](https://onnx.ai/onnx/operators/onnx__Round.html)
- [PyTorch torch.round](https://pytorch.org/docs/stable/generated/torch.round.html)
- ❔[coremltools.converters.mil.mil.ops.defs.iOS15.elementwise_unary.round](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.elementwise_unary.round) 🤔 \*From [Ningxin's investigation](https://github.com/webmachinelearning/webnn/issues/817#issuecomment-2833839764), CoreML's `round`ing mode behaves inconsistently depending on whether it goes through the NPU vs GPU, but it *should* be consistently emulatable via `elementwise_binary.sub(x, elementwise_binary.mod(x, 1.0f))`. 🤞

***
<a href="https://pr-preview.s3.amazonaws.com/fdwr/webnn/pull/859.html" title="Last updated on Jun 7, 2025, 2:35 AM UTC (697eaee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/859/61d4d5a...fdwr:697eaee.html" title="Last updated on Jun 7, 2025, 2:35 AM UTC (697eaee)">Diff</a>

PR #868, opened 2025-06-11
Add a non-normative table of operators by category

anssiko:
This non-normative table with functional grouping of operators complements the normative definition of operators that are currently in alphabetical order.

I tried to check we're covering all the operators, but it is very likely I may have missed some. I proactively added `roundEven` #859. Feedback welcome whether this makes sense and what would make for a good categorization.

The caveat is this table needs to be manually maintained alongside the normative definition, but that'd be just one extra line.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -736,6 +736,184 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
 
 For a history and rationale of this design, please see the <a href="https://github.com/webmachinelearning/webnn/blob/master/device-selection-explainer.md">device selection explainer</a>.
 
+## Operators ## {#programming-model-operators}
+
+*This section is non-normative.*
+
+The WebNN API defines a set of operators required by well-known CNN and RNN, transformer and generative models that address key [[#usecases-application]]. The details of each operator are defined in the normative sections of this specification, in alphabetical order by the operator name. These operators are grouped into categories based on their functionality in the following non-normative table to give a functional overview of the API surface.
+
+<table id='operators' class='data'>
+  <caption>Operators by category</caption>
+  <thead>
+    <tr>
+      <th>Category</th>
+      <th>Operators​</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Tensor creation</td>
+      <td>
+      {{MLGraphBuilder/input()}},
+      {{MLGraphBuilder/constant(tensor)|constant()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Tensor manipulation</td>
+      <td>
+      {{MLGraphBuilder/concat()}},
+      {{MLGraphBuilder/expand()}},
+      {{MLGraphBuilder/gather()}},
+      {{MLGraphBuilder/gatherElements()}},
+      {{MLGraphBuilder/scatterElements()}},
+      {{MLGraphBuilder/gatherND()}},
+      {{MLGraphBuilder/scatterND()}},
+      {{MLGraphBuilder/pad()}},
+      {{MLGraphBuilder/reshape()}},
+      {{MLGraphBuilder/slice()}},
+      {{MLGraphBuilder/split()}},
+      {{MLGraphBuilder/transpose()}},
+      {{MLGraphBuilder/resample2d()}},
+      {{MLGraphBuilder/reverse()}},
+      {{MLGraphBuilder/tile()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Tensor quantization</td>
+      <td>
+      {{MLGraphBuilder/quantizeLinear()}},
+      {{MLGraphBuilder/dequantizeLinear()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Tensor casting</td>
+      <td>
+      {{MLGraphBuilder/cast()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Mathematics</td>
+      <td>
+      {{MLGraphBuilder/add()}},
+      {{MLGraphBuilder/sub()}},
+      {{MLGraphBuilder/mul()}},
+      {{MLGraphBuilder/div()}},
+      {{MLGraphBuilder/max()}},
+      {{MLGraphBuilder/min()}},
+      {{MLGraphBuilder/pow()}},
+      {{MLGraphBuilder/abs()}},
+      {{MLGraphBuilder/ceil()}},
+      {{MLGraphBuilder/cos()}},
+      {{MLGraphBuilder/erf()}},
+      {{MLGraphBuilder/exp()}},
+      {{MLGraphBuilder/floor()}},
+      {{MLGraphBuilder/identity()}},
+      {{MLGraphBuilder/log()}},
+      {{MLGraphBuilder/neg()}},
+      {{MLGraphBuilder/reciprocal()}},
+      {{MLGraphBuilder/sin()}},
+      {{MLGraphBuilder/sqrt()}},
+      {{MLGraphBuilder/tan()}},
+      {{MLGraphBuilder/sign()}},
+      {{MLGraphBuilder/roundEven()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Logical</td>
+      <td>
+      {{MLGraphBuilder/equal()}},
+      {{MLGraphBuilder/notEqual()}},
+      {{MLGraphBuilder/greater()}},
+      {{MLGraphBuilder/greaterOrEqual()}},
+      {{MLGraphBuilder/lesser()}},
+      {{MLGraphBuilder/lesserOrEqual()}},
+      {{MLGraphBuilder/logicalNot()}},
+      {{MLGraphBuilder/where()}},
+      {{MLGraphBuilder/logicalAnd()}},
+      {{MLGraphBuilder/logicalOr()}},
+      {{MLGraphBuilder/logicalXor()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Convolution</td>
+      <td>
+      {{MLGraphBuilder/conv2d()}},
+      {{MLGraphBuilder/convTranspose2d()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Matrix multiplication</td>
+      <td>
+      {{MLGraphBuilder/matmul()}},
+      {{MLGraphBuilder/gemm()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Pooling</td>
+      <td>
+      {{MLGraphBuilder/averagePool2d()}},
+      {{MLGraphBuilder/l2Pool2d()}},
+      {{MLGraphBuilder/maxPool2d()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Activation</td>
+      <td>
+      {{MLGraphBuilder/clamp()}},
+      {{MLGraphBuilder/elu()}},
+      {{MLGraphBuilder/gelu()}},
+      {{MLGraphBuilder/hardSigmoid()}},
+      {{MLGraphBuilder/hardSwish()}},
+      {{MLGraphBuilder/leakyRelu()}},
+      {{MLGraphBuilder/linear()}},
+      {{MLGraphBuilder/prelu()}},
+      {{MLGraphBuilder/relu()}},
+      {{MLGraphBuilder/sigmoid()}},
+      {{MLGraphBuilder/softmax()}},
+      {{MLGraphBuilder/softplus()}},
+      {{MLGraphBuilder/softsign()}},
+      {{MLGraphBuilder/tanh()}},
+      {{MLGraphBuilder/triangular()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Normalization</td>
+      <td>
+      {{MLGraphBuilder/batchNormalization()}},
+      {{MLGraphBuilder/instanceNormalization()}},
+      {{MLGraphBuilder/layerNormalization()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Reduction</td>
+      <td>
+      {{MLGraphBuilder/argMin()}},
+      {{MLGraphBuilder/argMax()}},
+      {{MLGraphBuilder/reduceL1()}},
+      {{MLGraphBuilder/reduceL2()}},
+      {{MLGraphBuilder/reduceLogSum()}},
+      {{MLGraphBuilder/reduceLogSumExp()}},
+      {{MLGraphBuilder/reduceMax()}},
+      {{MLGraphBuilder/reduceMean()}},
+      {{MLGraphBuilder/reduceMin()}},
+      {{MLGraphBuilder/reduceProduct()}},
+      {{MLGraphBuilder/reduceSum()}},
+      {{MLGraphBuilder/reduceSumSquare()}},
+      {{MLGraphBuilder/cumulativeSum()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Recurrent Neural Networks</td>
+      <td>
+      {{MLGraphBuilder/gruCell()}},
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -736,6 +736,183 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+      </td>
+    </tr>
+    <tr>
+      <td>Pooling</td>
+      <td>
+      {{MLGraphBuilder/averagePool2d()}},
+      {{MLGraphBuilder/l2Pool2d()}},
+      {{MLGraphBuilder/maxPool2d()}}
+      </td>
+    </tr>
```
fdwr, line 863: It makes sense that convolution and pooling categories are adjacent, since they are both *windowed reduction* with very similar parameters (the only conceptual difference is that pooling has an imaginary filter of all 1's), and I like that convolution and matmul are adjacent (since you can map between GEMM <-> Conv).  How about this order?

- Matrix multiplication
- Convolution
- Pooling

index.bs
```diff
@@ -736,6 +736,183 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+  <tbody>
+    <tr>
+      <td>Tensor creation</td>
+      <td>
+      {{MLGraphBuilder/input()}},
+      {{MLGraphBuilder/constant(tensor)|constant()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Tensor manipulation</td>
```
fdwr, line 764: "Tensor manipulation" is a very broad name (one could say a `cast` operator manipulates tensors too by casting them), whereas a name like "Data reorganization"/"Tensor rearrangement" feels more apt, as they all rearrange/stretch/crop/reinterpret the data.

index.bs
```diff
@@ -736,6 +736,183 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+    </tr>
+    <tr>
+      <td>Tensor quantization</td>
+      <td>
+      {{MLGraphBuilder/quantizeLinear()}},
+      {{MLGraphBuilder/dequantizeLinear()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Tensor casting</td>
```
fdwr, line 793: I like that the casting and quantization categories are adjacent, since quantization is a form of adjusted casting. 👍

index.bs
```diff
@@ -736,6 +736,183 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
 
 For a history and rationale of this design, please see the <a href="https://github.com/webmachinelearning/webnn/blob/master/device-selection-explainer.md">device selection explainer</a>.
 
+## Operators ## {#programming-model-operators}
+
+*This section is non-normative.*
+
+The WebNN API defines a set of operators required by well-known CNN and RNN, transformer and generative models that address key [[#usecases-application]]. The details of each operator are defined in the normative sections of this specification, in alphabetical order by the operator name. These operators are grouped into categories based on their functionality in the following non-normative table to give a functional overview of the API surface.
```
fdwr, line 743: The hardest challenge with putting them all into mutually distinct groups is that some operators belong to multiple sets. For example, if we were to add `sinh` & `cosh` to WebNN, then surely they would go along with all the other math operators, but `sinh`/`cosh`/`tanh` are a trio, and so it would be weird to have `tanh` in a completely different category. Similarly, `clamp` is definitely a math function along with its siblings `min` and `max`, but quite rarely it's also useable for activation. So is it more a math function or an activation function? Neither - it's both. I tried putting operators into distinct categories in my [listing here](https://fdwr.github.io/MachineLearningOperators/OperatorFormulas.html) and eventually realized it was futile 😉, instead giving them multiple tags.

index.bs
```diff
@@ -736,6 +736,183 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+      <td>Logical</td>
+      <td>
+      {{MLGraphBuilder/equal()}},
+      {{MLGraphBuilder/notEqual()}},
+      {{MLGraphBuilder/greater()}},
+      {{MLGraphBuilder/greaterOrEqual()}},
+      {{MLGraphBuilder/lesser()}},
+      {{MLGraphBuilder/lesserOrEqual()}},
+      {{MLGraphBuilder/logicalNot()}},
+      {{MLGraphBuilder/where()}},
```
fdwr: 🤔 I suppose the selection operator (aka TOSA select) could be thought of as a logical operator, given it takes boolean input (all the others produce boolean *output*), but I always thought of it as more of a data reorganization operator like `gather`, gathering data respectively from input a or b, depending on the boolean indices (e.g. pseudocode kinda like `gather(concat([a, b], axis:rank), cast(indices, "uint32"), axis:rank)`), and `where` isn`t under the [Element-wise logical operations](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-logical) list either.

index.bs
```diff
@@ -736,6 +736,183 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+      {{MLGraphBuilder/leakyRelu()}},
+      {{MLGraphBuilder/linear()}},
+      {{MLGraphBuilder/prelu()}},
+      {{MLGraphBuilder/relu()}},
+      {{MLGraphBuilder/sigmoid()}},
+      {{MLGraphBuilder/softmax()}},
+      {{MLGraphBuilder/softplus()}},
+      {{MLGraphBuilder/softsign()}},
+      {{MLGraphBuilder/tanh()}},
+      {{MLGraphBuilder/triangular()}}
```
fdwr: The triangular matrix masking function doesn't feel like [activation](https://en.wikipedia.org/wiki/Activation_function), but I don't know what category it fits in either 🤔. Are there cases where trilu acts as activation that I'm unaware of?

fdwr (commented):
Thanks Anssi. Yes, I think it's useful to have an upfront table that categorizes operators like this.

index.bs
anssiko: This makes sense. Updated the order accordingly.

index.bs
anssiko: Updated to "Tensor rearrangement".

Would "Data organization" be a more established name? I chose this on a quest to have some symmetry with "Tensor ..." prefixed categories first.

index.bs
anssiko: I have to admin I struggled with this part too. Some of the groupings inherited from another exercise with @huningxin where we had to simplify things to another audience.

For web specs, dynamic elements might not be appropriate for a11y, printing and other reasons. Otherwise I'd happily port over your dynamic table to the spec (I had forgotten about this fantastic resource!).

Would a reasonable tradeoff for a static representation be to add those operators that are _clearly_ multipurpose to multiple categories? I experimented with this idea using `clamp`. For those operators that are _primarily_ associated with a particular category e.g. for historical reasons, say math functions, but are also used for other purposes, say for masking, I'd still keep them in maths. Not perfect, but...

index.bs
anssiko: I parked `where` to "Tensor rearrangement" for now.

index.bs
anssiko: Does it make sense to add a new "Tensor masking" / "Matrix masking" category for operators that are primarily(?) used for masking?

Experimented with this idea using `triangular`.

anssiko:
@fdwr, much thanks for your suggestions. I pushed an update with a new iteration to test some new ideas based on your feedback.

index.bs
huningxin: FYI, CoreML puts `select` into "control flow" category: https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.control_flow.select

index.bs
huningxin: Also "tensor rearrangement" op? FYI, CoreML puts `band_part` into "tensor operation" category: https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.tensor_operation.band_part

index.bs
```diff
@@ -736,6 +736,191 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+      {{MLGraphBuilder/scatterElements()}},
+      {{MLGraphBuilder/gatherND()}},
+      {{MLGraphBuilder/scatterND()}},
+      {{MLGraphBuilder/where()}},
+      {{MLGraphBuilder/pad()}},
+      {{MLGraphBuilder/reshape()}},
+      {{MLGraphBuilder/slice()}},
+      {{MLGraphBuilder/split()}},
+      {{MLGraphBuilder/transpose()}},
+      {{MLGraphBuilder/resample2d()}},
```
huningxin, line 779: I feel `resample2d` is not a tensor rearrangement op, because it doesn't just rearrange the existing values, but generate new values. FYI, CoreML puts "resample" into "image resizing" category: https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS16.image_resizing.resample

index.bs
fdwr: Yes, that's what I was thinking - just include `tanh` and `clamp` under math but also activation.

index.bs
fdwr: > because it doesn't just rearrange the existing values, but generate new values

That is true for linear interpolation.

Notice for *nearest neighbor* that `resample` is functionally identical to integral upsampling (or blockwise expansion, like that used in dequantizeLinear). One can even implement `dequantizeLinear`'s blockwise expansion step via a backend's resampling operator + NN (and I was thinking to augment `expand` to accept block sizes for better functional decomposition of Q/DQ). So, there is clearly a familial relationship between spatial expansion and resampling, along with other spatial interpolation functions like [ROI alignment](https://onnx.ai/onnx/operators/onnx__RoiAlign.html) and [warped grid sampling](https://onnx.ai/onnx/operators/onnx__GridSample.html).

> into "image resizing" category:

It's true that these spatial interpolation functions are *often* used with imagery, but 1D resampling is useful for audio rescaling too (so I wouldn't limit it to just "image resizing"). Maybe there's a category name that better encompasses all these spatial remappings/reprojections? Anssi *originally* had the category name of "Tensor manipulation", which he changed per my comment because I found it a little too generic, but would you find that amenable? "Spatial modification"? "Spatial manipulation"? "Tensor reprojection"?

index.bs
fdwr: > FYI, CoreML puts select into "control flow" category:

I can see why someone might think that, given it feels like a C ternary operator (a ? b : c), but functionally it's more like a `gatherElements` with 0/1 indices, as it doesn't cause any control flow change in the graph execution (unlike say ONNX `If`/`While`/`Scan` or CoreML `cond`/`while_loop`), and webNN doesn't support any true control flow operators currently, meaning it would be a weirdly isolated category.

index.bs
fdwr: > Does it make sense to add a new "Tensor masking"
> Also "tensor rearrangement" op? 

Content with either.

index.bs
fdwr: > I chose this on a quest to have some symmetry with "Tensor ..." prefixed categories first.

Do these help your quest? 😉

"Tensor spatial manipulation"?
"Tensor rearrangement"?
"Tensor reprojection"?
"Data reorganization"?
"Data organization"?
"Spatial modification"?
"Spatial manipulation"?
...

fdwr (approved):
Thanks for the updates AK. Looks good to me after tanh/clamp dual citizenship and deciding on a name for the spatial remapping category.

index.bs
huningxin: >  webNN doesn't support any true control flow operators currently, meaning it would be a weirdly isolated category.

That's true.

TOSA places `select` into "Elementwise Ternary Operators" category: https://www.mlplatform.org/tosa/tosa_spec.html#_elementwise_ternary_operators

Should we use "element-wise binary", "element-wise logical" and "element-wise unary" categories? They are already used in the normative text.

index.bs
huningxin: > but 1D resampling is useful for audio rescaling too (so I wouldn't limit it to just "image resizing")

Agreed. For 2D resmaple/resize, TOSA also places it under "image operators" category: https://www.mlplatform.org/tosa/tosa_spec.html#_image_operators

index.bs
```diff
@@ -736,6 +736,191 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+    <tr>
+      <td>Tensor rearrangement</td>
+      <td>
+      {{MLGraphBuilder/concat()}},
+      {{MLGraphBuilder/expand()}},
+      {{MLGraphBuilder/gather()}},
+      {{MLGraphBuilder/gatherElements()}},
+      {{MLGraphBuilder/scatterElements()}},
+      {{MLGraphBuilder/gatherND()}},
+      {{MLGraphBuilder/scatterND()}},
```
huningxin, line 772: Both CoreML and TOSA have a dedicated "gather/scatter" category:
https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#module-coremltools.converters.mil.mil.ops.defs.iOS15.scatter_gather
https://www.mlplatform.org/tosa/tosa_spec.html#_scattergather_operators

index.bs
```diff
@@ -736,6 +736,191 @@ Depending on the underlying platform, the user agent <span class=allow-2119>may<
...
+    </tr>
+    <tr>
+      <td>Tensor quantization</td>
+      <td>
+      {{MLGraphBuilder/quantizeLinear()}},
+      {{MLGraphBuilder/dequantizeLinear()}}
+      </td>
+    </tr>
+    <tr>
+      <td>Tensor casting</td>
```
huningxin, line 793: FYI, TOSA groups `cast` and `scale` (equivalent to quantizeLinear/dequantizeLinear) to "Type Conversion" category: https://www.mlplatform.org/tosa/tosa_spec.html#_type_conversion

index.bs
anssiko: OK, let's go with this simple solution. Updated to include `tanh` and `clamp` under math too.

index.bs
anssiko: This table currently proposes a higher-level functional grouping, and as such abstracts out "element-wise" from "logical" ops and groups both "element-wise binary" and "element-wise unary" ops under "Mathematics".

I'd propose we try to go with a higher-level grouping first, and let the normative text offer more detail where appropriate.

index.bs
anssiko: Moved `triangular` to "Tensor manipulation" (was "Tensor rearrangement") so we can drop the "Tensor masking" category, which I did.

index.bs
anssiko: Switched back to the more generic "Tensor manipulation" name for now so that `resample` feels welcome.

anssiko:
@fdwr @huningxin thanks for your suggestions. I pushed another update. PTAL if I missed any feedback: https://github.com/webmachinelearning/webnn/pull/868/commits/fed3ae8dd74ed22d2826b8b0bdf192c6e2772222

(This has been a fun exercise and a reminder naming in this field is hard and does not always agree with other fields, statistics, physics, information theory... While "Web API for chains of differentiable, parameterized geometric functions” would be a more accurate name, we call this spec the "Web Neural Network API" ;-))

fdwr (approved):
👍

index.bs
fdwr: I'm happy either way on this, having Q/DQ in their own category and putting them as part of data type conversion.

anssiko:
I will merge this first stab to shorten the open PR queue. Thank you for your insights @fdwr and @huningxin!

Please feel free to continue adjust this table as you see fit. A natural check point for adjustment is when new ops are being added or existing removed, i.e. the moment when this table needs to be updated manually. I hope that won't become a chore but is actually a useful exercise to reason about our op coverage.

Issue #870, opened 2025-06-24
Improve accessibility of examples


anssiko:
This feedback was provided as part of the accessibility review #105:

The [example graph](https://www.w3.org/TR/webnn/#examples) uses 'ASCII art' to convey its structure. This isn't accessible to ATs.

>This should be regarded as low priority compared to the other issues; if we get the time, we may submit a PR. My current best thought on how to make the graph (which is in this case a tree) more accessible would be to express it as a nested list, which would share the tree structure, but we would root the list on the output, so it would be 'backwards' relative to the visual version. [It could be done via SVG (using a technique Léonie Watson proposed)](https://tink.uk/accessible-svg-flowcharts/) though we'd want to test this out before making anyone do work!

tomayac:
Here's a Mermaid diagram:

```mermaid
flowchart TD
  constant1 --- A1
  input1    --- A1
  A1[Add] --> intermediateOutput1

  constant2 --- A2
  input2    --- A2
  A2[Add] --> intermediateOutput2

  intermediateOutput1 --> M[Mul]
  intermediateOutput2 --> M
  M --> output
```

Here's an SVG of that diagram:

![Image](https://github.com/user-attachments/assets/027785d9-3b11-4e28-a620-49b9b7a56383)

And here's some corresponding alternative text:

```
A computational graph with the following structure:
constant1 and input1 are both inputs to an "Add" operation, whose result is labeled intermediateOutput1.
constant2 and input2 are inputs to a second "Add" operation, producing intermediateOutput2.
intermediateOutput1 and intermediateOutput2 feed into a "Mul" (multiply) operation.
The result of the "Mul" operation is the final output.
```

tomayac:
SVG got cropped. Here's the text

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?xml-stylesheet href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.7.2/css/all.min.css" type="text/css"?>
<svg
  aria-roledescription="flowchart-v2"
  role="graphics-document document"
  style="
    overflow: hidden;
    max-width: 100%;
    touch-action: none;
    user-select: none;
    -webkit-user-drag: none;
    -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
  "
  class="flowchart"
  xmlns="http://www.w3.org/2000/svg"
  width="100%"
  id="graph-47"
  height="100%"
[...]

PR #871, opened 2025-06-24
Update WebNN architecture diagram SVG

anssiko:
Issue #870 reminded me we'd need a new SVG export for the architecture diagram too, so I took a plunge into the world of SVG.

It seems GH diff UI (below) does not render the SVG properly, but it does render with all major browser after applying the patch (see the commit message): https://raw.githubusercontent.com/webmachinelearning/webnn/arch-diagram-svg/content/webnn_arch.svg

I kept the Mermaid in the explainer for now, and put a note in there so we remember this SVG exists.

PTAL @tomayac for cross-checking if there's a better way to export this without errors.

FYI @matatk, here's another diagram we want to make screen reader friendly.


tomayac:
Those labels are a bit hard to see. 

Screenshot 2025-06-24 at 17 44 02

FWIW, my flow in https://github.com/webmachinelearning/webnn/issues/870#issuecomment-3000593932 was to paste the existing diagram in ChatGPT, ask it to create a Mermaid diagram out of it, and then fact-check it and polish it in the [online editor](https://mermaid.live/edit).

anssiko:
Those subgraph labels were also bugging me. IIRC we found out positioning them was not possible without CSS hacks that come with side effects. Please check with your favourite LLM to see if there's indeed a non-hacky solution :-)

tomayac:
Your diagram looks like drawn with Excalidraw. Maybe just move the labels up so they are between boxes next to the arrows?

anssiko:
It's drawn with Mermaid, the source code is in the commit message https://github.com/webmachinelearning/webnn/pull/871/commits/1bc34db64ec6ad90c2b324cc8bd3988cdce9cd97 and if you drop that source into https://mermaid.live/ it should render as in https://github.com/webmachinelearning/webnn/blob/main/explainer.md

I'd like to have clean Mermaid source without hacks that either as is could be used as an alt text, or used as input to LLM to produce a good quality alt text.

anssiko:
I updated this PR to make this a clean drop-in replacement from the current script-generated graph to an identical SVG, retaining the source. To be iterated further to improve accessibility. Copilot suggested an alt text that I adopted for now.

It looks like we cannot test this without deploying to production, so I'll hit the green button. 🤞

anssiko:
It appears embedding SVG into Markdown did not work as expected, so I deployed a hotfix https://github.com/webmachinelearning/webnn/commit/093be92ca87fbfdd912fe6654dd07349e1ab8d5b to revert back to the script-generated graph. Sorry for the noise.

PR #881, opened 2025-08-21
Add opSupportLimits() privacy considerations for fingerprinting

### LINKED ISSUE #875: Future-proofing privacy considerations of New API feature to identify feature support per operator
Making a note here for completeness (see also discussion at https://github.com/w3cping/privacy-request/issues/156) to ensure:

(1) the spec explicitly communicates that this capability doesn't add entropy, that it is just a more convenient way of discovering feature support information that can be inferred from the OS and browser version already

(2) To ensure the extension point for privacy controls are in place to prevent the capability from being used in the future for fingerprinting, etc.

anssiko:
Discussed at https://www.w3.org/2025/08/14-webmachinelearning-minutes.html#7993

Seeking initial review from @RafaelCintron and @reillyeon to ensure I interpreted and translated your meeting comments the right way into this spec note.

Fix #875

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1416,6 +1416,8 @@ Note: Similar to `dispatch()`, `writeTensor()` itself provides no signal that th
 ### {{MLContext/opSupportLimits()}}  ### {#api-mlcontext-opsupportlimits}
 The {{MLContext/opSupportLimits()}} exposes level of support that differs across implementations at operator level. Consumers of the WebNN API are encouraged to probe feature support level by using {{MLContext/opSupportLimits()}} to determine the optimal model architecture to be deployed for each target platform.
 
+Note: The {{MLContext/opSupportLimits()}} API is intended to not be a source of additional entropy for browser fingerprinting. The API is a more explicit mechanism for discovering the feature support information that can be inferred from the OS and browser version alone. If the diversity of future implementations warrants it, this explicit API allows a future version of this specification to add new privacy mitigations e.g. use configurations or buckets similar to WebGPU to reduce entropy.
+
 #### {{MLOpSupportLimits}} dictionary #### {#api-mlcontext-opsupportlimits-dictionary}
 The {{MLOpSupportLimits}} has the following top level members, aside from these, each [=operator=] has a corresponding member defined in its builder method.
 <script type="idl">


philloooo:
It seems to express similar thing as in https://github.com/webmachinelearning/webnn/blob/main/security-privacy.md#questions-to-consider, should we de-dup and reference the spec in the security-privacy.md?

anssiko:
@philloooo, slight preference to have this in context of the specification close to the algorithm so that more people will read the note. Happy to borrow that text verbatim or adapt. Then add a link to the respective section in the specification. Suggestions for concrete text welcome!

index.bs
```diff
@@ -1416,6 +1416,8 @@ Note: Similar to `dispatch()`, `writeTensor()` itself provides no signal that th
 ### {{MLContext/opSupportLimits()}}  ### {#api-mlcontext-opsupportlimits}
 The {{MLContext/opSupportLimits()}} exposes level of support that differs across implementations at operator level. Consumers of the WebNN API are encouraged to probe feature support level by using {{MLContext/opSupportLimits()}} to determine the optimal model architecture to be deployed for each target platform.
 
+Note: The {{MLContext/opSupportLimits()}} API is intended to not be a source of additional entropy for browser fingerprinting. The API is a more explicit mechanism for discovering the feature support information that can be inferred from the OS and browser version alone. If the diversity of future implementations warrants it, this explicit API allows a future version of this specification to add new privacy mitigations e.g. use configurations or buckets similar to WebGPU to reduce entropy.
```
reillyeon:
```suggestion
Note: The {{MLContext/opSupportLimits()}} API is not intended to provide additional entropy for browser fingerprinting. In current implementations this feature support information can be inferred from the OS and browser version alone. If the diversity of future implementations warrants it, this API allows future implementations to add new privacy mitigations e.g. to bucket capabilities similar to WebGPU to reduce entropy.

See [#privacy] for additional discussion of fingerprinting considerations.
```

anssiko: @reillyeon thank you, this is a great improvement.

It is a good idea to refer to https://www.w3.org/TR/webnn/#privacy

We should make sure applicable content from https://github.com/webmachinelearning/webnn/blob/main/security-privacy.md is migrated to https://www.w3.org/TR/webnn/#privacy that will arguably be read by more people. The main purpose of security-privacy.md is to facilitate privacy and security reviews and I suppose that doc is often a snapshot in time while the spec is expected to reflect the latest.

reillyeon (approved):
These edits look good, but please follow up by revisiting the general privacy section.

anssiko:
I will merge this PR now. A follow-up issue to revise privacy considerations: #886

PR #894, opened 2025-09-30
Correct pad scalar rank inconsistency

### LINKED ISSUE #893: Pad operator algorithm steps should not throw error for scalar input
@fdwr raised this issue when reviewing Chromium CL https://chromium-review.googlesource.com/c/chromium/src/+/6972256/comment/7552dd15_61137e37/

> The WebNN spec says in the table that ranks 0 to N are allowed for [pad's input](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-pad), including scalars which is mathematically reasonable (given beginningPadding and endingPadding are both empty to satisfy the rule that they have the same size as the input rank) and also consistent with tile and transpose as nops, and you even have an implementation of it below. Though, currently [one of the algorithm steps say](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-pad) "If input’s rank is 0, then throw a TypeError", and so I'd be happy to update the spec to remove that 1 line in the algorithm steps.

fdwr:
Fixes #893. The spec says pad's `opSupportLimits` support input ranks from 0 to N, and the Chromium implementation [see PR](https://chromium-review.googlesource.com/c/chromium/src/+/6972256/9/services/webnn/ort/context_impl_ort.cc#224) supports scalar `pad` (acts as nop, like an empty `transpose` or `tile` would), but this one algorithm step says it doesn't. So, resolve the inconsistency by deleting that line.

***
<a href="https://pr-preview.s3.amazonaws.com/fdwr/webnn/pull/894.html" title="Last updated on Sep 30, 2025, 6:02 AM UTC (43fe777)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/894/a3e8163...fdwr:43fe777.html" title="Last updated on Sep 30, 2025, 6:02 AM UTC (43fe777)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -7580,7 +7580,6 @@ partial dictionary MLOpSupportLimits {
   </summary>
     1. If [=this=] [=MLGraphBuilder/can not build=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and |input| returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If |input|'s [=MLOperand/rank=] is 0, then [=exception/throw=] a {{TypeError}}.
     1. If |beginningPadding|'s [=list/size=] and |endingPadding|'s [=list/size=] are not both equal to |input|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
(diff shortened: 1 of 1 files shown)


huningxin (approved):
LGTM, thanks @fdwr !

PR #884, opened 2025-08-27
Update explainer with new proposal for simple accelerator mapping

zolkis:
Added a solution proposal that is simpler than using fine grained device level op support limits, as discussed [here](https://github.com/webmachinelearning/webnn/issues/815#issuecomment-3198261369).

Adds some more examples and cleans up other sections affected by this proposal.

PR #895, opened 2025-10-08
Add a simple accelerator selection mechanism.

### LINKED ISSUE #815: Post-context creation query mechanism for supported devices

zolkis:
Fixes #815 

As explained in #884, add context options and attributes/internal slots that can be used for conveying application hints wrt the preferred acceleration type (CPU or massively parallel processing, i.e. NPU or GPU).

This is a minimal change, and we might want to refine more the algorithms wrt. context power preferences and acceleration options (currently not addressed). These could be done in this PR, or in a separate subsequent PR.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -748,7 +748,7 @@ An {{MLContext}} interface represents a global state of neural network execution
 In a situation when a GPU context executes a graph with a constant or an input in the system memory as an {{ArrayBufferView}}, the input content is automatically uploaded from the system memory to the GPU memory, and downloaded back to the system memory of an {{ArrayBufferView}} output buffer at the end of the graph execution. This data upload and download cycles will only occur whenever the execution device requires the data to be copied out of and back into the system memory, such as in the case of the GPU. It doesn't occur when the device is a CPU device. Additionally, the result of the graph execution is in a known layout format. While the execution may be optimized for a native memory access pattern in an intermediate result within the graph, the output of the last operation of the graph must convert the content back to a known layout format at the end of the graph in order to maintain the expected behavior from the caller's perspective.
 
 <div class="note">
-When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options, currently only the {{MLPowerPreference}} option.
+When an {{MLContext}} is created with {{MLContextOptions}}, the user agent selects and creates the underlying execution device by taking into account these options.
 
 Depending on the underlying platform, the user agent <span class=allow-2119>may</span> select different combinations of CPU, NPU and GPU devices.
 </div>
@@ -978,6 +978,7 @@ enum MLPowerPreference {
 
 dictionary MLContextOptions {
   MLPowerPreference powerPreference = "default";
+  boolean accelerated = true;
 };
 
 [SecureContext, Exposed=(Window, Worker)]
@@ -1001,6 +1002,8 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
 <dd>Prioritizes power consumption over other considerations such as execution speed.</dd>
 </dl>
 
+The <dfn dfn-for=MLContextOptions dfn-type=dict-member>accelerated</dfn> option indicates the application's preference as related to massively parallel acceleration. When set to `true` (by default), the underlying platform will attempt to use the available massively parallel accelerators, such as GPU or NPU, also depending on the {{MLContextOptions/powerPreference}}. When set to `false`, the application hints to prefer CPU inference.
+
 ### {{ML/createContext()}} ### {#api-ml-createcontext}
 
 <div dfn-for="ML/createContext(options), ML/createContext(gpuDevice)" dfn-type=argument>
@@ -1018,11 +1021,16 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
     1. If |options| is a {{GPUDevice}} object, then:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/webgpu=]".
         1. Set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
+        1. Set |context|.{{MLContext/[[accelerated]]}} to `true`.
+        1. Set |context|.{{MLContext/[[cpuFallbackActive]]}} to `false`.
     1. Otherwise:
         1. Set |context|.{{MLContext/[[contextType]]}} to "[=context type/default=]".
         1. Set |context|.{{MLContext/[[lost]]}} to [=a new promise=] in |realm|.
         1. If |options|["{{MLContextOptions/powerPreference}}"] [=map/exists=], then set |context|.{{MLContext/[[powerPreference]]}} to |options|["{{MLContextOptions/powerPreference}}"].
         1. Otherwise, set |context|.{{MLContext/[[powerPreference]]}} to {{MLPowerPreference/"default"}}.
+        1. If |options|["{{MLContextOptions/accelerated}}"] [=map/exists=], then set |context|.{{MLContext/[[accelerated]]}} to |options|["{{MLContextOptions/accelerated}}"].
+        1. Otherwise, set |context|.{{MLContext/[[accelerated]]}} to `true`.
+        1. Set |context|.{{MLContext/[[cpuFallbackActive]]}} to `false`.
     1. If the user agent cannot support |context|.{{MLContext/[[contextType]]}}, then return failure.
     1. Return |context|.
 </details>
@@ -1082,6 +1090,8 @@ interface MLContext {
 
   undefined destroy();
 
+  readonly attribute boolean accelerated;
+  readonly attribute boolean cpuFallbackActive;
   readonly attribute Promise<MLContextLostInfo> lost;
 };
 </script>
@@ -1095,6 +1105,12 @@ interface MLContext {
     : <dfn>\[[powerPreference]]</dfn> of type {{MLPowerPreference}}.
     ::
         The {{MLContext}}'s {{MLPowerPreference}}.
+    : <dfn>\[[accelerated]]</dfn> of type {{boolean}}.
+    ::
+        The {{MLContext}}'s processing type (CPU or massively parallel processing).
+    : <dfn>\[[cpuFallbackActive]]</dfn> of type {{boolean}}.
+    ::
+        The {{MLContext}}'s status for CPU fallback type (CPU or massively parallel processing).
     : <dfn>\[[lost]]</dfn> of type {{Promise}}<{{MLContextLostInfo}}>.
     ::
         A {{Promise}} that is resolved when the {{MLContext}}'s underlying execution device is no longer available.
@@ -1178,7 +1194,8 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
     1. If [=validating tensors with descriptors=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
-            1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
+            1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|, as well as |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[powerPreference]]}} and |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[accelerated]]}}.
+            1. If |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[accelerated]]}} is `true` and the underlying platform can only do CPU inference at the moment, then set |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[cpuFallbackActive]]}} to `true`, otherwise set it to `false`.
 
                 Issue(778): Add a mechanism for reporting errors during graph execution.
 
@@ -1730,7 +1747,7 @@ typedef (bigint or unrestricted double) MLNumber;
     : <dfn>\[[operator]]</dfn> of type [=operator=]
     ::
         Reference to {{MLOperand}}'s corresponding [=operator=].
-  
+
     : <dfn>\[[constantTensor]]</dfn> of type {{MLTensor}}
     ::
         The {{MLOperand}}'s tensor (only for constant operands).
@@ -2151,7 +2168,7 @@ Build a composed graph up to a given output operand into a computational graph a
         1. If |name| is empty, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
         1. If [=MLGraphBuilder/validating operand=] given [=this=] and |operand| returns false, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
         1. If |operand| is in [=this=]'s [=MLGraphBuilder/graph=]'s [=computational graph/inputs=] or [=computational graph/constants=], then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
-        1. If |operand|.{{MLOperand/[[constantTensor]]}} exists and |operand|.{{MLOperand/[[constantTensor]]}}.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}. 
+        1. If |operand|.{{MLOperand/[[constantTensor]]}} exists and |operand|.{{MLOperand/[[constantTensor]]}}.{{MLTensor/[[isDestroyed]]}} is true, then return [=a new promise=] in |realm| [=rejected=] with a {{TypeError}}.
     1. Let |operands| be a new empty [=/set=].
     1. Let |operators| be a new empty [=/set=].
     1. Let |inputs| be a new empty [=/set=].


anssiko:
@zolkis thank you for formalizing the group’s current thinking into this PR!

@huningxin @RafaelCintron, this spec PR is on the [WebML WG Teleconference – 23 October 2025](https://github.com/webmachinelearning/meetings/blob/main/telcons/2025-10-23-wg-agenda.md#-query-supported-devices) agenda. Reviews, comments, questions prior in this PR appreciated.

@handellm to check we remain aligned with Google Meet requirements.

FYI @mtavenrath who expressed interest in this space.

handellm:
Seems good!

index.bs
```diff
@@ -1001,6 +1002,8 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>powerPreference</dfn> opt
 <dd>Prioritizes power consumption over other considerations such as execution speed.</dd>
 </dl>
 
+The <dfn dfn-for=MLContextOptions dfn-type=dict-member>accelerated</dfn> option indicates the application's preference as related to massively parallel acceleration. When set to `true` (by default), the underlying platform will attempt to use the available massively parallel accelerators, such as GPU or NPU, also depending on the {{MLContextOptions/powerPreference}}. When set to `false`, the application hints to prefer CPU inference.
```
reillyeon:
```suggestion
The <dfn dfn-for=MLContextOptions dfn-type=dict-member>accelerated</dfn> option indicates the application's preference as related to massively parallel acceleration. When set to `true` (by default), the underlying platform will attempt to use the available massively parallel accelerators, such as a GPU or NPU, also depending on the {{MLContextOptions/powerPreference}}. When set to `false`, the application indicates it prefers unaccelerated CPU inference.
```

index.bs
```diff
@@ -1178,7 +1216,8 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
     1. If [=validating tensors with descriptors=] given |outputs| and |graph|.{{MLGraph/[[outputDescriptors]]}} returns false, then [=exception/throw=] a {{TypeError}}.
     1. Enqueue the following steps to |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[timeline]]}}:
         1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
-            1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|.
+            1. Issue a compute request to |graph|.{{MLGraph/[[implementation]]}} given |inputs| and |outputs|, as well as |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[powerPreference]]}} and |graph|.{{MLGraph/[[context]]}}.{{MLContext/[[accelerated]]}}.
```
huningxin: I suppose `powerPreference` and `accelerated` options should be used by `build` steps rather than `dispatch`?

index.bs
```diff
@@ -1178,7 +1216,8 @@ Note: `dispatch()` itself provides no signal that graph execution has completed.
...
+            1. Run the steps to [=poll CPU fallback status=] for |graph|.{{MLGraph/[[context]]}}.
```
huningxin: This step seems to be unnecessary because `cpuFallbackActive` getter already runs it?

index.bs
```diff
@@ -1114,6 +1130,28 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
...
+        1. Return [=this=].{{MLContext/[[cpuFallbackActive]]}}.
+</div>
+
+<details open algorithm>
+  <summary>
+    To <dfn>poll CPU fallback status</dfn>, run the following steps.
+  </summary>
+    1. If [=this=].{{MLContext/[[accelerated]]}} is `false`, then:
+        1. Set [=this=].{{MLContext/[[cpuFallbackActive]]}} to `true` and return.
+    1. If the underlying execution device is available, then:
```
huningxin: Is it worth adding a definition for "underlying execution device"?

index.bs
```diff
@@ -1095,6 +1105,12 @@ interface MLContext {
     : <dfn>\[[powerPreference]]</dfn> of type {{MLPowerPreference}}.
     ::
         The {{MLContext}}'s {{MLPowerPreference}}.
+    : <dfn>\[[accelerated]]</dfn> of type {{boolean}}.
+    ::
+        The {{MLContext}}'s processing type (CPU or massively parallel processing).
+    : <dfn>\[[cpuFallbackActive]]</dfn> of type {{boolean}}.
+    ::
+        The {{MLContext}}'s status for CPU fallback type (CPU or massively parallel processing).
```
huningxin: AFAIK, the major native ML runtimes, including Core ML, Windows ML (ONNX Runtime) and TFLite, enable CPU fallback by default. Some runtimes, e.g. ONNX Runtime, allow developers to disable CPU fallback explicitly through a session option `disable_cpu_ep_fallback` [1]. Without CPU fallback, model compilation may fail if the accelerator cannot execute all ops. Chromium prototype has a switch for that only for debugging purpose [2]. What are the other cases that a WebNN implementation may set this to false?

[1]: https://onnxruntime.ai/docs/genai/reference/config.html#sessionoptions
[2]: https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/webnn_switches.h;l=80;drc=fbd51a1eac8e9e8f6d0925730319fd2402b3ef47

index.bs
zolkis: Right, these would only be needed if there was an event (discussed earlier and agreed that polling is enough for now).

index.bs
zolkis: These steps were meant for the dispatch phase, when the _actual_ accelerators are selected.
If underlying accelerators cannot be modified during dispatch, then yes, these could be in the build steps, for static preparation.
However, if supported, they should also be included in the dispatch steps, which is the final decision point in dynamic execution.
I guess for now we could just move it to the build phase.

index.bs
zolkis: They are mentioned in the [device selection section](https://webmachinelearning.github.io/webnn/#programming-model-device-selection), though no formal definition is given.

If we wanted to give one, it's important to stress it's not a single device, but the final, eventually heterogeneous execution plan that maps specific parts of the model graph to the best available combination of accelerators at the exact moment of inference. 

During the build phase, we should not  _select_ a device, but define preferences (e.g. prioritized list of execution providers/delegates), which the runtime / underlying platform uses for the actual decisions.

index.bs
zolkis: Setting the CPU fallback option to false is when the application wants to have an (error) indication if massively parallel execution is not guaranteed with high chance (not an exact thing, but among many contradicting options, it's good enough). The use case is laid out in issue #815, see e.g. [comment](https://github.com/webmachinelearning/webnn/issues/815#issuecomment-2995936461), and the following discussion.
(Feel free to suggest other solutions.)

EDIT (w.r.t. where to check for CPU fallback): this use case would prefer early warning of CPU fallback likelihood (to be able to choose another inference path), so for that the checks make more sense in the build steps, indeed.

huningxin: > application wants to have an (error) indication if massively parallel execution is not guaranteed with high chance

How could an application indicate that? Should `MLContextOptions` add another property, something like `boolean cpuFallback`, default to true? An application can set `contextOptions.cpuFallback` to false for this use case.

zolkis: That was discussed in earlier calls (in the explainer related discussions): exposing a context option for setting CPU fallback to false hits some constraints and could be accomplished with the `accelerated` option, hence was discarded as an approach.

In #884 there is a code example for this use case:

```js
// create a context that should use massive parallel processing (e.g. GPU/NPU)
context = await navigator.ml.createContext({accelerated: true});
if (context.accelerated) {
    // the context will mostly use GPU/NPU, but CPU fallback may happen
} else {
    // the platform tells it likely cannot provide NPU or GPU, so try something else
}

// create a context that should preferably use NPU
context = await navigator.ml.createContext({accelerated: true, powerPreference: 'low-power'});
if (context.accelerated) {
    // NPU is likely used -- further requirements could be set by opSupportLimitsPerDevice
} else {
    // NPU is likely not available, and since GPU needs high power, it is not used
}
```

huningxin: Thanks for the code example. I understand an implementation should preferably use GPU/NPU if `accelerated` option is set to true. However, as I shared, the CPU fallback is enabled by default by major native ML runtimes. It's not clear to me how an implementation can tell an application wants to disable the CPU fallback. 

> could be accomplished with the accelerated option

Do you mean the implementation should disable CPU fallback if `accelerated` option is set to true? Then how could an application indicate it is fine with CPU fallback while preferring GPU/NPU execution?

zolkis: > It's not clear to me how an implementation can tell an application wants to disable the CPU fallback. 

No, that would not be supported (cannot be guaranteed). The use case is not to tell "disable CPU fallback", but to tell "I prefer massive parallel processing". So we don't expose disabling CPU fallback as a context option.

zolkis: But applications can find out if CPU fallback is being used, so the original Meet use case is covered by this, and by the context option for `accelerated` processing.

index.bs
zolkis: This opens again the discussion on the relationship between context and underlying execution device(s). I think we should not refer to a single device here, in the light of past discussions. 

In general we should bind the context not to a device, but to the execution plan (prioritized list of execution providers) mentioned above. Then, a separate concept (internal slot) would be the actual execution plan in the moment of inference. The text formulation should allow for a single device per context to supporting heterogeneous sub-graph execution on different devices. 

I think we could track that in a separate issue. In this PR, I have just removed the text "currently only the {{MLPowerPreference}} option" in line 751, and used the term from the device selection section in this algorithm. 

For this PR, I modify the text so that it's compatible with my explanation above.

index.bs
huningxin: Talked to @reillyeon offline today, the possible next step is: keep the `accelerated` option and device selection change in this PR, leave the `cpuFallbackActive` attribute change to a sperate issue for more discussion and clarification.

@zolkis , WDYT?

anssiko: This sounds reasonable to me.

@zolkis if you agree and are available, please open a separate issue for `cpuFallbackActive` and seed it with your insights. If you also update this PR accordingly we should be able to merge this PR by the end of the week.

zolkis: We have already removed the context option for preventing CPU fallback.

I'd like to understand the concerns with the `cpuFallbackActive` attribute. If it is it because of the polling steps, I already removed calling them from `graph.dispatch()` and didn't include them in `build()`, so there is only the getter, for which @handellm said would be good enough for Meet (instead of an event, which would present more issues).

Are there any further issues to be clarified, @huningxin , @reillyeon?

index.bs
```diff
@@ -1114,6 +1130,30 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
...
+        1. Return [=this=].{{MLContext/[[cpuFallbackActive]]}}.
+</div>
+
+<details open algorithm>
+  <summary>
+    To <dfn>poll CPU fallback status</dfn>, run the following steps.
+  </summary>
+    1. If [=this=].{{MLContext/[[accelerated]]}} is `false`, then:
+        1. Set [=this=].{{MLContext/[[cpuFallbackActive]]}} to `true` and return.
+    1. Issue a request to check whether the underlying platform uses CPU as the main underlying execution device for inference. If yes, then implementations <span class=allow-2119>should</span> set [=this=].{{MLContext/[[cpuFallbackActive]]}} to `true` and return.
```
handellm: Is "issue a request" achievable in a synchronous manner? If not perhaps this should be an async method.

index.bs
huningxin: According to the offline discussion, `cpuFallbackActive` seems to be a useful attribute of `MLGraph` (maybe coordinating with @philloooo 's proposal https://github.com/webmachinelearning/webnn/pull/854) rather than `MLContext`. I'll let @reillyeon and @philloooo chime in and share more thoughts.

index.bs
zolkis: True, the parlance needs changing, in order to avoid confusions. The underlying platform already knows this information - so the assumption is that it can be obtained synchronously (even if the operation would be asynchronous on OS level). Again, this speaks for an event based design. 
Nevertheless, looks like we are going to drop `cpuFallbackActive` for now.

index.bs
zolkis: > `cpuFallbackActive` seems to be a useful attribute of `MLGraph`

I agree, that makes a lot of sense. Even more, actually a sub-graph or individual ops might fall back to CPU (as mentioned before, a context / graph should be associated with an execution plan, not only with underlying execution devices).

For this PR, exposing `cpuFallbackActive` on context was chosen for the "simplicity" argument, also because a context still is associated with an underlying execution device -- for which we opened another issue in #897. Once we relax that and work with [these terms](https://github.com/webmachinelearning/webnn/issues/897#issuecomment-3455923178), I think we should properly address CPU fallback as well.

This is a good development, so I will remove it from this PR.

index.bs

zolkis:
@huningxin your feedbacks have been addressed:
- removed `cpuFallbackActive` from this PR (for later introduction);
- moved the `accelerated` consideration from the `dispatch()` steps to the `build()` steps.
When merging this PR, please keep the `cpuFallbackActive` commit in history.

index.bs
reillyeon: I agree with @zolkis's comments above that an `MLContext` represents a preferred order of execution providers (determined by power/acceleration preference) while only when you construct an `MLGraph` do you know what the actual execution plan for a given graph will look like.

huningxin (approved):
LGTM, thanks @zolkis !

anssiko:
I will now merge this and open a new issue #900 for more narrowly scoped discussion wrt the CPU fallback mechanism.

Again, thank you @zolkis and everyone who contributed to this feature through issue discussions, design proposals and review. This establishes a baseline for future device selection hints and enhancements.

fdwr:
> This establishes a baseline for future device selection hints and enhancements.

👍

I've mostly been observing the chatter on this one, but it seems we have multiple independent attributes that can contribute to device selection, including aspects like: how big a workload is, how immediately you want the result (a GPU can finish a large load faster than the CPU, but when desiring small latency with small workloads, the CPU is often better), whether it's continuous repeating work or single shot, desired power usage...

- **acceptable latency**: low, medium, high
- **continuity**: continuous, moderate, single shot
- **workload size**: small, medium, large
- **power usage**: low, medium, high
- ...

e.g.
| App scenarios             | preferred latency | workload size   | continuity | ideal device
|---------------------------|-------------------|-----------------|------------|--------------
| Realtime audio filtering  | low               | small | continuous | CPU/NPU (not GPU due to overhead of readback time)
| Audio to text             | ignorable         | medium | continuous | Any
| Realtime video processing | low               | large | continuous | GPU (since likely to be displayed afterward anyway)
| Offline video processing  | ignorable         | large | continuous | GPU/NPU (since large continuous workload)
| Image generation          | medium            | large | multiple executions | GPU/NPU (much faster than CPU)
| Image recognition         | ignorable         | small | single execution | Any (CPU is fast enough for simple recognition)
| Code editor prediction    | low               | small  | single execution | CPU since small workload and want to avoid pegging the GPU while typing

It's tempting to try to smash those all down into a concise set of enums, but I don't know if we can. With separate attributes, it reminds me conceptually of the font selection problem (family & weight & width & slope ...). 🤔

**Update 2025-11-09**; Issue created https://github.com/webmachinelearning/webnn/issues/902

anssiko:
@fdwr, thank you, this is great input. I agree this is analogous to font selection.

Would you mind opening a new issue for this, seed it with your comment? I'd like to put this on our F2F agenda to capture feedback on use case-driven priorities.

mtavenrath:
@fdwr I really like that list. 

Still I am curious how to really determine the perfect device and latency requirements? What if someone wants to launch a low latency, high compute intensive audio network? How is the WebNN backend supposed to know which dGPU to use if there are multiple dGPUs? What if the dGPU is slower than the iGPU, but has just been added to add certain display features like full sync or 4/8/16 displays? What if the dGPU is used by a WebGPU rendering process which would affect everything? What if a website wants to run 2 large models at once on two different dGPU?

Do we want WebNN to be limited to the simple use cases only or do we want to allow the complex use cases as well? There are ISVs in the professional space who port their software to the web and might be interested to choose a specific device. For those use cases it'd be great if it's possible to do an explicit device selection.

This leaves the question, how to not make such an API available to generic fingerprinting? With the risk of slightly lowering usability, would it make sense to make this feature available upon user-request only like webcams, audio or the GPU position?

By default only the simple API is available. If a more complex web application requires the advanced API it could request access and make the user aware that he is exposing more information from his system.

Issue #879, opened 2025-08-05
Propose to remove WPT conformance tests with negative scale of quantizeLinear operator


BruceDai:
According to [Quantize ONNX Models](https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html), there's below message

> scale is a positive real number used to map the floating point numbers to a quantization space.

And we also investigate current LLM Quantize Models, there's no case of using negative scale for `quantizeLinear` operator.

So I propose to remove [those WPT conformance tests with negative scale](https://github.com/web-platform-tests/wpt/blob/master/webnn/conformance_tests/quantizeLinear.https.any.js#L108).

@fdwr @huningxin @reillyeon Any suggestions? Thanks!

fdwr:
I guess we need to answer a few questions:

- Are they mathematically valid?
- Are they found in the wild?
- Do all backends support them?

## Are they mathematically valid?

Yes, and it's elegant for an operator's output to be equivalent to that operator's decomposition (and useful for conformance testing and a ground truth).

## Are they found in the wild?

Evidently they *can* appear in generated models (https://github.com/microsoft/onnxruntime-genai/issues/1051):

> When quantizing to INT4 precision, the model builder assumes that the scales are symmetric. When is_symmetric = True, the scales can be positive or negative.

So positive and negatives scales could be meaningful in certain contexts, such as symmetric quantization around zero or when modeling signed data distributions. Granted, ONNX says they should be positive, but the fact remains that they can be emitted and successfully execute with some ORT EP's.

## Do all backends support them?

- ORT, from the [above Phi-3 mini issue](https://github.com/microsoft/onnxruntime-genai/issues/1051) fails with some execution providers, whereas others work correctly like the DML EP. OpenVINO EP?
- CoreML rejects negative scale values (https://github.com/webmachinelearning/webnn/issues/822). If CoreML just removed the validation and permitted them through, it would probably compute the correct result.
- TF accepts negative scales fine (see below).

## Verdict

Positive and negatives scales are evidently useful in certain contexts, but also they're not widely supported by all the backends which WebNN targets, requiring either we:
- leave them as implementation defined. We can't really *validate* negative scales ahead of time because we'd have to inspect every scale value in the uploaded GPU resource, and we can't prevent them from reaching backends unless we relu them.
- decompose the `dequantizeLinear` call if a backend/EP doesn't support negative values, but that slows performance compared to the built-in operator to always decompose just because there *might* be some negative values.

[...]

PR #906, opened 2025-11-25
Restrict scale's values of dequantizeLinear and quantizeLinear to be positive and nonzero

BruceDai:
This is a Spec change PR for #879 where we're discussing about the lack of widespread support for the current negative scale of `dequantizeLinear` and `quantizeLinear` across all backend.

@fdwr @reillyeon @huningxin PTAL, thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/BruceDai/webnn/pull/906.html" title="Last updated on Jan 16, 2026, 8:38 AM UTC (eab6604)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/906/be76e5d...BruceDai:eab6604.html" title="Last updated on Jan 16, 2026, 8:38 AM UTC (eab6604)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -4106,7 +4106,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/dequantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else the behavior is [=implementation-defined=] (e.g. correct results, incorrect results or compilation failure).
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else the behavior is [=implementation-defined=] (e.g. correct results, incorrect results, or compilation failure).
         - <dfn>zeroPoint</dfn>: an {{MLOperand}}. The zero point tensor to subtract from each input value. It has the same [=MLOperand/shape=] as the scale.
         - <dfn>options</dfn>: an {{MLOperatorOptions}}. Specifies the optional parameters of the operation.
 
@@ -4280,7 +4280,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/quantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else behaviors are implementation dependent (e.g. correct results, incorrect results, compilation failure).
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else behaviors are implementation dependent (e.g. correct results, incorrect results, or compilation failure).
         - <dfn>zeroPoint</dfn>: an {{MLOperand}}. The zero point tensor to add to each rescaled input value. It has the same [=MLOperand/shape=] as the scale.
         - <dfn>options</dfn>: an {{MLOperatorOptions}}. Specifies the optional parameters of the operation.
 


index.bs
```diff
@@ -4333,6 +4334,7 @@ partial dictionary MLOpSupportLimits {
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input|, |scale|, and |zeroPoint| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/dataType=] is not one of its [=/allowed data types=] (according to [this table](#tensor-limits-quantizelinear)), then [=exception/throw=] a {{TypeError}}.
     1. If |scale|'s [=MLOperand/dataType=] is not one of its [=/allowed data types=] (according to [this table](#tensor-limits-quantizelinear)), then [=exception/throw=] a {{TypeError}}.
+    1. If |scale|'s values contain non-positive values, then [=exception/throw=] a {{TypeError}}.
```
philloooo: given scale is an MLOperand, not sure if we can synchronously throw a TypeError, . Is it ok to say if the scale contains non-positive values, it's implementation dependent on how the error is handled?

fdwr: Yeah, the spec should just say they are unsupported in prose. We can't actually validate this until runtime.

reillyeon: ...unless we introduce the concept of an `MLOperand` parameter that must be a build-time constant.

fdwr: > unless we introduce the concept of an MLOperand parameter that must be a build-time constant

There still remain dynamic cases when these inputs can be computed from another operator's output, but a constant property in the `MLOperand` may be informative for optimization purposes - for some precedent, there's already the `isConstant` slot and `constant` property getter from [`CreateConstantTensor`](https://webmachinelearning.github.io/webnn/#api-mlcontext-createconstanttensor).

BruceDai: Thanks for your reviewing! I've updated these descriptions in new commit, please take another look! ☕

index.bs
```diff
@@ -4106,7 +4106,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/dequantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input.
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. It must be positive.
```
fdwr: > It must be positive.

What about zero, which isn't really positive or negative 🤔? I'd either say "Negative scales are unsupported" if we're only concerned about negative values, or clearly say "They must be positive nonzero values" if zero was to be disallowed too. I think zero should be disallowed too because division by zero doesn't make sense for `quantizeLinear` (0 *would* be okay for `dequantizeLinear`, but that would be asymmetric).

index.bs
```diff
@@ -4170,6 +4170,7 @@ partial dictionary MLOpSupportLimits {
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any of |input|, |scale|, and |zeroPoint| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |input|'s [=MLOperand/dataType=] is not one of its [=/allowed data types=] (according to [this table](#tensor-limits-dequantizelinear)), then [=exception/throw=] a {{TypeError}}.
     1. If |scale|'s [=MLOperand/dataType=] is not one of its [=/allowed data types=] (according to [this table](#tensor-limits-dequantizelinear)), then [=exception/throw=] a {{TypeError}}.
+    1. If |scale|'s values contain non-positive values, they are unsupported.
```
fdwr: I'd just remove this line from the algorithm steps, since it's not an action.

fdwr (commented):
👀

index.bs
fdwr:
```suggestion
        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero.
```

index.bs
```diff
@@ -4280,7 +4280,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/quantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input.
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. It must be positive.
```
fdwr:
```suggestion
        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero.
```

index.bs
fdwr: > it's implementation dependent on how the error is handled

@philloooo: 👀I see one existing case with argMin/argMax, "In case of ties, the identity of the return value is implementation dependent". So we could update the current text "Values must be positive and nonzero" to something more like "Values must be positive and nonzero, or else results are implementation dependent"?

huningxin: > or else results are implementation dependent

It could be wrong results (i.e. OpenVINO) or compilation failure (i.e. CoreML), how could we cover both cases?

fdwr: > how could we cover both cases?

Ningxin - do you mean you want to list those possibilities out in the prose, such as:

> Values must be positive and nonzero, or else results are implementation dependent (e.g. correct results, incorrect results, compilation failure).

Or did you mean testing it in WPT? I don't think we need to test implementation dependent cases.

huningxin: > Ningxin - do you mean you want to list those possibilities out in the prose, such as:

Yes. But "wrong results" sounds like just to be incorrect output data. How about using "behavior", for example

> Values must be positive and nonzero, or else behaviors are implementation dependent (e.g. correct results, incorrect results, compilation failure).

BruceDai: Thanks! I've updated new commit accordingly.

fdwr (approved):
👍
(awaiting Phillis and Reilly before merging?)

index.bs
```diff
@@ -4106,7 +4106,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/dequantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input.
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else behaviors are implementation dependent (e.g. correct results, incorrect results, compilation failure).
```
reillyeon: Use the phrase "implementation-defined".

```suggestion
        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else the behavior is [=implementation-defined=] (e.g. correct results, incorrect results or compilation failure).
```

BruceDai: Thanks @reillyeon !👍
Updated in new commit. Please take another look  ☕

index.bs
```diff
@@ -4106,7 +4106,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/dequantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input.
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else the behavior is [=implementation-defined=] (e.g. correct results, incorrect results or compilation failure).
```
fdwr:
```suggestion
        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to multiply each input value by after adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else the behavior is [=implementation-defined=] (e.g. correct results, incorrect results, or compilation failure).
```

index.bs
```diff
@@ -4280,7 +4280,7 @@ partial dictionary MLOpSupportLimits {
 <div dfn-for="MLGraphBuilder/quantizeLinear(input, scale, zeroPoint, options)" dfn-type=argument>
     **Arguments:**
         - <dfn>input</dfn>: an {{MLOperand}}. The input tensor.
-        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input.
+        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else behaviors are implementation dependent (e.g. correct results, incorrect results, compilation failure).
```
fdwr:
```suggestion
        - <dfn>scale</dfn>: an {{MLOperand}}. The scale tensor to divide each input value by before adjusting by the zero point. It must be [=blockwise broadcastable=] with the input. Values must be positive and nonzero, or else behaviors are implementation dependent (e.g. correct results, incorrect results, or compilation failure).
```

BruceDai:
Thanks @fdwr for your update👍

fdwr:
@reillyeon - look good to you now?

reillyeon (approved):

Issue #904, opened 2025-11-10
2D, or not 2D, that is the question


fdwr:
## Issue

Multiple WebNN operators still have limited ranks (convolve, pool, and resample are all fixed to 4D tensor rank and 2D axes) which was [historically](https://github.com/webmachinelearning/webnn/issues/28) done for backends that might be more limited, but actually current backends (CoreML, DML, LiteRT, ORT) generally support more ranks than that, and this WebNN limitation has been problematic (e.g. Whisper uses 1D convolution, and so ORT's WebNN EP needed to [add extra reshape calls](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/webnn/builders/impl/conv_op_builder.cc#L132-L157) that prepend 1's to the inputs):

### Backend operator rank support

- Convolve (filtered windowed reduction)
  - [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.conv.conv) supports 3D-5D input with corresponding 1-3 reduction axes
  - [DirectML](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_convolution_operator_desc) supports 3D-5D with corresponding 1-3 reduction axes
  - [ONNX](https://onnx.ai/onnx/operators/onnx__Conv.html) supports 3D-5D input with corresponding 1-3 reduction axes
  - [TFLite](https://www.tensorflow.org/mlir/tfl_ops#tflconv_2d_tflconv2dop)/LiteRT support 4D-5D (3D is implementable via reshape)
- Pool* (windowed reduction)
  - [CoreML](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#module-coremltools.converters.mil.mil.ops.defs.iOS15.pool) supports 3D-5D input with corresponding 1-3 reduction axes
  - [DirectML](https://learn.microsoft.com/en-us/windows/win32/api/directml/ns-directml-dml_average_pooling_operator_desc) supports 4D-5D with corresponding 2-3 reduction axes (3D supported with some leading 1's)
  - [ONNX](https://onnx.ai/onnx/operators/onnx__AveragePool.html) supports 3D-5D input with corresponding 1-3 reduction axes
[...]

Issue #853, opened 2025-05-28
The minimum data type set


huningxin:
As a follow-up for the discussion at [WebML WG Teleconference – 22 May 2025](https://www.w3.org/2025/05/22-webmachinelearning-minutes.html#b53b), I checked the data types (and rank range) of all ops via opSupportLimits across DirectML, TFLite and CoreML backends in Chromium prototype. 

The result is promising: all ops have data type overlap except two ops:
* `l2Pool2d`: it is not implemented in TFLite backend yet ([Chromium issue](https://issues.chromium.org/issues/361717758)), because of a TFLite issue which is discussed in https://github.com/webmachinelearning/webnn/issues/278#issuecomment-2887950826. Once it is fixed, the overlap should be "float32" and "float16"
* `sign`: it is not implemented the CoreML backend yet ([Chromium issue](https://issues.chromium.org/issues/420724838)). [CoreML sign supports fp16, fp32, i32](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS15.elementwise_unary.sign). Once it is supported, the input data types intersection should be "float32", "float16" and "int32".

Please find the details in the following opSupportLimits intersection JOSN dump. /cc @reillyeon @fdwr 
<details>
  <summary>The opSupportLimits intersection of DirectML, TFLite and CoreML backends</summary>

The JOSN output
```json
{
  "abs": {
    "input": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    },
    "output": {
      "dataTypes": [ "float32", "float16", "int32" ]
    }
  },
  "add": {
    "a": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    },
    "b": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    },
    "output": {
      "dataTypes": [ "float32", "float16", "int32" ]
    }
  },
  "argMax": {
    "input": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 1, "max": 5 }
    },
    "output": {
      "dataTypes": [ "int32" ]
    }
  },
  "argMin": {
    "input": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 1, "max": 5 }
    },
[...]

Issue #278, opened 2022-07-01
How to define the algorithm of L2_Pool2d? 


mingmingtasd:
As you know, the algorithm of  L2_Pool2d is based on the 
Lp-normalization function which should be Y = **(X1^P + X2^P + ... + Xn^P) ^ (1/P)**.

But for L2_pool2d, I am not sure whether need to average the sum of elements as 
**Y =( (X1^2 + X2^2 + ... + Xn^2)/2) ^ (1/2)** or directly use Lp-normalization function as 
Y = **(X1^2 + X2^2 + ... + Xn^2) ^ (1/2)**.

I find two papers: https://sci-hub.yncjkj.com/10.1109/cvpr.2011.5995370 and https://sci-hub.yncjkj.com/10.1109/tcsvt.2015.2461978, they describe the Lp-normalization as below:

<img width="189" alt="Capture2" src="https://user-images.githubusercontent.com/52519066/163949663-af26d5ec-9aeb-4bb3-9fe2-a68bd517ca90.PNG">

<img width="205" alt="Capture" src="https://user-images.githubusercontent.com/52519066/163949688-3688d27b-29de-42e5-8398-08ba9e05d1e1.PNG">

So I confirm that **Lp-normalization function should be Y = (X1^P + X2^P + ... + Xn^P) ^ (1/P)**, but I am still not sure whether need to averge the sum of elements for L2_pool2d. I go through some framwork API spec and find the description as below:

**1. NNAPI [ANEURALNETWORKS_L2_POOL_2D](https://developer.android.com/ndk/reference/group/neural-networks#group___neural_networks_1ggaabbe492c60331b13038e39d4207940e0a2fb636e30d8853f9fa1a395e30660e92):**
output[b, i, j, c] =
    sqrt(sum_{di, dj} pow(input[b, strides[1] * i + di, strides[2] * j + dj, c], 2) /
         sum(1))
**2. ONNX [LpPool](https://github.com/onnx/onnx/blob/main/docs/Operators.md#lppool):**
LpPool consumes an input tensor X and applies Lp pooling across the tensor according to kernel sizes, stride sizes, and pad lengths. Lp pooling consisting of computing the Lp norm on all values of a subset of the input tensor according to the kernel size and downsampling the data into the output tensor Y for further processing.
**3. OpenVINO: Not Supported**
[...]

PR #910, opened 2025-12-10
add minimum data types and rank range for operations

### LINKED ISSUE #896: Add minimum data type set and rank range for input, constant, output and each operator into Spec
Current in Chromium code, we test against minimum data type set and rank range  using a workaround with [minimum_datatype_set.json](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/webnn/resources/minimum_datatype_set.json) which is based on @huningxin's comments in #853.

If WebNN API Spec documented this information, we can extract that from webnn.idl and generate target `minimum_datatype_set.json` for testing conveniently.

BruceDai:
This PR is to fix #896. 
@huningxin @fdwr @reillyeon PTAL, thanks!

***
<a href="https://pr-preview.s3.amazonaws.com/BruceDai/webnn/pull/910.html" title="Last updated on Jun 13, 2026, 3:27 PM UTC (8bf1fb2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/910/db1e51c...BruceDai:8bf1fb2.html" title="Last updated on Jun 13, 2026, 3:27 PM UTC (8bf1fb2)">Diff</a>


index.bs
```diff
@@ -2774,18 +2806,24 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{inputs}}'s [=list/items=]</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
     <td>1 to [=/any rank|N=]</td>
+    <td>1 to 5</td>
```
BruceDai, line 2828: @huningxin Last you commented on https://github.com/webmachinelearning/webnn/issues/853, the minimum ranks are `[0, 5]` as below, while the allowed ranks are `[1, N]`, so I added this information here as being `[1, 5]`. Am I right? Please help confirm it, thanks!

```
  "concat": {
    "inputs": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    },
    "output": {
      "dataTypes": [ "float32", "float16", "int32" ]
    }
  },
```

index.bs
```diff
@@ -9617,18 +10233,24 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
     <td>1 to [=/any rank|N=]</td>
+    <td>1 to 5</td>
```
BruceDai, line 10312: @huningxin please also confirm this information, it's similar to above `concat` operator, thanks!
You commented below on https://github.com/webmachinelearning/webnn/issues/853.
```
  "split": {
    "input": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    },
    "outputs": {
      "dataTypes": [ "float32", "float16", "int32" ]
    }
  },
```

index.bs
```diff
@@ -7776,24 +8244,57 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed data types=]</th>
+      <th>[=/minimum allowed data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>specified as part of operation steps</td>
+    <td></td>
```
BruceDai: According to @fdwr's comments on https://github.com/webmachinelearning/webnn/issues/278#issuecomment-2887950826, l2Pool2d is still not implemented in TFLite backend yet, so here I just followed @huningxin's comments on https://github.com/webmachinelearning/webnn/issues/853.

```
  "l2Pool2d": {
    "input": {
      "dataTypes": [],
      "rankRange": {}
    },
    "output": {
      "dataTypes": []
    }
  },
```

index.bs
```diff
@@ -3871,27 +4072,86 @@ partial dictionary MLOpSupportLimits {
...
+      <th>[=/allowed data types=]</th>
+      <th>[=/minimum allowed data types=]</th>
+      <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int64"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}</td>
+    <td></td>
```
BruceDai: This minimum information for `sign` operator is following @huningxin's comments on https://github.com/webmachinelearning/webnn/issues/853. 

```
  "sign": {
    "input": {
      "dataTypes": [],
      "rankRange": { "min": 0, "max": 0 }
    },
    "output": {
      "dataTypes": []
    }
  },
```

fdwr:
> I checked the unchanged Line 8987, it seems correct by using current 8 indents for </dl> to match front <dl> of also 8 indents. Any suggestions? Thanks!

@BruceDai : Remerge with main. My PR adds an extra blank line that pacifies the build error (and yes, that line was already correct, and nothing changed there for months, but bikeshed changed).

index.bs
```diff
@@ -1777,13 +1777,13 @@ The <dfn attribute for=MLOperand>shape</dfn> [=getter steps=] are to return [=th
...
 
-If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}.
+If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}. The <dfn>minimum allowed data types</dfn> is an intersection data types via {{MLContext/opSupportLimits()}} method across DirectML, TFLite and CoreML backends in Chromium prototype.
```
fdwr:
```suggestion
If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}. The <dfn>minimum allowed data types</dfn> is an intersection data types via {{MLContext/opSupportLimits()}} method across DirectML, ONNX Runtime, LiteRT, and CoreML backends in Chromium prototype.
```

index.bs
```diff
@@ -4401,7 +4401,7 @@ partial dictionary MLOpSupportLimits {
   <tr>
     <td>{{input}}</td>
     <td>{{MLOperandDataType/"uint8"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"uint32"}}, {{MLOperandDataType/"int32"}}</td>
-    <td>{{MLOperandDataType/"uint8"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"int32"}}</td>
+    <td>{{MLOperandDataType/"uint8"}}, {{MLOperandDataType/"int8"}}</td>
```
BruceDai, line 4406: This change is based on https://chromium-review.googlesource.com/c/chromium/src/+/7244911/4/third_party/blink/web_tests/external/wpt/webnn/resources/minimum_datatype_set.json#b182.

index.bs
```diff
@@ -1777,13 +1777,13 @@ The <dfn attribute for=MLOperand>shape</dfn> [=getter steps=] are to return [=th
...
+If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}. The <dfn>minimum allowed data types</dfn> is an intersection data types via {{MLContext/opSupportLimits()}} method across DirectML, TFLite and CoreML backends in Chromium prototype.
 
 Implementations may support fewer data types for operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
```
reillyeon: I think this is a clearer way to explain the difference between "allowed data types" and "minimum allowed data types". The same text should be updated for ranks.

```suggestion
If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}.

Implementations MAY support fewer data types for operands than specified, but MUST support at least the specified <dfn>minimum allowed data types</dfn>. Support can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.

The set of [=minimum data types=] has been determined based on implementation experience across a breadth of platforms to ensure that developers can produce interoperable content by designing models to only use these data types.
```

reillyeon (changes requested):

index.bs
BruceDai: Updated in new commit. Please take another look, thanks!

index.bs
```diff
@@ -1779,13 +1779,17 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
...
 
-Implementations may support fewer data types for operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
+Implementations MAY support fewer data types for operands than specified, but MUST support at least the specified <dfn>minimum allowed data types</dfn>. Support can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
+
+The set of [=/minimum allowed data types=] has been determined based on implementation experience across a breadth of platforms to ensure that developers can produce interoperable content by designing models to only use these data types.
 
 Issue: Should we specify the subset of data types that must be supported for each operator?
```
reillyeon: Remove this since this PR resolves the issue.

index.bs
```diff
@@ -1779,13 +1779,17 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
...
 
-Implementations may support fewer data types for operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
+Implementations MAY support fewer data types for operands than specified, but MUST support at least the specified <dfn>minimum allowed data types</dfn>. Support can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
+
+The set of [=/minimum allowed data types=] has been determined based on implementation experience across a breadth of platforms to ensure that developers can produce interoperable content by designing models to only use these data types.
```
reillyeon: Add, "The Web Platform Tests for this specification use this capability detection mechanism to validate correct behavior for all [=/allowed data types=] but can be passed with only support for the [=/minimum allowed data types=]."

reillyeon (approved):
General wording looks good. I have not validated the specific operator limits specified.

index.bs
BruceDai: Done. Thanks @fdwr !

index.bs
BruceDai: Updated! Thanks @reillyeon !

index.bs
```diff
@@ -1779,13 +1779,15 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
 
 If an operation supports only a subset of {{MLOperandDataType}}s, the <dfn>allowed data types</dfn> for each of the operation's input operands, including both positional arguments and options, are given as either an explicit list of {{MLOperandDataType}}s, or a constraint that the operand's [=MLOperand/dataType=] must be the <dfn lt="same type as">same as</dfn> the [=MLOperand/dataType=] of another input operand, or <dfn lt="any data type">any</dfn> to allow any {{MLOperandDataType}}.
 
-Implementations may support fewer data types for operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
+Implementations MAY support fewer data types for operands than specified, but MUST support at least the specified <dfn>minimum allowed data types</dfn>. Support can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
```
fdwr: > MUST support at least the specified <dfn>minimum allowed data types</dfn>

This adds an interesting implication back for ORT as a backend, since data type support varies *per EP*. Though, so long as the CPU EP exists, then fallback should satisfy that expectation (even if not the best performance due to device copies).

(no action expected - resolve me)

index.bs
fdwr: Concat requires >=1D input (so yes, "1 to 5"), since it functions on an axis, and the axiom must remain true that axis < rank, which would be invalid for rank 0.

index.bs
```diff
@@ -3435,29 +3501,71 @@ partial dictionary MLOpSupportLimits {
...
+      <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{a}}</td>
+    <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
```
fdwr: Which ML API is the outlier for pow being 0 to 4 instead of 0 to 5? It will be problematic for callers if mul/div/pow have inconsistent ranks from each other, as certain substitution operations might expect that they can safely chain these or switch from one to another (e.g. `mul(a,a) == pow(a,2)`) without surprise. I propose the outlier API fix this in their Chromium backend with flanking reshapes (flatten to 1D if > 4D, `pow`, then back to the original shape). Then `pow` can be remerged back with the rest above.

index.bs
```diff
@@ -2939,27 +2979,37 @@ partial dictionary MLOpSupportLimits {
...
+      <th>[=/minimum allowed data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>4</td>
```
fdwr, line 2998: This in correct for now, but likely relaxed in the future https://github.com/webmachinelearning/webnn/issues/904

(no action expected - resolve me)

index.bs
```diff
@@ -2389,38 +2397,52 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
     <td>1 to [=/any rank|N=]</td>
+    <td>3 to 5</td>
```
fdwr, line 2417: We probably want to support lower bound cases by just making the backends reshape them (basically a free operation), so that `batchNormalization` can consistently support 1-5 regardless of backend.

(no action expected now - resolve me)

index.bs
```diff
@@ -3622,32 +3730,127 @@ partial dictionary MLOpSupportLimits {
...
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 5</td>
+  </tr>
+  <tr>
+    <td>*output*</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 5</td>
```
fdwr, line 3763: It's not good for `logicalNot` to have a different rank from all the other logical operators. Otherwise it won't chain nicely with them, and there will be WebNN caller surprises. So either `logicalNot` should range 0 to 4 too like `logicalAnd` and `notEqual` (restricting it further), or the limiting backend should reshape > 4D cases (either directly to 1D, or explicitly `expand`ed if bidirectional broadcasting is needed, then to 1D), apply the logical operation, then reshape to the computed output shape.

*Ideally* for WebNN caller's sanity, we'd normalize all these simple elementwise operators (add, mul, pow, equal, logicalNot, identity...) consistently to the general 0 to 5 with backend reshapes, but at least the logical comparison family should be match to start with.

index.bs
fdwr: ? 0 doesn't make sense. Did you mean 0 to 5?

index.bs
```diff
@@ -3622,32 +3730,127 @@ partial dictionary MLOpSupportLimits {
...
 
+<table id=tensor-limits-elementwise-logical-part2 class='data' link-for="MLGraphBuilder/logicalNot(a, options)">
+  <caption>Tensor limits for {{MLGraphBuilder/logicalNot()}}</caption>
+  <thead>
+    <tr>
+      <th>operand</th>
+      <th>[=/allowed data types=]</th>
+      <th>[=/minimum allowed data types=]</th>
+      <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
```
fdwr:
```suggestion
      <th>[=/required ranks=]</th>
```

Calling something the "minimum allowed rank" is a little confusing, because the column actually includes both the minimum rank (0) *and* the maximum rank (5).

For consistency, "minimum allowed data types" -> "required data types".

index.bs
fdwr: Also empty data type? Shouldn't at least `float32` be supported by them all?

index.bs
```diff
@@ -4511,18 +4799,24 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed data types=]</th>
+      <th>[=/minimum allowed data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
```
fdwr, line 4812: 🤔 It's interesting that `gather` takes uint8 in all the API's apparently, but `expand` does not? Request double check? I'd generally expect in ML API's that any pure data movement operators take consistent data types.

index.bs
```diff
@@ -6232,27 +6610,37 @@ partial dictionary MLOpSupportLimits {
...
+      <th>[=/minimum allowed data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>4</td>
```
fdwr, line 6631: It's weird that `instanceNormalization` requires exactly rank 4, whereas `batchNormalization` takes a variable rank input and an axis, and `layerNormalization` takes a variable rank input axes. I guess it's just an old artifact of the original design (starting out with 4D requirements) that's never been relaxed.

(no action requested - just resolve, as I'll consider opening an issue for this separately)

index.bs
fdwr: That TFLite l2Pool2d issue has been pending since 2022-06-30 😳. The Chromium backend that calls TFLite could just emulate it for now to unblock us. e.g.:

```
powResult = pow(input, exponent)
poolSumResult = poolSum(powResult, axes, windowDimensions, padding, strides, dilations)
// where poolSum can be emulated via poolAverage(input * product(windowDimensions)...)
// or by convolution with a filter of 1's.
result = root(poolSumResult, exponent)
```

(no action expected in this PR - resolve me)

index.bs
```diff
@@ -7957,22 +8460,30 @@ partial dictionary MLOpSupportLimits {
...
     <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int64"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
   <tr>
     <td>{{slope}}</td>
     <td>[=/same type as|same as=] {{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
```
fdwr: elu and relu both show 0 to 5. Double check?

index.bs
```diff
@@ -4765,23 +5067,31 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed data types=]</th>
+      <th>[=/minimum allowed data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/minimum allowed ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"uint8"}}</td>
```
fdwr: `gatherElements` includes int8/uint8, but its symmetric counterpart `scatterElements` does not 😳 I don't know which API has this surprising inconsistency, but I'd rather restrict `gatherElements`'s required data types to exclude int8/uint8 than introduce this asymmetry.

index.bs
fdwr: Yep, 1 to 5.

index.bs
fdwr: Hmm, it's fishy to see `slice` take int8/uint8, whereas its counterpart `concat` does not. Either there is a typo, or I'd rather restrict `slice` to exclude int8/uint8 consistently with `concat`.

fdwr (commented):
👀 Thanks for the updates Bruce - I'm sure they were tedious :b.

index.bs
huningxin: > so I added this information here as being `[1, 5]`

It should be correct, thanks!

index.bs
huningxin: I suppose it was not implemented on every backend when I commented.

index.bs
BruceDai: Does Chromium implementation allow rank of `0` for the inputs of `concat` operator? If so, we need update it.

index.bs
BruceDai: I just checked these implementations
[https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/coreml/graph_builder_coreml…](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/coreml/graph_builder_coreml.cc;l=1263)

[https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/dml/context_impl_dml.cc;l=2…](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/dml/context_impl_dml.cc;l=254)

[https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/tflite/graph_builder_tflite…](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/tflite/graph_builder_tflite.cc;l=625)

[https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/ort/context_impl_ort.cc;l=2…](https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/ort/context_impl_ort.cc;l=217)

it should be
```
  "sign": {
    "input": {
      "dataTypes": ['float32', 'float16', 'int32'],
      "rankRange": { "min": 0, "max": 5 }
    },
    "output": {
      "dataTypes": ['float32', 'float16', 'int32'],
      "rankRange": { "min": 0, "max": 5 }
    }
  },
```

index.bs
BruceDai: > Which ML API is the outlier for pow being 0 to 4 instead of 0 to 5? 

It's limited by TFLite backend. 
Please refer to https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/tflite/graph_builder_tflite.cc;l=568

```c++
       // Limited to 4D when broadcasting is required:
       // https://source.chromium.org/chromium/chromium/src/+/main:third_party/tflite/src/tensorflow/lite/kernels/pow.cc
       /*pow_input=*/{kFloat16To32AndInt32, SupportedRanks::UpTo(4)},
```

index.bs
BruceDai: I just checked implementations, there're such these problems for CoreML, TFLite and DML backend, I submitted a [CL-7268347](https://chromium-review.googlesource.com/c/chromium/src/+/7268347) to fix them.

index.bs
BruceDai: `expand` by CoreML backend doesn't support `uint8` data type. Please review details.

Refer to https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/coreml/graph_builder_coreml.cc;l=1273

```c++
/*expand_input=*/{kFloatsAndInt32, kMaxRank},
```
and https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/coreml/graph_builder_coreml.cc;drc=7666bc1983c2a5b98e5dc6fa6c28f8f53c07d06f;l=1138
```c++
  static constexpr SupportedDataTypes kFloatsAndInt32{OperandDataType::kFloat16,
                                                      OperandDataType::kFloat32,
                                                      OperandDataType::kInt32};
```

index.bs
BruceDai: Updated in new commit, please take another look, thanks!

index.bs
BruceDai: Updated in new commit with this suggestion. Please take another look, thanks!

index.bs
BruceDai: Updated in new commit. Please take another look, thanks!

index.bs
huningxin: > I just checked implementations, there're such these problems for CoreML, TFLite and DML backend, I submitted a [CL-7268347](https://chromium-review.googlesource.com/c/chromium/src/+/7268347) to fix them.

`WebNNContextImpl::IntersectWithBaseProperties()` already excludes the scalar, please see https://source.chromium.org/chromium/chromium/src/+/main:services/webnn/webnn_context_impl.cc;l=399;drc=bcd9a1a07dd97bd7930d0950294ee1a47778db05

I also checked the intersection of concat limits of latest Chrome canary for CoreML, TFLite and ORT backends, the result is 
```json
"concat": {
    "inputs": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 1, "max": 5 }
    },
    "output": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 1, "max": 5 }
    }
  }
```

index.bs
huningxin:
```json
"split": {
    "input": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 1, "max": 5 }
    },
    "outputs": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 1, "max": 5 }
    }
  }
```

index.bs
huningxin:
```json
  "sign": {
    "input": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    },
    "output": {
      "dataTypes": [ "float32", "float16", "int32" ],
      "rankRange": { "min": 0, "max": 5 }
    }
  }
```

index.bs
huningxin: > @huningxin Last you commented on #853,

I updated the issue with latest opSupportLimits intersection of ORT, TFLite and CoreML backends of Chrome Canary: https://github.com/webmachinelearning/webnn/issues/853#issuecomment-3664995950

index.bs
```diff
@@ -2261,18 +2263,24 @@ partial dictionary MLOpSupportLimits {
...
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
     <td>1 to [=/any rank|N=]</td>
+    <td>1 to 5</td>
   </tr>
   <tr>
     <td>*output*</td>
     <td>{{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int64"}}</td>
+    <td>{{MLOperandDataType/"int32"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>1 to 5</td>
```
huningxin:
```suggestion
    <td>0 to 5</td>
```

index.bs
```diff
@@ -3435,29 +3501,71 @@ partial dictionary MLOpSupportLimits {
         - *pow*: Compute the values of the values of the first input tensor to the power of the values of the second input tensor, element-wise.
 </div>
 
-<table id=tensor-limits-elementwise-binary class='data' link-for="MLGraphBuilder/add(a, b, options), MLGraphBuilder/sub(a, b, options), MLGraphBuilder/mul(a, b, options), MLGraphBuilder/div(a, b, options), MLGraphBuilder/max(a, b, options), MLGraphBuilder/min(a, b, options), MLGraphBuilder/pow(a, b, options)">
-  <caption>Tensor limits for element-wise binary options</caption>
+<table id=tensor-limits-elementwise-binary-part1 class='data' link-for="MLGraphBuilder/add(a, b, options), MLGraphBuilder/sub(a, b, options), MLGraphBuilder/mul(a, b, options), MLGraphBuilder/div(a, b, options), MLGraphBuilder/max(a, b, options), MLGraphBuilder/min(a, b, options)">
```
huningxin:
```suggestion
<table id=tensor-limits-elementwise-binary-add-sub-mul-div-max-min class='data' link-for="MLGraphBuilder/add(a, b, options), MLGraphBuilder/sub(a, b, options), MLGraphBuilder/mul(a, b, options), MLGraphBuilder/div(a, b, options), MLGraphBuilder/max(a, b, options), MLGraphBuilder/min(a, b, options)">
```

index.bs
```diff
@@ -3435,29 +3501,71 @@ partial dictionary MLOpSupportLimits {
...
   <tr>
     <td>*output*</td>
     <td>[=/same type as|same as=] {{a}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>0 to 5</td>
+  </tr>
+</table>
+
+<table id==tensor-limits-elementwise-binary-part2 class='data' link-for="MLGraphBuilder/pow(a, b, options)">
```
huningxin:
```suggestion
<table id==tensor-limits-elementwise-binary-pow class='data' link-for="MLGraphBuilder/pow(a, b, options)">
```

index.bs
```diff
@@ -3622,32 +3730,127 @@ partial dictionary MLOpSupportLimits {
     **Returns:** an {{MLOperand}}. The output tensor that contains the result of element-wise comparison of the two input tensors.
 </div>
 
-<table id=tensor-limits-elementwise-logical class='data' link-for="MLGraphBuilder/equal(a, b, options), MLGraphBuilder/notEqual(a, b, options), MLGraphBuilder/greater(a, b, options), MLGraphBuilder/greaterOrEqual(a, b, options), MLGraphBuilder/lesser(a, b, options), MLGraphBuilder/lesserOrEqual(a, b, options), MLGraphBuilder/logicalNot(a, options), MLGraphBuilder/logicalAnd(a, b, options), MLGraphBuilder/logicalOr(a, b, options), MLGraphBuilder/logicalXor(a, b, options), MLGraphBuilder/isNaN(a, options), MLGraphBuilder/isInfinite(a, options)">
-  <caption>Tensor limits for element-wise logical options</caption>
+<table id=tensor-limits-elementwise-logical-part1 class='data' link-for="MLGraphBuilder/equal(a, b, options), MLGraphBuilder/notEqual(a, b, options), MLGraphBuilder/greater(a, b, options), MLGraphBuilder/greaterOrEqual(a, b, options), MLGraphBuilder/lesser(a, b, options), MLGraphBuilder/lesserOrEqual(a, b, options)">
```
huningxin:
```suggestion
<table id=tensor-limits-elementwise-logical-equal-notequal-greater-greaterorequal-lesser-lesserorequal class='data' link-for="MLGraphBuilder/equal(a, b, options), MLGraphBuilder/notEqual(a, b, options), MLGraphBuilder/greater(a, b, options), MLGraphBuilder/greaterOrEqual(a, b, options), MLGraphBuilder/lesser(a, b, options), MLGraphBuilder/lesserOrEqual(a, b, options)">
```

index.bs
```diff
@@ -3622,32 +3730,127 @@ partial dictionary MLOpSupportLimits {
...
   <tr>
     <td>*output*</td>
     <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
   </tr>
 </table>
 
+<table id=tensor-limits-elementwise-logical-part2 class='data' link-for="MLGraphBuilder/logicalNot(a, options)">
```
huningxin:
```suggestion
<table id=tensor-limits-elementwise-logical-logicalnot class='data' link-for="MLGraphBuilder/logicalNot(a, options)">
```

index.bs
```diff
@@ -3622,32 +3730,127 @@ partial dictionary MLOpSupportLimits {
...
+  <tr>
+    <td>*output*</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 5</td>
+  </tr>
+</table>
+
+<table id=tensor-limits-elementwise-logical-part3 class='data' link-for="MLGraphBuilder/logicalAnd(a, b, options), MLGraphBuilder/logicalOr(a, b, options), MLGraphBuilder/logicalXor(a, b, options)">
```
huningxin:
```suggestion
<table id=tensor-limits-elementwise-logical-logicaland-logicalor-logicalxor class='data' link-for="MLGraphBuilder/logicalAnd(a, b, options), MLGraphBuilder/logicalOr(a, b, options), MLGraphBuilder/logicalXor(a, b, options)">
```

index.bs
```diff
@@ -3622,32 +3730,127 @@ partial dictionary MLOpSupportLimits {
...
+  <tr>
+    <td>*output*</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>{{MLOperandDataType/"uint8"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
+  </tr>
+</table>
+
+<table id=tensor-limits-elementwise-logical-part4 class='data' link-for="MLGraphBuilder/isNaN(a, options), MLGraphBuilder/isInfinite(a, options)">
```
huningxin: Do the same for other table id.
```suggestion
<table id=tensor-limits-elementwise-logical-isnan-isinfinite class='data' link-for="MLGraphBuilder/isNaN(a, options), MLGraphBuilder/isInfinite(a, options)">
```

index.bs
```diff
@@ -3871,27 +4074,86 @@ partial dictionary MLOpSupportLimits {
...
+      <th>operand</th>
+      <th>[=/allowed data types=]</th>
+      <th>[=/required data types=]</th>
+      <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>specified as part of operation steps</td>
```
huningxin: As we already make separate tables, should we reflect the data types in the corresponding table instead?

index.bs
```diff
@@ -4133,28 +4395,38 @@ partial dictionary MLOpSupportLimits {
...
     <td>{{input}}</td>
     <td>{{MLOperandDataType/"uint8"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"uint32"}}, {{MLOperandDataType/"int32"}}</td>
+    <td>{{MLOperandDataType/"uint8"}}, {{MLOperandDataType/"int8"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>0 to 5</td>
   </tr>
   <tr>
     <td>{{scale}}</td>
     <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>{{MLOperandDataType/"float32"}}</td>
```
huningxin: also support float16

index.bs
```diff
@@ -4133,28 +4395,38 @@ partial dictionary MLOpSupportLimits {
...
     <td>{{zeroPoint}}</td>
     <td>[=/same type as|same as=] {{input}}</td>
+    <td>{{MLOperandDataType/"uint8"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"int32"}}</td>
     <td>[=/same rank as|same as=] {{input}}</td>
+    <td>0 to 5</td>
   </tr>
   <tr>
     <td>*output*</td>
     <td>[=/same type as|same as=] {{scale}}</td>
+    <td>{{MLOperandDataType/"float32"}}</td>
```
huningxin: also support float16

index.bs
```diff
@@ -4307,28 +4579,38 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed data types=]</th>
+      <th>[=/required data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>{{MLOperandDataType/"float32"}}</td>
```
huningxin: also supports float16

index.bs
```diff
@@ -4307,28 +4579,38 @@ partial dictionary MLOpSupportLimits {
...
     <td>[=/any rank|N=]</td>
+    <td>0 to 5</td>
   </tr>
   <tr>
     <td>{{scale}}</td>
     <td>[=/same type as|same as=] {{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}</td>
```
huningxin: also supports float16

index.bs
huningxin: tflite prelu has 0 to 4.

index.bs
```diff
@@ -7586,18 +8050,24 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>1 to 5</td>
```
huningxin: The latest data shows 0 to 5.

index.bs
```diff
@@ -7586,18 +8050,24 @@ partial dictionary MLOpSupportLimits {
...
   </tr>
   <tr>
     <td>*output*</td>
     <td>[=/same type as|same as=] {{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
     <td>[=/same rank as|same as=] {{input}}</td>
+    <td>1 to 5</td>
```
huningxin: 0 to 5

index.bs
```diff
@@ -7776,24 +8246,57 @@ partial dictionary MLOpSupportLimits {
...
+      <th>operand</th>
+      <th>[=/allowed data types=]</th>
+      <th>[=/required data types=]</th>
+      <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{input}}</td>
+    <td>specified as part of operation steps</td>
```
huningxin: Should we reflect allowed data types in table?

index.bs
```diff
@@ -9264,18 +9858,24 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"uint8"}}</td>
     <td>[=/any rank|N=]</td>
+    <td>1 to 5</td>
```
huningxin: 0 to 5, scalar is allowed when starts and ends are empty.

index.bs
```diff
@@ -9264,18 +9858,24 @@ partial dictionary MLOpSupportLimits {
...
   </tr>
   <tr>
     <td>*output*</td>
     <td>[=/same type as|same as=] {{input}}</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}, {{MLOperandDataType/"int8"}}, {{MLOperandDataType/"uint8"}}</td>
     <td>[=/same rank as|same as=] {{input}}</td>
+    <td>1 to 5</td>
```
huningxin: 0 to 5

index.bs
```diff
@@ -9988,18 +10630,24 @@ partial dictionary MLOpSupportLimits {
...
       <th>[=/allowed data types=]</th>
+      <th>[=/required data types=]</th>
       <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
     </tr>
   </thead>
   <tr>
     <td>{{input}}</td>
     <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
```
huningxin: ORT doesn't support int32

index.bs
fdwr: True, if `keepDimensions` is false, output can be a scalar.

index.bs
fdwr: > CoreML backend doesn't support uint8 data type

:( Understood.

index.bs
fdwr: Gotcha. I'll likely create CR bugs to rationalize some of these outliers and reduce caller surprises elu=5, relu=5, prelu=4.

index.bs
fdwr: True, same with `pad`.

index.bs
fdwr: 🤔 An ONNX [`BitCast` operator](https://github.com/onnx/onnx/issues/6887) could help with cases of pure data movement (since no actual math happens) like this one when the operator supports a different data type of the same bit size (sizeof(int32) == sizeof(uint32) == sizeof(float32)). For example, we could call BitCast(int32input, type=float32) -> Triangular -> BitCast(input, type=int32). I'll comment on that ONNX PR.

(update) BitCast operator is pending: https://github.com/onnx/onnx/pull/7527

index.bs
BruceDai: Thanks @huningxin! 
Resolved this comment by already adding this information in new commit.

index.bs
BruceDai: @huningxin In operation steps, the input of `averagePool2d` supports "float32", "float16" two data types, and the input of `maxPool2d` supports `any` data types, so I keep pervious descriptions here.

index.bs
BruceDai: @huningxin In operation steps, the input of `abs` and `neg` supports "float32", "float16", "int64", "int32" and "int8" five data types, the input of `identity` supports `any` data types, so I keep previous descriptions here.

BruceDai:
Thanks @huningxin and @fdwr for review and identifications. 👍
I submitted a new commit to address your comments. Please take another look, thanks!☕

index.bs
fdwr: Can we open a TFLite issue? Alternately we open a Chromium issue to {reshape, expand if needed, pow} until TFLite can address it.

BruceDai: Filed a Chromium issue - https://issues.chromium.org/issues/471017637.

index.bs
huningxin: I meant making a separate table for `identity`, so that we can list data types in the table rather than embedding them into the algorithm steps.

index.bs
huningxin: > I don't know which API has this surprising inconsistency

It's CoreML: https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS17.scatter_gather.scatter_along_axis

index.bs
huningxin: I meant by making separate table, you can list the supported data types in the table cell rather than embedding them into the algorithm steps.

index.bs
huningxin: Updated https://issues.chromium.org/issues/361717758#comment4 with @fdwr 's emulation code. Thanks!

huningxin: l2Pool2d is supported in Chrome 145.0.7626.0 canary, the updated minimum data type is
```json
 "l2Pool2d": {
    "input": {
      "dataTypes": [ "float32", "float16" ],
      "rankRange": { "min": 4, "max": 4 }
    },
    "output": {
      "dataTypes": [ "float32", "float16" ],
      "rankRange": { "min": 4, "max": 4 }
    }
  }
```

index.bs
BruceDai: Done by new commit of [b6b300f](https://github.com/webmachinelearning/webnn/pull/910/commits/b6b300f3a16e9e5b8fbec71e49cf58e2227ae380). Please take another look, thanks!

index.bs
BruceDai: Done by new commit of [b6b300f](https://github.com/webmachinelearning/webnn/pull/910/commits/b6b300f3a16e9e5b8fbec71e49cf58e2227ae380). Please take another look, thanks!

index.bs
BruceDai: Updated by new commit of [b6b300f](https://github.com/webmachinelearning/webnn/pull/910/commits/b6b300f3a16e9e5b8fbec71e49cf58e2227ae380). Please take another look, thanks!

index.bs
BruceDai: I filed a chromium issue [[TFLite] Support the maximum ranks of input for logicalAnd, logicalOr and logicalXor be greater than 4](https://issues.chromium.org/issues/478988222) for it.

index.bs
BruceDai: I filed a chromium issue [
[Core ML] The supported data types of input should be matched for gatherElements and scatterElements](https://issues.chromium.org/issues/478988225) for it.

index.bs
BruceDai: I filed a chromium issue [[TFLite] Support the maximum ranks of input for prelu be greater than 4](https://issues.chromium.org/issues/478888160) for it.

fdwr (approved):
👍

index.bs
BruceDai: Above issue [[Core ML] The supported data types of input should be matched for gatherElements and scatterElements](https://issues.chromium.org/issues/478988225) has been fixed by my [CL-7726068: exclude int8/uint8 for input of gatherElements in CoreML](https://chromium-review.googlesource.com/c/chromium/src/+/7726068). 
Now the required data types for `gatherElements` and `scatterElements` are the same data types as `float32`, `float16` and `int32`.

fdwr: Thank you sir 🙏.

index.bs
fdwr: (update) BitCast is in ONNX https://onnx.ai/onnx/operators/onnx__BitCast.html, now enabling callers of the ORT backend to support all data types of the same bit size for data movement operators, so long as the EP supports `BitCast` and at least one data type of a given bit size (e.g. supporting `int16` with `Gather` also enables `uint16`, `float16`, `bfloat16`...).

index.bs
BruceDai: My [CL-7768249 - Emulate rank-5 prelu's input tensors in TFLite](https://chromium-review.googlesource.com/c/chromium/src/+/7768249) was merged to fix above  [[TFLite] Support the maximum ranks of input for prelu be greater than 4](https://issues.chromium.org/issues/478888160) issue. So , the maximum rank of input tensors for `prelu`, `relu`, and `elu` operators are unified with support up to rank 5 now. 
@fdwr @huningxin I submitted a new commit to address that. PTAL, thanks!

mwyrzykowski (approved):
Conflict with the WG meeting, but ✅

fdwr (approved):
👍

index.bs
```diff
@@ -1786,13 +1786,15 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
...
 
-Implementations may support fewer data types for operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation.
+Implementations MAY support fewer data types for operands than specified, but MUST support at least the specified <dfn>required data types</dfn>. Support can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/dataTypes}} value of the corresponding member for the operation across DirectML, ONNX Runtime, LiteRT, and CoreML backends in Chromium prototype now.
```
huningxin: Chromium impl has removed DirectML backends.

index.bs
```diff
@@ -1786,13 +1786,15 @@ Since the {{MLOperand/[[builder]]}} object is bound by the {{MLGraphBuilder/cons
...
 
-Implementations may impose a more restricted lower bound and/or upper bound on the [=MLOperand/rank=] of operands than specified. This can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/rankRange}}.{{MLRankRange/min}} and {{MLTensorLimits/rankRange}}.{{MLRankRange/max}} values of the corresponding member for the operation.
+Implementations MAY impose a more restricted lower bound and/or upper bound on the [=MLOperand/rank=] of operands than specified, but MUST support at least the specified <dfn>required ranks</dfn>. Support can be queried for each operation using the {{MLContext/opSupportLimits()}} method on {{MLContext}} and inspecting the {{MLTensorLimits/rankRange}}.{{MLRankRange/min}} and {{MLTensorLimits/rankRange}}.{{MLRankRange/max}} values of the corresponding member for the operation across DirectML, ONNX Runtime, LiteRT, and CoreML backends in Chromium prototype now.
```
huningxin: Same here.

index.bs
huningxin: > Filed a Chromium issue - https://issues.chromium.org/issues/471017637.

@BruceDai , what's your plan for fixing this issue?

index.bs
huningxin: > I filed a chromium issue [[TFLite] Support the maximum ranks of input for logicalAnd, logicalOr and logicalXor be greater than 4](https://issues.chromium.org/issues/478988222) for it.

@BruceDai , what's your plan for fixing this issue? Generalizing them to 0 to 5D with backend reshapes sounds good to me.

BruceDai:
> @BruceDai , what's your plan for fixing this issue? Generalizing them to 0 to 5D with backend reshapes sounds good to me.

@huningxin Thanks for your kindly reminder! Please review the fixing [CL-7848774](https://chromium-review.googlesource.com/c/chromium/src/+/7848774) .

index.bs
BruceDai: Done by [CL-7848774](https://chromium-review.googlesource.com/c/chromium/src/+/7848774) being landed.

BruceDai:
Thanks @reillyeon @huningxin and @fdwr for reviewing and approving [CL-7848774](https://chromium-review.googlesource.com/c/chromium/src/+/7848774), now it has been landed. 
Please take another review the new commit. Thanks! ☕

fdwr (approved):
👍 Thank you for enabling this.

index.bs
huningxin: > > Filed a Chromium issue - https://issues.chromium.org/issues/471017637.
> 
> [@BruceDai](https://github.com/BruceDai) , what's your plan for fixing this issue?

@BruceDai , I guess this issue can be fixed in a similar way of [7848774: webnn: align logical op rank limits at 5 on the TFLite backend](https://chromium-review.googlesource.com/c/chromium/src/+/7848774)?

index.bs
```diff
@@ -3629,32 +3737,127 @@ partial dictionary MLOpSupportLimits {
...
+      <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{a}}</td>
+    <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}, {{MLOperandDataType/"int32"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
```
huningxin: Should we also generalize them to 0 to 5D? I already see a CL relies on 5D lesser in TFLite backend: [webnn: Support >4D indices for gather/gatherND/scatterND](https://chromium-review.googlesource.com/c/chromium/src/+/7866073)

index.bs
```diff
@@ -3629,32 +3737,127 @@ partial dictionary MLOpSupportLimits {
...
+      <th>[=/allowed ranks=]</th>
+      <th>[=/required ranks=]</th>
+    </tr>
+  </thead>
+  <tr>
+    <td>{{a}}</td>
+    <td>[=/any data type|any=]</td>
+    <td>{{MLOperandDataType/"float32"}}, {{MLOperandDataType/"float16"}}</td>
+    <td>[=/any rank|N=]</td>
+    <td>0 to 4</td>
```
huningxin: Do the same? Then we will have all logical operators supporting 0 to 5D.

fdwr: > Then we will have all logical operators supporting 0 to 5D.

Hmm, I wonder how far we are from all elementwise ops ranging 0 through 5?

huningxin: I guess two or three CLs for Chromium TFLite backend?
1. pow and logical binary ops
2. isNan/isInfinite

[7848774: webnn: align logical op rank limits at 5 on the TFLite backend](https://chromium-review.googlesource.com/c/chromium/src/+/7848774) already landed necessary helpers.

fdwr: ✅ pow and logical binary ops

BruceDai: Done by below three CLs: 

- [x] [CL-7899814](https://chromium-review.googlesource.com/c/chromium/src/+/7899814) - emulate rank-5 pow's input tensors in TFLite backend
- [x] [CL-7914923](https://chromium-review.googlesource.com/c/chromium/src/+/7914923) - emulate rank-5 input tensors for six logical comparison ops on TFLite
- [x] [CL-7927875](https://chromium-review.googlesource.com/c/chromium/src/+/7927875) - raise input rank limit to 5 for isNaN and isInfinite on TFLite

index.bs
BruceDai: Done by landed [CL-7899814](https://chromium-review.googlesource.com/c/chromium/src/+/7899814/) - emulate rank-5 pow's input tensors in TFLite backend.

BruceDai:
@huningxin @fdwr Now those remaining issues are all resolved by landing CLs into Chromium. Please take another look, thanks!

index.bs
BruceDai: Done.

fdwr (approved):
👍

huningxin (approved):
LGTM, @BruceDai thanks for your relentless effort!

huningxin:
I suppose this PR is ready to merge. Any further comments? @reillyeon and @anssiko

huningxin:
I am going to merge this PR. Thanks all again for the PR and review!

PR #916, opened 2026-01-13
fix reduceLogSumExp, softplus, and tanh decompositions to not overflow so easily

vmukhachev:
test case:
reduceLogSumExp([89]) should give 89, not Infinity

***
<a href="https://pr-preview.s3.amazonaws.com/vmukhachev/webnn/pull/916.html" title="Last updated on Jan 13, 2026, 9:46 PM UTC (5e297be)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/916/91745cf...vmukhachev:5e297be.html" title="Last updated on Jan 13, 2026, 9:46 PM UTC (5e297be)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -9487,8 +9487,7 @@ partial dictionary MLOpSupportLimits {
     </summary>
     <pre highlight="js">
     function softplus(builder, input) {
-      return builder.log(
-        builder.add(builder.exp(input), builder.constant(input.dataType, 1)));
+      return builder.add(input, builder.log(builder.add(builder.constant(input.dataType, 1), builder.exp(builder.neg(input)))));
     }
     </pre>
   </details>
--- a/index.bs
+++ b/index.bs
@@ -9778,13 +9778,13 @@ partial dictionary MLOpSupportLimits {
     </summary>
     <pre highlight="js">
     function tanh(builder, input) {
-      return builder.div(
+      return builder.mul(builder.sign(input), builder.div(
         builder.sub(
           builder.constant(input.dataType, 1),
-          builder.exp(builder.mul(builder.constant(input.dataType, -2), input))),
+          builder.exp(builder.mul(builder.constant(input.dataType, -2), builder.abs(input)))),
         builder.add(
           builder.constant(input.dataType, 1),
-          builder.exp(builder.mul(builder.constant(input.dataType, -2), input))));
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -8270,7 +8270,8 @@ partial dictionary MLOpSupportLimits {
     }
 
     function reduceLogSumExp(builder, input, options) {
-      return builder.log(builder.reduceSum(builder.exp(input), options));
+      const maxX = builder.reduceMax(input, {axes: options.axes, keepDimensions: true});
+      return builder.add(maxX, builder.log(builder.reduceSum(builder.exp(builder.sub(input, maxX)), options)));
```
fdwr, line 8274: > reduceLogSumExp([89]) should give 89, not Infinity

That is certainly nicer output 🙂 (similarly one would expect `pow(pow(89, 2), 0.5) == 89`), and indeed, that alternate expression is a common formula for increased numerical stability 👍, but I would still keep the old formula around as the most straight-forward mapping for readers, then add a comment that this alternate optimized form improves precision. e.g.

```js
    function reduceLogSumExp(builder, input, options) {
      return builder.log(builder.reduceSum(builder.exp(input), options));
      // Note the decomposition below improves numerical precision:
      // const maxX = builder.reduceMax(input, {axes: options.axes, keepDimensions: true});
      // return builder.add(maxX, builder.log(builder.reduceSum(builder.exp(builder.sub(input, maxX)), options)));
    }
```

And likewise below.

vmukhachev: @fdwr , Chrome implementation has this bug on Windows, may be I should add a test instead to the test repo.

fdwr: Or maybe we reverse that (after https://github.com/webmachinelearning/webnn/issues/947), and make the simpler decomposition the comment:

```js
    function reduceLogSumExp(builder, input, options) {
      const maxX = builder.reduceMax(input, {axes: options.axes, keepDimensions: true});
      return builder.add(maxX, builder.log(builder.reduceSum(builder.exp(builder.sub(input, maxX)), options)));

      // Note the decomposition above improves numerical precision, versus the simpler
      // direct decomposition:
      //
      // return builder.log(builder.reduceSum(builder.exp(input), options));
    }
```

vmukhachev: It would be easier to remove it altogether then

PR #923, opened 2026-03-18
Refactor device selection: Rename to computePolicy, remove accelerated, and add fallback

### LINKED ISSUE #911: accelerated should be prior to powerPreference for device selection
According to https://webmachinelearning.github.io/webnn/#dictdef-mlcontextoptions, the current WebNN specification states that the `accelerated` option has lower priority than `powerPreference`. This design creates logical contradictions and implementation difficulties, particularly in two scenarios:

**low-power + accelerated = true**: 
If no low-power device NPU/iGPU is available, it seems we have to select CPU to satisfy the "low-power" requirement. This violates the developer's explicit intent to use hardware acceleration.   

**high-performance/low-power  + accelerated = false**: 
The spec suggests choosing a "high-performance/low-power CPU mode". However, web browsers typically lack the granular control to force a CPU into a high-performance/low-power state (e.g., pinning to performance cores or boosting clock speeds), making this practically impossible to implement

So the logic should be inverted: `accelerated` has higher priority than `powerPreference`.

Rationale: `accelerated` defines the category of hardware (Dedicated Accelerator vs. CPU), while `powerPreference` should act as a hint for selecting within that category. If accelerated is true, the user agent MUST return an accelerated context (GPU/NPU) if available, regardless of power preference.
See the following table, this will make device selection implementation easier and avoid some confusions from contradictory combinations.

accelerated | powerPreference | device should be selected
-- | -- | -- 
true | "default"/"higher-performance" | GPU 
true | "low-power" | NPU; if no NPU, try to select iGPU
false | ignored | CPU

@huningxin @fdwr

mingmingtasd:
To fix #911 

**Description:**
This PR refactors the device selection preference API to establish a more extensible framework by replacing MLPowerPreference with MLComputePolicy in MLContextOptions.

**Key changes included:**

1. **Renamed preference enum**: MLPowerPreference has been refactored to MLComputePolicy (and the corresponding MLContextOptions dictionary member updated to computePolicy).
2. **Clarified extensibility**: The spec now explicitly states that the policy is designed to be extensible to accommodate future execution profiles, moving beyond just power and performance metrics.
3. **Removed "accelerated" option**: Removed the accelerated value to eliminate ambiguity, as performance and power semantics are better covered by other profiles.
4. **Introduced "fallback" policy**: Added a new "fallback" compute policy. This option prioritizes maximum compatibility over other considerations (typically running on a CPU) and is highly useful for verifying a model's numeric behavior without utilizing parallel accelerators like GPUs or NPUs.

The corresponding chromium CL is https://chromium-review.googlesource.com/c/chromium/src/+/7513189

PTAL, thanks! @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/mingmingtasd/webnn/pull/923.html" title="Last updated on May 7, 2026, 1:50 AM UTC (be6da78)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/923/a4a7eb6...mingmingtasd:be6da78.html" title="Last updated on May 7, 2026, 1:50 AM UTC (be6da78)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1095,7 +1095,6 @@ interface MLContext {
 
   undefined destroy();
 
-  readonly attribute MLComputePolicy computePolicy;
   readonly attribute Promise<MLContextLostInfo> lost;
 };
 </script>
@@ -1127,10 +1126,6 @@ The <dfn>context type</dfn> is the type of the execution context that manages th
 <dt>"<dfn>webgpu</dfn>"</dt>
 <dd>Context created from WebGPU device.</dd>
 </dl>
-
-<div algorithm>
-The <dfn attribute for=MLContext>computePolicy</dfn> getter steps are to return [=this=].{{MLContext/[[computePolicy]]}}.
-</div>
 
 <details open algorithm>
   <summary>
--- a/index.bs
+++ b/index.bs
@@ -1011,7 +1011,7 @@ The <dfn dfn-for=MLContextOptions dfn-type=dict-member>computePolicy</dfn> optio
 <dt>"<dfn enum-value>low-power</dfn>"</dt>
 <dd>Prioritizes power consumption over other considerations such as execution speed.</dd>
 <dt>"<dfn enum-value>fallback</dfn>"</dt>
-<dd>Prioritizes maximum compatibility over other considerations, typically running on a CPU. This is useful for testing a model's numeric behavior without utilizing parallel accelerators like GPUs or NPUs.</dd>
+<dd>Prioritizes wider compatibility, more predictable behavior, or improved privacy over other considerations such as execution speed.</dd>
 </dl>
 
 ### {{ML/createContext()}} ### {#api-ml-createcontext}
--- a/index.bs
+++ b/index.bs
@@ -980,15 +980,15 @@ WorkerNavigator includes NavigatorML;
 
 ## {{ML}} interface ## {#api-ml}
 <script type=idl>
-enum MLPowerPreference {
+enum MLComputePolicy {
   "default",
   "high-performance",
-  "low-power"
+  "low-power",
+  "fallback"
 };
 
 dictionary MLContextOptions {
-  MLPowerPreference powerPreference = "default";
-  boolean accelerated = true;
+  MLComputePolicy computePolicy = "default";
 };
 
 [SecureContext, Exposed=(Window, Worker)]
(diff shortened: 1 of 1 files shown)


mwyrzykowski (approved):

index.bs
```diff
@@ -1002,18 +1002,18 @@ interface ML {
...
+<dl dfn-for="MLComputePolicy">
 <dt>"<dfn enum-value>default</dfn>"</dt>
 <dd>Let the user agent select the most suitable behavior.</dd>
 <dt>"<dfn enum-value>high-performance</dfn>"</dt>
-<dd>Prioritizes execution speed over power consumption.</dd>
+<dd>Prioritizes execution speed over other considerations such as power consumption.</dd>
 <dt>"<dfn enum-value>low-power</dfn>"</dt>
 <dd>Prioritizes power consumption over other considerations such as execution speed.</dd>
+<dt>"<dfn enum-value>fallback</dfn>"</dt>
+<dd>Prioritizes maximum compatibility over other considerations, typically running on a CPU. This is useful for testing a model's numeric behavior without utilizing parallel accelerators like GPUs or NPUs.</dd>
```
handellm: This policy is useful for audio processing that doesn't want to depend on accelerators due to latency reasons. Can the text be updated to mention this case?

zolkis (approved):
Thanks!

index.bs
fdwr: (just naming) Hmm, reading the *description* of this enum ("Prioritizes maximum *compatibility*...", "useful for testing a model's *numeric behavior*..."), some more immediately enlightening names to readers would be something like "compatible"/"compatibility"/"stable"/"precision". I would have not guessed that description from the word "fallback", which implies you fell back to a less capable device than the one you really wanted, when actually CPU may have been *exactly* what you wanted (that is, it was the primary preference, not a fallback). e.g.

```js
enum MLComputePolicy {
  "default",
  "high-performance",
  "low-power",
  "compatible", // or stable/local...
};
```
> This policy is useful for audio processing that doesn't want to depend on accelerators due to latency reasons

Totally, as I recall some teams in the past complaining about GPU overhead for background audio filtering in chat apps, preferring to keep compute more *local* on the CPU. Perhaps as a separate PR, we could add an explicit "low-latency" option too, which would be even clearer in intent for that scenario.

Also "precision" would be a useful preference, as some NPU's and GPU's chop off low bits (also, separate PR).

index.bs
```diff
@@ -1002,18 +1002,18 @@ interface ML {
...
 </dl>
 
-The <dfn dfn-for=MLContextOptions dfn-type=dict-member>accelerated</dfn> option indicates the application's preference as related to massively parallel acceleration. This option has less priority than {{MLContextOptions/powerPreference}}. When set to `true` (by default), the underlying platform will attempt to use the available massively parallel accelerators, such as a GPU or NPU, also depending on the {{MLContextOptions/powerPreference}}. When set to `false`, the application indicates it prefers CPU inference. If there is contradictory input, for instance when {{MLContextOptions/powerPreference}} is {{MLPowerPreference/"high-performance"}} and {{MLContextOptions/accelerated}} is `false`, then the implementation will choose the best available match in the underlying platform (for instance a high performance CPU mode, or will ignore {{MLContextOptions/accelerated}} as it has less priority than {{MLContextOptions/powerPreference}}).
```
fdwr, line 1015: 🤔 Yeah, I like `accelerated` going away, as it felt like a weird side knob.

fdwr (approved):
Consider clearer naming, otherwise LGTM sir.

index.bs
zolkis: +1 for both the "compatible" (in this PR), adding "low-latency" in a next PR, and adding a preference/hint for "precision" (to `MLContextOptions`?).

handellm: Let me throw in "non-accelerated" here or even "cpu" if we don't like "fallback". Reading "compatible" makes me be mildly suspecting an accelerator could still be in the mix :)

mwyrzykowski: Not tied to `"fallback"` but it does align with existing WebGPU naming. 

Some opposition to `"cpu"` since it describes an implementation detail and not a policy. Definition of fallback as used in the WebGPU specification https://www.w3.org/TR/webgpu/#fallback-adapter:
> An [adapter](https://www.w3.org/TR/webgpu/#adapter) may be considered a fallback adapter if it has significant performance caveats in exchange for some combination of wider compatibility, more predictable behavior, or improved privacy.

whatever name chosen should help describe the behavior and not the implementation. The spec does not require cpu execution so having that in the name would seem incorrect.

mwyrzykowski: I think this wording is too implementation specific:

> Prioritizes maximum compatibility over other considerations, typically running on a CPU. This is useful for testing a model's numeric behavior without utilizing parallel accelerators like GPUs or NPUs

and something more general similar to the [WebGPU wording](https://www.w3.org/TR/webgpu/#fallback-adapter) where either "compatibility, more predictable behavior, or improved privacy" can be prioritized would be more consistent with existing web specification.

mwyrzykowski (changes requested):
Would like to see the references to CPU removed from the non-normative wording https://github.com/webmachinelearning/webnn/pull/923/changes#r3010808209 but otherwise no objections to the API surface changes ✅

mingmingtasd:
I have no preference for the naming; I'll wait for the decision. For the description wording, I'm OK with not mentioning CPU to avoid being implementation-specific.

mwyrzykowski (approved):
✅ with the current
 ```
  "low-power",
  "fallback"
```
values

**Comment by anssiko** [comment]
>[RESOLUTION](https://www.w3.org/2026/04/23-webmachinelearning-minutes.html#5f2e): Use MLComputePolicy "fallback" as the name for maximum compatibility preference. (PR [#923](https://github.com/webmachinelearning/webnn/issues/923))

**Comment by anssiko** [comment]
We also acknowledged the following enhancements in our [discussion today](https://www.w3.org/2026/04/23-webmachinelearning-minutes.html#2998):

>Anssi: other enhancements from this PR, to be addressed in a separate PR:
>… - "low-latency" MLComputePolicy, example use case audio processing
>… - "precision" MLContextOptions, to signal chopping off low bits is not preferred for this context

**Comment by mingmingtasd** [comment]
> I have no preference for the naming; I'll wait for the decision. For the description wording, I'm OK with not mentioning CPU to avoid being implementation-specific.

Thanks for your discussion! Let's keep the "fallback".

And in the latest commit, I removed the "CPU" word in the description for "fallback". @mwyrzykowski

**Review by anssiko — APPROVED** [review]
This is ready to be merged by the editors. Thank you all.

The editors can track the discussed enhancements https://github.com/webmachinelearning/webnn/pull/923#issuecomment-4305974696 as deemed appropriate.

File: index.bs [hunk 3]
```diff
@@ -1098,7 +1095,7 @@ interface MLContext {
 
   undefined destroy();
 
-  readonly attribute boolean accelerated;
+  readonly attribute MLComputePolicy computePolicy;
```
huningxin: Not sure whether compute policy should remain internal slot as previous power preference. It is not like accelerated attribute that can indicate whether the native AI runtime does fallback or not. @handellm , is there any use case of exposing it?

zolkis: The intent for exposing it was that the implementation can change it (as it was a context creation hint). The create algorithm should contain clauses for how to do this. Also, after context creation this might be changed by the impl, right? One use case could be to check if the context still has the same policy it was created with.

But one could also argue that the impl got the hint, thank you, it will translate it into a set of internal states, and will take it into account in the best possible way henceforth, so no need for the app meddles with it later. In that case, it could be only an internal slot. The question is what problems might be if exposing it (other than impl complexity to translating _back_ an internal state to the MLComputePolicy).

handellm: I don't think exposure is needed on the MLContext. What's important is to understand the effective policy after compilation IMO.

anssiko: @mingmingtasd, you can proceed to update this PR accordingly to remove exposure on MLContext. After this update the editors can merge this PR. I believe we have consensus on that design point.

@handellm, I will schedule a separate group discussion for the effective policy exposure topic informed by your use case. The earlier attempt at post-compile exposure was device-centric graph.devices discussed in https://github.com/webmachinelearning/webnn/issues/836 and in now closed PR https://github.com/webmachinelearning/webnn/pull/854. I believe the group now thinks the effective policy is a more appropriate abstraction to address this use case. No open issue for that yet, let's discuss first on our call this week.

mingmingtasd: > @mingmingtasd, you can proceed to update this PR accordingly to remove exposure on MLContext. After this update the editors can merge this PR. I believe we have consensus on that design point.

Removed exposure on MLContext in latest commit.

fdwr (approved):
👍

anssiko (approved):

huningxin (approved):
LGTM, thanks @mingmingtasd !

PR #933, opened 2026-05-06
validate concat inputs and split splits

### LINKED ISSUE #931: Define an upper bound on number of concat inputs and split outputs
In the same spirit as https://github.com/webmachinelearning/webnn/issues/928, do we need to allow splitting/concatting up to 2**32 tensors? 
Is there a more reasonable lower upper bound?
@fdwr  curious to learn from your model experience.

in this [CL](https://chromium-review.git.corp.google.com/c/chromium/src/+/7785709/comment/59c4e951_6e505cbd/) we are fuzzing just 1-10 inputs for split

philloooo:
Fixes #931 
Add validation that checks the concat inputs size is not greater than 8192 and split splits size is not greater than 8192.

@fdwr @huningxin

***
<a href="https://pr-preview.s3.amazonaws.com/philloooo/webnn/pull/933.html" title="Last updated on May 7, 2026, 6:48 PM UTC (56864ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/933/db1e51c...philloooo:56864ce.html" title="Last updated on May 7, 2026, 6:48 PM UTC (56864ce)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1711,6 +1711,10 @@ An {{MLOperandDescriptor}} |A| is <dfn for=MLOperandDescriptor>equal</dfn> to an
 A <dfn>valid dimension</dfn> is an integer greater than zero and in the range of {{long}}. Implementations may impose a smaller upper bound.
 </p>
 
+<p>
+A <dfn>valid tensor count</dfn> is an integer greater than zero and less or equal to 8192. Implementations may impose a smaller upper bound.
+</p>
+
 Issue(391): Should 0-size dimensions be supported?
 
 <details open algorithm>
@@ -2816,7 +2820,7 @@ partial dictionary MLOpSupportLimits {
   </summary>
     1. If [=this=] [=MLGraphBuilder/can not build=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any [=list/item=] in |inputs| returns false, then [=exception/throw=] a {{TypeError}}.
-    1. If |inputs| [=list/is empty=], then [=exception/throw=] a {{TypeError}}.
+    1. If |inputs| [=list/size=] is not a [=valid tensor count=], then [=exception/throw=] a {{TypeError}}.
     1. Let |first| be |inputs|[0].
     1. If |axis| is greater than or equal to |first|'s [=MLOperand/rank=], then [=exception/throw=] a {{TypeError}}.
     1. Let |desc| be the result of [=creating an MLOperandDescriptor=] given |first|'s [=MLOperand/dataType=] and |first|'s [=MLOperand/shape=].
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -2817,6 +2817,7 @@ partial dictionary MLOpSupportLimits {
     1. If [=this=] [=MLGraphBuilder/can not build=], then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
     1. If [=MLGraphBuilder/validating operand=] with [=this=] and any [=list/item=] in |inputs| returns false, then [=exception/throw=] a {{TypeError}}.
     1. If |inputs| [=list/is empty=], then [=exception/throw=] a {{TypeError}}.
+    1. If |inputs| [=list/size=] is greater than 8192, then [=exception/throw=] a {{TypeError}}.
```
fdwr: Rather than burying the literal 8192 directly into `concat` and `split`, should we add a shared definition somewhere for the maximum operator tensor count, like what we do for the [valid dimension](https://webmachinelearning.github.io/webnn/#valid-dimension)? "A valid dimension is an integer ... Implementations may impose a smaller upper bound..." ->"A valid tensor count is ...".

fdwr (approved):
LGTM, with one thought.

index.bs
philloooo: good idea!

index.bs
```diff
@@ -1711,6 +1711,10 @@ An {{MLOperandDescriptor}} |A| is <dfn for=MLOperandDescriptor>equal</dfn> to an
 A <dfn>valid dimension</dfn> is an integer greater than zero and in the range of {{long}}. Implementations may impose a smaller upper bound.
 </p>
 
+<p>
+A <dfn>valid tensor count</dfn> is an integer greater than zero and less or equal to 8192. Implementations may impose a smaller upper bound.
```
fdwr, line 1715: > greater than zero

One interesting consideration is for operators that have no tensor inputs, which only *generate* output, like sequential fills or random fills. In that case, 0 is a perfectly valid input tensor count.

> less or equal to 8192

Huh, I didn't notice this before, but I see the [ONNX spec for Split](https://onnx.ai/onnx/operators/onnx__Split.html) explicitly states "Output Between 1 and 2147483647 outputs", and [ONNX Concat](https://onnx.ai/onnx/operators/onnx__Concat.html) states "Inputs Between 1 and 2147483647 inputs". I'm not saying we should mimic that, but it's interesting the spec *explicitly* calls that out.

(no action expected from my musing here - just noticing. Resolve me after reading.)

fdwr (approved):
👍

huningxin (approved):
LGTM

fdwr:
Are we awaiting the next W3C meeting to merge? (I forget)

anssiko:
We have this on our agenda today. My expectation is we'll resolve this today so you can then merge with clear consensus.

Issue #930, opened 2026-04-21
RFE:  Add support for more floating point low-precision ML data types (`bfloat16`, `fp8`, `nvfp4`)


mtavenrath:
**Problem Statement**
Currently, the WebNN API supports standard data types like `float32`, `float16`, `int8`, and `uint8`. However, the machine learning hardware ecosystem has rapidly migrated toward specialized, low-precision floating-point formats to handle the extreme memory bandwidth and compute demands of modern Large Language Models (LLMs) and diffusion models. 

Developers bringing state-of-the-art models to the web are currently forced into a strict dilemma:
1. **Upcast to keep precision:** Converting native `bfloat16` or `fp8` models to `float32` respective `float16` preserves the model's accuracy and dynamic range, but immediately bloats the memory footprint and exacerbates memory bandwidth bottlenecks, which can be fatal for client-side inference.
2. **Quantize to integer to save memory:** Converting to `int8` or `int4` achieves the desired performance and footprint, but sacrifices precision. Integers space values uniformly, which crushes the activation outliers prevalent in LLMs, leading to a degradation in reasoning quality and perplexity.

Without native WebNN support for modern low-precision floating-point formats, the API forces developers to choose between performance and accuracy, losing out on the hardware-accelerated "best of both worlds" that these new formats were designed to provide.

**The Case for Each Format**

1. **`bfloat16` (Brain Floating Point)**
   * **Context:** Widely used as the default format for training modern AI models because its 8-bit exponent matches `float32`, preventing gradient underflow/overflow. 
   * **Issue:** Porting a natively `bfloat16`-trained model to WebNN currently requires an offline conversion to `float32` (doubling the model size) or `float16` (which requires careful scaling to avoid out-of-bounds errors).
   
2. **`fp8` (OCP 8-bit Floating Point - E4M3 & E5M2)**
   * **Context:** Now standardized by the Open Compute Project and natively supported by current-gen hardware (NVIDIA Hopper/Blackwell, Ryzen AI, RDNA4, Intel XE2 (Lunar Lake, ARC B-series)). 
   * **Issue:** `fp8` acts as the perfect middle ground between the two extremes mentioned above. It provides the memory footprint of `int8` while maintaining the dynamic range of a floating-point exponent, allowing it to handle activation outliers without precision loss.
   
3. **`nvfp4` (4-bit Microscaled Floating Point)**
[...]

Issue #779, opened 2024-11-06
Support block-wise quantization


huningxin:
[Block-wise quantization](https://arxiv.org/abs/2110.02861) divides input tensors into smaller blocks that are independently quantized, resulting in faster optimization and high precision quantization. It is used for popular language models, such as [phi-3 mini int4 quantized model](https://huggingface.co/microsoft/Phi-3-mini-4k-instruct).

### Native ML API's support
DML `DML_OPERATOR_QUANTIZE` and `DML_OPERATOR_DEQUANTIZE` introduced in [Feature Level 6.3](https://learn.microsoft.com/en-us/windows/ai/directml/dml-feature-level-history#dml_feature_level_6_3)
CoreML [constexpr_blockwise_shift_scale](https://apple.github.io/coremltools/source/coremltools.converters.mil.mil.ops.defs.html#coremltools.converters.mil.mil.ops.defs.iOS18.compression.constexpr_blockwise_shift_scale)
TFLite: ?

### Proposal
No API signature changes regarding to @fdwr 's [proposal](https://github.com/webmachinelearning/webnn/issues/375#issuecomment-2292466613) of `dequantizeLinear` and `quantizeLinear` ops. 
```js
MLOperand dequantizeLinear(MLOperand input, MLOperand scale, MLOperand zeroPoint, optional MLOperatorOptions options = {});
MLOperand quantizeLinear(MLOperand input, MLOperand scale, MLOperand zeroPoint, optional MLOperatorOptions options = {});
```

The `block_size` is an integer and implied by `block_size = input_size / scale_size` (where `input_size % scale_size == 0`) along a dimension. `zeroPoint` and `scale` should have the same shape.

fdwr:
Thanks for the paper link. I'd be surprised if TFLite didn't have some blockwise support *somewhere*, but if not, it might need decomposition (e.g. `scale` and `zeroPoint` blockwise expanded up to the input shape via tf.tile or tf.repeats or some other similar function, then dq = (input - zeroPoint) * scale).

fdwr:
[...]

PR #938, opened 2026-06-29
Add low-precision floating point data types explainer

mklimenko-nv:
This PR is a first step to add low-precision floating point data types to the WebNN specification. It's a follow-up to #930 with a conservative addition of `bfloat16` and `float8` types only, with a potential follow-up for microscaled data types.

### DIFF
--- a/low-precision-data-types-explainer.md
+++ b/low-precision-data-types-explainer.md
@@ -1,0 +1,173 @@
+# Low Precision Data Types Explainer
+
+## Introduction
+This explainer summarizes the importance of low-precision floating point data types for web-based inference and proposes the expansion of MLOperandDataType to support emerging AI data type formats.                
+
+- [Low Precision Data Types Explainer](#low-precision-data-types-explainer)
+  - [Introduction](#introduction)
+  - [Motivation](#motivation)
+  - [Proposed data types overview and use cases](#proposed-data-types-overview-and-use-cases)
+    - [1.	`bfloat16`](#1bfloat16)
+    - [2.	`float8`](#2float8)
+  - [Proposed changes](#proposed-changes)
+    - [1.	Extend MLOperandDataType](#1extend-mloperanddatatype)
+    - [2.	New scale descriptor for quantization](#2new-scale-descriptor-for-quantization)
+    - [3.	Extend per-operator type tables](#3extend-per-operator-type-tables)
+    - [4.	Extend buffer validation mechanism](#4extend-buffer-validation-mechanism)
+  - [Data types and Q/DQ nodes handling](#data-types-and-qdq-nodes-handling)
+  - [Accuracy implications](#accuracy-implications)
+  - [Pros and cons](#pros-and-cons)
+    - [Pros](#pros)
+    - [Cons](#cons)
+  - [Security and privacy considerations](#security-and-privacy-considerations)
+
+
+## Motivation
+The latest developments in both specialized hardware, as well as AI and large language models (LLMs) have radically shifted the bottlenecks of machine learning. Specifically in inference, such models are rarely compute-bound nowadays, they're limited by memory bandwidth and capacity. Loading and storing billions of parameters in 32-bit (`float8`) or even 16-bit (`float16`) floating-point formats results in a massive memory footprint and overwhelms the memory bandwidth, causing high latency and power consumption.
+
+Hardware vendors in cooperation with software researchers proved, that in most cases model weights and activations can leverage lower-precision data types with negligible accuracy impact. This led to development of native silicon support in GPUs and NPUs for such data types, allowing to both save memory footprint by compressing the weights and activations, as well as increase the compute throughput.
+
+Currently, WebNN supports `float32` and `float16`. To bring the desktop-class AI performance to the browser, WebNN must support the data types that modern hardware is designed to accelerate. This would enable developers to deploy massive models directly in the browser with native performance, dramatically reduced download times and accelerated inference speeds.
+
+## Proposed data types overview and use cases
+### 1.	`bfloat16`
+Bfloat16 is a 16-bit floating-point format designed by Google Brain for training purposes. The main purpose of this format was to provide a full `float32` range at the expense of the precision. This was mainly used for the backpropagation passes of the training process to avoid an “exploding gradients” issue during the first several iterations. Combined with `float32` master weights this allowed to reduce the memory traffic and later leverage the hardware accelerated blocks with no influence on the quality of the trained model.
+
+However, this format has been getting more traction lately in multiple application fields in inference as well, thanks to the high range of values. To name a few examples:
+1.	StyleGAN-like models use modulated convolutions with renormalized weights at runtime. This weights normalization produces large intermediate values that can result in `float16` overflow, forcing the usage of higher-precision data types.
+2.	In case of LLMs running with long contexts, KV-cache tensors and attention intermediates can span a wide dynamic range across layers and sequence length. `float16` range increases the risk of clipping and accumulated drift over long prefills and generation, degrading the output quality and causing inconsistencies.
+3.	When using Mixture-of-Experts (MoE) models, different experts have sparse and uneven activation scales, sometimes varying by orders of magnitude. Extended range of `bfloat16` allows for more robust expert routing.
+
+### 2.	`float8`
+Float8 is a family of 8-bit floating point formats standardized under the OFP8 and Microscaling Formats (MX) Specification, extending the idea of using lower-precision data types and designed to roughly double the compute throughput relative to 16-bit floating point data types. Unlike 16-bit floating point data types, `float8` have a limited range and rely on scaling to stay numerically stable. Depending on the model, this scaling can be performed on a per-tensor, per-channel or per-block level.
+
+OCP specification lists two main formats of `float8`, rearranging the bit allocation between the exponent and mantissa, named E5M2 (typical for gradients in training) and E4M3 (typical for weights and activations in inference), with one bit reserved for the sign. Those data types are primarily used in LLMs to speed up matrix multiplications and compress KV-cache but are also quite useful for diffusion and classical computer vision models, providing a hardware-friendly alternative to integer quantization, with competitive accuracy when properly scaled.
+
+Despite easier upcast to `float16`, E5M2 subformat is primarily used in training and provides limited benefits for inference. In this explainer we'll focus on E4M3 format only.
+
+## Proposed changes
+We propose extending the MLOperandDataType enum and define, how these types interact with typed arrays, since JavaScript lacks native `bfloat16` and `float8` arrays.
+
+### 1.	Extend MLOperandDataType
+```javascript
+enum MLOperandDataType {
+  "float32",
+  "float16",
+  "int32",
+  "uint32",
+  "int64",
+  "uint64",
+  "int8",
+  "uint8",
+  // Proposed additions:
+  "bfloat16",
+  "float8"	// Synonymous to float8e4m3 in this scenario
+};
+```
+To address the lack of those data types in JavaScript, developers will pass those weights as `Uint16Array` for `bfloat16`, and `Uint8Array` for `float8`.
+
+### 2.	New scale descriptor for quantization
+Current Q/DQ approach is designed with affine integer semantics in mind. The following changes extend it to address block-wise and microscaling formats.
+
+```javascript
+enum MLQuantizationScheme {
+  "affine",           // scale * (x – zp) // existing
+  "symmetric-float",  // x * scale, zp = 0
+  "blockwise-float"  // per-block scaling tensor
+};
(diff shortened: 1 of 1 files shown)


low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+  - [Data types and Q/DQ nodes handling](#data-types-and-qdq-nodes-handling)
+  - [Accuracy implications](#accuracy-implications)
+  - [Pros and cons](#pros-and-cons)
+    - [Pros](#pros)
+    - [Cons](#cons)
+  - [Security and privacy considerations](#security-and-privacy-considerations)
+
+
+## Motivation
+The latest developments in both specialized hardware, as well as AI and large language models (LLMs) have radically shifted the bottlenecks of machine learning. Specifically in inference, such models are rarely compute-bound nowadays, they're limited by memory bandwidth and capacity. Loading and storing billions of parameters in 32-bit (`float8`) or even 16-bit (`float16`) floating-point formats results in a massive memory footprint and overwhelms the memory bandwidth, causing high latency and power consumption.
```
fdwr: > parameters in 32-bit (`float8`) 

Did you mean 32-bit (`float32`) here?

low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Introduction
+This explainer summarizes the importance of low-precision floating point data types for web-based inference and proposes the expansion of MLOperandDataType to support emerging AI data type formats.                
+
+- [Low Precision Data Types Explainer](#low-precision-data-types-explainer)
+  - [Introduction](#introduction)
+  - [Motivation](#motivation)
+  - [Proposed data types overview and use cases](#proposed-data-types-overview-and-use-cases)
+    - [1.	`bfloat16`](#1bfloat16)
+    - [2.	`float8`](#2float8)
```
fdwr: How certain are we that this specific float8 is *the* float8 bit allocation that we want to codify into the API without any clarifying suffix? Will we need another float8 in the future, one that would need a distinguishing suffix (making it odd to have `float8` and `float8suffixed`)? After all, we already determined that IEEE `float16` (`float16m10e5s1`) was insufficient for all models, warranting `float16m7e8s1`.

Just to double check, the flavor of float8 proposed is *this* one ⭐, being IEEE-compliant, with full infinity, NaN, and distinct positive/negative zero?

```c++
float8m3e4s1_t ⭐     { uint8_t mantissa: 3; uint8_t exponent: 4; uint8_t sign: 1;} // infinities, NaNs
float8m2e5s1_t        { uint8_t mantissa: 2; uint8_t exponent: 5; uint8_t sign: 1;} // infinities, NaNs
float8m3e4s1fnuz_t    { uint8_t mantissa: 3; uint8_t exponent: 4; uint8_t sign: 1;} // no infinities, NaN as -0
float8m2e5s1fnuz_t    { uint8_t mantissa: 2; uint8_t exponent: 5; uint8_t sign: 1;} // no infinities, NaN as -0
float8m4e3s1_t        { uint8_t mantissa: 4; uint8_t exponent: 3; uint8_t sign: 1;} // infinities, NaN's
float8m0e8s0fn_t      { uint8_t mantissa: 8; uint8_t exponent: 0; uint8_t sign: 0;} // no infinities, NaN via all ones (purely range scaling)
float8m3e4b11s1fnuz_t { uint8_t mantissa: 3; uint8_t exponent: 4; uint8_t sign: 1;} // exp bias 11, no infinities, NaN as -0
float8m3e4s1fn_t      { uint8_t mantissa: 3; uint8_t exponent: 4; uint8_t sign: 1;} // infinities, NaNs
```

low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+To address the lack of those data types in JavaScript, developers will pass those weights as `Uint16Array` for `bfloat16`, and `Uint8Array` for `float8`.
+
+### 2.	New scale descriptor for quantization
+Current Q/DQ approach is designed with affine integer semantics in mind. The following changes extend it to address block-wise and microscaling formats.
+
+```javascript
+enum MLQuantizationScheme {
+  "affine",           // scale * (x – zp) // existing
+  "symmetric-float",  // x * scale, zp = 0
+  "blockwise-float"  // per-block scaling tensor
```
fdwr: [dequantizeLinear](https://webmachinelearning.github.io/webnn/#api-mlgraphbuilder-dequantizelinear) already supports blockwise scales? See emulated decomposition:

```js
function dequantizeLinear(builder, input, scale, zeroPoint, options)
{
    // output = (input - zeroPoint) * scale
    const floatInput = builder.cast(input, scale.dataType);
    const floatZeroPoint = builder.cast(zeroPoint, scale.dataType);
    const upsampledScale = blockwiseExpand(builder, scale, input.shape);
    const upsampledZeroPoint = blockwiseExpand(builder, floatZeroPoint, input.shape);
    return builder.mul(builder.sub(floatInput, upsampledZeroPoint), upsampledScale);
}
```

I think we had a Phi-3 WebNN demo running some time ago that used block scales (though, it's [now Phi 4](https://github.com/microsoft/webnn-developer-preview/tree/main/demos/text-generation/models/microsoft/Phi-4-mini-instruct-onnx-webnn), and I'm not sure if it's been kept up to date with the API) .

If this was unclear from the documentation, we should clarify it.

low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+};
+
+dictionary MLQuantizationOptions : MLOperatorOptions {
+  MLQuantizationScheme scheme = "affine";
+  unsigned long blockSize;
```
fdwr: We didn't add `blockSize` after becoming apparent that it was redundant from the tensor shapes, and given the ratio of the input tensor's shape and zero-point shape would have needed to be validated anyway, it's cleaner to just remove a potential failure point. Plus it's more flexible than a single block size since multiple flavors of dequantization shapes exist (along rows, columns, square blocks, asymmetric...). Of course, performance is best when callers use shapes that align to whatever the underlying backend's have been optimized for (and it's actually worse performance if it's incompatible and you have to emulate things), but it's hard to know what backend you'll run on, and the current design doesn't lock us into so rigid a box when it changes later.

low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+```
+To address the lack of those data types in JavaScript, developers will pass those weights as `Uint16Array` for `bfloat16`, and `Uint8Array` for `float8`.
+
+### 2.	New scale descriptor for quantization
+Current Q/DQ approach is designed with affine integer semantics in mind. The following changes extend it to address block-wise and microscaling formats.
+
+```javascript
+enum MLQuantizationScheme {
+  "affine",           // scale * (x – zp) // existing
+  "symmetric-float",  // x * scale, zp = 0
```
fdwr: There's a [pending issue](https://github.com/webmachinelearning/webnn/issues/943) to make zeroPoint optional, which would be equivalent to symmetric, right?

low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+  MLQuantizationScheme scheme = "affine";
+  unsigned long blockSize;
+};
+```
+
+### 3.	Extend per-operator type tables
+Every operator specification provides a tensor limits table with allowed data types for different operands. With an introduction of additional low-precision floating-point data types those tables need to be reviewed and extended.
+
+### 4.	Extend buffer validation mechanism
+Currently, buffer validation is only done for strongly typed buffers. In case of low-precision floating-point data types, this check is significantly simplified and would always return True for `ArrayBufferView` due to usage of `Uint8Array`. The algorithm should consider that the low-precision data types are represented using non-native JavaScript data types, but the dimensions and the element count still needs to be validated.
```
fdwr: > the low-precision data types are represented using non-native JavaScript data types

Note there's precedent for using a different underlying buffer type than the tensor's data type, as we had to use `Uint16Array` for float16 tensors, before `Float16Array` existed.

fdwr (commented):
Tis a nice document, thanks for preparing, and I support adding low precision data types (I may just have "are we sure xyz" questions 🙂).

low-precision-data-types-explainer.md
fdwr: Aha, supporting links:
- https://github.com/webmachinelearning/webnn/issues/779
- https://chromium-review.googlesource.com/c/chromium/src/+/5964816

low-precision-data-types-explainer.md
mklimenko-nv: Yes, that's a typo, thanks for noticing!

low-precision-data-types-explainer.md
mklimenko-nv: I agree with the `float16` vs `bfloat16` use cases, but with `float8` I find it important to maintain a balance between being explicit and user-friendly. We see that the `e4m3` variant is used primarily in inference with `e5m2` being more popular in training. For the sake of completeness, it is possible to specify the suffix upfront, I'll update the explainer. 

On the `_fnuz` subvariants and other flavors my personal stance is that they're vendor-specific and would overcomplicate the specification.

low-precision-data-types-explainer.md
mklimenko-nv: I missed it, thanks for pointing out

low-precision-data-types-explainer.md
mklimenko-nv: Yes, that's correct. I'd like to add to that that the 3% performance improvement is a conservative estimation based on my previous experiments.

low-precision-data-types-explainer.md
mklimenko-nv: I'll rephrase it for the next iteration of the explainer, thanks!

low-precision-data-types-explainer.md
fdwr: > with float8 I find it important to maintain a balance between being explicit and user-friendly

Yeah, I'm okay with no suffix (I don't want web devs to have to type out cryptic suffixes either :b), but I'm just double checking the confidence percentage that this float8 is *the* float8, and if the answer is "yes, because this is what most current/upcoming hardware prefers, and this is what the models need for inference", then so be it. All the ML [libraries I surveyed](https://github.com/webmachinelearning/webnn/issues/930#issuecomment-4410036511) showed support for that type (along with others).

> On the _fnuz subvariants and other flavors my personal stance...

I too would prefer the more IEEE-like float8, consistent with WebNN's existing float16 and float32 (which have infinity, NaN, and dual-sign zero), and though there's no *official* IEEE float8, m3e4s1 is one of the first examples [shown here](https://en.wikipedia.org/wiki/Minifloat#8-bit_(1.4.3)).

low-precision-data-types-explainer.md
fdwr: Might some wording like this help?

https://github.com/webmachinelearning/webnn/compare/main...fdwr:webnn:user/dwayner/dequantizeLinearClarifyBlockwiseNature?expand=1

anssiko:
I observe general support and no concerns, see also [meeting minutes](https://www.w3.org/2026/08/13-webmachinelearning-minutes.html#b45c).

@mklimenko-nv, any pending changes you'd still like to integrate into this v1 PR? Otherwise, editors are free to merge this by our next meeting.

low-precision-data-types-explainer.md
```diff
@@ -0,0 +1,173 @@
...
+
+OCP specification lists two main formats of `float8`, rearranging the bit allocation between the exponent and mantissa, named E5M2 (typical for gradients in training) and E4M3 (typical for weights and activations in inference), with one bit reserved for the sign. Those data types are primarily used in LLMs to speed up matrix multiplications and compress KV-cache but are also quite useful for diffusion and classical computer vision models, providing a hardware-friendly alternative to integer quantization, with competitive accuracy when properly scaled.
+
+Despite easier upcast to `float16`, E5M2 subformat is primarily used in training and provides limited benefits for inference. In this explainer we'll focus on E4M3 format only.
+
+## Proposed changes
+We propose extending the MLOperandDataType enum and define, how these types interact with typed arrays, since JavaScript lacks native `bfloat16` and `float8` arrays.
+
+### 1.	Extend MLOperandDataType
+```javascript
```
huningxin: Label the IDL snippet as `idl` or `webidl`?

mklimenko-nv:
@anssiko, I've made some minor modifications to the explainer based on the discussion above (thanks @fdwr!).

anssiko:
@mklimenko-nv thank you, it looks like all feedback has been addressed. The editor are free to merge this explainer PR.

@mklimenko-nv are you interested in taking the first stab at the spec PR for this?

mklimenko-nv:
@anssiko, yeah, sounds good to me, let me prepare a spec PR for this. Thanks!

fdwr (approved):
👍

PR #942, opened 2026-07-13
Add initial WebNN-WebGPU Interop specification

bbernhar:
For resolution https://www.w3.org/2026/06/18-webmachinelearning-minutes.html
@fdwr @huningxin PTAL

***
<a href="https://pr-preview.s3.amazonaws.com/bbernhar/webnn/pull/942.html" title="Last updated on Jul 13, 2026, 6:11 PM UTC (aa58a2b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/942/c515210...bbernhar:aa58a2b.html" title="Last updated on Jul 13, 2026, 6:11 PM UTC (aa58a2b)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1522,6 +1522,10 @@ Note: There is no guarantee that exporting an {{MLTensor}} is zero-copy. If the 
     1. Return |buffer|.
 </details>
 
+The user agent MUST ensure that no operation enqueued to |gpuDevice|'s {{GPUQueue}} which reads from or writes to |buffer| executes before the steps enqueued above to ensure the contents of |buffer| reflect |tensor|.{{MLTensor/[[data]]}} have completed.
+
+Note: This ordering guarantee is internal to the user agent; it does not require any observable synchronization on the part of the application. An application may safely record and submit {{GPUQueue}} work using |buffer| immediately after {{MLContext/exportToGPU()}} returns.
+
 When the {{GPUBuffer}} returned by {{MLContext/exportToGPU()}} is destroyed, the user agent MUST [=return an exported MLTensor=] given the {{MLTensor}} it was exported from.
 
 Note: Destroying the {{MLTensor}} or its {{MLContext}} while it [=MLTensor/is exported=] does not invalidate the returned {{GPUBuffer}}; the {{GPUBuffer}} remains usable by the {{MLTensor/[[gpuDevice]]}} until it is destroyed.
@@ -2123,6 +2127,8 @@ An {{MLTensor}} is created by its associated {{MLContext}}.
                 Note: If the export was zero-copy, then |buffer| and |tensor|.{{MLTensor/[[data]]}} are backed by the same memory and no copy is performed.
 
 </details>
+
+The user agent MUST ensure that the enqueued steps above to ensure |tensor|.{{MLTensor/[[data]]}} reflects the contents of |buffer| do not execute before all operations previously enqueued to |gpuDevice|'s {{GPUQueue}} which read from or write to |buffer| have completed.
 
 ### {{MLTensor/destroy()}} ### {#api-mltensor-destroy}
 
--- a/index.bs
+++ b/index.bs
@@ -1494,7 +1494,7 @@ Exports an [=MLTensor/is exported|exportable=] {{MLTensor}} to its associated {{
 
 While an {{MLTensor}} [=MLTensor/is exported=], the {{MLTensor/[[gpuDevice]]}} has exclusive access to the {{MLTensor}}'s contents, and any WebNN operation which depends on the {{MLTensor}} — including {{MLContext/dispatch()}}, {{MLContext/readTensor(tensor)}}, {{MLContext/readTensor(tensor, outputData)}}, and {{MLContext/writeTensor()}} — will fail.
 
-The returned {{GPUBuffer}} is created with a {{GPUBuffer/usage}} of `GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST` and a {{GPUBuffer/size}} equal to the {{MLTensor}}'s {{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=]. When the returned {{GPUBuffer}} is destroyed, the {{MLTensor}} is [=return an exported MLTensor|returned=] to the {{MLContext}} and may be used with WebNN operations again.
+The returned {{GPUBuffer}} is created with a {{GPUBuffer/usage}} of <code>GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST</code> and a {{GPUBuffer/size}} equal to the {{MLTensor}}'s {{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=]. When the returned {{GPUBuffer}} is destroyed, the {{MLTensor}} is [=return an exported MLTensor|returned=] to the {{MLContext}} and may be used with WebNN operations again.
 
 Note: There is no guarantee that exporting an {{MLTensor}} is zero-copy. If the memory backing the {{MLTensor}} is not accessible to the {{MLTensor/[[gpuDevice]]}}, exporting may require copying the {{MLTensor}}'s contents into the returned {{GPUBuffer}}, and copying them back once the {{GPUBuffer}} is destroyed.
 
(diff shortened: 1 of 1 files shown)


bbernhar:
@anssiko please assign to editors? Won't let me for some reason.

index.bs
```diff
@@ -1439,6 +1481,90 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
...
+<div dfn-for="MLContext/exportToGPU(tensor)" dfn-type=argument>
+    **Arguments:**
+      - <dfn>tensor</dfn>: an {{MLTensor}}. The tensor to export.
+
+    **Returns:** a {{GPUBuffer}}.
+</div>
+
+While an {{MLTensor}} [=MLTensor/is exported=], the {{MLTensor/[[gpuDevice]]}} has exclusive access to the {{MLTensor}}'s contents, and any WebNN operation which depends on the {{MLTensor}} — including {{MLContext/dispatch()}}, {{MLContext/readTensor(tensor)}}, {{MLContext/readTensor(tensor, outputData)}}, and {{MLContext/writeTensor()}} — will fail.
+
+The returned {{GPUBuffer}} is created with a {{GPUBuffer/usage}} of <code>GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST</code> and a {{GPUBuffer/size}} equal to the {{MLTensor}}'s {{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=]. When the returned {{GPUBuffer}} is destroyed, the {{MLTensor}} is [=return an exported MLTensor|returned=] to the {{MLContext}} and may be used with WebNN operations again.
```
mwyrzykowski, line 1498: It would be nice to allow custom usages, e.g., read only storage, when the WebGPU operations are read only, but that could be added later if needed

mwyrzykowski (approved):

index.bs
```diff
@@ -1439,6 +1481,90 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
...
+    1. Let |usage| be <code>GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST</code>.
+    1. Let |buffer| be a new {{GPUBuffer}} in |realm| associated with |gpuDevice| whose {{GPUBuffer/size}} is |tensor|.{{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=] and whose {{GPUBuffer/usage}} is |usage|.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
+            1. Ensure the contents of |buffer| reflect |tensor|.{{MLTensor/[[data]]}}.
+
+                Note: If the export is zero-copy, then |buffer| and |tensor|.{{MLTensor/[[data]]}} are backed by the same memory and no copy is performed.
+
+    1. Set |tensor|.{{MLTensor/[[exportedBuffer]]}} to |buffer|.
+    1. Return |buffer|.
```
huningxin, line 1522: Should we specify the explicit cross-timeline ordering, e.g. the buffer is unavailable/pending on the WebGPU queue until the ML timeline step completes?

index.bs
```diff
@@ -1949,9 +2089,38 @@ An {{MLTensor}} is created by its associated {{MLContext}}.
...
+  <summary>
+    To <dfn>return an exported MLTensor</dfn> given {{MLTensor}} |tensor|, run the following steps:
+  </summary>
+    1. If |tensor| [=MLTensor/is not exported=], then abort these steps.
+    1. Let |buffer| be |tensor|.{{MLTensor/[[exportedBuffer]]}}.
+    1. Set |tensor|.{{MLTensor/[[exportedBuffer]]}} to null.
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then abort these steps.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] |tensor|.{{MLTensor/[[context]]}} [=MLContext/is lost=]:
+            1. Ensure |tensor|.{{MLTensor/[[data]]}} reflects the contents of |buffer|.
```
huningxin, line 2125: Same here for cross-timeline ordering, waiting for outstanding `gpuDevice` queue work touching `buffer` to finish?

index.bs
```diff
@@ -1439,6 +1481,90 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
...
+    1. If |tensor|.{{MLTensor/[[isDestroyed]]}} is true, then [=exception/throw=] a {{TypeError}}.
+    1. If |tensor|.{{MLTensor/[[exportableToGPU]]}} is false, then [=exception/throw=] a {{TypeError}}.
+    1. Let |realm| be [=this=]'s [=relevant realm=].
+    1. Let |gpuDevice| be |tensor|.{{MLTensor/[[gpuDevice]]}}.
+    1. If |gpuDevice| is lost, then [=exception/throw=] an "{{InvalidStateError}}" {{DOMException}}.
+    1. If |tensor| [=MLTensor/is exported=], then return |tensor|.{{MLTensor/[[exportedBuffer]]}}.
+    1. Let |usage| be <code>GPUBufferUsage.STORAGE | GPUBufferUsage.COPY_SRC | GPUBufferUsage.COPY_DST</code>.
+    1. Let |buffer| be a new {{GPUBuffer}} in |realm| associated with |gpuDevice| whose {{GPUBuffer/size}} is |tensor|.{{MLTensor/[[descriptor]]}}'s [=MLOperandDescriptor/byte length=] and whose {{GPUBuffer/usage}} is |usage|.
+    1. Enqueue the following steps to |tensor|.{{MLTensor/[[context]]}}.{{MLContext/[[timeline]]}}:
+        1. Run these steps, but [=/abort when=] [=this=] [=MLContext/is lost=]:
```
huningxin, line 1516: What happens if GPUDevice is lost?

index.bs
```diff
@@ -1439,6 +1481,90 @@ Writes data to the {{MLTensor/[[data]]}} of an {{MLTensor}} on the {{MLContext}}
...
+Exports an [=MLTensor/is exported|exportable=] {{MLTensor}} to its associated {{GPUDevice}} as a {{GPUBuffer}}, and rents it to that {{GPUDevice}} until the returned {{GPUBuffer}} is destroyed.
+
+<div dfn-for="MLContext/exportToGPU(tensor)" dfn-type=argument>
+    **Arguments:**
+      - <dfn>tensor</dfn>: an {{MLTensor}}. The tensor to export.
+
+    **Returns:** a {{GPUBuffer}}.
+</div>
+
+While an {{MLTensor}} [=MLTensor/is exported=], the {{MLTensor/[[gpuDevice]]}} has exclusive access to the {{MLTensor}}'s contents, and any WebNN operation which depends on the {{MLTensor}} — including {{MLContext/dispatch()}}, {{MLContext/readTensor(tensor)}}, {{MLContext/readTensor(tensor, outputData)}}, and {{MLContext/writeTensor()}} — will fail.
```
huningxin: Be specific about "will fail", e.g. say "throw a TypeError"?

bbernhar: Done.

index.bs
bbernhar: Added.

index.bs
bbernhar: Added.

index.bs
bbernhar: Nothing implicit, stays exported. Added note.

index.bs
bbernhar: Ack. Happy to take a look if you need me to.

bbernhar:
@fdwr still missing your feedback, PTAL.

index.bs
```diff
@@ -1086,6 +1086,8 @@ interface MLContext {
   undefined dispatch(MLGraph graph, MLNamedTensors inputs, MLNamedTensors outputs);
 
   Promise<MLTensor> createTensor(MLTensorDescriptor descriptor);
+  Promise<MLTensor> createExportableTensor(
+    MLTensorDescriptor descriptor, GPUDevice gpuDevice);
```
fdwr, line 1090: (mostly verb naming questions) Is `export` common nomenclature for this use? It may well be, and I'm just unfamiliar with it, but I see `export`, and I think the tensor is being serialized to a file or into a memory blob, not that it's simply shared with with a device/imported into a device's memory. Additionally "export" has the notion of sending information *away* from the current area of interest (like from a paint program out to a file or onto the network), whereas here the area of interest *is* the GPU (since that's where the model will execute), and so you're sending information *toward* the thing of interest (it's more like you're importing it or uploading it into GPU memory).

- `createDeviceTensor`? // Because it's for the device
- `createShareableTensor`? // Because of similarity to Allow(Shared)BufferSource
- `createSharedTensor`
- `createAcceleratedTensor`?
- `createInteropTensor`?

For `exportToGPU`, more often I've heard `uploadToGPU` and `copyToGPU`, but again, maybe this is just a term I've not heard (my graphics history dwells more in D3D/OpenGL).

Since you can overload on arity, do you think we could just have a two-parameter overload of `createTensor`, where passing the device (or would you rather name it separately to reserve future potential createTensor overloads with multiple parameters)?

```
const outputTensor = await context.createTensor({dataType: 'float32', shape: [4], readable: true}, gpuDevice);
```

Or rather, considering that WebGPU generally prefers capability flags ([GPUBufferUsage.COPY_SRC](https://developer.mozilla.org/en-US/docs/Web/API/GPUBuffer/usage), GPUBufferUsage.COPY_DST, GPUBufferUsage.MAP_READ...) over distinct creation methods, maybe use them? e.g.

`createTensor(descriptor, {usage: MLTensorUsage.WEBGPU_INTEROP}, device);`
`createTensor(descriptor, {usage: MLTensorUsage.SHARED}, device);`

(if in the end, we stick with createExportableTensor, I'm not opposed to it, but it's not my first intuition when I see the function name to guess what it does)

fdwr (approved):
👍 I just have one consideration about naming perspective, but the overall flow (a new or updated tensor creation function, MLTensor property getter, and tensor copying+rental function) looks good to me.

index.bs
bbernhar: Thanks @fdwr. Some background: "export" was chosen because web developers are already familiar with WebGPU's `importExternalTexture()` terminology.  The WebGPU operation is an import of a resource exported by another API.  I'd be hesitant to use "upload" or "copy" since this operation is preferably neither or zero-copy.

I agree that usage flags are useful and would support adding them, but I'd prefer to keep that as a separate/new issue since we've already established APIs without usage flags in WebNN (e.g. createConstantTensor()).

fdwr: So if created on a UMA system or in GPU-visible memory, no copy occurs? Yeah, in that case, calling the sharing functionality "upload" or "copy" would be a misnomer.

> WebGPU's importExternalTexture()

Aha, I Ctrl+F'd for "export" in WebGPU and found no results, but didn't search for "import" :b. I'll have to [read this](https://www.w3.org/TR/webgpu/#dom-gpudevice-importexternaltexture) more about creating *external* textures. 

> I agree that usage flags are useful and would support adding them, but I'd prefer to keep that as a separate/new issue

👍

huningxin (approved):
LGTM, thanks!

anssiko:
>[RESOLUTION](https://www.w3.org/2026/09/10-webmachinelearning-minutes.html#a562): Add initial WebNN-WebGPU interop support per PR [#942](https://github.com/webmachinelearning/webnn/issues/942) to WebNN API spec.

Editors are free to merge this PR at will. Thank you all for your contributions.

anssiko (approved):

Issue #883, opened 2025-08-27
Support flexible input sizes


huningxin:
There are models that may require flexible input sizes that are determined at inference time.

Examples include some vision models that may need to work on multiple resolutions, e.g., [MODNet](https://huggingface.co/Xenova/modnet) (a model for real-time portrait matting), the [model](https://netron.app/?url=https://huggingface.co/Xenova/modnet/resolve/main/onnx/model_fp16.onnx) accepts input in shape of `[batch_size,3,height,width]`. 

Transformers may work on arbitrary input lengths. When using KV cache, they may also need to increase the KV cache length by 1 per each inference.

For speech recognition model, e.g., [Whisper](https://huggingface.co/Xenova/whisper-tiny.en), the [encoder](https://netron.app/?url=https://huggingface.co/Xenova/whisper-tiny.en/resolve/main/onnx/encoder_model.onnx) needs to work on arbitrary input lengths, its input shape is `[batch_size, feature_size, encoder_sequence_length]`.  The [decoder (with past)](https://netron.app/?url=https://huggingface.co/Xenova/whisper-tiny.en/resolve/main/onnx/decoder_with_past_model.onnx) needs to increase the KV cache length by 1 per each inference. i.e. the shape of input named "past_key_values.0.decoder.key" is `[batch_size,6,past_decoder_sequence_length,64]`, the shape of corresponding output named "present.0.decoder.key" is `[batch_size,6,past_decoder_sequence_length + 1,64]`.

For other language models, e.g., [Qwen2.5-0.5B-Instruct](https://huggingface.co/onnx-community/Qwen2.5-0.5B-Instruct), the [model](https://netron.app/?url=https://huggingface.co/onnx-community/Qwen2.5-0.5B-Instruct/resolve/main/onnx/model_q4f16.onnx) takes input in shape `[batch_size,sequence_length]`. Similarly, it also increases KV cache length at inference time, i.e. the shape of input "past_key_values.0.key" is `[batch_size,2,past_sequence_length,64]`, the corresponding output "present.0.key" shape is `[batch_size,2,past_sequence_length + 1,64]`

Lack of the support for flexible input sizes increases the complexity of using WebNN for those models. For example, apps need to modify the model and fix the input size when compiling the WebNN graph. At the inference time, apps need to resize the image input, or pad the language input to maximum length before passing the input to WebNN graph.

[...]

PR #945, opened 2026-08-07
Add dynamic shape explainer

miaobin:
This is the initial draft to summarize the discussion on dynamic shapes #883 .

The explainer covers named/unnamed dynamic dimensions, deferred (dispatch-time) shape validation, `computeShapes()` API, and a new family of shape-as-data (`*Dynamic`) operators.

Open questions and considered alternatives are called out explicitly in the explainer and feedback on this doc would be very welcome.

### DIFF
--- a/dynamic-shape-explainer.md
+++ b/dynamic-shape-explainer.md
@@ -96,7 +96,7 @@ const padded = builder.padDynamic(x, builder.constant(..., [0, 0]), builder.conc
 This motivates the **shape-as-data operators** described below.
 
 ### Getting output shapes before dispatch
-A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs, it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs - it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
 
 ```js
 const outShapes = await graph.computeShapes({'attention_mask': [1, 37]});
--- a/dynamic-shape-explainer.md
+++ b/dynamic-shape-explainer.md
@@ -158,9 +158,10 @@ dictionary MLOperandDescriptor {
 `computeShapes()` runs the same shape inference and shape computation as dispatch, but early and without executing the graph: one forward pass over the graph resolves every operand, and it returns the concrete shape of each output for the given input shapes:
 
 ```webidl
+typedef record<USVString, sequence<[EnforceRange] unsigned long>> MLNamedShapes;
+
 partial interface MLGraph {
-  record<DOMString, sequence<[EnforceRange] unsigned long>> computeShapes(
-      record<DOMString, sequence<[EnforceRange] unsigned long>> inputShapes);
+  MLNamedShapes computeShapes(MLNamedShapes inputShapes);
 };
 ```
 
@@ -279,12 +280,18 @@ bool DimensionsAreDefinitelyUnequal(Dimension a, Dimension b) {
 
 The same predicate is applied uniformly across operators that impose cross-dimension constraints, e.g. `matmul`'s contraction dimension, concat's non-concatenated axes, `reshape`'s element-count product, broadcasting, and so on.
 
+This invariant is one-sided, and only its upper bound is a contract: an implementation **must not** reject a graph that some assignment of concrete input shapes would make valid. Nothing requires it to be good at detecting the rest. The predicate above is the weakest way to honor that — it proves a contradiction only between two static dimensions — and an implementation carrying symbolic *expressions* may reject more, and earlier. Concatenating two `[batch, seq]` tensors along axis 1 doubles the element count, for instance, so reshaping that result back to `[batch, seq]` cannot hold for any input: our model sees only an opaque name for the concatenated axis and defers, while an implementation that knows the axis is `2 * seq` can prove the graph dead and may fail `build()` on it — as some backends performing this analysis internally already do. That is legitimate and costs nothing in portability, since such a graph would fail at every dispatch in any case. It does mean that a successful `build()` is not a portable promise that a graph will ever run — only that this implementation could not prove otherwise.
+
 ### Shape computation at dispatch
 *Shape computation* is performed by a **shape interpreter** in the user agent, which evaluates a `shape()`-rooted chain down to the concrete integer values a shape parameter needs. Shape inference ([Deferred validation](#deferred-validation)) invokes it at dispatch each time it reaches a `*Dynamic` operator, and uses the values it returns as that operator's output shape.
 
-There is no blessed list of operators that may appear on a shape chain. In principle any operator can compute a shape — `matmul` reducing two 1-D vectors to an element count is a legitimate, if unusual, way to do it.
+Two questions hide behind "which operators may appear on a shape chain": what a graph may *express*, and what an implementation must be able to *resolve*.
 
-The set the prototype implements — arithmetic and structural transforms on shape tensors, with a little floating-point support for cases such as `reciprocal`, is therefore an implementation-cost and performance trade-off, not a design boundary, and the right balance is something to settle as the feature develops.
+The first is not restricted by operator identity. In principle any operator can compute a shape — `matmul` reducing two 1-D vectors to an element count is a legitimate, if unusual, way to do it. What is restricted is the chain's **root**, described below, and that restriction is what keeps a shape chain from becoming the model.
+
+Interoperability lives in the second question, and it needs a floor: shapes that resolve in one user agent should resolve in another. ONNX answers the same question with a deliberately narrow closed set — the operators carrying a data-propagation function (`Add`, `Cast`, `Concat`, `Gather`, `Mul`, `Shape`, `Size`, `Slice`, `Squeeze`, `Sub`, `Unsqueeze`) — while ORT's [`symbolic_shape_infer.py`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/symbolic_shape_infer.py) covers a much wider de-facto set. A required minimum for WebNN belongs between the two, and naming it is work this proposal still owes; the prototype today implements a strict superset of the ONNX set — arithmetic and structural transforms on shape vectors, plus enough floating-point for cases such as `reciprocal`.
+
+Above that floor, a stronger interpreter is a quality-of-implementation matter, with one requirement: an implementation that cannot evaluate a chain must fail with a clean "cannot resolve this shape" error, rather than executing the chain on the accelerator or guessing a value. Heavy operators — `conv2d`, `matmul` on real tensors, attention — are not expected to be resolvable, and an implementation may reject them on a shape chain. Evaluation is integer work on small vectors in the user agent; it is not an execution of the graph, and it never reads a tensor's data.
 
 A small chain makes this concrete. To reshape a `[1, 'seqlen', 512]` tensor into `[1, 'seqlen', 8, 64]` (splitting the static hidden size into 8 heads × 64) while keeping the dynamic sequence length:
 
@@ -393,7 +400,9 @@ Currently, the `shape()` operator returns an operand's entire shape as a 1D tens
 - `dimension(axis)` → Returns a 0D scalar tensor representing the size of a specific dimension (similar to StableHLO's `get_dimension_size`).
 
 ## Open Questions
-- **Tensor-Derived sizes.** Whether, and how, to admit dimensions that depend on tensor *values*, which this proposal places out of scope.
+- **Tensor-Derived sizes.** Whether, and how, to admit dimensions that depend on tensor *values* — an output whose extent is `NonZero`-shaped rather than a function of the input shapes. This proposal places them out of scope, and draws that boundary structurally rather than by classifying tensors: follow a shape chain back and it bottoms out either at `shape()` outputs and build-time constants, in which case it resolves before dispatch, or at a tensor's data, in which case it is rejected. Admitting the second class needs more than interpreter coverage. Something has to allocate the output before its size is known, which requires an upper bound to allocate against (see [Bounded (min/max) dimensions](#bounded-minmax-dimensions)) and a way to report the size actually produced; and `computeShapes()` could not answer for such an output at all, since the answer does not exist until the graph runs.
+
+- **Making the shape subgraph explicit.** The values on a shape chain are already separate from tensor data by construction: a chain's roots are `shape()` outputs and build-time constants, so nothing on it can read a tensor, and an implementation resolves it on CPU without executing the graph. That separation is *implicit* today — it follows from the root restriction rather than from anything in the type system, and `MLOperand` is reused for both — which has already led readers to expect that `computeShapes()` might have to run the model. Options for making it visible: expose an operand property such as `isShape`, which documents the split and adds no constraints; or introduce a distinct `MLShapeOperand` type, making the guarantee structural.
 
 - **Bidirectional broadcasting for Expand.** In dynamic shape scenarios, the shape operand of `expand` may contain flexible dimensions (e.g., `[1, 1, 1]`) that require bidirectional broadcasting against the input tensor. The original WebNN specification was limited to unidirectional broadcasting. Extending the `expand` operator to support bidirectional broadcasting aligns with the ONNX `Expand` operator and ensures correct handling of these dynamic shape cases.
 
--- a/dynamic-shape-explainer.md
+++ b/dynamic-shape-explainer.md
@@ -13,6 +13,10 @@
 4. [Use Cases](#use-cases)
 5. [Proposed API](#proposed-api)
 6. [Design Discussion](#design-discussion)
+    - [Dimension semantics](#dimension-semantics)
+    - [Deferred validation](#deferred-validation)
+    - [Shape computation at dispatch](#shape-computation-at-dispatch)
+    - [Backends mapping](#backends-mapping)
 7. [Considered Alternatives](#considered-alternatives)
 8. [Privacy & Security Considerations](#privacy--security-considerations)
 9. [Future Consideration](#future-consideration)
@@ -96,10 +100,10 @@ const padded = builder.padDynamic(x, builder.constant(..., [0, 0]), builder.conc
 This motivates the **shape-as-data operators** described below.
 
 ### Getting output shapes before dispatch
-A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs - it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs — it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
 
 ```js
-const outShapes = await graph.computeShapes({'attention_mask': [1, 37]});
+const outShapes = graph.computeShapes({'attention_mask': [1, 37]});
 // => {'output': [1, 74]}
 
 // Allocate output MLTensors of the resolved size, then dispatch.
@@ -238,7 +242,7 @@ For example, when the target shape is a runtime operand (values unknown at build
 *(Optimization: This chain may be skippable if the framework can prove the operand is already sentinel-free.)*
 
 ## Design Discussion
-This section describes the runtime behavior that gives it meaning — the two mechanisms that carry dynamism at dispatch, **shape inference** and **shape computation**, plus the dimension semantics they honor. The two are easy to confuse by name, so to be precise: *shape inference* propagates each operand's **shape** forward across the whole graph, while *shape computation* evaluates one `shape()`-rooted chain down to the concrete **integer values** a shape parameter needs.
+This section describes the runtime behavior that gives it meaning, plus the dimension semantics it honors. Two mechanisms carry dynamism at dispatch: **shape inference**, which walks the whole graph and computes each operand's *shape* ([Deferred validation](#deferred-validation)), and **shape computation**, which evaluates a single `shape()`-rooted chain to the *values* a shape parameter needs ([Shape computation at dispatch](#shape-computation-at-dispatch)). The second is a step inside the first.
 
 ### Dimension semantics
 - **Shared names are constraints.** Two dynamic dimensions with the **same name** are guaranteed to take the **same** concrete value throughout the graph (e.g. "query and key sequence lengths are equal"); the implementation enforces this across inputs and uses it to cancel dimensions in operations such as `reshape`. A name that appears only once constrains nothing. A dynamic dimension has **no min/max bound** — anything not provably static simply defers.
@@ -252,9 +256,9 @@ This section describes the runtime behavior that gives it meaning — the two me
 ### Deferred validation
 Validation splits across two phases:
 
-- **At build time**, we validate only what is knowable without concrete shapes: data-type compatibility, rank constraints (e.g. `conv2d` needs rank 4), same-name symbolic consistency, and any **definite static contradiction** (e.g. reshaping a static `[2, 3]` to `[7]`).
+- **At build time**, shapes propagate symbolically — every operand gets a shape expressed in names and static sizes — and we validate only what is knowable without concrete shapes: data-type compatibility, rank constraints (e.g. `conv2d` needs rank 4), same-name symbolic consistency, and any **definite static contradiction** (e.g. reshaping a static `[2, 3]` to `[7]`).
 
-- **At dispatch time** (and at `computeShapes()`), once concrete input shapes are known, we run **shape inference** — a forward propagation of each operand's concrete *shape* — over the whole graph, and validate the resulting concrete shapes against every constraint, including buffer sizes.
+- **At dispatch time** (and at `computeShapes()`), once concrete input shapes are known, we run **shape inference**: the same forward propagation, now carrying each operand's concrete *shape* across the whole graph. We then validate the resulting concrete shapes against every constraint, including buffer sizes.
 
 Deferring this work is the inherent trade-off of dynamic shapes: the shape resolution and validation a static graph completes once at build time now runs at inference time — as a gatekeeper on every `dispatch`, before the graph executes. This adds per-inference overhead, so a user agent should optimize the common cases: because the result is a pure function of the input shapes, it can skip re-validation when a dispatch repeats a set of input shapes it has already validated.
 
@@ -272,7 +276,7 @@ bool DimensionsAreDefinitelyUnequal(Dimension a, Dimension b) {
 The same predicate is applied uniformly across operators that impose cross-dimension constraints, e.g. `matmul`'s contraction dimension, concat's non-concatenated axes, `reshape`'s element-count product, broadcasting, and so on.
 
 ### Shape computation at dispatch
-*Shape computation* is the dispatch-time evaluation of a `shape()`-rooted chain down to the concrete values a shape parameter needs. It is the value-computing counterpart to *shape inference* ([Deferred validation](#deferred-validation)), which propagates operand shapes across the whole graph.
+*Shape computation* is performed by a **shape interpreter** in the user agent, which evaluates a `shape()`-rooted chain down to the concrete integer values a shape parameter needs. Shape inference ([Deferred validation](#deferred-validation)) invokes it at dispatch each time it reaches a `*Dynamic` operator, and uses the values it returns as that operator's output shape.
 
 There is no blessed list of operators that may appear on a shape chain. In principle any operator can compute a shape — `matmul` reducing two 1-D vectors to an element count is a legitimate, if unusual, way to do it.
 
@@ -288,9 +292,9 @@ const newShape = builder.concat([batchSeq, heads], 0);    // [1, seqlen, 8, 64]
 const y = builder.reshapeDynamic(x, newShape);            // y: [1, 'seqlen', 8, 64]
 ```
 
-At dispatch with `seqlen = 37`, shape computation walks the `newShape` chain: `shape(x) `→` [1, 37, 512], slice `→` [1, 37], concat([1, 37], [8, 64]) `→` [1, 37, 8, 64]`. That computed value becomes reshapeDynamic's inferred output shape, `[1, 37, 8, 64]`. Note what it read: `x`'s **shape** and the **constant** `[8, 64]` — never `x`'s data.
+At dispatch with `seqlen = 37`, the interpreter walks the `newShape` chain: `shape(x) `→` [1, 37, 512], slice `→` [1, 37], concat([1, 37], [8, 64]) `→` [1, 37, 8, 64]`. That computed value becomes reshapeDynamic's inferred output shape, `[1, 37, 8, 64]`. Note what it read: `x`'s **shape** and the **constant** `[8, 64]` — never `x`'s data.
 
-What shape computation may read is bounded not by the operators on the chain but by its **root**: `shape()` outputs and build-time constants only. A chain that reaches an input's tensor **data** is unresolvable by design and is rejected — the out-of-scope [Tensor-Derived](#open-questions) case.
+What the interpreter may read is bounded not by the operators on the chain but by its **root**: `shape()` outputs and build-time constants only. A chain that reaches an input's tensor **data** is unresolvable by design and is rejected — the out-of-scope [Tensor-Derived](#open-questions) case.
 
 ### Backends mapping
 The dimension model maps cleanly onto all three backends, as shown below. The **shape-as-data operator family** is currently implemented only on the ORT backend.
@@ -346,23 +350,23 @@ Dynamic shapes add no new fingerprinting or cross-origin surface: they expose no
 
 The renderer is untrusted, so the service must remain safe on any graph a compromised renderer can construct, including ill-formed dynamic graphs:
 
-- **No dereference of an absent rank.** Unranked operands (e.g. from a no-axes squeeze) are handled uniformly by each shared validator — propagate, resolve, or cleanly reject — and unranked *graph inputs* are rejected at build time (a graph input always has a known rank). A dispatch-time exit gate fails cleanly if any operand is still unranked after inference, so no unranked operand ever reaches a backend.
+- **No dereference of an absent rank.** Unranked operands (e.g. from a no-axes squeeze) are handled uniformly by each shared validator — propagate, resolve, or cleanly reject — and unranked *graph inputs* are rejected at build time (a graph input always has a known rank). A dispatch-time exit gate fails cleanly if any operand is still unranked after shape inference, so no unranked operand ever reaches a backend.
 
 - **Shape computation never reads input data**, as described above, which also bounds what a graph can make the interpreter do.
 
-- **Bounded constant collection.** The interpreter seeds only from the shape operands of the dynamic operators and walks back along the shape-computation chain, copying only the constants actually on that chain (and skipping oversized constants). Weight tensors are never on a shape chain and are not copied, bounding both memory and shape-computation work (a DoS/OOM guard).
+- **Bounded constant collection.** The constants that the interpreter may need are gathered ahead of time: the collector seeds from the shape operands of the dynamic operators, walks back along the chain, and copies only the constants it finds there. Current weight tensors are not on a shape chain and are not copied, which keeps the memory and work involved independent of the model's size (a DoS/OOM guard).
 
 ## Future Consideration
 
 ### Bounded (min/max) dimensions
 Currently, the proposed model is intentionally unbounded; a dynamic dimension is either provably static or deferred without a specific size range. As a future enhancement, we plan to allow dynamic dimensions to optionally declare a `minSize` and `maxSize` bound (and potentially an "optimal" size). For example: `{name: 'seqlen', minSize: 1, maxSize: 2048}`.
 
-Rather than introducing a new source of dynamism, these bounds will serve as critical implementation hints to underlying runtimes, unlocking ahead-of-time memory allocations and graph optimizations that an unbounded model cannot achieve. Some runtime already accepts a constraint of this shape and acts on it. TensorRT's optimization profiles (min / opt / max), Core ML's `RangeDim`, and OpenVINO's bounded partial shapes, so the hint has somewhere to go rather than terminating in the user agent.
+Rather than introducing a new source of dynamism, these bounds will serve as critical implementation hints to underlying runtimes, unlocking ahead-of-time memory allocations and graph optimizations that an unbounded model cannot achieve. Several runtimes already accept a constraint of this shape and act on it — TensorRT's optimization profiles (min / opt / max), Core ML's `RangeDim`, and OpenVINO's bounded partial shapes — so the hint has somewhere to go rather than terminating in the user agent.
 
 ### Shape specialization and preparation
-Knowing the shapes is not the same as being ready to run them: a backend may still have to plan memory, select kernels, or recompile. What that costs varies a great deal between runtimes (some absorb a shape change almost for free), while others may re-compile the graph. So on some backends a caller that changes shape often pays a real price.
+Knowing the shapes is not the same as being ready to run them: a backend may still have to plan memory, select kernels, or recompile. What that costs varies a great deal between runtimes: some absorb a shape change almost for free, while others re-compile the graph. So on some backends a caller that changes shape often pays a real price.
 
-`computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. That helps when a shape is then reused, but not when a caller keeps switching between shapes, and a user agent cannot tell which of them are worth keeping ready. One direction worth exploring is to let the caller say so: build with difference input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.
+`computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. That helps when a shape is then reused, but not when a caller keeps switching between shapes, and a user agent cannot tell which of them are worth keeping ready. One direction worth exploring is to let the caller say so: build with different input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.
 
 ### Fine-grained Shape Queries
 Currently, the `shape()` operator returns an operand's entire shape as a 1D tensor. To provide more fine-grained shape retrieval, we may consider introducing two additional operators:
--- a/dynamic-shape-explainer.md
+++ b/dynamic-shape-explainer.md
@@ -242,7 +242,11 @@ For example, when the target shape is a runtime operand (values unknown at build
 *(Optimization: This chain may be skippable if the framework can prove the operand is already sentinel-free.)*
 
 ## Design Discussion
-This section describes the runtime behavior that gives it meaning, plus the dimension semantics it honors. Two mechanisms carry dynamism at dispatch: **shape inference**, which walks the whole graph and computes each operand's *shape* ([Deferred validation](#deferred-validation)), and **shape computation**, which evaluates a single `shape()`-rooted chain to the *values* a shape parameter needs ([Shape computation at dispatch](#shape-computation-at-dispatch)). The second is a step inside the first.
+This section describes the runtime behavior that gives it meaning, plus the dimension semantics it honors. Two mechanisms carry dynamism at dispatch:
+
+`Shape Inference` is the whole-graph pass: it computes each operand's shape, over symbolic dimensions at build time and over concrete ones at dispatch.
+
+`Shape Computation` isn't a superset of it — it's a step inside it, and it computes a different kind of thing. When inference reaches reshapeDynamic(x, newShape), newShape's own shape (a 1-D tensor of length 4, say) tells it nothing about the output; it needs newShape's values. So inference pauses, evaluates the chain that produces newShape down to concrete integers, and resumes with that as the output shape. Whole graph vs. one chain and inference is the caller.
 
 ### Dimension semantics
 - **Shared names are constraints.** Two dynamic dimensions with the **same name** are guaranteed to take the **same** concrete value throughout the graph (e.g. "query and key sequence lengths are equal"); the implementation enforces this across inputs and uses it to cancel dimensions in operations such as `reshape`. A name that appears only once constrains nothing. A dynamic dimension has **no min/max bound** — anything not provably static simply defers.
--- a/dynamic-shape-explainer.md
+++ b/dynamic-shape-explainer.md
@@ -10,13 +10,13 @@
 1. [Introduction](#introduction)
 2. [Goals](#goals)
 3. [Non-goals](#non-goals)
-4. [Use Cases](#non-goals)
+4. [Use Cases](#use-cases)
 5. [Proposed API](#proposed-api)
 6. [Design Discussion](#design-discussion)
 7. [Considered Alternatives](#considered-alternatives)
 8. [Privacy & Security Considerations](#privacy--security-considerations)
 9. [Future Consideration](#future-consideration)
-10. [Open Questions](#fine-grained-shape-queries)
+10. [Open Questions](#open-questions)
 11. [References & Acknowledgements](#references--acknowledgements)
 
 ## Introduction
@@ -35,7 +35,7 @@ This proposal introduces **Dynamic Shape** for WebNN, enabling graph dimensions 
 ## Goals
 - Allow a single compiled `MLGraph` to execute across varying runtime input sizes, without rebuilding.
 
-- Model dynamism the way the underlying runtimes already do: a dimension is either a **static size**, a **named dynamic** dimension (a symbolic name), or an **unnamed dynamic** dimension (fully unconstrained).
+- A dimension is either a static size or a named dynamic dimension (a symbolic name). Requiring every dynamic dimension to carry a name costs no expressive power and gives every dimension a debuggable symbol.
 
 - Defer shape validation that cannot be decided at build time to the point where concrete input shapes are known, without weakening the checks that can be decided at build time.
 
@@ -96,7 +96,7 @@ const padded = builder.padDynamic(x, builder.constant(..., [0, 0]), builder.conc
 This motivates the **shape-as-data operators** described below.
 
 ### Getting output shapes before dispatch
-A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs (it is derived inside the subgraph). To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs, it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
 
 ```js
 const outShapes = await graph.computeShapes({'attention_mask': [1, 37]});
@@ -106,23 +106,27 @@ const outShapes = await graph.computeShapes({'attention_mask': [1, 37]});
 ```
 
 ## Proposed API
-The web-facing surface changes fall into three groups: 1) the dimension model and descriptors, 2) new `computeShapes()` methods and 3) new operators. The runtime behavior behind these — the two mechanisms that carry dynamism at dispatch, **shape inference** and **shape folding** — is described under [Design Discussion](#design-discussion).
+The web-facing surface changes fall into three groups: 1) the dimension model and descriptors, 2) new `computeShapes()` methods and 3) new operators. The runtime behavior behind these — the two mechanisms that carry dynamism at dispatch, **shape inference** and **shape computation** — is described under [Design Discussion](#design-discussion).
 
 ### 1. Dimension model and descriptors
-A graph input's dimension may now be a number (static integer), a string (a **named dynamic** dimension mapped to a runtime symbolic name), or — as a `null` element of the shape sequence — an **unnamed dynamic** dimension. This is expressed with a new `MLDimension` type and a new `MLInputOperandDescriptor`:
+A graph input's dimension is either a number (a static integer) or a string (a **named dynamic** dimension mapped to a runtime symbolic name). There is no third state. This is expressed with a new `MLDimension` type and a new `MLInputOperandDescriptor`:
 
 ```webidl
 // A single input dimension: a number is static; a string is a named dynamic
-// dimension; a null element (below) is an unnamed dynamic dimension.
+// dimension. Every dynamic dimension carries a name.
 typedef ([EnforceRange] unsigned long or DOMString) MLDimension;
 
 dictionary MLInputOperandDescriptor {
   required MLOperandDataType dataType;
-  required sequence<MLDimension?> shape;
+  required sequence<MLDimension> shape;
 };
 ```
 
-Reading a shape back, `MLOperand.shape` is widened accordingly: its elements may now be strings, an unnamed dynamic dimension surfaces as `"?"` (a provisional representation — see [Open Questions](#fine-grained-shape-queries)), and the whole attribute is `null` for an **unranked** operand — one whose rank is not yet known (see [Dimension semantics](#dimension-semantics)):
+The empty string is **not** a valid dimension name; `input()` throws a `TypeError`. A name exists to relate dimensions to one another, and an empty name relates nothing — admitting it would either silently force every `""` dimension in the graph to resolve to the same value, or create a "half-named" dimension whose meaning differs from layer to layer. ONNX uses `dim_param == ""` to mean *anonymous*, so a caller reaching for `""` is asking for the state this design deliberately does not have.
+
+A framework that receives anonymous dimensions from its own model format synthesizes names before calling WebNN. It already knows a stable identity for each one — the tensor's name and the axis index — so a name such as `"encoder_input_axis1"` is both free to produce and more useful in a diagnostic than an anonymous placeholder. (This is caller-side naming; it is distinct from the names a user agent synthesizes for *derived* dimensions, described in [Dimension semantics](#dimension-semantics).)
+
+Reading a shape back, `MLOperand.shape` is widened accordingly: its elements may now be strings, and the whole attribute is `null` for an **unranked** operand — one whose rank is not yet known (see [Dimension semantics](#dimension-semantics)):
 
 ```webidl
 interface MLOperand {
@@ -131,6 +135,10 @@ interface MLOperand {
   readonly attribute FrozenArray<MLDimension>? shape;
 };
 ```
+
+Because every dynamic dimension is named, an element of this array is always either a number or a non-empty string — there is no sentinel value to interpret. The outer `null` is unrelated: it reports an unranked operand, not an unknown dimension.
+
+Names that the user agent synthesized for derived dimensions are observable here, which is what makes them useful for debugging, but their content and format are implementation-defined and **not stable**: rebuilding the same graph, or a change in how the implementation walks it, may produce different names. They are diagnostic labels, not identifiers to be read back and fed into `input()` as a cross-tensor constraint.
 
 The plain `MLOperandDescriptor` is deliberately **unchanged** (static-only): it describes `constant()` data, and `computeShapes()` likewise returns fully concrete (static) shapes. Dynamism thus lives only on graph *inputs* and propagates from there.
 
@@ -143,7 +151,7 @@ dictionary MLOperandDescriptor {
 ```
 
 ### 2. `computeShapes()`
-`computeShapes()` runs the same shape inference and folding as dispatch, but early and without executing the graph, returning the concrete output shape for each output given concrete input shapes:
+`computeShapes()` runs the same shape inference and shape computation as dispatch, but early and without executing the graph, returning the concrete output shape for each output given concrete input shapes:
 
 ```webidl
 partial interface MLGraph {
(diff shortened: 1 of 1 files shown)


miaobin:
A [POC](https://chromium-review.googlesource.com/c/chromium/src/+/8176983) based on the current explainer is WIP. And it has been validated against some real-world Transformer, LLM and image generation models.
With credits to @Honry  for the ORT WebNN EP POC and assistance with the explainer, @huningxin and @fdwr  for the initial review of this explainer.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,361 @@
...
+
+```webidl
+// Unchanged — always concrete.
+dictionary MLOperandDescriptor {
+  required MLOperandDataType dataType;
+  required sequence<[EnforceRange] unsigned long> shape;
+};
+```
+
+### 2. `computeShapes()`
```
fdwr, line 157: I love that `computeShapes` exists, because even though the *primary* intention is for returning the shape, it might also offer a means for the backends to precompute needed memory allocations, so that the later `dispatch` is immediate. Under the hood, model execution has a few key steps between knowing the shape (like planning memory and recompiling shaders) and actually executing, and unfortunately many ML libraries fail to expose that key stage -_- (and deferring all the way to dispatch is bad because then it means the first dispatch will be slow, and later ones will be fast, but also any shape changes will cause stutters).

One performance concern I have is the potential ping-pong of shape vacillation. If the model uses one shape the first iteration, and another shape the next 10 iterations, then every transition might require reallocating and replanning memory again. If however the caller could just say "build this graph, and return me two MLGraphs for these two input shapes, with all weights shared between them", the backends could execute more efficiently (no need to *see the future* or cache things behind your back).

Additionally, using `computeShapes` catches errors early rather than more confusingly during `dispatch` when it's too late to do much about it. So, there are a few good reasons to push shape evaluation earlier than dispatch.

(mostly just thinking aloud - resolve me)

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,361 @@
...
+
+Existing framework workarounds — such as frequent graph recompilation or falling back to slower, CPU-based execution environments (e.g., ONNX Runtime Web via WebAssembly) — introduce significant performance bottlenecks, particularly in latency-sensitive autoregressive decoding tasks.
+
+This proposal introduces **Dynamic Shape** for WebNN, enabling graph dimensions to remain unresolved during compilation and instead derive concrete values from input tensors at dispatch time. This mechanism permits a single compiled graph to accommodate a diverse range of runtime input sizes. This document details the design explored in the Chromium implementation and serves as a technical reference for the Working Group's resolution of [Issue #883, "Support flexible input sizes"](https://github.com/webmachinelearning/webnn/issues/883).
+
+## Goals
+- Allow a single compiled `MLGraph` to execute across varying runtime input sizes, without rebuilding.
+
+- Model dynamism the way the underlying runtimes already do: a dimension is either a **static size**, a **named dynamic** dimension (a symbolic name), or an **unnamed dynamic** dimension (fully unconstrained).
```
fdwr: Do you think we even need unnamed dimensions? A uniquely named dynamic dimension and an unnamed dynamic dimension are identical after all, and it's preferable to have debuggable symbols. The only reason to support them would be because existing callers may have them (like ORTWeb calling WebNN), but we could always just synthesize a name on the fly like "inputTensorName2_axis3" 🤔. I mean, if I was debugging and hit a shape inference error, I'd rather see that than just `null` for a name. Alternately using "" instead of null could be less problematic (no need to check for null first before trying to use/print the string).

Actually, seeing generated names would probably help *you* too during WebNN/Chromium debugging, seeing where pass-through fails during shape inference.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,361 @@
...
+// dimension; a null element (below) is an unnamed dynamic dimension.
+typedef ([EnforceRange] unsigned long or DOMString) MLDimension;
+
+dictionary MLInputOperandDescriptor {
+  required MLOperandDataType dataType;
+  required sequence<MLDimension?> shape;
+};
+```
+
+Reading a shape back, `MLOperand.shape` is widened accordingly: its elements may now be strings, an unnamed dynamic dimension surfaces as `"?"` (a provisional representation — see [Open Questions](#fine-grained-shape-queries)), and the whole attribute is `null` for an **unranked** operand — one whose rank is not yet known (see [Dimension semantics](#dimension-semantics)):
```
fdwr: > one whose rank is not yet known

Interestingly this has also come up for `MLOperandDataType dataType` too, not just shape, when you want to infer the type based on the input tensor's type so you can reuse the same mini-model for float16 and float32.
(just comment, no action expected - resolve me)

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,361 @@
...
+
+  // Dynamic variants: shape parameters are operands, evaluated at dispatch.
+  MLOperand reshapeDynamic(MLOperand input, MLOperand newShape, optional MLOperatorOptions options = {});
+  MLOperand expandDynamic(MLOperand input, MLOperand newShape, optional MLOperatorOptions options = {});
+  MLOperand sliceDynamic(MLOperand input, MLOperand starts, MLOperand sizes, optional MLSliceDynamicOptions options = {});
+  MLOperand padDynamic(MLOperand input, MLOperand beginningPadding, MLOperand endingPadding, optional MLOperatorOptions options = {});
+
+  // Equal-split (scalar count) and explicit-splits (operand) forms.
+  sequence<MLOperand> splitDynamic(MLOperand input, [EnforceRange] unsigned long splits, optional MLSplitOptions options = {});
+  sequence<MLOperand> splitDynamic(MLOperand input, MLOperand splits, optional MLSplitOptions options = {});
```
fdwr, line 218: Are we going with two separate functions rather than `([EnforceRange] unsigned long or sequence<[EnforceRange] unsigned long>) splits` like normal split?

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,361 @@
...
+    if (IsStatic(a) && IsStatic(b))
+      return StaticValue(a) != StaticValue(b);
+    return false;  // unknown -> defer
+}
+```
+
+The same predicate is applied uniformly across operators that impose cross-dimension constraints, e.g. `matmul`'s contraction dimension, concat's non-concatenated axes, `reshape`'s element-count product, broadcasting, and so on.
+
+### Shape folding at dispatch
+*Shape folding* is the dispatch-time evaluation of a `shape()`-rooted chain down to the concrete values a shape parameter needs. It is deliberately narrow — the shape-calculation operations only: arithmetic and structural transforms on shape tensors (usually integer, sometimes float, e.g. reciprocal) — and never the data-producing operators such as `conv2d` or `matmul`, which do not sit on a shape chain. It is the value-computing counterpart to *shape inference* ([Deferred validation](#deferred-validation)), which propagates operand shapes across the whole graph.
```
fdwr: > and never the data-producing operators such as conv2d or matmul

It sounds like this is saying the output of matmul cannot be a shape input to an op like `tileDynamic`? If so, I don't think we should have an explicit list of blessed operators that can modify a shape, but rather, any WebNN operator can modify a shape tensor along the chain piping to a `fooDynamic` operator, so long as those ops are placed on the CPU EP/backend (because you would not want to read back from the GPU and stall). Plus, MatMul is really an arithmetic op (mul and add), and maybe a model wants to compute a shape using MatMul as a dot product (reducing two 1D vectors into a single scalar element count, equivalent to `add` and `reduceProduct`). I haven't seen that personally, but it seems plausible. 🤔 Though, I certainly *could* understand type restrictions (like only supporting integers), or dimension restrictions (like only working with 1D tensors).

Though, if the reason for this restriction is because we're doing these operators ourselves on the Chromium side for shape inference (rather than using the backend), then I could understand op restrictions. Still, it will be weird if WebNN can only use specific shape-compatible ops.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,361 @@
...
+
+- **Folding never reads input data**, as described above, which also bounds what a graph can make the interpreter do.
+
+- **Bounded constant collection.** The interpreter seeds only from the shape operands of the dynamic operators and walks back along the shape-computation chain, copying only the constants actually on that chain (and skipping oversized constants). Weight tensors are never on a shape chain and are not copied, bounding both memory and folding work (a DoS/OOM guard).
+
+## Future Consideration
+
+### Bounded (min/max) dimensions
+Currently, the proposed model is intentionally unbounded; a dynamic dimension is either provably static or deferred without a specific size range. As a future enhancement, we plan to allow dynamic dimensions to optionally declare a `minSize` and `maxSize` bound (and potentially an "optimal" size). For example: `{name: 'seqlen', minSize: 1, maxSize: 2048}`.
```
fdwr, line 386: > optionally declare a `minSize` and `maxSize` bound

Another potentially useful constraint is a "multiple of", because models like StableDiffusion will fail if you try to bind an input whose width/height isn't a multiple of 8. (resolve me)

fdwr (approved):
👍 I have some thoughts, but it's 95% 👌.

anssiko:
I cleared the automatic IPR check for this PR, since explainer documents are consider non-substantive from the W3C Patent Policy point of view.

<details><summary>Details</summary>This IPR check is to ensure normative portions of the specification come from organizations who participate the WG. Furthermore, in this case, I can attest the authors of this PR are affiliated with Intel, and as such, any normative portions are reusable in the spec PR, as appropriate.</details>

dynamic-shape-explainer.md
miaobin: In fact, we are considering removing `sequence<MLOperand> splitDynamic(MLOperand input, [EnforceRange] unsigned long splits, optional MLSplitOptions options = {});`, because the existing `split` operator can already express the case where `splits` is an `unsigned long`.

dynamic-shape-explainer.md
miaobin: Good catch! ONNX Runtime itself employs a similar processing mechanism; I developed the POC by referencing ORT [`symbolic_shape_infer.py`](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/symbolic_shape_infer.py#L130) and `MatMul` is indeed among the supported operations. The description here is admittedly somewhat problematic; my initial concern was that supporting certain complex operators would result in a significant performance penalty.
I think that as we enter the development phase, we need to strike a balance in the level of support provided.

By the way, we’ve collectively agreed that the term "shape folding" isn't quite right, it sounds like a graph optimization process. We’re calling this part "shape compute" now, WDYT?

dynamic-shape-explainer.md
miaobin: Some history, since we did have derived names early on and removed them deliberately. We generated things like "height-3" and "broadcast_1_hight". The names weren't unique and two unrelated conv/slice calls would collide on the same name and the backend would conflate two independent dimensions. We patched that with a per-builder counter. Additionally, when we created the POC on the ORT backend at the begining, we only called `OrtApi::SetDimensions`. Regardless of the name, any `dim` that was dynamic was ultimately set to `-1`. Then we dropped derived names entirely in the rewrite and narrowed the rule to "a name survives only on 1:1 pass-through".

A synthesized name that is unique and implies no relationship is semantically identical to an anonymous dim, and it's strictly better to look at while debugging. I agree with the direction and will take it.

dynamic-shape-explainer.md
fdwr: > I think that as we enter the development phase, we need to strike a balance in the level of support provided.

Understandable. ⚖️

> By the way, we’ve collectively agreed that the term "shape folding" isn't quite right

Folding implies a reduction in rank, but the shape can grow too. So yeah, that doesn't sound right.

More often, I hear terms "shape inference" and "shape propagation". "shape inference" always sounded a bit funny to me, because computing a shape is a deterministic process (and a fairly simple one when you consider each node individually and its shape equations), rather than a Sherlock Holmes-level "conclusion reached on the basis of evidence and reasoning". Plus, there is no real "inference" (extrapolating, deduction, guessing) of unknowns happening here, but rather very *known* computations. I suppose someone just co-opted "inference" from normal execution (which *is* fuzzy deduction) to *also* apply to shape computation, without thinking about the actual meaning of the word.

"shape computation" works too, as each node certainly computes a new shape, whereas "shape propagation" describes the overall process for what's happening across the graph, with all the computations collectively contributing to it. "shape computation" also matches the proposed `computeShapes` method.

I'd be comfortable with any of those 3 names, except shape folding 😉.

dynamic-shape-explainer.md
miaobin: Done, the explainer now requires every dynamic dimension to carry a name, and the unnamed state is gone. `MLInputOperandDescriptor.shape` drops its nullable elements entirely (`sequence<MLDimension>`), so `null` is no longer accepted at all and the empty string is rejected too.

@huningxin  Do you have any thoughts on this point?

dynamic-shape-explainer.md
miaobin: > One performance concern I have is the potential ping-pong of shape vacillation.

This is a point worth considering. I once ran a batch dynamic image classification demo on an OpenVINO CPU EP; when the batch size was between 1 and 16, execution was very fast after the initial run—suggesting that compilation only occurred the first time. However, once the batch size exceeded 16, the subsequent run slowed down abruptly, likely due to recompilation.

I have added a "Shape specialization and preparation" section. I recorded your suggestion as the open direction — build with different sizes , receive several MLGraphs each bound to a set of concrete shapes, weights shared.

dynamic-shape-explainer.md
miaobin: Went with shape computation.

dynamic-shape-explainer.md
miaobin: I investigated various runtime environments and found none that support declarative stride constraints for dimensions:
- Core ML has RangeDim and EnumeratedShapes
- TensorRT has min/opt/max optimization profiles
- OpenVINO has bounded partial shapes
- ORT has symbolic names plus free-dimension overrides
- LiteRT has only -1

> models like StableDiffusion will fail if you try to bind an input whose width/height isn't a multiple of 8

The 8-multiple requirement comes from the VAE's downsampling stages. I believe that if the dimensions do not meet the requirements, a shape mismatch error will be reported at a specific operator (such as `reshape`, `concat`, or others) during the validation.

anssiko:
@reillyeon you had good questions comments on our [last call](https://www.w3.org/2026/08/13-webmachinelearning-minutes.html#b1bc) for this. Do you have some other Googlers in mind who should review this explainer PR? 

When this PR lands, the team will start landing the implementation in smaller chunks for further validation of this approach.

reillyeon (approved):
Overall I think this approach is reasonable. My main concern is that by making shapes dynamic there are more opportunities for memory safety issues in the implementation.

anssiko:
With two approvals, editors are welcome to merge this explainer PR at will.

Thank you @miaobin for this contribution and Reilly, Dwayne, everyone for your review.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,392 @@
...
+const padded = builder.padDynamic(x, builder.constant(..., [0, 0]), builder.concat([pad, zero], 0));
+```
+
+This motivates the **shape-as-data operators** described below.
+
+### Getting output shapes before dispatch
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs, it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+
+```js
+const outShapes = await graph.computeShapes({'attention_mask': [1, 37]});
```
huningxin: computeShapes() is sync in WebIDL but awaited in the sample code.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Shape specialization and preparation
+Knowing the shapes is not the same as being ready to run them: a backend may still have to plan memory, select kernels, or recompile. What that costs varies a great deal between runtimes (some absorb a shape change almost for free), while others may re-compile the graph. So on some backends a caller that changes shape often pays a real price.
+
+`computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. That helps when a shape is then reused, but not when a caller keeps switching between shapes, and a user agent cannot tell which of them are worth keeping ready. One direction worth exploring is to let the caller say so: build with difference input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.
```
huningxin: typo: "build with difference input sizes" → "different"

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,392 @@
...
+const heads = builder.constant(/*uint32*/ ..., [8, 64]);  // static tail
+const newShape = builder.concat([batchSeq, heads], 0);    // [1, seqlen, 8, 64]
+const y = builder.reshapeDynamic(x, newShape);            // y: [1, 'seqlen', 8, 64]
+```
+
+At dispatch with `seqlen = 37`, shape computation walks the `newShape` chain: `shape(x) `→` [1, 37, 512], slice `→` [1, 37], concat([1, 37], [8, 64]) `→` [1, 37, 8, 64]`. That computed value becomes reshapeDynamic's inferred output shape, `[1, 37, 8, 64]`. Note what it read: `x`'s **shape** and the **constant** `[8, 64]` — never `x`'s data.
+
+What shape computation may read is bounded not by the operators on the chain but by its **root**: `shape()` outputs and build-time constants only. A chain that reaches an input's tensor **data** is unresolvable by design and is rejected — the out-of-scope [Tensor-Derived](#open-questions) case.
+
+### Backends mapping
```
huningxin, line 310: Add a corresponding entry to the table of contents.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,392 @@
...
+
+The renderer is untrusted, so the service must remain safe on any graph a compromised renderer can construct, including ill-formed dynamic graphs:
+
+- **No dereference of an absent rank.** Unranked operands (e.g. from a no-axes squeeze) are handled uniformly by each shared validator — propagate, resolve, or cleanly reject — and unranked *graph inputs* are rejected at build time (a graph input always has a known rank). A dispatch-time exit gate fails cleanly if any operand is still unranked after inference, so no unranked operand ever reaches a backend.
+
+- **Shape computation never reads input data**, as described above, which also bounds what a graph can make the interpreter do.
```
huningxin, line 379: What is the "interpreter"? It's an undefined term, only appears here and following "Bounded constant collection". From context it means the dispatch-time evaluator that performs shape computation. Consider introducing it once in "Shape computation at dispatch" — e.g. "Shape computation is performed by a shape interpreter in the user agent, which evaluates a shape()-rooted chain…"

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,392 @@
...
+Dynamic shapes add no new fingerprinting or cross-origin surface: they expose no device or environment information a static graph does not, and shape computation and inference read only shapes the page itself supplied. The synthesized names for derived dimensions are newly observable output, but they are derived from the graph the page itself constructed (operation type, index, and axis) and carry nothing about the device or the environment; they are also explicitly unstable and not part of the API contract, so nothing may be inferred from a change in one. The considerations below are therefore about security.
+
+The renderer is untrusted, so the service must remain safe on any graph a compromised renderer can construct, including ill-formed dynamic graphs:
+
+- **No dereference of an absent rank.** Unranked operands (e.g. from a no-axes squeeze) are handled uniformly by each shared validator — propagate, resolve, or cleanly reject — and unranked *graph inputs* are rejected at build time (a graph input always has a known rank). A dispatch-time exit gate fails cleanly if any operand is still unranked after inference, so no unranked operand ever reaches a backend.
```
huningxin: In "after inference", does the "inference" mean "shape inference", better to be explicit.

dynamic-shape-explainer.md
```diff
@@ -96,7 +96,7 @@ const padded = builder.padDynamic(x, builder.constant(..., [0, 0]), builder.conc
 This motivates the **shape-as-data operators** described below.
 
 ### Getting output shapes before dispatch
-A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs (it is derived inside the subgraph). To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs, it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
```
fdwr:
```suggestion
A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs - it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
```

dynamic-shape-explainer.md
```diff
@@ -106,23 +106,27 @@ const outShapes = await graph.computeShapes({'attention_mask': [1, 37]});
 ```
 
 ## Proposed API
-The web-facing surface changes fall into three groups: 1) the dimension model and descriptors, 2) new `computeShapes()` methods and 3) new operators. The runtime behavior behind these — the two mechanisms that carry dynamism at dispatch, **shape inference** and **shape folding** — is described under [Design Discussion](#design-discussion).
+The web-facing surface changes fall into three groups: 1) the dimension model and descriptors, 2) new `computeShapes()` methods and 3) new operators. The runtime behavior behind these — the two mechanisms that carry dynamism at dispatch, **shape inference** and **shape computation** — is described under [Design Discussion](#design-discussion).
```
fdwr, line 113: > The two are easy to confuse by name

Indeed, as I'm now confused :b. I presume shape propagation (or shape inference) would include *both* concrete and symbolic shapes through the graph, and shape computation would be a superset of shape propagation with *concrete values*?

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,392 @@
...
+
+The renderer is untrusted, so the service must remain safe on any graph a compromised renderer can construct, including ill-formed dynamic graphs:
+
+- **No dereference of an absent rank.** Unranked operands (e.g. from a no-axes squeeze) are handled uniformly by each shared validator — propagate, resolve, or cleanly reject — and unranked *graph inputs* are rejected at build time (a graph input always has a known rank). A dispatch-time exit gate fails cleanly if any operand is still unranked after inference, so no unranked operand ever reaches a backend.
+
+- **Shape computation never reads input data**, as described above, which also bounds what a graph can make the interpreter do.
+
+- **Bounded constant collection.** The interpreter seeds only from the shape operands of the dynamic operators and walks back along the shape-computation chain, copying only the constants actually on that chain (and skipping oversized constants). Weight tensors are never on a shape chain and are not copied, bounding both memory and shape-computation work (a DoS/OOM guard).
```
huningxin: Re "(and skipping oversized constants)", what are the oversized constants? Do you mean its length is over maximum rank limit? Should we reject rather than skip it?

dynamic-shape-explainer.md
```diff
@@ -314,22 +342,27 @@ The first cut loosened each operator's build-time validator independently to tol
...
+
+### Shape specialization and preparation
+Knowing the shapes is not the same as being ready to run them: a backend may still have to plan memory, select kernels, or recompile. What that costs varies a great deal between runtimes (some absorb a shape change almost for free), while others may re-compile the graph. So on some backends a caller that changes shape often pays a real price.
+
+`computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. That helps when a shape is then reused, but not when a caller keeps switching between shapes, and a user agent cannot tell which of them are worth keeping ready. One direction worth exploring is to let the caller say so: build with difference input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.
```
fdwr: Thanks for mentioning this scenario.

fdwr (approved):
👍 Cool.

dynamic-shape-explainer.md
miaobin: "Backends mapping" is currently a sub-item of "Design Discussion," so all the sub-items have been listed together in the table of contents.

dynamic-shape-explainer.md
miaobin: Done. Introduced it in "Shape computation at dispatch".

dynamic-shape-explainer.md
miaobin: Done.

dynamic-shape-explainer.md
miaobin: This is actually an implementation level design choice. To prevent the collection of large constants, we limit the maximum number of elements. Here, "skip" refers to the practice of not collecting a constant when an oversized one is encountered during the collection phase.

The description of this passage has been refined.

dynamic-shape-explainer.md
miaobin: `Shape Inference` is the whole-graph pass: it computes each operand's shape, over symbolic dimensions at build time and over concrete ones at dispatch.

`Shape Computation` isn't a superset of it — it's a step inside it, and it computes a different kind of thing. When inference reaches reshapeDynamic(x, newShape), newShape's own shape (a 1-D tensor of length 4, say) tells it nothing about the output; it needs newShape's values. So inference pauses, evaluates the chain that produces newShape down to concrete integers, and resumes with that as the output shape. Whole graph vs. one chain and inference is the caller.

huningxin: @miaobin , your explanation sounds good, could it be embedded into the explainer?

dynamic-shape-explainer.md
huningxin: > we limit the maximum number of elements

What's the limit? Do you mean validate against [valid dimension](https://www.w3.org/TR/webnn/#valid-dimension)?

dynamic-shape-explainer.md
miaobin: > Could it be embedded into the explainer?

Of course, it was mentioned in the introduction to the `Design Discussion`.

dynamic-shape-explainer.md
miaobin: > What's the limit? Do you mean validate against [valid dimension](https://www.w3.org/TR/webnn/#valid-dimension)?

The limit is the rank of a constant. A constant with more elements than the limit is not a shape related vector, it is a weight, and we will not copy it.

miaobin:
@huningxin The latest [Patchset](https://chromium-review.googlesource.com/c/chromium/src/+/8176983/4) of the dynamic shape POC runs dynamic shape inference and validation in Blink at dispatch time and also makes `MLGraph.computeShapes()` synchronous. PTAL, thanks.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,400 @@
...
+const input = builder.input('attention_mask', {dataType: 'float32', shape: [1, 128]});
+// ...a different compiled graph is required for each sequence length.
+```
+
+With dynamic shapes, the sequence length is declared as a named dynamic dimension and the graph is compiled once:
+
+```js
+// After: 'sequence_length' is a named dynamic dimension.
+const input = builder.input('attention_mask', {dataType: 'float32', shape: [1, 'sequence_length']});
+const graph = await builder.build({output});
```
theHamsta: It could help implementations to have guaranteed bounds min,max or typical expected on the dynamic sizes. Otherwise, compilation of the model would need to assume worst cases which might be harmful for runtime or do post-compilation at runtime when it encounters an unexpected size the first time on dispatch.

theHamsta, line 71: the post-compilation on dispatch might be finger printable (below discussion suggest `computeShapes` is the natural place for post-compilation and shape specialization). Also, the implementation hitting an very unfortunate shape might have a finger printable cost that allows guesses about the used HW and implementation.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,400 @@
...
+
+- **At build time**, shapes propagate symbolically — every operand gets a shape expressed in names and static sizes — and we validate only what is knowable without concrete shapes: data-type compatibility, rank constraints (e.g. `conv2d` needs rank 4), same-name symbolic consistency, and any **definite static contradiction** (e.g. reshaping a static `[2, 3]` to `[7]`).
+
+- **At dispatch time** (and at `computeShapes()`), once concrete input shapes are known, we run **shape inference**: the same forward propagation, now carrying each operand's concrete *shape* across the whole graph. We then validate the resulting concrete shapes against every constraint, including buffer sizes.
+
+Deferring this work is the inherent trade-off of dynamic shapes: the shape resolution and validation a static graph completes once at build time now runs at inference time — as a gatekeeper on every `dispatch`, before the graph executes. This adds per-inference overhead, so a user agent should optimize the common cases: because the result is a pure function of the input shapes, it can skip re-validation when a dispatch repeats a set of input shapes it has already validated.
```
theHamsta: I would not agree that deferring the work is inherent to dynamic shapes, but it is a design decision in this document.

The alternative would be to do symbolic shape inference that would require stricter validation on graph build. This would require a stricter definition on what graphs are allowed, as this API allows to define networks that don't yield consistent or valid shapes for all or some input shapes (mathematically provable).

Having symbolic expressions for all output shapes (which might get complex expressions) would make computeShapes comparably cheap. Especially, when bounds on input shapes would allow the implementation to have provable bounds on all used operators of the network during MLGraphBuilder.build

A challenge with symbolic validation is that malicious inputs could try to craft computationally complex shapes for validation to prove. An implementation could limit the maximum time allowed for validation.

The deferred validation leaves users a bit in doubt for which shapes their dynamic shape model is actually valid. Currently, ONNX runtime can't really perform shape inference in most cases when any of the inputs is dynamic. In contrast, TensorRT whichs implements most of its validation by symbolically determining output shapes and intermediate (min,max,typical shapes).

For developer tooling, it might be valuable to have a WebNN definition of dynamic shapes, which requires to know the shape of each operand symbolically at build time. This can flag problems, allows them to see restrictions like divisibility, min or max shapes of the input and ultimately for GraphBuilder.build to leverage this knowledge during compilation, without requiring post-compilation on computeShapes. Having predictable constraints would actually be a benefit over ONNX/onnxruntime where one can have onnx without knowing whether it is actually valid for some/all/or under which constraints.

dynamic-shape-explainer.md
theHamsta: if computeShapes can trigger a post-compilation of the model, wouldn't it be more reasonable to defer optimization of the model for a shape to the dispatch step? The web app might call `computeShape` just to get information to decide for which shape actually perform dispatches.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,400 @@
...
+This motivates the **shape-as-data operators** described below.
+
+### Getting output shapes before dispatch
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs — it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+
+```js
+const outShapes = graph.computeShapes({'attention_mask': [1, 37]});
+// => {'output': [1, 74]}
+
+// Allocate output MLTensors of the resolved size, then dispatch.
```
theHamsta, line 109: did I understand this correctly? There will be no way to resize tensors, one would delete and reallocate tensors?

The validation that the WebNN implementation would perform would run multiple times, for each output on `computeShapes` and then again on actual dispatch?

dynamic-shape-explainer.md
theHamsta: This doubt is partially addressed by

> `computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. That helps when a shape is then reused, but not when a caller keeps switching between shapes, and a user agent cannot tell which of them are worth keeping ready. > One direction worth exploring is to let the caller say so: build with different input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,400 @@
...
+
+### Shape specialization and preparation
+Knowing the shapes is not the same as being ready to run them: a backend may still have to plan memory, select kernels, or recompile. What that costs varies a great deal between runtimes: some absorb a shape change almost for free, while others re-compile the graph. So on some backends a caller that changes shape often pays a real price.
+
+`computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. That helps when a shape is then reused, but not when a caller keeps switching between shapes, and a user agent cannot tell which of them are worth keeping ready. One direction worth exploring is to let the caller say so: build with different input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.
```
theHamsta: > `computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named the shapes it is about to run. 

Wouldn't it be more reasonable to do most of the preparation only in dispatch?  With a call to `computeShapes()` the user does not promise to call dispatch afterwards. They might just to computeShapes to select to among different output shapes.

theHamsta: the post-compilation on dispatch for specialization might be finger printable. Also, the implementation hitting an very unfortunate shape might have a finger printable cost that allows guesses about the used HW and implementation.

dynamic-shape-explainer.md
theHamsta: https://github.com/KhronosGroup/NNEF-Tools/tree/v2.0/skriptnd-pyproject/skriptnd always performs symbolic shape inference for each intermediate expression. Each intermediate expression has an symbolic expression tree for each of its shape components. SkirptND can therefore very early at build time notify on inconsistencies in a model

miaobin: You're right that it's a design decision rather than something inherent, and I've fixed that sentence. What is inherent is only that a constraint depending on a concrete value can't be settled before the value exists; how much else gets deferred is our choice, and we chose the far end — a dimension is an opaque name, not an expression.

I've added a `Symbolic shape expressions` alternative recording your point and why we start where we do:

Shape-as-data operators put some shapes beyond a closed-form expression. A target shape is a runtime operand produced by an arbitrary chain, including data-dependent selection such as the where that lowers ONNX's -1, and a no-axes squeeze leaves even the rank data-dependent. Symbolic inference could cover a large subset of graphs, but it could not replace the dispatch-time gate.
Backends can't consume expressions. ONNX takes a dim_param string, LiteRT a -1, Core ML a RangeDim, so expressions stay a user-agent-internal enrichment, valuable for diagnostics but invisible below.
On your own caveat about crafted complexity: the graph comes from an untrusted renderer, so a wall-clock limit would make build outcomes machine-dependent (and is itself a timing signal). It would have to be a deterministic budget on expression size and depth.

I agree on the developer-tooling value, and it pairs with the bounded-dimension work already under `Future Consideration` — bounds are the missing precondition for most build-time proofs. Recording it as an additive layer rather than a replacement for now.

dynamic-shape-explainer.md
miaobin: > There will be no way to resize tensors, one would delete and reallocate tensors?

Yes there is no resize. An `MLTensor` carries a concrete descriptor from `createTensor()`, so a new shape means a new tensor.

> The validation that the WebNN implementation would perform would run multiple times, for each output on computeShapes and then again on actual dispatch?

It is not per output. It is a single forward pass over the whole graph that resolves every operand and returns all output shapes together, so the cost scales with the number of operations — not with the number of outputs, and not with tensor sizes. It is integer work over shapes and never reads weights tensor data.

It does run again at dispatch(), for two reasons: `computeShapes()` is optional and the tensors actually bound may differ from the shapes it was asked about, and an implementation cannot take an earlier answer on trust — it has to derive shapes from the tensors actually bound. Since the result is a pure function of the input shapes it can be cached, so the repeat is a lookup rather than a second pass. I have clarified both points in the `computeShapes()` section.

dynamic-shape-explainer.md
miaobin: > Wouldn't it be more reasonable to do most of the preparation only in dispatch?

Fair point — the wording was assuming something the API doesn't guarantee, and I've fixed that.
I would still keep `computeShapes()` as a natural place for preparation rather than moving all of it to dispatch. In the dominant pattern the call is a good predictor: a framework calls it precisely to size the output tensors for a dispatch it is about to make, since dispatch() takes caller-allocated outputs. Your case — comparing candidate shapes and running none of them — is real too, and an implementation cannot distinguish the two, so the section now frames this as a signal whose strength depends on why it was called, and notes that the call is synchronous, which bounds what can reasonably be done there.

> The post-compilation on dispatch for specialization might be finger printable. 

You're right, and timing is device information. I've reworked that opening to separate the two surfaces:

Declarative (unchanged): no value the API returns carries device or environment information a static graph does not.
Timing (widened): one graph dispatched over many shapes exposes a cost curve rather than the single build() duration, and its discontinuities can suggest the backend or the class of hardware.

I don't think there is a WebNN-level mitigation — constant-time execution across shapes isn't realistic for this API. Two things bound it: the work WebNN itself adds at dispatch is data-independent and cached per set of input shapes, so the variable part is the backend's own re-specialization, which the underlying runtime performs with or without WebNN; and explicit specialization would make that cost something the caller asked for rather than an implicit cost at dispatch.

fdwr: > I would still keep computeShapes() as a natural place for preparation rather than moving all of it to dispatch. 

👍 The more deferred work we squeeze into `dispatch`, the less `dispatch` behaves as originally intended - a lightweight execution of what has already been prepared. It's very common in graphics workloads (especially games) to want to frontload such blocking operations that would trigger memory planning and allocation. Unfortunately very few ML libraries nicely delineate between memory allocation and execution steps. WebNN at least offering (but not requiring) that step would afford implementations to the chance frontload things, reducing the stutter of the costly first frame.

theHamsta: At the moment, ONNX runtime would not do much shape analysis. ONNX, the library around the serialization format (e.g. onnx.shape_inference.infer_shapes_path), does some propagation of dim_param strings:
- it propagates the dim_param string to other intermediates and outputs where it can guarantee the equality of the values
- it usually generates new dim_params when an input dim_param is transformed in some way (e.g. a shape is padded)

PyTorch tries to do some symbolic shape inference via SymPy. Even more with dynamo, it would then even embed the symbolic restrictions into the ONNX shape strings. We could have similar capabilities in WebNN if we want and also guarantee that dertermining those expressions is always possible

image

WebNN would be free to impose some additional restrictions and validation on the input graph for which it will later require certain runtime behavior even if inference framework themselves don't perform that kind of analysis.

For onnxruntime with a TRT backend, TRT performs especially on the min,max,opt bounds symbolic shape inference and will report inconsistency errors on graph build. Especially, implementation that do actual kernel compilation (like some torch.compile backends) do symbolic inference because they can't afford to recompile on any shape change and want to have as much a-priori knowledge for compilation.

For the current proposal, would it be legitimate for an implementation to return errors from their symbolic analysis on MLGraphBuilder.build or would then need to defer their error reporting for `computeShapes`/`dispatch`?

[script.py](https://github.com/user-attachments/files/32006954/script.py)

theHamsta: To be clear I'm not necessarily against the current proposal. I'm also still undecided which of the approaches would be best for implementations or users to have.

As a ML devs, mismatching shapes in pytorch and ONNX just have always haunted me

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,413 @@
...
+};
+```
+
+### 2. `computeShapes()`
+`computeShapes()` runs the same shape inference and shape computation as dispatch, but early and without executing the graph: one forward pass over the graph resolves every operand, and it returns the concrete shape of each output for the given input shapes:
+
+```webidl
+partial interface MLGraph {
+  record<DOMString, sequence<[EnforceRange] unsigned long>> computeShapes(
+      record<DOMString, sequence<[EnforceRange] unsigned long>> inputShapes);
```
theHamsta: WebNN has type aliases for similar dicts: e.g. MLNamedTensors or MLNamedOperands.

Does it make sense to name this dict something like MLNamedShapes? It is not very visible to the JS developer, but I would give it some name in my implementation of this draft in RustNN.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,413 @@
...
+  MLOperand squeeze(MLOperand input, optional MLSqueezeOptions options = {});
+  MLOperand unsqueeze(MLOperand input, sequence<[EnforceRange] unsigned long> axes,
+                       optional MLOperatorOptions options = {});
+  MLOperand reshapeTo2d(MLOperand input, optional MLReshapeTo2dOptions options = {});
+
+  // Dynamic variants: shape parameters are operands, evaluated at dispatch.
+  MLOperand reshapeDynamic(MLOperand input, MLOperand newShape, optional MLOperatorOptions options = {});
+  MLOperand expandDynamic(MLOperand input, MLOperand newShape, optional MLOperatorOptions options = {});
+  MLOperand sliceDynamic(MLOperand input, MLOperand starts, MLOperand sizes, optional MLSliceDynamicOptions options = {});
+  MLOperand padDynamic(MLOperand input, MLOperand beginningPadding, MLOperand endingPadding, optional MLOperatorOptions options = {});
```
theHamsta, line 216: for those operators, symbolic shape inference would be inherently more difficult as the symbolic expression propagation would no longer be just shapes of tensors but also allowed to be actual data of the tensors.

The consequence of such operators is also that `computeShape` might be that it is possible to construct networks where `computeShape` is not simpler by any amount than actual dispatch:

- you pass you data as shapes
- you transform shape into data via the `shape` operator
- you run inference on the data (convolutions, matrix multiply)
- you transform data back into shape via those dynamic variants

So, your WebNN implementation is required to run actual inference during `computeShapes` and therefore also needs to have access to the weights. You could do that with the actual backend framework, but this framework will do its own kind of validation and inference behavior. How can a separate WebNN validation and shape inference logic from the validation and shape inference logic of my backend? I also know that many implementation will not be able to perform arbitrary meta on shape-to-data-and-back. Some stuff will work other things not. What does WebNN actually require implementation to work?

Another open question here is then for `computeShape` or shape inference in general, is it expected that the runtime on CPU or the accelerator runs the shape computation? For ONNX runtime, it is something weird in-between. TensorRT performs some analysis and tries to separate data paths where some tensors are shape tensors which ought to be interpreted by the runtime and others are GPU tensors which will be calculated by the runtime. Others, are kind of shape tensors, but they depend on inputs (`computeShapes` does not allow to depend on input data, but it still allows to transform input shapes into data) or are truely dispatch dependent (NonZero). Such tensors can just be interpreted by an accelerator by actually performing the calculation on a worst case maximum of memory and then output the calculated output side. A runtime has to classify those cases and it will work for most constructed network. However, there exist constructible networks that will just not work with most runtime. Which of those are required by WebNN to work?

When both `computeShape` and `dispatch` may use network constants where should I put them? On the accelerator or the runtime? Also this has to be determined by some kind of analysis. Some of those constant are required both for shape inference and inference. Would it make sense to separate those data paths and have one type MLOperand for data evaluated during dispatch and another MLShapeOperand which is only interpreted on CPU by the runtime? I can go from input shapes into calculations of MLShapeOperand which I interpret, but I have separate them for the data that is calculated during dispatch and on the accelerator.  I would prefer to be able to reason tensor shapes. MLShapeOperand would be interpreted during `computeShapes` (or the validation of `dispatch`), `MLOperand` is interpreted from the accelerator during dispatch.

In RustNN, we currently construct the graph and translate it to the semantics of the backend framework. We preferably put the constant to the accelerator and the accelerator framework. We don't want to keep a copy of the constant for the RustNN runtime. If it comes now to `dispatch` or `computeShape`, we can either rely on semantics of the implementation framework to do `computeShape`/`dispatch` which might not agree with WebNN semantics (or whether it agrees is determined on build) or I duplicate all inference logic into something that just does shape inference conforming to WebNN.

reillyeon: I had a similar concern that computing shapes could effectively require executing the model but I convinced myself that the intent here is that the compute graph for shapes is separate from the compute graph for tensors. However the reuse of the `MLOperand` type makes that hard to see. We probably need to make this explicit by either adding an `isShape` property to an operand (forbidding non-shape operators from being used to compute shapes) or by using a completely different operand type.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,413 @@
...
+const padded = builder.padDynamic(x, builder.constant(..., [0, 0]), builder.concat([pad, zero], 0));
+```
+
+This motivates the **shape-as-data operators** described below.
+
+### Getting output shapes before dispatch
+A framework such as ONNX Runtime Web partitions a model and hands WebNN a subgraph. That subgraph's output can carry a dynamic dimension that is not present on any of its inputs — it is derived inside the subgraph, and so carries a name the user agent synthesized rather than one the framework supplied. To allocate the output tensor, the framework needs the concrete output shape before it dispatches:
+
+```js
+const outShapes = graph.computeShapes({'attention_mask': [1, 37]});
```
theHamsta, line 106: `MLGraph.computeShapes` is quite similar to `MLContext.dispatch`. Though, `ccomputeShapes` is on graph and `dispatch` is on MLContext.

dynamic-shape-explainer.md
miaobin: > Would it be legitimate for an implementation to return errors from their symbolic analysis on MLGraphBuilder.build or would then need to defer their error reporting for computeShapes/dispatch?

I apologize that some of the descriptions caused confusion. I was likely preparing the explainer while simultaneously working on the POC, and I ended up conflating the implementation details with the specification. Because the section read as though the predicate was the rule. I've split the two apart.

The invariant is one-sided: an implementation must not reject a graph that some assignment of concrete input shapes would make valid. Nothing obliges it to be any good at detecting the rest. The three-valued predicate in the explainer is the weakest way to honor that — it only proves a contradiction between two static dimensions — so an implementation carrying symbolic expressions may reject strictly more, and earlier. Concatenating two [batch, seq] tensors along axis 1 doubles the element count, so reshaping that result back to [batch, seq] can never hold; our deferred-validation model sees only an `unknown` for the concatenated axis and defers, while a symbolic one knows the axis is `2 * seq`, proves the graph invalid, and may fail build(). The one consequence worth stating is that a successful build() is then not a portable promise that a graph will ever run, only that this implementation could not prove otherwise — which I've now written down.

dynamic-shape-explainer.md
miaobin: Good suggestion and adopted as MLNamedShapes.

dynamic-shape-explainer.md
miaobin: > What does WebNN actually require implementation to work?

That was a real gap. The explainer said "no blessed list of operators" without saying what must therefore work, which reads as unbounded. It conflated two questions, and I've split them:

What a graph may express is not restricted by operator identity. Restricting it that way is what @fdwr objected to earlier, and I still think that's right. What is restricted is the chain's root: shape() outputs and build-time constants only. That restriction is the load-bearing one. It is why a shape chain cannot become the model.

ONNX answers the same question with a deliberately narrow closed set — the 11 operators carrying a data-propagation function (Add, Cast, Concat, Gather, Mul, Shape, Size, Slice, Squeeze, Sub, Unsqueeze) while ORT's `symbolic_shape_infer.py` covers a much wider set. A required minimum for WebNN belongs between the two. Naming it is work this proposal still owes, and I've written that down rather than leaving it implied; for calibration, the prototype today is a strict superset of the ONNX set.

> Is it expected that the runtime on CPU or the accelerator runs the shape computation?

On CPU. A shape chain is integer work on vectors of at most rank length, and an implementation that cannot evaluate one fails cleanly rather than dispatching it.

> Which of those are required by WebNN to work?

No classification is needed because the boundary is very clear: a chain bottoms out either at `shape()` and `constants`, or at a tensor's `data`. And that route isn't merely an implementation choice for us: there is no worst case to allocate (the model is deliberately unbounded, so bounded dimensions are a precondition), dispatch() takes caller-allocated outputs so the produced size would need a way back, and computeShapes() could not answer for such an output at all. I've expanded the Tensor-Derived open question to record that.

> When both computeShape and dispatch may use network constants where should I put them?

That is a good question. Similar to the approach you described for RustNN, WebNN currently uploads constants to the service side as soon as they are created. Indeed, to perform validation based on actual input shapes, we need to retain certain constants related to shape inference. And the POC additionally caps each collected constant at a shape vector's worth of elements (16, covering Pad's 2 * max_rank). So you keep a handful of ≤16-element vectors, not anything model-sized, and the constants that are needed for both paths cost about a hundred bytes of duplication rather than a second copy of the model.
About the compute logic what we did is much smaller than "duplicate all inference logic". It is an integer interpreter over roughly twenty operators on vectors of at most rank length — about 880 lines in our implementation.

> About `MLShapeOperand`

I added a section titled "Making the shape subgraph explicit" to the "Open Questions" section to facilitate further discussion.

theHamsta: > That is a good question. Similar to the approach you described for RustNN, WebNN currently uploads constants to the service side as soon as they are created. Indeed, to perform validation based on actual input shapes, we need to retain certain constants related to shape inference.

Ah, so the intent is that `computeShape` is performed by the browser WebNN runtime on a certain sub-section of the graph and the rest of the graph is then performed by some accelerator framework. So, an implementation would try to split the graph into those two parts. Both for the shape graph and for the accelerator graph I need an implementation of WebNN to run all (or a fraction) the operators.

> On CPU. A shape chain is integer work on vectors of at most rank length, and an implementation that cannot evaluate one fails cleanly rather than dispatching it.

Is there any guidance on how complete such a shape chain should be in comparison to a full WebNN implementation? Or would a typical approach be to just run it with the CPU backend of your WebNN implementation? Or just an ad-hoc interpreter for what you think that networks would typically use?

theHamsta: TRT has a function to get the output shapes given certain input shapes. ONNX runtime can only try to evaluate symbolic inference if it has been performed before on the ONNX, for the usual case, inference is required to know the output shapes as they also allow operators like NonZero where only execution determines the final output shape.

For the shape interpreter, I could create a new hand-written implementation of it or run ONNX runtime inference only on the shape graph part of the overall graph.

dynamic-shape-explainer.md
```diff
@@ -0,0 +1,422 @@
...
+
+- **Making the shape subgraph explicit.** The values on a shape chain are already separate from tensor data by construction: a chain's roots are `shape()` outputs and build-time constants, so nothing on it can read a tensor, and an implementation resolves it on CPU without executing the graph. That separation is *implicit* today — it follows from the root restriction rather than from anything in the type system, and `MLOperand` is reused for both — which has already led readers to expect that `computeShapes()` might have to run the model. Options for making it visible: expose an operand property such as `isShape`, which documents the split and adds no constraints; or introduce a distinct `MLShapeOperand` type, making the guarantee structural.
```
theHamsta, line 405: Regarding the separations of constants: we could separate MLGraphBuilder.constant from MLGraphBuilder.shape_constant 

- computeShape would only be allowed to read from shape_constant but not from a regular constant
- regular inference would be allowed to read both

This would allow to immediately send `.constant` to the accelerator without needing to wait for the user to construct the full graph and be able to analyze it.

To me it seems that actually `computeShape` never needs to read the contents of constants unless they are input to one of the rank changing or dynamic operator variants. Those operators basically transform the input chain to their shape argument from the `dispatch` part of the graph to the `computeShape` part of the computation.

theHamsta: Please feel free to close my comments! If you feel that there is nothing actionable. Most of the comments are mainly to clarify our understanding of this proposal. Please excuse if this generates any kind of noise for you!

dynamic-shape-explainer.md
miaobin: `dispatch()` binds `MLTensors` on the context, so it belongs to `MLContext` and takes the graph as an argument. 
`computeShapes()` touches neither tensors nor the device. It reads the graph's topology and the caller's integer shapes. So it sits on `MLGraph`, and that is also what lets it be synchronous.

dynamic-shape-explainer.md
miaobin: To be precise about what the prototype does: we identify the *shape subgraph* by walking backward from the shape operands of the `*Dynamic` operators. That walk bottoms out at `shape()` outputs and build-time constants — it cannot bottom out anywhere else, because the interpreter has no access to tensor data (an input
operand simply evaluates to "unknown"). That root restriction is what keeps the shape subgraph a small subgraph rather than the whole model.

A standalone interpreter evaluates that subgraph on the CPU. What it implements in the POC is the set (`Add`, `Cast`, `Concat`, `Gather`, `Mul`, `Shape`, `Size`, `Slice`, `Squeeze`, `Sub`, `Unsqueeze`): arithmetic and structural transforms over small 1-D vectors, plus enough floating-point for cases like `floor(dim * scale)`. It grew from what the models we tested actually needed rather than from a designed list; `[symbolic_shape_infer.py](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/symbolic_shape_infer.py#L130)` is the wider reference we point at when asking where a required floor should sit.

Worth stressing: this does **not** partition the graph or move any work off the accelerator. The graph handed to the backend is unchanged — `shape()` and the `*Dynamic` operators are emitted as ordinary nodes, and the backend executes the whole graph however it sees fit. The CPU evaluation is a side computation, used for validation and for answering `computeShapes()`.

**Two things make it necessary:**

1. **Allocation ordering.** `dispatch()` takes caller-allocated output `MLTensor`s, and requires each one's descriptor to equal the graph's output descriptor — same data type, same shape, dimension by dimension. An `MLTensor`'s shape is fixed at creation, and `dispatch()` returns `undefined` synchronously, so it has no way to hand back tensors it allocated itself. The caller — an application, or a framework such as ORT Web — therefore has to know the output shapes for the given input shapes *before* it calls. Without `computeShapes()` it would have to reimplement WebNN's shape inference just to size those allocations.

2. **Validation has to happen before work reaches the backend.** The renderer is untrusted, so the privileged service re-runs the same inference independently rather than taking the renderer's word for it. And `dispatch()` has no error-reporting channel: a failure that reaches the backend cannot be reported per-dispatch, only escalated to context loss — in our implementation, a session-run failure currently tears down every WebNN context in the GPU process. So "can these shapes be resolved, and do they match what the caller allocated?" has to be answered up front, on the CPU, before anything is submitted.

dynamic-shape-explainer.md
miaobin: Not noise at all! I really appreciate the thorough read. A number of your comments led directly to changes in the explainer, and your comments are genuinely useful for pressure-testing whether the explanation actually holds up. Thanks for taking the time!

dynamic-shape-explainer.md
theHamsta: > It grew from what the models we tested actually needed rather than from a designed list; 

Could an actual design for WebNN have a similar list of limited operators that might grow over time? It would give devs and implementations some orientation on what is expected to work and allow to make certain assumptions on shape

That subgraphs for shape inference for dynamic operators always originate at `shape` nodes or other dynamic operators and that their outputs are always the shape inputs of dynamic operators, is something I didn't realize during the read of the document but only when thinking about it over the weekend. It simplifies what dynamic shapes can do. It falls naturally out of this document that `computeShape` only depends on input shapes. Could it be mentioned in this document?

miaobin: > Could an actual design for WebNN have a similar list of limited operators that might grow over time?

Yes, that is the intent, and I'll make it explicit. A required minimum set, normative, that later revisions can grow the same way the operator set itself grows. A floor rather than a ceiling: an implementation may resolve more, and one that cannot resolve a chain must say so cleanly rather than guess. That is also what keeps it compatible with @fdwr's earlier point: the list constrains what an implementation must be able to *resolve*, not what a graph may *express*.

> Could it be mentioned in this document?

Yes, the invariant is currently one sentence near the end of that section, and only one of its two ends is stated. I'll state both, up front:

- every chain **ends** at the shape argument of a `*Dynamic` operator — that is what makes it a shape chain, and the only way a computed value can influence a shape;
- every chain **begins** at a `shape()` output or a build-time **constant** — those constants being exactly the ones your `shapeConstant` proposal would have the caller declare;

Therefore the interpreter never reads tensor data, and `computeShapes()` is a pure function of the input shapes.

miaobin:
Hi @theHamsta, thank you for the depth of this review; it changed the document in several places. I'd like to check whether the current state is one you'd be comfortable approving.

anssiko:
Last call for comments. Editors will merge this by our next meeting this Thu.

dynamic-shape-explainer.md
```diff
@@ -264,7 +265,7 @@ Validation splits across two phases:
...
+Some deferral is inherent: a constraint whose truth depends on a concrete value cannot be settled before that value exists. How *much* is deferred is a design choice, and this proposal defers nearly all of it — a dynamic dimension is an opaque name rather than a symbolic expression, so build time can only reject contradictions provable between names ([Symbolic shape expressions](#symbolic-shape-expressions) covers the alternative). So the shape resolution and validation a static graph completes once at build time now runs at inference time — as a gatekeeper on every `dispatch`, before the graph executes. This adds per-inference overhead, so an implementation should optimize the common cases: because the result is a pure function of the input shapes, it can skip re-validation when a dispatch repeats a set of input shapes it has already validated.
```
fdwr: > ...so build time can only reject contradictions provable between names (Symbolic shape expressions covers the alternative). So the shape resolution and validation a static graph completes once at build time now runs at inference time...

I'm having a hard time parsing this inverted wording - it sounds like the complicated word salad an LLM might produce :b (sorry if you actually wrote this sentence 😅). How about:

> So inference now computes the shape resolution and shape validation that would have previously completed at build time, preceding every `dispatch`.

dynamic-shape-explainer.md
```diff
@@ -279,12 +280,24 @@ bool DimensionsAreDefinitelyUnequal(Dimension a, Dimension b) {
...
 
+This invariant is one-sided, and only its upper bound is a contract: an implementation **must not** reject a graph that some assignment of concrete input shapes would make valid. Nothing requires it to be good at detecting the rest. The predicate above is the weakest way to honor that — it proves a contradiction only between two static dimensions — and an implementation carrying symbolic *expressions* may reject more, and earlier. Concatenating two `[batch, seq]` tensors along axis 1 doubles the element count, for instance, so reshaping that result back to `[batch, seq]` cannot hold for any input: our model sees only an opaque name for the concatenated axis and defers, while an implementation that knows the axis is `2 * seq` can prove the graph dead and may fail `build()` on it — as some backends performing this analysis internally already do. That is legitimate and costs nothing in portability, since such a graph would fail at every dispatch in any case. It does mean that a successful `build()` is not a portable promise that a graph will ever run — only that this implementation could not prove otherwise.
```
fdwr: > This invariant is one-sided, and only its upper bound is a contract ...

Huh? 🤔 If this from GPT, please avoid that model in favor of Claude or pure human writing that generates clearer wording. English is my native language, and I can't figure out what it's trying to say.

```diff
- This invariant is one-sided, and only its upper bound is a contract: an implementation **must not** reject a graph that some assignment of concrete input shapes would make valid. Nothing requires it to be good at detecting the rest. ... it proves a contradiction only between two static dimensions ...
+ An implementation **must not** reject a graph based on validly assigned concrete input shapes. ... Build time would only fail with static dimensions ...
```

dynamic-shape-explainer.md
```diff
@@ -349,8 +362,21 @@ The first cut loosened each operator's build-time validator independently to tol
...
+Dynamic shapes add no new **declarative** fingerprinting or cross-origin surface: no value the API returns carries device or environment information a static graph does not, and shape computation and inference read only shapes the page itself supplied. The synthesized names for derived dimensions are newly observable output, but they are derived from the graph the page itself constructed (operation type, index, and axis) and carry nothing about the device or the environment; they are also explicitly unstable and not part of the API contract, so nothing may be inferred from a change in one.
```
fdwr:
```diff
- Dynamic shapes add no new **declarative** fingerprinting or cross-origin surface: no value the API returns carries device or environment information a static graph does not ...
+ Dynamic shapes add no new fingerprinting or cross-origin surface: the API returns no value that carries device or environment information which a static graph wouldn't already ...
```

Let's humanize this with contractions, transitionals, relative pronouns (that/which), and more direct clause order.

dynamic-shape-explainer.md
```diff
@@ -365,12 +391,12 @@ The renderer is untrusted, so the service must remain safe on any graph a compro
...
+`computeShapes()` gives an implementation a natural place to do that preparation, since the caller has just named a set of concrete shapes. How good a predictor that is depends on why it was called: a framework sizing output tensors for a dispatch it is about to make is one thing, a caller comparing candidate output shapes and running none of them is another, and an implementation cannot distinguish them. The call is also synchronous, which bounds what it can reasonably do there. Even where preparation is warranted it pays off only if the shape is reused, and an implementation has no way to know which shapes are worth keeping ready. One direction worth exploring is to let the caller say so: build with different input sizes returning several `MLGraph`s, each bound to a set of concrete shapes and all sharing one copy of the weights.
```
fdwr: > How good a predictor that is depends on why it was called: a framework sizing output tensors for a dispatch it is about to make is one thing, a caller comparing candidate output shapes and running none of them is another, and an implementation cannot distinguish them. 

What is the above sentence saying/adding?

```diff
- Even where preparation is warranted it pays off only if the shape is reused, and an implementation has no way to know which shapes are worth keeping ready.
+ Shape preparation only pays off if the shapes are reused, and an implementation has doesn't know which shapes are worth keeping ready.
```

dynamic-shape-explainer.md
```diff
@@ -380,7 +406,9 @@ Currently, the `shape()` operator returns an operand's entire shape as a 1D tens
...
 
 ## Open Questions
-- **Tensor-Derived sizes.** Whether, and how, to admit dimensions that depend on tensor *values*, which this proposal places out of scope.
+- **Tensor-Derived sizes.** Whether, and how, to admit dimensions that depend on tensor *values* — an output whose extent is `NonZero`-shaped rather than a function of the input shapes. This proposal places them out of scope, and draws that boundary structurally rather than by classifying tensors: follow a shape chain back and it bottoms out either at `shape()` outputs and build-time constants, in which case it resolves before dispatch, or at a tensor's data, in which case it is rejected. Admitting the second class needs more than interpreter coverage. Something has to allocate the output before its size is known, which requires an upper bound to allocate against (see [Bounded (min/max) dimensions](#bounded-minmax-dimensions)) and a way to report the size actually produced; and `computeShapes()` could not answer for such an output at all, since the answer does not exist until the graph runs.
```
fdwr:
```diff
- **Tensor-Derived sizes.** Whether, and how, to admit dimensions that depend on tensor *values* — an output whose extent is `NonZero`-shaped rather than a function of the input shapes. This proposal places them out of scope, and draws that boundary structurally rather than by classifying tensors ...
+ **Tensor-Derived sizes.** Dimensions that depend on tensor *values* are out of scope for this proposal ...
```

fdwr (commented):
Some of the newly added sentences need polish for our human readers. Otherwise the latest iteration seems fine to me.

dynamic-shape-explainer.md
miaobin: Ha, this sentence did come out of GPT editing pass and clearly needed a human sanity check. Updated to your phrasing, much better. Thanks for catching it!

dynamic-shape-explainer.md
miaobin: This paragraph was genuinely hard to follow😂. Rewrote it into shorter sentences and fixed an ambiguous reference ("our model" → "this proposal", which was confusable with an actual model). Should read a lot cleaner now and let me know if anything's still unclear! (Maybe Claude is a better choise🧐)

dynamic-shape-explainer.md
miaobin: Done.

dynamic-shape-explainer.md
miaobin: This description has been simplified.

> What is the above sentence saying/adding?

The above sentence was originally intended to address theHamsta's concern that `dispatch` is not necessarily called after `computeShapes()`.

dynamic-shape-explainer.md
miaobin: Done.

miaobin:
> Some of the newly added sentences need polish for our human readers. Otherwise the latest iteration seems fine to me.

@fdwr  I've pushed two commits:
On one hand I've fixed the five places you flagged. I used your wording where it kept the meaning. On the other hand, I've applied the same cleanup to the other sentences added since your last approval: the Design Discussion intro, the shape chain part of Shape computation at dispatch, Symbolic shape expressions, the timing paragraph, and the Making the shape subgraph explicit open question. It changes wording only, not the design. And I also fixed a broken code span in the reshape example. PTAL, thanks!

anssiko:
Resolved as https://github.com/webmachinelearning/webnn/issues/883#issuecomment-5817718631

fdwr (approved):
👍 Thanks Bin, humans happy 😁.

PR #953, opened 2026-09-04
Bugfix: Allow 0-D output for gather()

### LINKED ISSUE #946: `gather` should allow 0D output
The spec allows a scalar (0-D) `indices` operand. Since `output rank = rank(input) + rank(indices) − 1`, a scalar index over a 1-D input yields a 0-D output:

```
input   : [4]   (rank 1)
indices : []    (rank 0, scalar)
axis    : 0
=> output: []   (rank 0)
```

But `gather`'s `output` `rankRange.min` is currently 1, which rejects this valid case. The minimum output rank should be 0.

huningxin:
The indices operand already allows rank 0, and the output rank is rank(input) + rank(indices) - 1. So a rank-1 input gathered with scalar indices produces a rank-0 output.

The algorithm steps needed no change; they already produce an empty output shape for this case.

Fixes #946


fdwr (approved):
👍

PR #955, opened 2026-09-11
Enforce validation when creating an MLOperandDescriptor

### LINKED ISSUE #949: Validate LSTM/GRU full-sequence output regardless of the `returnSequence` flag
Some underlying backends allocate the full-sequence output tensor internally regardless of the `returnSequence` flag but do not validate its size, which might cause an out-of-bounds write. We should always validate the full-sequence output descriptor [steps, numDirections, batchSize, hiddenSize] so that oversized shapes are rejected before reaching the backend.

shiyi9801:
Fix https://github.com/webmachinelearning/webnn/issues/949

This PR makes `create an MLOperandDescriptor` validate the resulting shape via checking dimensions, so oversized outputs are rejected during graph building. `concat()`, `pad()`, and the pooling ops are restructured to construct their descriptor from the final computed shape instead of mutating one created earlier.

This PR also adds unconditional validation for the full-sequence output of `gru()`/`lstm()`, which some platforms produce regardless of the `returnSequence` flag.

***
<a href="https://pr-preview.s3.amazonaws.com/shiyi9801/webnn/pull/955.html" title="Last updated on Sep 14, 2026, 8:47 AM UTC (816b267)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/webmachinelearning/webnn/955/36224cf...shiyi9801:816b267.html" title="Last updated on Sep 14, 2026, 8:47 AM UTC (816b267)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1686,6 +1686,7 @@ An {{MLOperandDescriptor}} |A| is <dfn for=MLOperandDescriptor>equal</dfn> to an
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |dataType|.
     1. Set |descriptor|.{{MLOperandDescriptor/shape}} to a [=list/clone=] of |shape|.
+    1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
     1. Return |descriptor|.
 </details>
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1819,6 +1686,7 @@ An {{MLOperandDescriptor}} |A| is <dfn for=MLOperandDescriptor>equal</dfn> to an
     1. Let |descriptor| be a new {{MLOperandDescriptor}}.
     1. Set |descriptor|.{{MLOperandDescriptor/dataType}} to |dataType|.
     1. Set |descriptor|.{{MLOperandDescriptor/shape}} to a [=list/clone=] of |shape|.
+    1. If [=MLOperandDescriptor/checking dimensions=] given |descriptor| returns false, then [=exception/throw=] a {{TypeError}}.
```
huningxin, line 1689: This is a good fix — it ensures all output operand descriptors are validated, which was missing before.

Since this now applies to every computed output descriptor (not just GRU/LSTM), could you mention that scope in the PR title and commit log? A number of operators — conv2d, gemm, matmul, gather, tile, expand, where, quantize/dequantizeLinear, argMin/argMax, and the broadcasting element-wise ops — gain new TypeError paths here, and reviewers should know that going in.

Two gaps the constructor-side check can't reach:

`concat` constructs the descriptor from first's shape and then mutates shape[axis] to the accumulated size, so the validated shape isn't the final one. It needs a check after the mutation.

`reshape` copies input's descriptor and sets shape directly without calling the constructor. The element count is already covered by the existing product check, but the byte length / rank limits aren't.

Could you also audit for any other methods that bypass the constructor or change dimensions afterwards?

shiyi9801: > Could you also audit for any other methods that bypass the constructor or change dimensions afterwards?

Done! Fixed the constructor of `concat()`, `pad()`, and the pooling ops.

`reshape` already validates that the output tensor has the same element count as the input tensor, which is ensured to be valid, so it does not need additonal checks.

huningxin: Thanks — concat looks right now, and good catch on pad and pool2d.

reshape still copies input.[[descriptor]] and sets shape directly without going through the constructor, so it doesn't get the new validation. As you said, the element count is already covered by the existing product check, but the rank limit isn't — could you convert it to use creating an MLOperandDescriptor for consistency?

